PR #1, opened 2019-12-13
test-case: add verify-ucm-config.sh

libinyang:
This patch add the case of verifying the ucm configuration files.
This patch doesn't support UCMv2. And UCMv2 verifying patch will be
submitted after DUT has supported UCMv2.

Signed-off-by: Libin Yang <libin.yang@linux.intel.com>

### DIFF
--- a/test-case/verify-ucm-config.sh
+++ b/test-case/verify-ucm-config.sh
@@ -1,0 +1,192 @@
+#!/bin/bash
+
+##
+## Case Name: verify-ucm-config
+## Preconditions:
+##    N/A
+## Description:
+##    verify the ucm configuration
+## Case step:
+##    1. load the ucm config file
+##    2. check ucm can be run by alsaucm correctly or not:
+##       check the ucm configuration file meets the syntax or not
+##       check if the devices can be enabled or not
+##       check if the controls can be set correctly or not
+##       and so on. Details can be found in cmd 'alsaucm --help'
+## Expect result:
+##    command line check with $? without error
+## Example of alsaucm output of verbs:
+##   0: HiFi
+##    Play HiFi quality Music.
+## Example of alsaucm output of devices:
+##   0: Speaker
+##     description of Speaker.
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+func_opt_parse_option $*
+
+declare -A verb_array
+
+# get the cards all verbs. The verbs are the items of SectionUseCase in
+# UCM config file.
+# For example, There is:
+#  SectionUseCase."HiFi" {
+#      File "HiFi.conf"
+#      Comment "Play HiFi Music"
+#  }
+# Then we will get the below information after we run `alsaucm -c "$cardname" list _verbs`
+#   0: HiFi
+#    Play HiFi quality Music.
+func_get_verbs()
+{
+    dlogc "alsaucm -c '$cardname' list _verbs"
+    verbs=`alsaucm -c "$cardname" list _verbs 2>/dev/null`
+    [[ $? -ne 0 ]] && dloge "'$cardname' list _verbs failed!" && exit 1
+    [[ -z "$verbs" ]] && dloge "'$cardname' ucm has no verbs!" && exit 1
+
+    # save the verbs into array verb_array
+    local OLD_IFS="$IFS"
+    IFS=$'\n'
+    ((verbs_desc_line=0))
+
+    # verb counter
+    ((verb_i=0))
+    # $desc_line is used to define it is the even line (description line) or the odd line
+    # verb format: there are 2 lines in each verb.
+    # The first line  is the verb's identifier;
+    # The second line is the verb's descriptor
+    # For example, run `alsaucm -c cardname list _verbs` and suppose we get:
+    #   0: HiFi
+    #    Play HiFi quality Music.
+    # the verb is 'HiFi'
+    # the verb's description is "Play HiFi quality Music"
+    for verb in $verbs; do
+	if [ $verbs_desc_line -eq 0 ]; then
+	    # get the verb identifier. In the upper example, it is "HiFi"
+	    verb=`echo $verb|awk '{print $2}'`
+            [[ -z "$verb" ]] && dlogw "verb is null, skipping" && continue
+            dlogi "found verb: $verb"
+	    verb_array[$verb_i]="$verb"
+	    ((verb_i++))
+	    ((verbs_desc_line=1))
+	else
+            # Let's skip the description
+	    ((verbs_desc_line=0))
+	fi
+    done
+    IFS="$OLD_IFS"
+}
+
+func_verify_verb_device()
+{
+    local cardname=$1
+    local verb=$2
+    local -A dev_array
+
+    # 1. get all the devices used by the verb
+
+    # The device is defined in "SectionDevice". For example:
+    #  SectionDevice."Speaker" {
+    #      xxxx
+    #  }
+    # Then the device is "Speaker"
+    dlogc "alsaucm -c '$cardname' set _verb $verb list _devices"
+    devices=$(alsaucm -c "$cardname" set _verb $verb list _devices 2>/dev/null)
+    [[ $? -ne 0 ]] && dloge "list device failed" && return 1
+
+    local OLD_IFS="$IFS"
+    IFS=$'\n'
+    ((devs_desc_line=0))
+
+    # 2. save the devices into the array dev_array
+
+    # device counter
+    ((device_i=0))
+    # After run $(alsaucm -c "$cardname" set _verb $verb list _devices),
+    # it will show all the devices it supports. Each device takes 2 lines output.
+    # The first line is the device identifier;
+    # The second line is the device description.
+    # For example:
+    #   0: Speaker
+    #     description of Speaker.
+
+    for device in $devices; do
+	if [ $devs_desc_line -eq "0" ]; then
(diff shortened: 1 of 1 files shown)


test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+ucm_file=${ucm_folder}/${cardname}/${cardname}.conf
+
+[[ ! -e "${ucm_file}" ]] && dloge "ucm configuration: $ucm_file not exists!" && exit 2
+
+# currently we only supports old ucm configuration.
+# only after alsa-lib, alsa-utils and pulseaudio are updated, UCMv2 will be supported.
+# check ucm version
+[[ "$(grep "Syntax 2" "${ucm_file}")" ]] && dloge "not support UCMv2 currently" && exit 2
+
+# use alsaucm to check the ucm configurations
```
xiulipan: Can we dump some info of  the ucm_file here?

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+dlogc "alsaucm open '$cardname'"
+# maybe use cardname let alsaucm choose ucm file
+alsaucm open "$cardname"
+[[ $? -ne 0 ]] && dloge "open card '$cardname' failed!" && exit 1
+
+dlogc "alsaucm -c "$cardname" reload"
+alsaucm -c "$cardname" reload
+[[ $? -ne 0 ]] && dloge "card '$cardname' reload failed. Please check the card amixer setting." && exit 1
```
xiulipan: how to check amixer setting when ucm reload failed? If we could not give detail, please just remove the advice here.

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+
+dlogc "alsaucm -c '$cardname' list _verbs"
+verbs=`alsaucm -c "$cardname" list _verbs`
+[[ $? -ne 0 ]] && dloge "'$cardname' list _verbs failed!" && exit 1
+[[ -z "$verbs" ]] && dloge "'$cardname' ucm has no verbs!" && exit 1
+((i=0))
```
xiulipan: What is this i used for here? Can you give it a more verbose name

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+
+OLD_IFS="$IFS"
```
xiulipan: I also see `OLD_IFS` with local prefix in `func_verify_verb_device()`. Could you check if you are miss use or use in purpose. 
Maybe change a name to avoid confusion.

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+IFS=$'\n'
```
xiulipan: Do you want to play trick with IFS to get each line of `$verbs`? Let is very hard to understand.
@Bin-QA can you give some suggestion here to avoid play trick like this. It is very hard to understand here.

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+        ret=`alsaucm -c "$cardname" set _verb $verb get _verb`
+        [[ $? -ne 0 ]] && dloge "set verb: $verb failed!" && IFS="$OLD_IFS" && exit 1
+        ret=`echo $ret |grep "$verb" 2>/dev/null`
+        [[ -z "$ret" ]] && dloge "'$verb' is not the selected verb" && IFS="$OLD_IFS" && exit 1
+
+        func_verify_verb_device $cardname $verb
+        ret=$?
+        [[ $ret -ne 0 ]] && IFS="$OLD_IFS" && exit $ret
+
+        # verb format: the first line is verb and the second line is description.
```
xiulipan: Can you give some example of verb here? I think this kind of parsing is not very strong and easy be broken.

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+        # Let's skip the description
+        ((i=1))
+    else
+        ((i=0))
+    fi
+done
+IFS="$OLD_IFS"
```
xiulipan: What is this line used for? restore the IFS before exit?

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+        [[ -z "$verb" ]] && dlogw "verb is null, skipping" && continue
+        dlogi "found verb: $verb"
+
+        dlogc "alsaucm -c '$cardname' set _verb $verb get _verb"
+        ret=`alsaucm -c "$cardname" set _verb $verb get _verb`
+        [[ $? -ne 0 ]] && dloge "set verb: $verb failed!" && IFS="$OLD_IFS" && exit 1
+        ret=`echo $ret |grep "$verb" 2>/dev/null`
+        [[ -z "$ret" ]] && dloge "'$verb' is not the selected verb" && IFS="$OLD_IFS" && exit 1
+
+        func_verify_verb_device $cardname $verb
```
Bin-QA: 1. IFS change without export, just effect in your script, so line 106 restore can be ignore
2. IFS will effect the function, because without `local`, all variable is global, so there will conflict with `func_verify_verb_device ` define,

So you do the too many thing in your IFS, it will effect too many commands

I suggest you store the data into array, then loop the array like this [line to array](https://github.com/thesofproject/sof-test/blob/master/case-lib/hijack.sh#L27)

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+}
+
+ucm_folder=/usr/share/alsa/ucm
+
+sofcard=${SOFCARD:-0}
+cardname=$(sof-dump-status.py -s $sofcard)
+[[ $? -ne 0 ]] && exit 1
+
+# TODO: check the audio card dmi name (dmi name is not recommanded).
+ucm_file=${ucm_folder}/${cardname}/${cardname}.conf
```
Bin-QA: this case need libasound2-data package to install/some thing, I think you need update other file to add this package

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+        [[ -z "$ret" ]] && dloge "'$verb' is not the selected verb" && IFS="$OLD_IFS" && exit 1
+
+        func_verify_verb_device $cardname $verb
+        ret=$?
+        [[ $ret -ne 0 ]] && IFS="$OLD_IFS" && exit $ret
+
+        # verb format: the first line is verb and the second line is description.
+        # Let's skip the description
+        ((i=1))
+    else
```
Bin-QA: I think you try to use a flag to mark the case pass/failed  
But when you failed, you already exit the script, so the flag `i` is not necessary

test-case/verify-ucm-config.sh
libinyang: what infomation do you mean?

test-case/verify-ucm-config.sh
libinyang: OK. I will remove the advice.

test-case/verify-ucm-config.sh
libinyang: That one is local version. I will define this one as local too.

test-case/verify-ucm-config.sh
libinyang: I believe this is a very normal way to do such things to use IFS. Or any suggestions?

test-case/verify-ucm-config.sh
libinyang: Sorry, I can't understand what do you mean.

test-case/verify-ucm-config.sh
libinyang: Yes, restore it.

test-case/verify-ucm-config.sh
```diff
@@ -0,0 +1,123 @@
...
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+func_opt_parse_option $*
+
+func_verify_verb_device()
+{
+    local cardname=$1
+    verb=$2
```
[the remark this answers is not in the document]
libinyang: right. will fix it.

test-case/verify-ucm-config.sh
libinyang: we need updated libasound2.so. This is another feature. I'm doing it in another thread.

test-case/verify-ucm-config.sh
libinyang: No. $i is used for judge it is the even line or odd line

test-case/verify-ucm-config.sh
libinyang: yes, I have dumped it in previous line: dlogi "found verb: $verb"

test-case/verify-ucm-config.sh
libinyang: OK, I will add an example ahead of this section to let it be more readable.

libinyang:
It seems the alsaucm test case is not so easy to understand (it's a totally new case for us all). I will try to add as more comments as possible to help others to review the code easily.

libinyang:
patch updated

libinyang:
ping @xiulipan

libinyang:
ping for merge

xiulipan:
@Bin-QA @keqiaozhang @aiChaoSONG Please help to give a second approval.

libinyang:
ping for merge

PR #2, opened 2019-12-13
Card name

libinyang:
The patchset does:
1. fix the getting card short name issue
2. provide the card long name inferface

### DIFF
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -98,7 +98,7 @@ class clsSYSCardInfo():
         for idx in range(0, len(output), 2):
             card_info={}
             card_info['id'] = output[idx].strip().split(' ')[0]
-            card_info['short'] = output[idx].split(' ')[-1].strip()
+            card_info['short'] = output[idx].split(' - ')[-1].strip()
             card_info['type']=output[idx].strip().split()[1][1:]
             card_info['desc']=output[idx + 1].strip()
             card_info['codec']=[]
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -100,7 +100,7 @@ class clsSYSCardInfo():
             card_info['id'] = output[idx].strip().split(' ')[0]
             card_info['short'] = output[idx].split(' - ')[-1].strip()
             card_info['type']=output[idx].strip().split()[1][1:]
-            card_info['desc']=output[idx + 1].strip()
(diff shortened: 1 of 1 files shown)


tools/sof-dump-status.py
```diff
@@ -98,7 +98,7 @@ def _loadACard(self):
         for idx in range(0, len(output), 2):
             card_info={}
             card_info['id'] = output[idx].strip().split(' ')[0]
-            card_info['short'] = output[idx].split(' ')[-1].strip()
+            card_info['short'] = output[idx].split(' - ')[-1].strip()
```
xiulipan: A quick question here. is the spliter **space_dash_space** here? it looks very like to '-'
Please clarify. If there is an example is better.

tools/sof-dump-status.py
```diff
@@ -100,7 +100,7 @@ def _loadACard(self):
             card_info['id'] = output[idx].strip().split(' ')[0]
             card_info['short'] = output[idx].split(' - ')[-1].strip()
             card_info['type']=output[idx].strip().split()[1][1:]
-            card_info['desc']=output[idx + 1].strip()
+            card_info['longname']=output[idx + 1].strip()
```
xiulipan: Do you mean the old description is actually the longname?
I remember there is some requirement for alsalib to support longname? Do we need to check the alsalib here first?

tools/sof-dump-status.py
Bin-QA: so what is different between those change? you find the short name with the 'space' ?

libinyang: @xiulipan @Bin-QA 
Yes, it is space_dash_space. Our SOF is ok for space only because we don't use any space in our short card name. But as the test case is targeted to be used widely, we should consider some corner cases. I found this bug when I try to run it on my legacy hda driver.

tools/sof-dump-status.py
libinyang: OK. I will check it later

libinyang:
checked. It is longname. You can refer sound/core/init.c for details.

xiulipan:
> I can accept long name, but I couldn't understand why need to use ' - ' instead ' '
I think @libinyang have explained this with 

> I found this bug when I try to run it on my legacy hda driver.

@Bin-QA Please re-check and approval if you think this PR is OK.

PR #4, opened 2019-12-18
Feature: add env PATH into hijack sudo function

Bin-QA:
Now skip type 'sudo env PATH="PATH"' from case command

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -76,11 +76,11 @@ sudo()
             return $?
         ;;
         '1')    # sudo without passwd
-            eval $(echo "$SUDO_CMD $*")
+            eval $(echo "$SUDO_CMD env 'PATH=$PATH' $*")
             return $?
         ;;
         '2')    # sudo need passwd
-            eval $(echo "echo '$SUDO_PASSWD' | $SUDO_CMD -S $*")
+            eval $(echo "echo '$SUDO_PASSWD' | $SUDO_CMD -S env 'PATH=$PATH' $*")
             return $?
         ;;
         *)      # without sudo permission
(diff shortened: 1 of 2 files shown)


case-lib/hijack.sh
```diff
@@ -76,11 +76,11 @@ sudo()
             return $?
         ;;
         '1')    # sudo without passwd
-            eval $(echo "$SUDO_CMD $*")
+            eval $(echo "$SUDO_CMD env 'PATH=$PATH' $*")
             return $?
         ;;
         '2')    # sudo need passwd
-            eval $(echo "echo '$SUDO_PASSWD' | $SUDO_CMD -S $*")
+            eval $(echo "echo '$SUDO_PASSWD' | $SUDO_CMD -S env 'PATH=$PATH' $*")
```
xiulipan: will `-E` help here to pass the ENV into sudo

Bin-QA: try with -E option which is not work in our script, because we are not `export` the ENV

Bin-QA: try with option `--preserve-env=PATH` it will report `command not found`

it seems is hijack effect

xiulipan:
@aiChaoSONG @libinyang @keqiaozhang Please help to review.

aiChaoSONG:
When I submit patches to kernel, Pierre told me I should explain my patch clearly in commit message. It's a what, how and why steps. you should make it clear that what the problems is(why it need a fix),  how you solve it, and why you solve it this way (safe? clarity? or ?). 
I suggest to make our commit message more verbose, what do you think? @xiulipan @Bin-QA 

@Bin-QA I see what you did in this patch, but I didn't get the reason. could you please explain deeper?

PR #5, opened 2019-12-18
Feature: Add case stress help function support

Bin-QA:
case/lib.sh to add func_lib_trigger_stress
which detect current script run from local/remote
to keep script can run finish when lost the remote connect

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -17,8 +17,14 @@ if [ ! "$SOFCARD" ];then
     SOFCARD=$(grep '\]: sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
 fi
 
-declare -g DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+if [ ! "$DMESG_LOG_START_LINE" ];then
+    declare -g DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+fi
 declare -g SOF_LOG_COLLECT=0
+# To record output to STRESS_OUTPUT_LOG
+declare -g STRESS_OUTPUT_LOG
+# To record script status to STRESS_STATUS_LOG
+declare -g STRESS_STATUS_LOG
 
 func_lib_setup_kernel_last_line()
 {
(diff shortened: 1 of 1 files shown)


case-lib/lib.sh
```diff
@@ -85,3 +91,29 @@ func_lib_restore_pulseaudio()
...
+        # convert itself to nohup without ssh lost connect when case trigger by ssh connect
+        if [ $PPID -ne 1 ];then
+            dlogi "redirect the output please check for $STRESS_OUTPUT_LOG"
+            nohup $(cat $LOG_ROOT/cmd-orig.txt) > $STRESS_OUTPUT_LOG &
+            # delay for nohup command apply
+            sleep 1s
+            # now output current process with tail command
+            clear
+            tail -f $STRESS_OUTPUT_LOG
+	    # TODO: need to verify with S3 status to confirm script block
```
xiulipan: Intend here.

case-lib/lib.sh
```diff
@@ -85,3 +91,29 @@ func_lib_restore_pulseaudio()
...
+func_lib_trigger_stress()
+{
+    export LOG_ROOT=$LOG_ROOT
+    STRESS_OUTPUT_LOG="$LOG_ROOT/stress.txt"
+    STRESS_STATUS_LOG="$LOG_ROOT/status.txt"
+    [[ -f $STRESS_OUTPUT_LOG ]] && return
+
+    touch $STRESS_OUTPUT_LOG
+    touch $STRESS_STATUS_LOG
+    [ ! -f $LOG_ROOT/cmd-orig.txt ] && cp $LOG_ROOT/cmd.txt $LOG_ROOT/cmd-orig.txt
```
xiulipan: It may good to add some comments here to explain why cp is needed here.

case-lib/lib.sh
```diff
@@ -85,3 +91,29 @@ func_lib_restore_pulseaudio()
     unset PULSECMD_LST
     declare -ag PULSECMD_LST
 }
+
+func_lib_trigger_stress()
+{
+    export LOG_ROOT=$LOG_ROOT
+    STRESS_OUTPUT_LOG="$LOG_ROOT/stress.txt"
```
xiulipan: Can we have some iteration index here in the log files? Or this will contain full round stress test logs?

Bin-QA: because it have different stress type
So here I write for the reboot case, it will trigger this function after each time system reboot
But from the S3, I think this type case just to trigger this function once
So maybe add status control function in it?

Bin-QA:
This change is not the requirement with case design, so close it

PR #3, opened 2019-12-18
Bugfix: tools/sof-boot-once.sh builtin & relative path

Bin-QA:
builtin for fix typo  
relative path to add folder without start with '.'

Signed-off-by: Wu, BinX <binx.wu@intel.com>

PR #6, opened 2019-12-18
Case: add stress reboot test-case

Bin-QA:
Test case reboot for the stress test type

This case base on #3 #4 #5 change

### DIFF
--- a/test-case/stress-reboot.sh
+++ b/test-case/stress-reboot.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+##
+## Case Name: stress test reboot
+## Preconditions:
+##    N/A
+## Description:
+##    run reboot for the stress test
+##    default duration is 10s
+##    default loop count is 3
+## Case step:
+##    1. check system status is correct
+##    2. wait for the Random value
+##    3. trigger for the reboot
+## Expect result:
+##    Test execute without report error in the LOG
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=5
+
+func_opt_parse_option $*
+
+func_lib_check_sudo
+func_lib_trigger_stress
+
+loop_count=${OPT_VALUE_lst['l']}
+
+[[ $loop_count -le 0 ]] && exit 0
+# write the orig & status count to the status file
+echo $loop_count >> $STRESS_STATUS_LOG
+count=$(head -n 1 $STRESS_STATUS_LOG|awk '{print $1;}')
+# because write record by desc
+current=$[ $count - $loop_count + 1 ]
+
+dlogi "Round: $current/$count"|tee -a $STRESS_OUTPUT_LOG
+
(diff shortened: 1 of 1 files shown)


test-case/stress-reboot.sh
```diff
@@ -0,0 +1,87 @@
...
+sed -i '$s/$/ pass/' $STRESS_STATUS_LOG
+
+# run the script for the next round
+loop_count=$[ $loop_count - 1 ]
+
+# Store the environment for the next round load this script
+dlogi "Setup the environment for next round bootup" |tee -a $STRESS_OUTPUT_LOG
+sudo sof-boot-once.sh export LOG_ROOT=$LOG_ROOT
+# just keep the last dmesg bootup kernel line for the success
+sudo sof-boot-once.sh export DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
```
xiulipan: Why call `sof-boot-once.s` multiple times? Maybe some re-name for this helper function is needed.

test-case/stress-reboot.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# run this script for next boot up round
+dlogi "sof-boot-onece.sh ${BASH_SOURCE[0]} -l $loop_count" |tee -a $STRESS_OUTPUT_LOG
```
xiulipan:
```suggestion
dlogc "sof-boot-onece.sh ${BASH_SOURCE[0]} -l $loop_count" |tee -a $STRESS_OUTPUT_LOG
```

test-case/stress-reboot.sh
```diff
@@ -0,0 +1,87 @@
...
+# because for the 1st round those information write by nohup command redirect
+# but after system reboot the nohup redirect is missing
+dlogi "Check status begin" | tee -a $STRESS_OUTPUT_LOG
+for i in ${verify_lst[*]}
+do
+    #$(dirname ${BASH_SOURCE[0]})/$i
+    if [ 0 -ne 0 ];then
+        # last line add failed keyword
+        sed -i '$s/$/ fail/' $STRESS_STATUS_LOG
+        dlogi "$i: fail" |tee -a $STRESS_OUTPUT_LOG
```
xiulipan: Can you hijack the dlog* cmd to wrap the `|tee -a $STRESS_OUTPUT_LOG` when `STRESS_OUTPUT_LOG` is defined?

Bin-QA: that's the good idea, I will try it

test-case/stress-reboot.sh
Bin-QA: convert sof-boot-once.sh to the help function?  
but I think it maybe useful for the stand along, expect we just handle it for the reboot

here will cause other problem: how to clean up the rc.local file  
Now sof-boot-once.sh will do the write and cleanup rc.local it also is reason why this script is stand alone

test-case/stress-reboot.sh
Bin-QA: check with #5 , I don't think here need to modify  
because when run from remote, the `nohup` command will handle all output  
from the local it don't need to handle it  

here use `tee` command because after system reboot,
the `nohup` command is miss, Because  current script is run in `rc.local` which called by root permission

Bin-QA:
Will new design the test-case current close this PR

PR #7, opened 2019-12-18
tools: remove deprecated output fields for PCM

aiChaoSONG:
Output fields 'rate', 'channel' and 'dev' for PCM is inherited from previous topology reader tool, and is deprecated now. Remove these fields for clarity.

we will never use these fields in the future, and before add new features to this tool, we ensure the code is clean.


aiChaoSONG:
@xiulipan @Bin-QA Please help to review this change.

PR #8, opened 2019-12-18
test: add alsabat test

keqiaozhang:
Add a common script for alsabat test.

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+##
+## Case Name: check alsabat
+## Preconditions:
+##    This test case requires physical loopback between playback and capture.
+##    playback <=====>  capture
+##    nocodec : no need to use hw loopback cable, It support DSP loopback by quirk
+## Description:
+##    run alsabat test on the specified pipelines
+## Case step:
+##    1. Specify the pcm IDs for playback and catpure
+##    3. run alsabat test
+## Expect result:
+##    The return value of alsabat is 0
+##
+
+# remove the existing alsabat wav files
+rm -f /tmp/bat.wav.*
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['p']='pcm_p'     	OPT_DESC_lst['p']='pcm for playback. Example: hw:0,0'
+OPT_PARM_lst['p']=1          	OPT_VALUE_lst['p']=''
+
+OPT_OPT_lst['c']='pcm_c'      	OPT_DESC_lst['c']='pcm for capture. Example: hw:1,0'
+OPT_PARM_lst['c']=1             OPT_VALUE_lst['c']=''
+
+OPT_OPT_lst['f']='frequency'    OPT_DESC_lst['f']='target frequency'
+OPT_PARM_lst['f']=1             OPT_VALUE_lst['f']=997
+
+OPT_OPT_lst['n']='frames'     OPT_DESC_lst['n']='test frames'
+OPT_PARM_lst['n']=1             OPT_VALUE_lst['n']=240000
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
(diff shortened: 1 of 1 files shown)


test-case/check-alsabat.sh
```diff
@@ -0,0 +1,87 @@
...
+dlogi "check the PCMs before alsabat test"
+[[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && dloge "Failed to play on PCM: $pcm_p" && exit 1
+[[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && dloge "Failed to capture on PCM: $pcm_c" && exit 1
+
+# alsabat test
+dlogc "alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency"
+alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency &
+sleep 1
+dlogc "alsabat -Cplug$pcm_c -F $frequency"
+alsabat -Cplug$pcm_c -F $frequency
```
Bin-QA: I think you need to add comment/something to restrict pcm_p/pcm_c format  
because here you direct use it, so you need to let other body to know  
format is `hw:0,0` or `0,0`, it seems `0,0` will cause failed

keqiaozhang: Added an example in usage.

test-case/check-alsabat.sh
```diff
@@ -0,0 +1,87 @@
...
+    for file in /tmp/bat.wav.*
+    do
+	    size=`ls -l $file | awk '{print $5}'`
+	    if [[ $size -gt 0 ]]; then
+		    cp $file $LOG_ROOT/
+	    fi
+    done
+}
+
+dlogi "platform is $platform"
```
xiulipan: is the platform needed to output here?

test-case/check-alsabat.sh
```diff
@@ -0,0 +1,87 @@
...
+
+if [ "$pcm_p" = "" ]||[ "$pcm_c" = "" ];
+then
+	dloge "No playback or capture PCM is specified. Skip the alsabat test"
+	exit 2
+fi
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+platform=$(sof-dump-status.py -p)
```
xiulipan: see below about platform

test-case/check-alsabat.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# check the PCMs before alsabat test
+dlogi "check the PCMs before alsabat test"
+[[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && dloge "Failed to play on PCM: $pcm_p" && exit 1
+[[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && dloge "Failed to capture on PCM: $pcm_c" && exit 1
+
+# alsabat test
+dlogc "alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency"
+alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency &
+sleep 1
```
xiulipan: Please add comment to explain why the sleep is needed here

test-case/check-alsabat.sh
```diff
@@ -0,0 +1,87 @@
...
+		    cp $file $LOG_ROOT/
+	    fi
+    done
+}
+
+dlogi "platform is $platform"
+
+# check the PCMs before alsabat test
+dlogi "check the PCMs before alsabat test"
+[[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && dloge "Failed to play on PCM: $pcm_p" && exit 1
```
xiulipan: Please also add some comments about plug used here

test-case/check-alsabat.sh
keqiaozhang: Deleted. This output is not necessary in this case.

test-case/check-alsabat.sh
keqiaozhang: Updated.

test-case/check-alsabat.sh
keqiaozhang: Updated.

test-case/check-alsabat.sh
keqiaozhang: Updated.

PR #11, opened 2019-12-20
Add common opt at case-lib/opt.sh

Bin-QA:
1. As we have lot of same description option so move them
into the common function, so when modify/add can use it
this change is not refer the code logic

2. Change opt parse error exit code to exit 2
Current define exit code
0: pass; 1: fail;2: skip

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -87,7 +87,7 @@ func_opt_parse_option()
             echo -e '    -h |  --help'
             echo -e "\tthis message"
             _func_case_dump_descption
-            exit 0
+            exit 2
         }
 
     # generate the command to load 'getopt'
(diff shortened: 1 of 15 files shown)


wenqingfu:
I would say the code readability drops because of your 1). why bother?  

2) should be a separate PR, don't mix changes together.

Bin-QA:
> I would say the code readability drops because of your 1). why bother?
yes, you are right, I am not realize the readability problem  
Because I will try to add the common option for the almost script  
I will redesign the script process

> 1. should be a separate PR, don't mix changes together.
Got it.

case-lib/opt.sh
```diff
@@ -101,7 +101,7 @@ func_opt_parse_option()
         [ ! "$idx" ] && idx="${_op_long_lst[$1]}"
         if [ "$idx" ]; then
             if [ ${OPT_PARM_lst[$idx]} -eq 1 ]; then
-                [ ! "$2" ] && echo "option: $1 missing parameter, parsing error" && exit 1
+                [ ! "$2" ] && echo "option: $1 missing parameter, parsing error" && exit 2
```
xiulipan: can we use `dlog` here instead of echo

Bin-QA:
After check with requirement, the new PR to fix exit code
```
man bash :
All builtins return an exit status of 2  to  indicate  incorrect
usage, generally invalid options or missing arguments.
```

So I will check with script to add comment of exit 2 means

PR #12, opened 2019-12-24
Feature add: modify dlog* message output to more clean up

Bin-QA:
Current we support the test-case scripot load other test-case script
to finish the test-case, but from the test-case output information
the current script output will mix with sub-test-case script

Now add the 'Sub-Test' flag to the sub-test-case
when end-user catch the terminal output without confuse
for the script process

it will be helpful for maintenance the script

### DIFF
--- a/case-lib/logging_ctl.sh
+++ b/case-lib/logging_ctl.sh
@@ -28,9 +28,18 @@ _func_log_cmd()
     # writing functions
     shopt -s expand_aliases
     if [ "X1" ]; then
+        # PPID: The process ID of the shell's parent.
+        # get Current script parent process name
+        local ppcmd=$(ps -p $PPID -o cmd --noheader|awk '{print $2;}') ext_message=""
+        # confirm this script load by the script, Add the flag for it
+        [[ "$(file $ppcmd 2>/dev/null |grep 'shell script')" ]] && ext_message=" Sub-Test:"
+        _func_logcmd_add_timestamp()
+        {
+            echo $(date -u '+%Y-%m-%d %T %Z') $*
+        }
         for key in ${!LOG_LIST[@]};
         do
-            alias "$key"="echo \$(date -u '+%Y-%m-%d %T %Z') ${LOG_LIST[$key]}"
+            alias "$key"="_func_logcmd_add_timestamp $ext_message ${LOG_LIST[$key]}"
         done
     else
         _func_empty_function() { return 0; }
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -50,13 +50,13 @@ function exit()
 
     case $exit_status in
         0)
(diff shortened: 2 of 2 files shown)


case-lib/logging_ctl.sh
```diff
@@ -28,9 +28,16 @@ _func_log_cmd()
     # writing functions
     shopt -s expand_aliases
     if [ "X1" ]; then
+        local ppcmd=$(ps -p $PPID -o cmd --noheader|awk '{print $2;}') ext_message=""
```
libinyang: Where is PPID defined? 
awk '{print $2;}' in my environment, it is the parameter, not the cmdline if I run it directly locally.

case-lib/logging_ctl.sh
```diff
@@ -28,9 +28,16 @@ _func_log_cmd()
...
+        [[ "$(file $ppcmd 2>/dev/null |grep 'shell script')" ]] && ext_message=" Sub-Test:"
```
libinyang: I think sometimes it is not the sub-test even cmd name is not the $0. For example, the output from lib.sh. This should not a sub-test.

case-lib/logging_ctl.sh
Bin-QA: PPID define by bash  
`PPID   The process ID of the shell's parent.  This variable is readonly.`

case-lib/logging_ctl.sh
Bin-QA: compare with $0 because I want to use it with this pull request: https://github.com/thesofproject/sof-test/pull/5

So when trigger this function, it will load itself to avoid test case stop when lost ssh connect

case-lib/logging_ctl.sh
libinyang: Oh, right, PPID means the parent PID. Please ingore this comment

Bin-QA:
Force update for rebase

Bin-QA:
@libinyang @xiulipan @aiChaoSONG @keqiaozhang 
Please help to review this patch

case-lib/logging_ctl.sh
```diff
@@ -28,9 +28,18 @@ _func_log_cmd()
     # writing functions
     shopt -s expand_aliases
     if [ "X1" ]; then
+        # PPID: The process ID of the shell's parent.
+        # get Current script parent process name
+        local ppcmd=$(ps -p $PPID -o cmd --noheader|awk '{print $2;}') ext_message=""
+        # confirm this script load by the script, Add the flag for it
+        [[ "$(file $ppcmd 2>/dev/null |grep 'shell script')" ]] && ext_message=" Sub-Test:"
+        _func_logcmd_echo_overwrite()
```
xiulipan: I think we can rename this into `_func_logcmd_add_timestamp()`

PR #13, opened 2019-12-25
env_check: remove gcc check in env_check.sh

aiChaoSONG:
Since the binary tplgreader is replaced by tplgtool.py, we don't need gcc as prerequisit
any more. Remove gcc check for clarity


aiChaoSONG:
@xiulipan @Bin-QA @libinyang Please help to review

PR #14, opened 2019-12-25
env_check: fix sed command output error

Bin-QA:
sed command for group check process output error
have more "'" in display command line,
it will cause this command couldn't work by directory paste


aiChaoSONG:
The suggested sed command for group check to user has a not paired single quote, which will 
cause this command couldn't work by directly paste.

Remove redundant single quote for sed to run directly in command line by paste

I suggest change the commit message like this.@Bin-QA

Bin-QA:
> The suggested sed command for group check to user has a not paired single quote, which will
> cause this command couldn't work by directly paste.
> 
> Remove redundant single quote for sed to run directly in command line by paste
> 
> I suggest change the commit message like this.@Bin-QA

Thank you, commit already update

PR #16, opened 2019-12-25
Bug fix: sof-kernel-log-check DSP reset

Bin-QA:
After patch: thesofproject/linux#1676 merged
add "error: status" and "error: cl_dsp_init" into ignore string
to avoid catch DPS reset which can be ignored

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -3,7 +3,7 @@
 begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
-ignore_str="no reply expected,|error: debugfs write failed to idle -16"
+ignore_str="no reply expected,|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init"
 project_key="sof-audio"
 
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0


bardliao:
How about changing the message on kernel?
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda-loader.c#L331
Maybe use dev_warn and "warning:" instead of dev_err and "error:".

Bin-QA:
> How about changing the message on kernel?
> https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda-loader.c#L331
> Maybe use dev_warn and "warning:" instead of dev_err and "error:".

```
        # 0 error: cl_dsp_init
        # 1 error: status
        # 2 error: Error
        # 3 error: iteration
```
If you want modify the source code those information also need to modify:   
Line 0 error: cl_dsp_init  
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda-loader.c#L81
Line 1 error: status  
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda.c#L204
Line 2 error: Error  
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda-loader.c#L331
Line 3 error: iteration  
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda-loader.c#L336

Because from the kernel log, catch 4 error message when DSP reset  
On the other hand, I format the error message to avoid miss some really error
```
dsp_count=$(echo "$err"|grep 'error: cl_dsp_init' -c)
allow_count=$(echo "$err"|grep 'error: cl_dsp_init' -A3|grep 'iteration [01]' -c)
```

xiulipan:
@bardliao @Bin-QA Let wait for the final discussion about the dsp init message. I think this error is not user friendly.

wenqingfu:
> @bardliao @Bin-QA Let wait for the final discussion about the dsp init message. I think this error is not user friendly.

Which discussion do you refer to @xiulipan ?  

I don't see this discussion pattern is finalizing.  I agree with @bardliao. Similarly,  the unexpected reply shouldn't be called error, either. If they are not real errors, then don't call it error.  It freak people out. 

small fix, but please just act with your patch.  @bardliao , can you please fix it?

bardliao:
> > @bardliao @Bin-QA Let wait for the final discussion about the dsp init message. I think this error is not user friendly.
> 
> Which discussion do you refer to @xiulipan ?
> 
> I don't see this discussion pattern is finalizing. I agree with @bardliao. Similarly, the unexpected reply shouldn't be called error, either. If they are not real errors, then don't call it error. It freak people out.
> 
> small fix, big good, please just act with your patch. @bardliao , can you please fix it?

Yes, I will send a kernel PR to fix it.

xiulipan:
@wenqingfu It seems the vacation break the discussion. I also agree with @bardliao that we should not output error when we have retry method.
@ranj063 what is your opinion here, even the S0ix refine https://github.com/thesofproject/linux/pull/1631 may have benefit on this. But the error with retry will still make people get panic.

wenqingfu:
Thanks, @xiulipan @bardliao .  connect the dot with https://github.com/thesofproject/linux/pull/1676 ,  Seems we have overwhelmingly consensus of "let's don't freak people out" there. :)

Bin-QA:
Update for kernel patch merged
@xiulipan @libinyang please help to review this patch

Bin-QA:
`ignore_str="no reply expected,|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init|iteration [01]"`

How about remove 'failed' keyword in `err_str`
if here just use 'failed' to catch DSP reset

xiulipan:
@Bin-QA I think there will also be IPC failed, so let keep it to see if we got any other issues.
Please send the patch to the change 
`ignore_str="no reply expected,|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init|iteration [01]"`

PR #17, opened 2019-12-25
test: add xrun injection test

keqiaozhang:
This script intends to test the ability of XRUN handling for each pipeline.

### DIFF
--- a/test-case/check-xrun-injection.sh
+++ b/test-case/check-xrun-injection.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+
+##
+## Case Name: check-xrun-injection
+## Preconditions:
+##    Need to enable CONFIG_SND_PCM_XRUN_DEBUG in kenrel config
+## Description:
+##    check xrun injection during playback/capture
+##    default duration is 10s
+##    default interval time of xrun injection is 0.5s
+## Case step:
+##    1. Parse TPLG file to get pipeline
+##    2. Specify the audio parameters
+##    3. Run aplay or arecord on each pipeline with parameters
+##    4. do xrun injection during playback or capture
+## Expect result:
+##    The return value of aplay/arecord should be 0
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='aplay/arecord duration in second'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=10
+
+OPT_OPT_lst['i']='interval'     OPT_DESC_lst['i']='interval time of xrun injection'
+OPT_PARM_lst['i']=1         OPT_VALUE_lst['i']=0.5
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+func_opt_parse_option $*
+
+tplg=${OPT_VALUE_lst['t']}
+duration=${OPT_VALUE_lst['d']}
+interval=${OPT_VALUE_lst['i']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+func_lib_setup_kernel_last_line
+func_lib_check_sudo
+func_pipeline_export $tplg "type:playback,capture,both"
+
+declare -A CMD FILE
+CMD['playback,both']='aplay'
+FILE['playback,both']='/dev/zero'
+CMD['capture,both']='arecord'
+FILE['capture,both']='/dev/null'
+
+func_xrun_injection()
+{
+    count=1
+    while(true)
+    do
+        ps -ef |grep "$pid" |grep -v grep
+        if [ $? -eq 0 ]; then
+            dlogi "XRUN injection: $count"
+            sudo bash -c "'echo 1 > $xrun_injection'"
+            sleep $interval
+            let count++
+        else
+            break # aplay/arecord is finished, stop xrun injection
+        fi
+    done
+}
+
+func_test_pipeline_with_type()
+{
+    func_pipeline_export $tplg "type:$1"
+    for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+    do
+        channel=$(func_pipeline_parse_value $idx channel)
+        rate=$(func_pipeline_parse_value $idx rate)
+        fmt=$(func_pipeline_parse_value $idx fmt)
+        dev=$(func_pipeline_parse_value $idx dev)
+        pcm=$(func_pipeline_parse_value $idx pcm)
+        id=$(func_pipeline_parse_value $idx id)
+        pipeline_type=$(func_pipeline_parse_value $idx "type")
+        pcm=pcm${id}${2}
+        xrun_injection="/proc/asound/card0/$pcm/sub0/xrun_injection"
+
+        # check xrun injection file
+        [[ ! -e $xrun_injection ]] && dloge "XRUN DEBUG is not enabled in kernel, skip the test." && exit 2
+        dlogi "Testing: test xrun injection on PCM:$pcm,$pipeline_type. Interval time: $interval"
+        dlogc "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q
+        "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q &
+        pid=$!
+        # do xrun injection
+        dlogc "echo 1 > $xrun_injection"
+        func_xrun_injection
+        # check aplay/arecord return value
+        wait $pid
+        if [ $? != 0 ]; then
+            dloge "$pipeline_type on $pcm failed."
+            exit 1
+        fi
+    done
+}
+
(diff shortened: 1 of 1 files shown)


test-case/check_xrun_injection.sh
```diff
@@ -0,0 +1,129 @@
...
+func_lib_setup_kernel_last_line
+func_lib_check_sudo
+func_pipeline_export $tplg "type:playback,capture,both" 
+
+xrun_injection()
+{
+    count=1
+    while(true)
+    do
+    	ps -ef |grep "$pid"| grep -v grep
```
Bin-QA: 1.  `$pid` is parameter, here is not clear
2.  you can use `pidof` to confirm target command still in the process, you also can use sof-process-state do the more check

test-case/check_xrun_injection.sh
```diff
@@ -0,0 +1,129 @@
...
+    channel=$(func_pipeline_parse_value $idx channel)
+    rate=$(func_pipeline_parse_value $idx rate)
+    fmt=$(func_pipeline_parse_value $idx fmt)
+    dev=$(func_pipeline_parse_value $idx dev)
+    pcm=$(func_pipeline_parse_value $idx pcm)
+    id=$(func_pipeline_parse_value $idx id)
+    pipeline_type=$(func_pipeline_parse_value $idx "type")
+
+    # clean up dmesg
+    sudo dmesg -C
```
Bin-QA: why need `dmesg -c`, I not find you use dmesg again

keqiaozhang: At end of the script, it will use sof-kernel-log-check.sh to check the kernel log.

test-case/check_xrun_injection.sh
keqiaozhang: I think "pid=$!" is safer,  pidof will list all the pid num of cmdlines. Suppose there was more than one aplay process，the pid num will not be unique.
In this part, I only want to check if target pid is still running or not.  The return value will be checked by wait "$pid".

test-case/check_xrun_injection.sh
Bin-QA: sof-kernel-log-check.sh is check for the full kernel log after begin  
So if you want to check the step by step, you don't need to do this

test-case/check_xrun_injection.sh
```diff
@@ -0,0 +1,129 @@
...
+	else
+		break # aplay/arecord is finished, stop xrun injection
+    	fi
+    done
+}
+
+run_test()
+{
+    dlogc "$cmd -D$dev -r $rate -c $channel -f $fmt -d $duration $file -q"
+    $cmd -D$dev -r $rate -c $channel -f $fmt -d $duration $file -q &
```
aiChaoSONG: $file here seems always /dev/zero, is that ok for arecord?

test-case/check_xrun_injection.sh
keqiaozhang: Updated and thanks for your comment.

test-case/check_xrun_injection.sh
keqiaozhang: Updated. Good catch.

test-case/check_xrun_injection.sh
```diff
@@ -0,0 +1,156 @@
...
+        if [[ -z $capture_file ]]; then
+            dlogi "no capture file specified, use /dev/null as dummy capture output"
+            file=/dev/null
+        else
+            mkdir -p $output_dir
+            file=$output_dir/${capture_file}_${dev}.wav
+            dlogi "using $file as capture output"
+	fi
+    fi
+}
```
Bin-QA: 1.  line 112 can be move to the top, because this folder just need to create once
2.  after remove the line 112, you can find the capture & playback use the same logic

I guest use `/dev/zero` & `/dev/null` as the default value of `$playback_file` & `$capture_file`  
then here you just need to check target file whether exist, if not exist, you can direct exit as 2/1

from your code logic, I think this maybe help for your code more clear  
https://github.com/thesofproject/sof-test/blob/master/test-case/multiple-pipeline-capture.sh#L53  
https://github.com/thesofproject/sof-test/blob/master/test-case/multiple-pipeline-capture.sh#L84  

here use the associative array about bash, you need to `declare -A ` for your array

test-case/check_xrun_injection.sh
```diff
@@ -0,0 +1,156 @@
...
+{
+    count=1
+    while(true)
+    do
+        ps -ef |grep "$pid" |grep -v grep
+        if [ $? -eq 0 ]; then
+            dlogc "XRUN injection: $count"
+            sudo bash -c "'echo 1 > /proc/asound/card0/$pcm/sub0/xrun_injection'"
+            sleep $interval
+            let count++
```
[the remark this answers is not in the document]
keqiaozhang: Just a log. if case fails, it's easy to know where it failed.

test-case/check_xrun_injection.sh
keqiaozhang: Test script is updated. Good example.

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,117 @@
...
+            break # aplay/arecord is finished, stop xrun injection
+        fi
+    done
+}
+
+check_audio_file()
+{
+    if [[ ! -f ${FILE[$1]} ]]; then
+        dloge "${FILE[$1]} does not exist, skip the test."
+	return 1
```
Bin-QA: here have the function return

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,117 @@
...
+        fmt=$(func_pipeline_parse_value $idx fmt)
+        dev=$(func_pipeline_parse_value $idx dev)
+        pcm=$(func_pipeline_parse_value $idx pcm)
+        id=$(func_pipeline_parse_value $idx id)
+        pipeline_type=$(func_pipeline_parse_value $idx "type")
+        pcm=pcm${id}${2}
+
+        dlogi "Testing: test xrun injection on PCM:$pcm,$pipeline_type. Interval time: $interval"
+        check_audio_file
+        dlogc "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q
```
Bin-QA: No function return check

test-case/check-xrun-injection.sh
keqiaozhang: Let's make it in an easy way. Do not accept user specified audio files, but use /dev/zero and /dev/null as default.

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,102 @@
...
+FILE['capture,both']='/dev/null'
+
+func_xrun_injection()
+{
+    count=1
+    while(true)
+    do
+        ps -ef |grep "$pid" |grep -v grep
+        if [ $? -eq 0 ]; then
+            dlogc "XRUN injection: $count"
```
xiulipan: `dlogc` is used for cmd.
For this it should be `dlogi`

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,102 @@
...
+        func_xrun_injection
+        # check aplay/arecord return value
+        wait $pid
+        if [ $? != 0 ]; then
+            dloge "$pipeline_type on $pcm failed."
+            exit 1
+        fi
+    done
+}
+
```
xiulipan: We need a check for the xrun injection feature check. It is not default kconfig for most kernel.

keqiaozhang: Added xrun injection file check.

test-case/check-xrun-injection.sh
keqiaozhang: Solved.

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,106 @@
...
+        pipeline_type=$(func_pipeline_parse_value $idx "type")
+        pcm=pcm${id}${2}
+        xrun_injection="/proc/asound/card0/$pcm/sub0/xrun_injection"
+
+        dlogi "Testing: test xrun injection on PCM:$pcm,$pipeline_type. Interval time: $interval"
+        dlogc "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q
+        "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q &
+        pid=$!
+        # check xrun injection file
+        [[ ! -e $xrun_injection ]] && dloge "XRUN DEBUG is not enabled in kernel, skip the test." && return 2
```
xiulipan: @keqiaozhang, I am not sure if this cmd will work without root.
PS I am not sure if you should check the xrun injection after you start some aplay/arocred.
I would suggest to check this before we start the test.

keqiaozhang: Moved the xrun injection file check before aplay/arecord.
I confirmed that it works without root.

test-case/check-xrun-injection.sh
```diff
@@ -0,0 +1,106 @@
...
+        fmt=$(func_pipeline_parse_value $idx fmt)
+        dev=$(func_pipeline_parse_value $idx dev)
+        pcm=$(func_pipeline_parse_value $idx pcm)
+        id=$(func_pipeline_parse_value $idx id)
+        pipeline_type=$(func_pipeline_parse_value $idx "type")
+        pcm=pcm${id}${2}
+        xrun_injection="/proc/asound/card0/$pcm/sub0/xrun_injection"
+
+        # check xrun injection file
+        [[ ! -e $xrun_injection ]] && dloge "XRUN DEBUG is not enabled in kernel, skip the test." && return 2
```
xiulipan:
```suggestion
        [[ ! -e $xrun_injection ]] && dloge "XRUN DEBUG is not enabled in kernel, skip the test." && exit 2
```
We should exit the whole test here if we could not find xurn_injection debugfs.
PS: we do not check return value for this function.

keqiaozhang: I always confuse return with exit. Fixed.

PR #18, opened 2019-12-31
Bug fix: verify-pcm-list typo error

Bin-QA:
when find $tplg_file in current path should direct to use
the $f is not define so it will be empty value
which will cause the error:
when find $tplg_file but not use it, because tplg_file will
be overwrite with empty value


Bin-QA:
close the PR for modify the TPLG load process

PR #20, opened 2020-01-03
test: add stop/start pipeline test

keqiaozhang:
This script intends to test stop/start aplay/arecord on each pipeline.

### DIFF
--- a/test-case/check-signal-stop-start.sh
+++ b/test-case/check-signal-stop-start.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+
+##
+## Case Name: check-signal-stop-start
+## Preconditions:
+##    N/A
+## Description:
+##    Run aplay/arecord on each pipeline and use SIGSTOP and SIGCONT to
+##    simulate keyboard inputs ctrl+z and fg.
+##    kill -SIGSTOP $pid will stop the thread
+##    kill -SIGCONT $pid will start the thread
+## Case step:
+##    1. aplay/arecord on PCM
+##    2. send SIGSTOP/SIGCONT to stop/start
+## Expect result:
+##    no errors for aplay/arecord
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+OPT_OPT_lst['m']='mode'     OPT_DESC_lst['m']='test mode'
+OPT_PARM_lst['m']=1         OPT_VALUE_lst['m']='playback'
+
+OPT_OPT_lst['i']='sleep'    OPT_DESC_lst['i']='interval time for stop/start'
+OPT_PARM_lst['i']=1         OPT_VALUE_lst['i']=0.5
+
+OPT_OPT_lst['c']='count'    OPT_DESC_lst['c']='test count of stop/start'
+OPT_PARM_lst['c']=1         OPT_VALUE_lst['c']=10
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+func_opt_parse_option $*
+
+tplg=${OPT_VALUE_lst['t']}
+count=${OPT_VALUE_lst['c']}
+interval=${OPT_VALUE_lst['i']}
+test_mode=${OPT_VALUE_lst['m']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+case $test_mode in
+    "playback")
+        cmd=aplay
+        dummy_file=/dev/zero
+    ;;
+    "capture")
+        cmd=arecord
+        dummy_file=/dev/null
+    ;;
+    *)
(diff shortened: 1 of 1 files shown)


xiulipan:
@keqiaozhang can you change the test case name with signal stop start? I think stop start is very confusing.

keqiaozhang:
Changed the case name to "check-signal-stop-start.sh".

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
+##
+## Case Name: check-stop-start
```
xiulipan: case name also need update

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
...
+FILE['playback,both']='/dev/zero'
+CMD['capture,both']='arecord'
+FILE['capture,both']='/dev/null'
+
+func_stop_start_pipeline()
+{
+    count=1
+    while(true)
+    do
+        ps -ef |grep "$pid" |grep -v grep
```
xiulipan: can we use some other check logic? like `kill -0 $pid`?
@Bin-QA any suggestion?

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
...
+            dlogi "Stop/start count: $count"
+            # stop the pipeline
+            kill -SIGSTOP $pid
+            sleep $interval
+            # start the pipeline
+            kill -SIGCONT $pid
+            sleep $interval
+            let count++
+        else
+        break # aplay/arecord is finished, terminate stop/start test
```
xiulipan: intend here.

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
...
+
+declare -A CMD FILE
+CMD['playback,both']='aplay'
+FILE['playback,both']='/dev/zero'
+CMD['capture,both']='arecord'
+FILE['capture,both']='/dev/null'
+
+func_stop_start_pipeline()
+{
+    count=1
```
xiulipan: can we make this `local`

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
...
+        pcm=$(func_pipeline_parse_value $idx pcm)
+        pipeline_type=$(func_pipeline_parse_value $idx "type")
+
+        dlogi "Testing: run stop/start test on PCM:$pcm,$pipeline_type. Interval time: $interval"
+        dlogc "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q &
+        "${CMD[$1]}" -D$dev -r $rate -c $channel -f $fmt -d $duration "${FILE[$1]}" -q & pid=$!
+        # do stop/start test
+        func_stop_start_pipeline
+        # check aplay/arecord return value
+        wait $pid
```
xiulipan: @keqiaozhang @Bin-QA some question here.
We will only exit the `func_stop_start_pipeline` when aplay/arecord exit, can we wait for it after the exit?

Bin-QA: I think the function loop can direct load form here look like this, here `wait` use to query process exit code,.
```
while [ "$(ps -p $pid)"]
do
            dlogi "Stop/start count: $count"
            # stop the pipeline
            kill -SIGSTOP $pid
            sleep $interval
            # start the pipeline
            kill -SIGCONT $pid
            sleep $interval
            let count++
done
wait $pid
```

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,99 @@
...
+            dloge "$pipeline_type on $pcm failed."
+            exit 1
+        fi
+    done
+}
+
+func_test_pipeline_with_type "playback,both"
+func_test_pipeline_with_type "capture,both"
+
+sof-kernel-log-check.sh $KERNEL_LAST_LINE
```
Bin-QA: @keqiaozhang, you need to load this function `func_lib_setup_kernel_last_line` to get `$KERNEL_LAST_LINE` value

test-case/check-signal-stop-start.sh
keqiaozhang: solved.

test-case/check-signal-stop-start.sh
keqiaozhang: @Bin-QA  good suggestion. Updated.

test-case/check-signal-stop-start.sh
keqiaozhang: Updated.

xiulipan:
@keqiaozhang one concern here, it seems we always allow user to select playback/capture for the test. Would it keep in that style or merge the playback and capture inside the test case?

test-case/check-signal-stop-start.sh
```diff
@@ -0,0 +1,103 @@
...
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='duration time'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=10
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+func_opt_parse_option $*
+
+tplg=${OPT_VALUE_lst['t']}
+count=${OPT_VALUE_lst['c']}
```
wenqingfu: 'c' is not defined. There is no use of it, either.  Please enable the support of $count number of start/stop.

In the meantime, do you need to specify duration in this test? can you let it run indefinitely and kill aplay/arecord once you reached the $count?

PR #21, opened 2020-01-03
kmod_scripts: add support for snd_soc_cml_rt1011_rt5682

aiChaoSONG:
Add snd_soc_cml_rt1011_rt5682 module removal support for CML with rt1011 and rt5682 codecs.


plbossart:
Can I ask why you are duplicating the files already supported in sof/tools/kmod_scripts?

This is not good practice. Either we remove the code from sof/tools/kmod_scripts and use this one, or you remove this one.

xiulipan:
@plbossart to proposal was to remove the test scripts from sof repo. Will try to send the PR to remove test scripts after we got all test scripts migration done

PR #22, opened 2020-01-03
Test Case: Add check suspend resume test

Bin-QA:
Add new test Case for the suspend resume

### DIFF
--- a/test-case/check-suspend-resume.sh
+++ b/test-case/check-suspend-resume.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+##
+## Case Name: check suspend/resume status
+## Preconditions:
+##    N/A
+## Description:
+##    Run the suspend/resume command to check device status
+## Case step:
+##    1. switch suspend/resume operation
+##    2. use rtcwake -m mem command to do suspend/resume
+##    3. check command return value
+##    4. check dmesg errors
+##    5. check wakeup increase
+## Expect result:
+##    suspend/resume recover
+##    check kernel log and find no errors
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+random_min=3    # wait time should >= 3 for other device wakeup from sleep
+random_max=20
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=5
+
+OPT_OPT_lst['t']='type'    OPT_DESC_lst['t']="suspend/resume type from /sys/power/mem_sleep"
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']=""
+
+OPT_OPT_lst['s']='sleep'    OPT_DESC_lst['s']='suspend/resume command:rtcwake sleep duration'
+OPT_PARM_lst['s']=1         OPT_VALUE_lst['s']=5
+
+OPT_OPT_lst['w']='wait'     OPT_DESC_lst['w']='idle time after suspend/resume wakeup'
+OPT_PARM_lst['w']=1         OPT_VALUE_lst['w']=5
+
+OPT_OPT_lst['r']='random'   OPT_DESC_lst['r']="Randomly setup wait/sleep time, range is [$random_min-$random_max], this option will overwrite s & w option"
+OPT_PARM_lst['r']=0         OPT_VALUE_lst['r']=0
+
+func_opt_parse_option $*
+func_lib_check_sudo
+func_lib_setup_kernel_last_line
+
+type=${OPT_VALUE_lst['t']}
+# switch type
+if [ "$type" ]; then
+    # check for type value effect
+    [[ ! "$(cat /sys/power/mem_sleep|grep $type)" ]] && dloge "useless type option" && exit 2
+    dlogc "sudo bash -c 'echo $type > /sys/power/mem_sleep'"
+    sudo bash -c "'echo $type > /sys/power/mem_sleep'"
+fi
+dlogi "Current suspend/resume type mode: $(cat /sys/power/mem_sleep)"
+
+loop_count=${OPT_VALUE_lst['l']}
+declare -a sleep_lst wait_lst
+
+if [ ${OPT_VALUE_lst['r']} -eq 1 ]; then
+    # create random number list
+    for i in $(seq 1 $loop_count)
+    do
+        sleep_lst[$i]=$(func_lib_get_random $random_max $random_min)
+        wait_lst[$i]=$(func_lib_get_random $random_max $random_min)
+    done
+else
+    for i in $(seq 1 $loop_count)
+    do
+        sleep_lst[$i]=${OPT_VALUE_lst['s']}
+        wait_lst[$i]=${OPT_VALUE_lst['w']}
+    done
+fi
+
+for i in $(seq 1 $loop_count)
+do
+    dlogi "Round($i/$loop_count)"
+    # cleanup dmesg befor run case
+    sudo dmesg --clear
+    sleep_count=$(cat /sys/power/wakeup_count)
+    dlogc "Run the command: rtcwake -m mem -s ${sleep_lst[$i]}"
+    sudo rtcwake -m mem -s ${sleep_lst[$i]}
+    [[ $? -ne 0 ]] && dloge "rtcwake return value error" && exit 1
+    dlogc "sleep for ${wait_lst[$i]}"
+    sleep ${wait_lst[$i]}
+    dlogi "Check for the kernel log status"
+    wake_count=$(cat /sys/power/wakeup_count)
+    # sof-kernel-log-check script parameter number is 0/Non-Number will force check from dmesg
+    sof-kernel-log-check.sh 0
+    [[ $? -ne 0 ]] && dloge "Catch dmesg error" && exit 1
+    # check wakeup count correct
+    [[ $wake_count -le $sleep_count ]] && dloge "suspend/resume didn't happen, because /sys/power/wakeup_count does not increase" && exit 1
+done
+
+# check full log
+sof-kernel-log-check.sh $KERNEL_LAST_LINE
+exit $?


Bin-QA:
@libinyang @keqiaozhang @xiulipan @aiChaoSONG 
please help to verify this case

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+##
+## Case Name: check S3 status
```
xiulipan: please make the name into `check-suspend-resume`
Not all platform are using S3 now.

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
...
+## Preconditions:
+##    N/A
+## Description:
+##    Run the S3 command to check device status
```
xiulipan: suspend/resume

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
...
+## Case step:
+##    1. switch S3 operation
+##    2. use rtcwake -m mem command to do S3
+##    3. check command return value
+##    4. check dmesg errors
+##    5. check wakeup increase
+## Expect result:
+##    S3 recover
```
xiulipan: suspend/resume

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
...
+
+##
+## Case Name: check S3 status
+## Preconditions:
+##    N/A
+## Description:
+##    Run the S3 command to check device status
+## Case step:
+##    1. switch S3 operation
+##    2. use rtcwake -m mem command to do S3
```
xiulipan: `to do suspend/resume`

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
...
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+# wait time should >= 3 for other device wakeup from sleep
+random_min=3
+random_max=20
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=5
+
+OPT_OPT_lst['p']='power'    OPT_DESC_lst['p']="S3 power type from /sys/power/mem_sleep"
```
xiulipan: s2idle is used for S0ix in Linux.
For s3 it is using` s2idle [deep]`

test-case/check-s3.sh
```diff
@@ -0,0 +1,93 @@
...
+dlogi "Current power mode: $(cat /sys/power/mem_sleep)"
+# check for power value effect
+[[ ! "$(cat /sys/power/mem_sleep|grep $power)" ]] && dloge "useless power option" && exit 2
+dlogc "sudo bash -c 'echo $power > /sys/power/mem_sleep'"
+sudo bash -c "'echo $power > /sys/power/mem_sleep'"
+
+loop_count=${OPT_VALUE_lst['l']}
+declare -a sleep_lst wait_lst
+
+if [ ${OPT_VALUE_lst['r']} -eq 1 ]; then
```
xiulipan: can we make the random number generator into a help function.

Bin-QA: OK, I will add this help function

test-case/check-s3.sh
Bin-QA: so make the deep as the default option?

xiulipan: @Bin-QA My suggestion is to keep the default value and output them.
@aiChaoSONG will those GLK have default s2idle setting?

case-lib/lib.sh
```diff
@@ -85,3 +85,20 @@ func_lib_restore_pulseaudio()
...
+        random_min=$1
+        random_scope=$(expr $random_max - $random_min)
+        expr $RANDOM % $random_scope + $random_min
+    elif [ $# -eq 1 ];then
+        random_max=$1
+        expr $RANDOM % $random_max
+    else
+        echo $RANDOM
+    fi
+}
```
xiulipan: Please make this change into a single commit.

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,96 @@
...
+        wait_lst[$i]=${OPT_VALUE_lst['w']}
+    done
+fi
+
+for i in $(seq 1 $loop_count)
+do
+    dlogi "Round($i/$loop_count)"
+    # cleanup dmesg befor run case
+    sudo dmesg --clear
+    bcount=$(cat /sys/power/wakeup_count)
```
xiulipan: can we use wake_count and sleep_count to replace acount and bcount

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,96 @@
...
+
+random_min=3    # wait time should >= 3 for other device wakeup from sleep
+random_max=20
+def_power=$(sed 's/ /\n/g' /sys/power/mem_sleep|grep '\['|sed 's/\[//g;s/\]//g')
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=5
+
+OPT_OPT_lst['p']='power'    OPT_DESC_lst['p']="suspend/resume type from /sys/power/mem_sleep"
+OPT_PARM_lst['p']=1         OPT_VALUE_lst['p']="$def_power"
```
xiulipan: can we make this default as empty string? And then only overwrite the setting when there are some values here.

test-case/check-s3.sh
keqiaozhang: @xiulipan , Due to the coreboot issue, S3 is not supported on GLK, we need to bypass S3 test on GLK.

xiulipan: @keqiaozhang one thing to confirm, will GLK setting default to be `s2idle `or `deep`?

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,94 @@
...
+OPT_OPT_lst['r']='random'   OPT_DESC_lst['r']="Random setup wait/sleep time, time range is [$random_min-$random_max], this option will overwait s & w option"
+OPT_PARM_lst['r']=0         OPT_VALUE_lst['r']=0
+
+func_opt_parse_option $*
+func_lib_check_sudo
+func_lib_setup_kernel_last_line
+
+power=${OPT_VALUE_lst['p']}
+# switch type
+if [ "$power" ]; then
```
libinyang: If user doesn't set the power mode, the susprend/resume mode will be random. Is this what we expected?

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,94 @@
...
+    [[ $? -ne 0 ]] && dloge "rtcwake return value error" && exit 1
+    dlogc "sleep for ${wait_lst[$i]}"
+    sleep ${wait_lst[$i]}
+    dlogi "Check for the kernel log status"
+    wake_count=$(cat /sys/power/wakeup_count)
+    # sof-kernel-log-check script parameter number is 0/Non-Number will force check from dmesg
+    sof-kernel-log-check.sh 0
+    [[ $? -ne 0 ]] && dloge "Catch dmesg error" && exit 1
+    # check wakeup count correct
+    [[ $wake_count -le $sleep_count ]] && dloge "Fake suspend/resume, because /sys/power/wakeup_count is not increase" && exit 1
```
libinyang: "is not increase" should be "does not increase"

test-case/check-s3.sh
keqiaozhang: Default value is deep on GLK.

test-case/check-suspend-resume.sh
Bin-QA: it is not the random, just use the system current setting for it  
This value will be dump by the test-case to run  
Current I am wait for confirm the default value of `power`

libinyang: Yes, this is what I mean. It depends on the current setting. But the current setting maybe s2idle or deep or standby. If we have a default value for power mode, it can fix this issue.

xiulipan: @libinyang @Bin-QA It would not be random. The system will always have default value. We should test with the system default value.
We may get into some trouble if we try to test in other modes.

libinyang: But the system's default value will be different on different platforms based on my understanding. So without appointing the mode, we can't tell we are doing S3 or s2idle for the case. What's more, someone may temporarily uses the DUT to test suspend/resume in another mode.

xiulipan:
@Bin-QA @keqiaozhang @libinyang So here I think we will keep the behavior to be
- use default mode to do test if no argument is passed.
- override the mode with the argument

keqiaozhang:
@xiulipan Agree, use deep as default and override it if it's specified.

test-case/check-suspend-resume.sh
wenqingfu: Fake suspend/resume, because /sys/power/wakeup_count is not increase -> suspend/resume didn't happen, because /sys/power/wakeup_count does not increase

test-case/check-suspend-resume.sh
wenqingfu: rename this parameter from 'power' to 'type'

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,94 @@
...
+random_min=3    # wait time should >= 3 for other device wakeup from sleep
+random_max=20
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=5
+
+OPT_OPT_lst['p']='power'    OPT_DESC_lst['p']="suspend/resume type from /sys/power/mem_sleep"
+OPT_PARM_lst['p']=1         OPT_VALUE_lst['p']=""
+
+OPT_OPT_lst['s']='sleep'    OPT_DESC_lst['s']='suspend/resume command:rtcwake sleep option'
```
wenqingfu: rtcwake sleep option  -> rtcwake sleep duration

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,94 @@
...
+OPT_PARM_lst['s']=1         OPT_VALUE_lst['s']=5
+
+OPT_OPT_lst['w']='wait'     OPT_DESC_lst['w']='wait time after suspend/resume wakeup'
```
wenqingfu: wait time -> idle time

test-case/check-suspend-resume.sh
```diff
@@ -0,0 +1,94 @@
...
+OPT_PARM_lst['w']=1         OPT_VALUE_lst['w']=5
+
+OPT_OPT_lst['r']='random'   OPT_DESC_lst['r']="Random setup wait/sleep time, time range is [$random_min-$random_max], this option will overwait s & w option"
```
wenqingfu: Random setup -> Randomly setup
time range -> range
overwait -> overwrite

stripes416:
I'm currently running the manual suspend-resume tests on ICL --- this looks good to me & I can't think of anything to add to the code. Good job @Bin-QA

PR #23, opened 2020-01-03
test-case: fix FW path for verify-firmware-presence.sh

aiChaoSONG:
Since linux 5.5.0-rc2, FW signed with comminuty key is moved to '/lib/firmware/intel/sof/community'.
fix path for verify-firmware-presence.sh

### DIFF
--- a/test-case/verify-firmware-presence.sh
+++ b/test-case/verify-firmware-presence.sh
@@ -18,7 +18,7 @@ source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
 
 func_opt_parse_option $*
 
-path="/lib/firmware/intel/sof"
+path=$(sof-dump-status.py -P)
 platform=$(sof-dump-status.py -p)
 fw="$path/sof-$platform.ri"
 dlogi "Checking file: $fw"
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -299,6 +299,7 @@ if __name__ == "__main__":
     parser.add_argument('-i', '--id', type=int, help='just dump the pcm infomation of target id sound card')
     parser.add_argument('-s', '--short', type=int, help='just dump the short name of target id sound card')
     parser.add_argument('-l', '--longname', type=int, help='just dump the longname name of target id sound card')
+    parser.add_argument('-P', '--fwpath', action='store_true', help='get firmware path according to DMI info')
     parser.add_argument('--version', action='version', version='%(prog)s 1.0')
 
     ret_args = vars(parser.parse_args())
@@ -338,6 +339,56 @@ if __name__ == "__main__":
         if card_info is None:
             exit(0)
         print(card_info['longname'])
+        exit(0)
+
+    # The kernel has changed the default firmware path when community
+    # key is used. Here we output firmware path according to kernel's
+    # match table.
+    if ret_args.get('fwpath') is True:
+        def is_community_board(board, community_boards):
+            for elem in community_boards:
+                if match_board(elem, board):
+                    return True
+            return False
+
+        # see if 'board' matches 'match', only match the key defined
+        # in community boards' matches field.
+        def match_board(match, board):
+            for key in match["matches"].keys():
+                if match["matches"][key] != board["matches"][key]:
+                    return False
+            return True
+
+        # The "community_boards" structure here is in accordance
+        # with "struct dmi_system_id community_key_platforms" in
+        # "sound/soc/sof/sof-pci-dev.c".
+        community_boards = [
+            {
+                "ident": "Up Squared",
+                "matches": {
+                    "board_name": "UP-APL01",
+                    "board_vendor": "AAEON"
+                }
+            },
+            {
+                "ident": "Chromebook",
+                "matches":  {
+                    "board_vendor": "Google",
+                }
+            }
+        ]
+        fw_path = "/lib/firmware/intel/sof"
+        sysinfo.loadDMI()
+        board = {
+                    "ident": sysinfo.dmi["board_name"],
+                    "matches": {
+                        "board_name": sysinfo.dmi["board_name"],
+                        "board_vendor": sysinfo.dmi["board_vendor"]
+                    }
+                }
+        if is_community_board(board, community_boards):
+            fw_path += "/community"
+        print(fw_path)
         exit(0)
 
     sysinfo.loadDMI()


Bin-QA:
@xiulipan 
how about `find /lib/firmware/intel/sof -name "sof-$platform.ri"`

test-case/verify-firmware-presence.sh
```diff
@@ -18,7 +18,7 @@ source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
 
 func_opt_parse_option $*
 
-path="/lib/firmware/intel/sof"
+path="/lib/firmware/intel/sof/community"
```
xiulipan: Not all platform is using this path, please update this part to change the path with DMI info to follow the kernel change.

aiChaoSONG:
@xiulipan @Bin-QA This solution is from Xiuli, I think this is the perfect solution to this issue. please review again.

tools/sof-dump-status.py
```diff
@@ -339,6 +340,17 @@ def _covertType(type_lst):
             exit(0)
         print(card_info['longname'])
         exit(0)
+    if ret_args.get('community') is True:
+        vender_list = ["AAEON", "Google"]
+        board_list = ["UP-APL01"]
+
+        sysinfo.loadDMI()
+        board = sysinfo.dmi["board_name"]
+        vendor = sysinfo.dim["board_vendor"]
```
Bin-QA: I suggest convert all string to lower/upper to compare to avoid some mistake

aiChaoSONG: I don't think this is needed, these string used here is from Linux Kernel, and the kernel will use the string from BIOS, all 3 places has the same form. btw, there is no case transform in the kernel for this purpose.  the string got by sof-dump-status.py is always the same as the kernel.

test-case/verify-firmware-presence.sh
```diff
@@ -20,7 +20,13 @@ func_opt_parse_option $*
 
 path="/lib/firmware/intel/sof"
 platform=$(sof-dump-status.py -p)
-fw="$path/sof-$platform.ri"
+is_community=$(sof-dump-status.py -C)
```
xiulipan: @aiChaoSONG  can you just make this tools to output default firmware path?

tools/sof-dump-status.py
```diff
@@ -299,6 +299,7 @@ def _covertType(type_lst):
     parser.add_argument('-i', '--id', type=int, help='just dump the pcm infomation of target id sound card')
     parser.add_argument('-s', '--short', type=int, help='just dump the short name of target id sound card')
     parser.add_argument('-l', '--longname', type=int, help='just dump the longname name of target id sound card')
+    parser.add_argument('-C', '--community', action='store_true', help='check if fw is signed with community key')
```
xiulipan: change into output firmware path.

tools/sof-dump-status.py
xiulipan: Please add some inline comments here to the file to help others keep this part synced with kernel.

aiChaoSONG:
@xiulipan @Bin-QA @keqiaozhang Please help to review again, thanks

tools/sof-dump-status.py
```diff
@@ -340,6 +341,31 @@ def _covertType(type_lst):
         print(card_info['longname'])
         exit(0)
 
+    # The kernel has changed the default firmware path when community
+    # key is used. Here we output firmware path according to kernel's
+    # match table.
+    if ret_args.get('fwpath') is True:
+
```
Bin-QA: empty line?

tools/sof-dump-status.py
```diff
@@ -340,6 +341,31 @@ def _covertType(type_lst):
...
+            else:
+                return ""
+
+        fw_path = "/lib/firmware/intel/sof"
+        sysinfo.loadDMI()
+        board = sysinfo.dmi["board_name"]
+        vendor = sysinfo.dmi["board_vendor"]
+        cpath = community_path(sysinfo.dmi, board, vendor)
+        if cpath != "":
+            fw_path = cpath
```
Bin-QA: why need the function?

aiChaoSONG: Xiuli said we may have more folders in the future, eg, production,  with a function, the program structure will be clearer

tools/sof-dump-status.py
```diff
@@ -340,6 +341,20 @@ def _covertType(type_lst):
...
+    # The kernel has changed the default firmware path when community
+    # key is used. Here we output firmware path according to kernel's
+    # match table.
+    if ret_args.get('fwpath') is True:
+        fw_path = "/lib/firmware/intel/sof"
+        sysinfo.loadDMI()
+        community_vendor_list = ["AAEON", "Google"]
+        community_board_list = ["UP-APL01"]
+        if sysinfo.dmi["board_name"] in community_board_list or \
+            sysinfo.dmi["board_vendor"] in community_vendor_list:
```
xiulipan: @aiChaoSONG I think this logic is different with what we got from the https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c

AAEON will have many boards, but only "UP-APL01" is using community fw.

aiChaoSONG:
@xiulipan Now change the match mechanism as the kernel did, and the code is extensible. if we need more keys and boards to match in the future, only need to add the key and board to board list. no need to change match mechanism.

tools/sof-dump-status.py
```diff
@@ -340,6 +341,55 @@ def _covertType(type_lst):
...
+                if match["match"][key] == "":
+                    continue
+                else:
+                    if match["match"][key] != board["match"][key]:
+                        return False
+            return True
+
+        community_boards = [
+            {
+                "name": "Up Squred",
```
stripes416: @aiChaoSONG I think you mean Up Squared here.

aiChaoSONG: @stripes416 thanks, fixed the typo.

tools/sof-dump-status.py
```diff
@@ -340,6 +341,55 @@ def _covertType(type_lst):
...
+                "name": "Up Squared",
+                "match": {
+                    "board_name": "UP-APL01",
+                    "board_vendor": "AAEON"
+                }
+            },
+            {
+                "name": "Chromebook",
+                "match":  {
+                    "board_name": "",
```
xiulipan: @aiChaoSONG Can we keep this filled not defined here? Then in the above check can we handle that there is no define in structure? This can save us some time if we need to add more DMI info here.

tools/sof-dump-status.py
```diff
@@ -340,6 +341,55 @@ def _covertType(type_lst):
...
+        def match_board(match, board):
+            for key in match["match"].keys():
+                if match["match"][key] == "":
+                    continue
+                else:
+                    if match["match"][key] != board["match"][key]:
+                        return False
+            return True
+
+        community_boards = [
```
xiulipan: Please add inline comments to show this table should sync with sof-pci-dev.c

aiChaoSONG: @xiulipan added comments to tell people the community_boards structure is from "sound/soc/sof/sof-pci-dev.c".

tools/sof-dump-status.py
aiChaoSONG: @xiulipan  "board_name" field removed for Chromebook, for it make no sense on classify boards type. we only use the identification fields to classify a board, eg: For google, "vendor_name", and for Up2: "board_name" and "vendor_name".

Thank you for showing me ideas on implementation of such a little feature, illuminating me a lot.

aiChaoSONG:
@xiulipan @Bin-QA @keqiaozhang Please help to review, I believe this is the last review of this pr

tools/sof-dump-status.py
```diff
@@ -340,6 +341,56 @@ def _covertType(type_lst):
...
+                if match["match"][key] != board["match"][key]:
+                    return False
+            return True
+
+        # The "community_boards" structure here is in accordance
+        # with "struct dmi_system_id community_key_platforms" in
+        # "sound/soc/sof/sof-pci-dev.c".
+        community_boards = [
+            {
+                "name": "Up Squared",
```
Bin-QA: why need `name`? I can understand here struct same with [community_key_platforms](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c#L40)  
But here `name` is not to use, so don't need the `name` filed

aiChaoSONG: Xiuli said we should give a name to every board, this name is for us to know which board it is.

Bin-QA: How about convert `name` to the comment

libinyang: I'm always having concern on the match_board(). Using this method, we will have to continue updating this file when the board list has updated. Or the board list won't be updated any more?

aiChaoSONG: @libinyang We will update board list according to sound/soc/sof/sof-pci-dev.c in the kernel. because we don't know how the kernel code change. maybe sometimes later, the kernel will add a field for matching, or add some other board, we should do the same change, too.

aiChaoSONG: @Bin-QA the kernel use "ident" , which means identity, how about this name?

Bin-QA: the really reason is this field is not in use, so here you struct is 
```
list[
    dict{
        # identify
        dict{
        # broad
        # vendor
        }
    }
]
```
But I think this struct is more clear & easy to query
```
dict{
    # broad 
    list [
        # vendor
    ]
}
```
python dict.get() can direct get the list

aiChaoSONG: with the first structure, we eliminate the "name" field, no need to use the second embedded dict. the second is not OK, because for some board, we need to match 2 fields, and other may be 1 fields. with a list, we need to keep field in it's place(index), but the list length is not fixed, this is error-prone.
The "name" field is not used for now, but later we may want to output which board is community board. @Bin-QA

PR #25, opened 2020-01-06
Feature add: Add random function in case-lib/lib.sh

Bin-QA:
Add func_lib_get_random go get random value
$# == 2: $1 is max, $2 is min
so the random value will in [ min - max ]
$# == 1: $1 is max
so the random value will in [ 0 - max ]
Other:
get $RANDOM which define by bash, this value will in [ 0 - 32767 ]


Bin-QA:
#22 Base on this patch to load the Random number

PR #26, opened 2020-01-07
tools: add multiple formats output for sof-tplgreader.py

aiChaoSONG:
Currenty, we only use the first supported format to run playback and capture test, this limits the test capacity. This patch add multiple formats output in sof-tplgreader.py, thus, enable multiple format support for test cases

I keep the 'fmt' in pipeline_dict, so, there is no influence on exist cases.  for cases want to use multiple format, use 'fmts'.
Example:
with `fmt = $(func_pipeline_parse_value $idx fmt)`, you got "S16_LE"
with  `fmts = $(func_pipeline_parse_value $idx fmts)`, you got "S16LE S24LE S32LE"

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -3,9 +3,9 @@
 import subprocess
 import os
 import re
-from tplgtool import TplgParser
+from tplgtool import TplgParser, TplgFormatter
 
-class clsTPLGeader:
+class clsTPLGReader:
     def __init__(self):
         self._pipeline_lst = []
         self._output_lst = []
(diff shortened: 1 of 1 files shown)


tools/sof-tplgreader.py
```diff
@@ -16,43 +16,40 @@ def __init__(self):
...
-                    pipeline_dict['type'] = 'playback'
-                elif pcm['capture'] == 1:
-                    pipeline_dict['type'] = 'capture'
-                else: # error pcm ??
-                    return 1
+                pipeline_dict['type'] = tplg_formatter._get_pcm_type(pcm)
                 cap = pcm["caps"][pcm['capture']]
-                pipeline_dict['fmt'] = self._fmt2str(cap)
+                formats = tplg_formatter._get_pcm_fmt(pcm)
+                pipeline_dict['fmts'] = self._fmt2str(formats, pipeline_dict['type'])
```
Bin-QA: this keyword will be used for many way from [func_pipeline_parse_value](https://github.com/thesofproject/sof-test/blob/master/case-lib/pipeline.sh#L71)  
So when test case load [**fmt**](https://github.com/thesofproject/sof-test/blob/master/test-case/check-capture.sh#L63) it will be apply

-  add new keywords
-  change refer test-case for it

OR you direct apply it, change test-case for it

tools/sof-tplgreader.py
```diff
@@ -16,43 +16,40 @@ def __init__(self):
...
-                elif pcm['playback'] == 1:
-                    pipeline_dict['type'] = 'playback'
-                elif pcm['capture'] == 1:
-                    pipeline_dict['type'] = 'capture'
-                else: # error pcm ??
-                    return 1
+                pipeline_dict['type'] = tplg_formatter._get_pcm_type(pcm)
                 cap = pcm["caps"][pcm['capture']]
-                pipeline_dict['fmt'] = self._fmt2str(cap)
+                formats = tplg_formatter._get_pcm_fmt(pcm)
```
Bin-QA: you already get the [string list](https://github.com/thesofproject/sof-test/blob/master/tools/tplgtool.py#L550)

why not direct use `formats = ",".join(tplg_formatter._get_pcm_fmt(pcm)` to covert to string

tools/sof-tplgreader.py
```diff
@@ -16,43 +16,40 @@ def __init__(self):
...
                 pipeline_dict['id'] = str(pcm["pcm_id"])
-                if pcm['playback'] == 1 and pcm['capture'] == 1:
-                    pipeline_dict['type'] = 'both'
-                elif pcm['playback'] == 1:
-                    pipeline_dict['type'] = 'playback'
-                elif pcm['capture'] == 1:
-                    pipeline_dict['type'] = 'capture'
-                else: # error pcm ??
-                    return 1
+                pipeline_dict['type'] = tplg_formatter._get_pcm_type(pcm)
```
Bin-QA: remove code without handle 'none', I split this code from TplgFormatter,  
because I think the none is ok for tplgreader, but for the test-case without process to handle the 'none'  
so here will catch the function return

tools/sof-tplgreader.py
```diff
@@ -16,43 +16,40 @@ def __init__(self):
...
-            return "S24_LE"
-        elif fmt & (1 << 10) != 0:
-            return "S32_LE"
-        elif fmt & (1 << 14) != 0:
-            return "FLOAT"
-        else: # without match
-            return ""
+    # supported formats of playback pipeline in formats[0]
+    # supported formats of capture pipeline in formats[1]
+    def _fmt2str(self, formats, pcm_type):
```
Bin-QA: for this tools, don't need get full list, so you can direct to access target list

xiulipan:
LGTM
@aiChaoSONG @Bin-QA let try to think where can we use this kind of fmts.
PS: we should split both into playback and capture with same pcm id.

aiChaoSONG:
@xiulipan what I want to do is to add a specific test case to test all support formats on all pipeline, and left others unchanged. No need to verify every formats in every test case.

Bin-QA:
@xiulipan @aiChaoSONG how about add option for current test-case script to decide whether to go though the **fmts** like [check-capture](https://github.com/thesofproject/sof-test/blob/master/test-case/check-capture.sh#L84)
```
OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='loop full fmt'
OPT_PARM_lst['f']=0         OPT_VALUE_lst['f']=0

if [ ${OPT_VALUE_lst['f']} -eq 1 ];then
    fmt=$(func_pipeline_parse_value $idx fmts)
fi

    for fmt_item in $(echo $fmt)
    do
        arecord -D$dev -r $rate -c $channel -f $fmt -d $duration $file -vv -q
    done
```

aiChaoSONG:
@Bin-QA  will send a new PR to modify test case later

xiulipan:
@aiChaoSONG @Bin-QA let try to make a new test case like `check-format`? I think this case will focus on if all format is support after we sure that the pipeline is working.

Bin-QA:
what the different with my opinion?   
Here do the script change  
Then load test case by different parameter  
So the new case will add at script load  
don't need to add new script to handle this

aiChaoSONG:
@xiulipan @Bin-QA  I vote on add a new test case rather than change current test case, the reasons are:
- if add an option, this option will goes to every test case, but we don't need to test all formats in all test cases(my opinion). we only expose the fmts to test case, test case should choose to use fmt or fmts.
- check-aplay / check-capture only require minimum condition, that is run with the first supported format.  but check all formats has a higher failure probability, we may take it as a higher level test cases.

Btw, We should firstly discuss if we need to test all formats in all test cases.

Bin-QA:
```
OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='loop full fmt'
OPT_PARM_lst['f']=0         OPT_VALUE_lst['f']=0
```
`OPT_PARM_lst['f']=0` means don't need parameter, so this option is a switch
`OPT_VALUE_lst['f']=0` means default value is close

So add this option don't have any change for our current case

Now for the case load you can add ` -f ` to check all formats, default just check the 1st, so it is an extend

xiulipan:
@Bin-QA @aiChaoSONG so I think we can make alignment here.
extend the same scripts but create a new test case in CI.

PR #31, opened 2020-01-09
test: add test mode option for xrun injection test

keqiaozhang:
Add test mode option to make it as a common case to support
playback and capture respectively.


keqiaozhang:
Close this PR and track in https://github.com/thesofproject/sof-test/pull/32

PR #32, opened 2020-01-09
test: add test mode and count parameter for xrun injection test

keqiaozhang:


wenqingfu:
seems you brought in the patch for #31 together accidentally?

keqiaozhang:
@wenqingfu , Yes,  I should submit these 2 patches as one patch set.

aiChaoSONG:
@keqiaozhang @wenqingfu Can we close #31, and leave #32 for review and merge? may be need to change title of #32

stripes416:
This also LGTM.

PR #35, opened 2020-01-10
Test Case: Add check reboot test

Bin-QA:
Add new Test Case for the reboot

### DIFF
--- a/test-case/check-reboot.sh
+++ b/test-case/check-reboot.sh
@@ -1,0 +1,154 @@
+#!/bin/bash
+
+##
+## Case Name: test reboot
+## Preconditions:
+##    N/A
+## Description:
+##    run reboot for the test
+## Case step:
+##    1. check system status is correct
+##    2. wait for the delay time
+##    3. trigger for the reboot
+## Expect result:
+##    Test execute without report error in the LOG
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=3
+
+OPT_OPT_lst['t']='timeout'  OPT_DESC_lst['t']='timeout after system boot up'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']=30
+
+OPT_OPT_lst['d']='delay'    OPT_DESC_lst['d']='delay time mapping to sub-case PM status'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=10
+
+func_opt_parse_option $*
+
+func_lib_check_sudo
+loop_count=${OPT_VALUE_lst['l']}
+delay=${OPT_VALUE_lst['d']}
+timeout=${OPT_VALUE_lst['t']}
+
+# write the total & current count to the status file
+status_log=$LOG_ROOT/status.txt
+echo "$loop_count $(uname -r)" >> $status_log
+count=$(head -n 1 $status_log|awk '{print $1;}')
+orig_kern=$(head -n 1 $status_log|awk '{print $2;}')
+current=$[ $count - $loop_count ]
+cur_kern=$(uname -r)
+
+# compare kernel version
+if [ "$(head -n 1 $status_log|awk '{print $2;}')" != "$(uname -r)" ]; then
+    sed -i '$s/$/ fail/' $status_log
+    dloge "Kernel version already been changed"
+    exit 1
+fi
+
+# delay timeout for wait SOF load finish
+# catch tplg file as the keyword
+load_time=0
+while [ ! "$(sof-get-default-tplg.sh)" ]
+do
+    sleep 1
+    load_time=$[ $load_time + 1 ]
+    # trigger timeout detect
+    if [ $load_time -ge $timeout ];then
+        dloge "Wait too long for SOF load: $load_time s"
+        exit 1
+    fi
+done
+
+dlogi "SOF load took $load_time s to load TPLG success"
+
+declare -a verify_lst
+verify_lst=(${verify_lst[*]} "verify-firmware-presence.sh")
+verify_lst=(${verify_lst[*]} "verify-kernel-module-load-probe.sh")
+verify_lst=(${verify_lst[*]} "verify-pcm-list.sh")
+verify_lst=(${verify_lst[*]} "verify-sof-firmware-load.sh")
+verify_lst=(${verify_lst[*]} "verify-tplg-binary.sh")
+verify_lst=(${verify_lst[*]} "check-runtime-pm-status.sh")
+
+declare -A verify_opt_lst
+verify_opt_lst['verify-pcm-list.sh']="-t $(sof-get-default-tplg.sh)"
+verify_opt_lst['check-runtime-pm-status.sh']="-t $(sof-get-default-tplg.sh) -l 1 -d $delay"
+
+dlogi "Round: $current/$count Check status begin"
+for i in ${verify_lst[*]}
+do
+    dlogc "$(dirname ${BASH_SOURCE[0]})/$i $(echo ${verify_opt_lst["$i"]})"
+    $(dirname ${BASH_SOURCE[0]})/$i $(echo ${verify_opt_lst["$i"]})
+    if [ $? -ne 0 ];then
+        # last line add failed keyword
+        sed -i '$s/$/ fail/' $status_log
+        # record failed case:
+        echo "$(dirname ${BASH_SOURCE[0]})/$i $(echo ${verify_opt_lst["$i"]})" >> $status_log
+        dloge "$i: fail"
+        exit 1
+    else
+        # last line add pass keyword
+        dlogi "$i: pass" 
+    fi
+done
+
+dlogi "Round $current: Status check finished"
+sed -i '$s/$/ pass/' $status_log
+[[ $loop_count -le 0 ]] && exit 0
+
+dlogi "Do the prepare for the next round bootup"
+# run the script for the next round
+next_count=$[ $loop_count - 1 ]
+
+full_cmd=$(ps -p $PPID -o args --no-header)
+# parent process have current script name
+# like: bash -c $0 .....
+# here is string compare, [] will cause compare failed
+if [[ "$full_cmd" =~ "bash -c" ]]; then
+    full_cmd=${full_cmd#bash -c}
+else
+    full_cmd=$(ps -p $$ -o args --no-header)
+    full_cmd=${full_cmd#\/bin\/bash}
+fi
+
+# load script default value for the really full command
+if [ $# -eq 0 ]; then
+    full_cmd=$(echo $full_cmd|sed "s:$0:& -l $loop_count:g")
+fi
+
+# convert relative path to absolute path
+full_cmd=$(echo $full_cmd|sed "s:$0:$(realpath $0):g")
+# some load will use '~' which is $HOME, but after system reboot, in rc.local $USER is root
+# so the '~' will lead to the error path
+full_cmd=$(echo $full_cmd|sed "s:~:$HOME:g")
+# now convert full current command to next round command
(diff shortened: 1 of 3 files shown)


test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+    else
+        # last line add pass keyword
+        dlogi "$i: pass" 
+    fi
+done
+dlogi "Round $current: Status check finished"
+sed -i '$s/$/ pass/' $status_log
+[[ $loop_count -le 0 ]] && exit 0
+
+dlogi "Do the perpare for the next round bootup"
```
stripes416: perpare -> prepare (or in this case, a better word might be preparation)

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+# run the script for the next round
+next_count=$[ $loop_count - 1 ]
+
+full_cmd=$(ps -p $PPID -o args --no-header)
+# parenet process have current script name
```
stripes416: parenet -> parent

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# because write too many information & complex to the /etc/rc.local file, so here direct operation it
+# normally suggest to use 'sudo sof-boot-once.sh command' to write it
+
+# change the file own & add write permission
+sudo chmod u+w $boot_file
+sudo chown $UID $boot_file
+# because hijack sudo command, so "sed -i '/^exit 0/d' $boot_file" have some problem in here
+old_content="$(cat $boot_file|grep -v '^exit')"
+# write the infromation to /etc/rc.local
```
stripes416: infromation -> information

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+OPT_OPT_lst['d']='delay'    OPT_DESC_lst['d']='delay after system bootup'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=2
+
+func_opt_parse_option $*
+
+func_lib_check_sudo
+loop_count=${OPT_VALUE_lst['l']}
+delay=${OPT_VALUE_lst['d']}
+
+# write the orig & status count to the status file
```
wenqingfu: from the code below, "orig & status" better phrased as  "total & current"

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+status_log=$LOG_ROOT/status.txt
+echo $loop_count >> $status_log
+count=$(head -n 1 $status_log|awk '{print $1;}')
+# because write record by desc
```
wenqingfu: not sure what does this comment mean, the code is self-explanatory enough, maybe just delete it?

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+        dlogi "$i: fail" 
+        exit 1
+    else
+        # last line add pass keyword
+        dlogi "$i: pass" 
+    fi
+done
+dlogi "Round $current: Status check finished"
+sed -i '$s/$/ pass/' $status_log
+[[ $loop_count -le 0 ]] && exit 0
```
wenqingfu: before exit the whole testing, do you need to restore the original /etc/rc.local that the script modified in later section?

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+
+dlogi "Do the perpare for the next round bootup"
+# run the script for the next round
+next_count=$[ $loop_count - 1 ]
+
+full_cmd=$(ps -p $PPID -o args --no-header)
+# parenet process have current script name
+# like: bash -c $0 .....
+if [[ "$full_cmd" =~ "bash -c" ]]; then
```
wenqingfu: The rest if statement you use [, so [[ -> [ to keep the style unified.

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# load script default value for the really full command
+if [ $# -eq 0 ]; then
+    full_cmd=$(echo $full_cmd|sed "s:$0:& -l $loop_count:g")
+fi
+
+# convert relative path to absolute path
+full_cmd=$(echo $full_cmd|sed "s:$0:$(realpath $0):g")
+# some load will use '~' which is $HOME, but after system reboot, in rc.local $USER is root
+# so the '~' will direct the error path
```
wenqingfu: direct -> lead to

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+full_cmd=$(echo $full_cmd|sed "s:~:$HOME:g")
+# now convert full current command to next round command
+full_cmd=$(echo $full_cmd|sed "s:-l $loop_count:-l $next_count:g")
+
+boot_file=/etc/rc.local
+# if miss rc.local file let sof-boot-once.sh to create it
+[[ ! -f $boot_file ]] && sudo sof-boot-once.sh
+
+# because write too many information & complex to the /etc/rc.local file, so here direct operation it
+# normally suggest to use 'sudo sof-boot-once.sh command' to write it
```
wenqingfu: this two lines of comment is not useful, suggest to just delete them.

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# change the file own & add write permission
+sudo chmod u+w $boot_file
+sudo chown $UID $boot_file
+# because hijack sudo command, so "sed -i '/^exit 0/d' $boot_file" have some problem in here
```
wenqingfu: ditto. I think you can delete the comment here.

test-case/check-reboot.sh
```diff
@@ -0,0 +1,135 @@
...
+old_content="$(cat $boot_file|grep -v '^exit')"
+# write the infromation to /etc/rc.local
+# LOG_ROOT to confirm all data include sub-case write log to target folder
```
wenqingfu: I think what you meant is,  "LOG_ROOT to make sure all tests, including sub-cases, write log to the same target folder"

test-case/check-reboot.sh
Bin-QA: the case logic base on `sof-boot-once.sh`  
here use this feature in linux:  
1. `/etc/rc.local` is the as load service  
2. `/etc/rc.local` loaded, modify the file, just effect the file not effect the memory(this round)   

so `sof-boot-once.sh` will add itself into `/etc/rc.local`,  
it means `/etc/rc.local` will load this script without parameter  
`sof-boot-once.sh` load without `/etc/rc.local` will qurey from `/etc/rc.local` for itself  
and delete from `sof-boot-once.sh` to the last line  
So it means boot once, because just effect next boot

test-case/check-reboot.sh
```diff
@@ -0,0 +1,130 @@
...
+echo $loop_count >> $status_log
+count=$(head -n 1 $status_log|awk '{print $1;}')
+current=$[ $count - $loop_count ]
+
+dlogi "Round: $current/$count"
+
+# verify-pcm-list.sh & verify-tplg-binary.sh need TPLG file
+export TPLG=$(sof-get-default-tplg.sh)
+
+declare -a verify_lst
```
aiChaoSONG: Here you use a list to note which case to run after reboot, is it better to use some configure file? Then if we want to add one more case to run after reboot, we only need to add an item in configuration file.

Bin-QA: 1. add configure file means all same type case need the configure file
2. mapping it as parameter need to handle too many option
3. if load the complex test-case, I think the case description also need to update

So I more like to keep this list/use the scan to run all `verify-*` script?

aiChaoSONG: for keep the list / run all `verify-*`, I agree on keep the list, this will limit the influence within the script, the latter will add a rule for script naming, I think that's not good. @xiulipan

test-case/check-reboot.sh
```diff
@@ -0,0 +1,130 @@
...
+##    Test execute without report error in the LOG
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=3
+
+OPT_OPT_lst['d']='delay'    OPT_DESC_lst['d']='delay after system bootup'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=2
```
fredoh9: in some case delaying 2 second is not enough, I think setting at least 5 seconds by default is better.

test-case/check-reboot.sh
```diff
@@ -0,0 +1,130 @@
...
+
+func_opt_parse_option $*
+
+func_lib_check_sudo
+loop_count=${OPT_VALUE_lst['l']}
+delay=${OPT_VALUE_lst['d']}
+
+# write the total & current count to the status file
+status_log=$LOG_ROOT/status.txt
```
fredoh9: When booting failed, it may boot to generic kernel. verify_list will fail with generic but should we remember kernel name/version being tested against?

stripes416: thats definitely a good point.

test-case/check-reboot.sh
Bin-QA: this delay is after `verify_lst`, so may be I move the delay before `verify_lst` is better also to load `check-runtime-pm-status` to check the PM status

test-case/check-reboot.sh
Bin-QA: check kernel version is good idea

test-case/check-reboot.sh
xiulipan: @Bin-QA @aiChaoSONG Configurable list would be better.
But how could we configure this list? by parameters or by files.

test-case/check-reboot.sh
xiulipan: @Bin-QA @fredoh9 On some device the SOF load will take up to 30s.
As sof is always work on module mode, it will be probed very late. We also need to wait for the tplg to load. So I will suggest to set a timeout and give a time about how long SOF fully loaded took

test-case/check-reboot.sh
xiulipan: Maybe we should keep a list for the verify test cases to compare if the env is changed.

Bin-QA: How about add more environment into [sof-dump-status.py](https://github.com/thesofproject/sof-test/blob/master/tools/sof-dump-status.py)  

so here direct load this script store into the file & compare the file content

Bin-QA:
Update for rebase

tools/sof-get-default-tplg.sh
```diff
@@ -3,9 +3,9 @@
 # Current: load it from dmesg
```
xiulipan: could you change the comment here?
I see it will fallback in below order:
```
dmesg
journalctl -k
/var/log/syslog
```

test-case/check-reboot.sh
```diff
@@ -0,0 +1,154 @@
...
+echo "$loop_count $(uname -r)" >> $status_log
+count=$(head -n 1 $status_log|awk '{print $1;}')
+orig_kern=$(head -n 1 $status_log|awk '{print $2;}')
+current=$[ $count - $loop_count ]
+cur_kern=$(uname -r)
+
+# compare kernel version
+if [ "$(head -n 1 $status_log|awk '{print $2;}')" != "$(uname -r)" ]; then
+    sed -i '$s/$/ fail/' $status_log
```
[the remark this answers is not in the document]
Bin-QA: > why not use ">>" to append the log?
`>>` will write the information to the new line
> PS: why not wrapper a function for this feature? I see few usage for this ` # last line add failed keyword`
Here just one command for `sed`, so I don't think it need to wrapper a function

tools/sof-get-default-tplg.sh
stripes416: Marking resolved as @Bin-QA made the change with the latest push.

fredoh9:
I think this is good test script to test reboot test. Let's merge this.

PR #41, opened 2020-01-14
Bugfix: check-pause-resume error expect script

Bin-QA:
1. Fix error 'sleep', tcl not support 'sleep', just support 'after'
2. Add random for the sleep time, it is more like human input
  remove "sleep" option for this change
3. remove "duration" option, it option is confluct with "count"
  after "count" trigger, target process will been quit
4. fix status check after expect script
5. add dmesg check after expect script

### DIFF
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -26,12 +26,6 @@ OPT_PARM_lst['m']=1         OPT_VALUE_lst['m']='playback'
 OPT_OPT_lst['c']='count'    OPT_DESC_lst['c']='pause/resume repeat count'
 OPT_PARM_lst['c']=1         OPT_VALUE_lst['c']=10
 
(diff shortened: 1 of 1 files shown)


test-case/check-pause-resume.sh
```diff
@@ -76,32 +68,45 @@ do
     fmt=$(func_pipeline_parse_value $idx fmt)
     dev=$(func_pipeline_parse_value $idx dev)
 
-    dlogi "Entering expect script with: $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i -d $duration $file_name -q"
+    dlogi "Entering expect script with: $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q"
 
+    # expect is tcl language script
+    #   expr rand(): produces random numbers between 0 and 1
+    #   after ms: Ms must be an integer giving a time in milliseconds. The command sleeps for ms milliseconds and then returns.
```
stripes416: I think you might have left a word off of the end of your comment here.

PR #42, opened 2020-01-14
Bugfix: add delay option for check-runtime-pm-status

Bin-QA:
convert sleep 5s to sleep ${delay} which is parameter of script


xiulipan:
@keqiaozhang @aiChaoSONG @ClarexZhou Please help to fix this hot fix for the runtime pm test case.

PR #46, opened 2020-01-16
tools: Update the sof-kernel-dump

Bin-QA:
1.  tools: add sof-get-kernel-line to catch line NO
this patch will helpful other usage
2. speed up the sof-kernel-dump
the large file still is big problem
3. update document for add new script
4. fix https://github.com/thesofproject/sof-test/issues/48

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,5 +98,8 @@ apl
 * sof-kernel-dump.sh
 <br> catch all kernel information after system boot up from /var/log/kern.log file
 
+* sof-get-kernel-line.sh
+<br> print all kernel versions and their line numbers from /var/log/kern.log, with the most recent <first/last>
+
 * sof-disk-usage.sh
 <br> check current disk size for avoid system problem without enough space
--- a/tools/sof-get-kernel-line.sh
+++ b/tools/sof-get-kernel-line.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# dump line number + kernel version
+# for example:
+# 22 4.15.0-74-generic
+# 1092 4.15.0-74-generic
+# 2156 4.15.0-74-generic
+# 3246 4.15.0-74-generic
+log_file=${1:-"/var/log/kern.log"}
+[[ ! -f $log_file ]] && builtin exit 1
+bootup_keyword='\[    0.000000\] Linux version '
+grep -na "$bootup_keyword" $log_file|awk -F ':' '{print $1 $(NF-2);}'|sed "s/$bootup_keyword//g"|awk '{print $1, $2;}'
--- a/tools/sof-kernel-dump.sh
+++ b/tools/sof-kernel-dump.sh
@@ -1,19 +1,26 @@
 #!/bin/bash
 
-log_file="/var/log/kern.log"
-bootup_keyword='\[    0.000000\] Linux version '"$(uname -r)"
-# tr '\0' ' ' to fix warning message: "command substitution: ignored null byte in input"
-base_info=$(grep -na "$bootup_keyword" $log_file|tr '\0' ' ')
-# rollback to kern.log.1, OS will write the log into the different file
-# so here when current log file: kern.log miss the system boot information
-# rollback catch from kern.log.1, normally just 1 file is enough to cache it
-[[ ! "$base_info" ]] && log_file="/var/log/kern.log.1"
-boot_line=$(grep -na "$bootup_keyword" $log_file |tr '\0' ' '|sed -n '$p'|awk -F ':' '{print $1;}')
-[[ ! "$boot_line"  ]] && boot_line=0
-#sed -n "$boot_line,\$p" $log_file |awk '{ for(i=1; i<=4; i++){ $i="" }; print $0 }'|sed 's/^    //g'
-sed -n "$boot_line,\$p" $log_file |cut -f5- -d ' '
-[[ "$base_info" ]] && exit
-# this logic means not find bootup information in the kern.log file, but find at kern.log.1
-# so some information write at kern.log, it also need to dump
-#cat /var/log/kern.log |awk '{ for(i=1; i<=4; i++){ $i="" }; print $0 }'|sed 's/^    //g'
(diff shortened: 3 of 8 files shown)


tools/sof-get-kernel-line.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+log_file=${1:-"/var/log/kern.log"}
+bootup_keyword='\[    0.000000\] Linux version '
+grep -na "$bootup_keyword" $log_file|awk -F ':' '{print $1 $(NF-2);}'|sed "s/$bootup_keyword//g"|awk '{print $1, $2;}'
```
stripes416: Please add a newline at the end of your file :)

tools/sof-kernel-dump.sh
```diff
@@ -1,19 +1,17 @@
 #!/bin/bash
```
stripes416: In this commit message, I think you meant campre -> compare.

README.md
```diff
@@ -98,5 +98,8 @@ apl
 * sof-kernel-dump.sh
 <br> catch all kernel information after system boot up from /var/log/kern.log file
 
+* sof-get-kernel-line.sh
+<br> catch all kernel line number from /var/log/kern.log file
```
xiulipan: After run a test I got what you mean here, I shows the start line number with the kernel version form oldest to latest. I will suggest to from latest to oldest.
@stripes416 Could you help to get some more accurate description for this script.
```
22 4.15.0-74-generic
1092 4.15.0-74-generic
2156 4.15.0-74-generic
3246 4.15.0-74-generic
```

tools/sof-get-kernel-line.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
xiulipan: I will suggest to have some inline description with example output here.

You do not only output line number but also kernel version info.
```
22 4.15.0-74-generic
1092 4.15.0-74-generic
2156 4.15.0-74-generic
3246 4.15.0-74-generic
```

tools/sof-kernel-dump.sh
```diff
@@ -1,19 +1,17 @@
...
-boot_line=$(grep -na "$bootup_keyword" $log_file |tr '\0' ' '|sed -n '$p'|awk -F ':' '{print $1;}')
-[[ ! "$boot_line"  ]] && boot_line=0
-#sed -n "$boot_line,\$p" $log_file |awk '{ for(i=1; i<=4; i++){ $i="" }; print $0 }'|sed 's/^    //g'
-sed -n "$boot_line,\$p" $log_file |cut -f5- -d ' '
-[[ "$base_info" ]] && exit
-# this logic means not find bootup information in the kern.log file, but find at kern.log.1
-# so some information write at kern.log, it also need to dump
-#cat /var/log/kern.log |awk '{ for(i=1; i<=4; i++){ $i="" }; print $0 }'|sed 's/^    //g'
-cat /var/log/kern.log |cut -f5- -d ' '
+reverse_order=${1:-0}
```
xiulipan: could you add some inline comments here?
or directly make the output from `sof-get-kernel-line.sh` reserved.

PS: you can try with `tac`

tools/sof-kernel-dump.sh
```diff
@@ -1,19 +1,17 @@
...
+reverse_order=$[ $reverse_order + 1 ]
+log_file=${2:-"/var/log/kern.log"}
+declare -a line_lst
+line_lst=($(sof-get-kernel-line.sh $log_file|tail -n $reverse_order|awk '{print $1;}'))
```
xiulipan: Also, please add some inline comments here to help understand what you are doing here.
I think this would be too complex to get the desired boot start line.

I do some test and I can get the begin line with:
`sof-get-kernel-line.sh $log_file | tac | sed -n "3, 1p" | awk '{print $1;}'`

No need to play with array.

PS: it seems we do not have error handling here. What if user input the order outside our range?EG. we only have one boot log, but user want the previous one.

tools/sof-kernel-dump.sh
```diff
@@ -1,19 +1,17 @@
...
+end_line=$(wc -l $log_file|awk '{print $1;}')
+begin_line=$[ $end_line - ${line_lst[0]} + 1 ]
```
xiulipan: Please see my above comments to get the begin line directly.
And this is actually total line number right?

tools/sof-get-kernel-line.sh
```diff
@@ -0,0 +1,6 @@
...
+log_file=${1:-"/var/log/kern.log"}
+bootup_keyword='\[    0.000000\] Linux version '
+grep -na "$bootup_keyword" $log_file|awk -F ':' '{print $1 $(NF-2);}'|sed "s/$bootup_keyword//g"|awk '{print $1, $2;}'
```
xiulipan:
```suggestion
grep -na "$bootup_keyword" $log_file|awk -F ':' '{print $1 $(NF-2);}'|sed "s/$bootup_keyword//g"|awk '{print $1, $2;}' | tac
```
use a simple tac to output in reserved order

Bin-QA: here you try the small file on the terminal,  
if you file is large like our BDW:  1.8G  
your line number will be >1000 so here your terminal output will been flash

tools/sof-kernel-dump.sh
Bin-QA: the basic logic is use `tail` command to dump the last part which is faster than `sed`  
So here must to know how many line to dump  
it is why need begin_line

tools/sof-kernel-dump.sh
Bin-QA: I will add some error handle for the `kern.log.1` file

README.md
stripes416:
```suggestion
<br> print all kernel versions and their line numbers from /var/log/kern.log, with the most recent <first/last>
```
first/last is just based on whichever way you choose to print it.

PR #53, opened 2020-01-21
[RFC] case-lib: add lock of sof-test

Bin-QA:
record current PID as lock file
from dlog* refer keyword to ignore sub-test case
when load exit will clean up the lock file for current PID

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -45,8 +45,19 @@ function exit()
         done
     fi
 
+    # check function already defined
     # when exit force check the pulseaudio whether disabled
-    func_lib_restore_pulseaudio
+    [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
+
+    if [ -f $SOF_LOCK ];then
+        # use string compare instead of int to confirm file content correct
+        # just remove the current pid lock file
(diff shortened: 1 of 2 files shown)


case-lib/hijack.sh
```diff
@@ -48,6 +49,11 @@ function exit()
         done
     fi
 
+    # just remote the current pid lock file
```
paulstelian97: remove?

Bin-QA:
Update for code rebase

stripes416:
@Bin-QA I'm having a hard time understanding what you are trying to accomplish here. Can you explain a little more about why you need to have locks on the pulseaudio pid?

Bin-QA:
> @Bin-QA I'm having a hard time understanding what you are trying to accomplish here. Can you explain a little more about why you need to have locks on the pulseaudio pid?

this patch try to lock the device to confirm current just have one script(sof-test) run on it  
it will to avoid CI/manual etc to run the script(sof-test) on the same DUT cause the result is not correct

Bin-QA:
@stripes416 @xiulipan @keqiaozhang @aiChaoSONG @libinyang @fredoh9 please help to review & whether need this patch to protect current DUT just run one sof-test

Issue #54, opened 2020-01-21
[BUG] random Device or resource busy in kmod test with audio


xiulipan:
In test **check-kmod-load-unload-after-playback.sh**
the first check to aplay random failed due to
`aplay: main:788: audio open error: Device or resource busy`

Full log
https://sof-ci.01.org/linuxpr/PR1737/build2947/devicetest/BDW_WSB_RT286/check-kmod-load-unload-after-playback.sh/check-kmod-load-unload-after-playback.sh.txt
https://sof-ci.01.org/linuxpr/PR1737/build2947/devicetest/BDW_WSB_RT286/check-kmod-load-unload-after-playback.sh/

It seems pulseaudio is working at that moment.

xiulipan:
Fix by #55

PR #55, opened 2020-01-21
case-lib: bugfix #54 for pulseaudio restore logic

Bin-QA:
1. move restore function after kill child process to avoid kill
pulseaudio again
2. add delay check for restore pulseaudio to confirm restore success
3. add nohup for restore pulseaudio to avoid pulseaudio killed
by ssh lost connect

fix issue #54 

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -80,8 +80,20 @@ func_lib_restore_pulseaudio()
     do
         user=${line%% *}
         cmd=${line#* }
-        sudo -u $user $cmd >/dev/null 2>&1 &
+        nohup sudo -u $user $cmd >/dev/null &
     done
+    # now wait for the pulseaudio restore in the ps process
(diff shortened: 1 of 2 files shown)


case-lib/lib.sh
```diff
@@ -80,8 +80,20 @@ func_lib_restore_pulseaudio()
...
+    while [ ! "$(ps -C pulseaudio --no-header)" ];
+    do
+        wait_t=$[ $wait_t + 1 ]
+        sleep 1s
+        if [ $wait_t -ge $timeout ]; then
+            dlogw "Restore pulseaudio wait too long: $timeout"
+            break
+        fi
+    done
+    dlogi "Restore pulseaudio spend $wait_t to wait"
```
stripes416:
```suggestion
    dlogi "Restoring pulseaudio took $wait_t seconds"
```

PR #56, opened 2020-01-22
test-case: add ability to loop ipc flood test

fredoh9:

### DIFF
--- a/test-case/check-ipc-flood.sh
+++ b/test-case/check-ipc-flood.sh
@@ -21,10 +21,14 @@ OPT_PARM_lst['c']=1         OPT_VALUE_lst['c']=10000
 OPT_OPT_lst['f']='dfs'      OPT_DESC_lst['f']='system dfs file'
 OPT_PARM_lst['f']=1         OPT_VALUE_lst['f']="/sys/kernel/debug/sof/ipc_flood_count"
 
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
+
 func_opt_parse_option $*
 
 lpc_loop_cnt=${OPT_VALUE_lst['c']}
 ipc_flood_dfs=${OPT_VALUE_lst['f']}
+loop_cnt=${OPT_VALUE_lst['l']}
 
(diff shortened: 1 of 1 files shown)


test-case/check-ipc-flood.sh
```diff
@@ -34,13 +38,17 @@ func_lib_check_sudo
...
-sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE
-[[ $? -ne 0 ]] && dloge "Catch error in kernel log" && exit 1
+for i in $(seq 1 $loop_cnt)
+do
+    dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
+    sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
+
+    sof-kernel-log-check.sh $KERNEL_LAST_LINE
```
Bin-QA: the `KERNEL_LAST_LINE` will trace from you case start,  
so if you put it into loop you should reset this value  
on the other hand, you can try `sof-kernel-log-check.sh 0` in your loop  
```
    # sof-kernel-log-check script parameter number is 0/Non-Number will force check from dmesg
    sof-kernel-log-check.sh 0
```

I will suggest check from `dmesg` in the loop,  
check full log `/var/log/syslog` at script quit to avoid miss some kernel error

fredoh9: just updated as you suggested.
`sof-kernel-log-check.sh 0`

test-case/check-ipc-flood.sh
```diff
@@ -34,13 +38,17 @@ func_lib_check_sudo
...
+    sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
+
+    sof-kernel-log-check.sh 0
+    [[ $? -ne 0 ]] && dloge "Catch error in dmesg" && exit 1
 
-dlogi "Dumping test logs!"
-dmesg | grep "IPC Flood count" -A 2
+    dlogi "Dumping test logs!"
+    dmesg | grep "IPC Flood count" -A 2
+done
```
Bin-QA: can you add this line ` sof-kernel-log-check.sh $KERNEL_LAST_LINE ` before exit

so your code will look like:
```
sof-kernel-log-check.sh $KERNEL_LAST_LINE
exit $?
```

PR #58, opened 2020-01-23
test-case: check return value of aplay or arecord

fredoh9:
If aplay or arecord is failed, no need to proceed to test.
Exit early to report for failure.

### DIFF
--- a/test-case/check-signal-stop-start.sh
+++ b/test-case/check-signal-stop-start.sh
@@ -94,7 +94,21 @@ do
 
     dlogi "Testing: run stop/start test on PCM:$pcm,$pipeline_type. Interval time: $interval"
     dlogc $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
-    $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q & pid=$!
+    $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
+    pid=$!
+
+    # If the process is terminated too early, this is error case.
+    # Typical root causes of the process early termination are,
+    #     1. soundcard is not enumerated
+    #     2. soundcard is enumerated but the PCM device($dev) is not available
+    #     3. the device is busy
+    #     4. set params fails, etc
+    sleep 0.5
+    if [[ ! -d /proc/$pid ]]; then
+        dloge "$cmd process[$pid] is terminated too early"
+        exit 1
+    fi
+
     # do stop/start test
     func_stop_start_pipeline
     # kill aplay/arecord process
--- a/test-case/check-xrun-injection.sh
+++ b/test-case/check-xrun-injection.sh
@@ -102,6 +102,19 @@ do
     dlogc $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q
     $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
     pid=$!
+
+    # If the process is terminated too early, this is error case.
+    # Typical root causes of the process early termination are
+    #     1. soundcard is not enumerated
+    #     2. soundcard is enumerated but the PCM device($dev) is not available
+    #     3. the device is busy
+    #     4. set params fails, etc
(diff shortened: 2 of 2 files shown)


test-case/check-signal-stop-start.sh
```diff
@@ -95,6 +95,8 @@ do
     dlogi "Testing: run stop/start test on PCM:$pcm,$pipeline_type. Interval time: $interval"
     dlogc $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
     $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q & pid=$!
+    [[ $? -ne 0 ]] && dloge "$cmd error without process" && exit 1
```
Bin-QA: the `$?` mapping to the last command result, so it will be `pid=$!`  
I don't think this check will work as your expect

you can try this command for $? usage
```
[[ "$(ls)" ]] && echo "here"
echo $?
```

stripes416: @Bin-QA Thanks the info re: `$(ls)` --- I didn't know of a way to handle this since the aplay / arecord command ends in ` & pid=$!`

fredoh9: @Bin-QA This was good comment. I was testing error cases in locally, it took sometime.
Catching return value for background process is not good idea. Rather we should check for the process id still running after short sleep. I will submit a patch shortly.

test-case/check-signal-stop-start.sh
```diff
@@ -94,7 +94,21 @@ do
 
     dlogi "Testing: run stop/start test on PCM:$pcm,$pipeline_type. Interval time: $interval"
     dlogc $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
-    $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q & pid=$!
+    $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
+    pid=$!
+
+    # If the process is terminated too early, this is error case.
+    # Typical root casuses of the process early termination are,
```
stripes416:
```suggestion
    # Typical root causes of the process early termination are,
```

test-case/check-xrun-injection.sh
```diff
@@ -102,6 +102,19 @@ do
     dlogc $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q
     $cmd -D$dev -r $rate -c $channel -f $fmt $dummy_file -q &
     pid=$!
+
+    # If the process is terminated too early, this is error case.
+    # Typical root casuses of the process early termination are
```
stripes416:
```suggestion
    # Typical root causes of the process early termination are
```

test-case/check-signal-stop-start.sh
aiChaoSONG: @Bin-QA  @fredoh9 will it better to assign the cmd execution result to some variable, and make a check?

Bin-QA: `$?` usage example:
```
[[ "string" ]] && echo "here"
echo $?
[[ "" ]] && echo "here"
echo $?
```

So here `$?` is use for `exit status`  
if you want to know background process exit status ,you can try `wait` command

xiulipan:
@fredoh9 Could you rebase the PR? I see some https://github.com/thesofproject/sof-test/pull/58/commits/6644846f70920e590cd78ef9b43fa1fa9780c0b0
You need merge this into previous commit, try `git rebase -i`
```
Apply suggestions from code review

Co-Authored-By: Kelly Ledford <kelly.ledford@linux.intel.com>
```

test-case/check-signal-stop-start.sh
```diff
@@ -98,7 +98,7 @@ do
     pid=$!
 
     # If the process is terminated too early, this is error case.
-    # Typical root casuses of the process early termination are,
+    # Typical root causes of the process early termination are,
```
paulstelian97: Squash?

PR #61, opened 2020-01-30
test-case/manual: DPMST test

stripes416:
* Manual walkthrough for DisplayPort Multi-Stream Transport testing

### DIFF
--- a/test-case/manual/DPMST.md
+++ b/test-case/manual/DPMST.md
@@ -1,0 +1,71 @@
+# DPMST test
+DisplayPort Multi-Stream Transport is a manual test.
+
+# Preconditions
+1. Hardware requirements met for each DPMST case
+
+# Test Description
+* Test output from one test unit to (minimum of) two separate monitors with audio
+  output via DP (as well as with USB-C and HDMI adapters if monitors have those
+  inputs available).
+
+# 1) Single DP -> Multi DP hub Test Case:
+## Hardware Requirements
+1. One DP-MST hub with single DP input -> multi-DP output
+2. Two monitors with DP input and audio output
+3. Two DP-USB-C adapters (if monitors also have USB-C input)
+4. Two DP-HDMI adapters (if monitors also have HDMI input)
+* If DP-MST hub has additional outputs, test can be extended to include
+  additional monitors as availble
+
+## Run Instructions
+1. Two monitors are connected via one DP hub using DP-DP cables.
+2. Run:
+   ```
+   amixer contents | grep -i jack -A 2
+   ```
+   to ensure DP jacks on DP hub are detected.
+3. Run aplay, using the 2 DP jacks found with `amixer contents` command.
+   ```
+   aplay -D $dev(jack1) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   aplay -D $dev(jack2) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   ```
+4. Connected monitors should both show display and play audio.
+5. Check dmesg for any unexpected errors.
+* If monitors have USB-C / HDMI input as well, test can also be run with:
+  * DP hub -> DP cable + DP-USB-C adapters -> monitors
+  * DP hub -> DP cable + DP-HDMI adapters -> monitors
+
+# 2) Daisy Chain Test Case:
+## Hardware Requirements
+1. One monitor with DP input **AND** ouput, as well as audio output (monitor1)
+2. An additional monitor with DP input and audio output (monitor2)
+3. Two DP-USB-C adapters (if monitors also have USB-C input)
+4. Two DP-HDMI adapters (if monitors also have HDMI input)
+* Test can be extended to include additional monitors if hardware is available.
+
+## Run Instructions
+1. Monitors are connected via:
+   test unit -> DP cable -> monitor1 -> DP cable -> monitor2
+* Make sure DP 1.2 is enabled in the monitor on screen display menu as well
+2. Run:
+   ```
+   amixer contents | grep -i jack -A 2
+   ```
+   to ensure DP jacks on monitors are detected.
+3. Run 2 aplay processes, using the 2 DP jacks found with `amixer contents` command.
+   ```
+   aplay -D $dev(jack1) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   aplay -D $dev(jack2) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   ```
+4. Connected monitors should both show display and play audio.
+5. Check dmesg for any unexpected errors.
+* If monitors have USB-C / HDMI input as well, test can also be run with:
+  * test unit -> DP cable + DP-USB-C adapter -> monitor1 -> DP cable + DP-USB-C
+  adapter -> monitor2
+  * test unit -> DP cable + DP-HDMI adapter -> monitor1 -> DP cable + DP-HDMI
+  adapter -> monitor2
+
+# Expect result (both cases)
+* Display and audio to monitors is as expected.
+* No unexpected dmesg errors on testing unit.


test-case/manual/DPMST.md
```diff
@@ -0,0 +1,67 @@
...
+   ```
+   aplay -D $dev(jack1) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   aplay -D $dev(jack2) -r $rate -c 2 -f $fmt -d 3 resource/raw/test<specified>.raw
+   ```
+4. Connected monitors should both show display and play audio.
+5. Check dmesg for any unexpected errors.
+* If monitors have USB-C / HDMI input as well, test can also be run with:
+  * DP hub -> DP cables + DP-USB-C adapters -> monitors
+  * DP hub -> DP cables + DP-HDMI adapters -> monitors
+
```
fredoh9: '## Expect result' for 1) Single DP ->/... is required
or make below '## Expect result' to '#Expect result', which means same expected results for all

stripes416: fixed

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,67 @@
...
+1. Hardware requirements met for each DPMST case
+
+# Test Description
+* Test output from one test unit to two separate monitors with audio output via
+  DP (as well as with USB-C and HDMI adapters if monitors have those inputs
+  available).
+
+# 1) Single DP -> Multi DP hub Test Case:
+## Hardware Requirements
+1. One DisplayPort hub = single DP -> multi-DP
```
xiulipan: Could you change this? It should be some DP-MST hub. Normal hub will not have this feature. @libinyang could you help to clarify?

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,67 @@
...
+2. Two monitors with DP input and audio output
```
xiulipan: DP-MST will have maximum to 3 DP output. So change to 3 would be better.

test-case/manual/DPMST.md
stripes416: @xiulipan The info I received from @ClarexZhou when I asked if there was a specific type of hub needed was "I think any DP hub (support one to x) should be ok." If this isn't correct, it will definitely need to be updated.

test-case/manual/DPMST.md
stripes416: @xiulipan are you sure? @ClarexZhou indicated the hub they were using was 1->4. 

I will however update this to include testing with more monitors if possible.

xiulipan: @stripes416 I just do a quick search on Amazon and found a lot of types of DP-MST hub with different output. So I think either will OK, but it must to be DP-MST hub.

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,67 @@
...
+3. Two DP-USB-C adapters (if monitors also have USB-C input)
+4. Two DP-HDMI adapters (if monitors also have HDMI input)
+
+## Run Instructions
+1. Two monitors are connected via one DP hub using DP-DP cables.
+2. Run:
+   ```
+   amixer contents | grep -i jack -A 2
+   ```
```
keqiaozhang: I think we also need some amixer settings for *pin/*port part.  Or we cannot get audio output from each DP monitor.

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,67 @@
...
+3. Two DP-USB-C adapters (if monitors also have USB-C input)
+4. Two DP-HDMI adapters (if monitors also have HDMI input)
+
+## Run Instructions
+1. Monitors are connected via:
+   test unit -> DP cable -> monitor1 -> DP cable -> monitor2
+2. Run:
+   ```
+   amixer contents | grep -i jack -A 2
+   ```
```
keqiaozhang: We also need amixer settings here.

test-case/manual/DPMST.md
stripes416: @ClarexZhou indicated that the process no longer includes the *pin/*port info.
> Sof-dev is using common HDMI solution now. Won’t need Pinx-Ports any more.

test-case/manual/DPMST.md
xiulipan: @keqiaozhang after switch to HDA HDMI machine driver, do we still need amixer setting

test-case/manual/DPMST.md
stripes416: Addressed in latest update.

test-case/manual/DPMST.md
stripes416: Addressed in latest update.

stripes416:
I've updated this PR based on all the information I have. If there is any additional information that needs to be updated / dispersed, then someone needs to give it to me. I pinged everyone involved with this PR again via email as well.

test-case/manual/DPMST.md
keqiaozhang: Just confirmed that there're no pin/port kcontrols in amixer setting. now we are using common HDMI, no need to change the amixer settings anymore.

test-case/manual/DPMST.md
keqiaozhang: Confirmed that there're no Pin x/Port x kcontrols in amixer setting. so we don't need amixer setting anymore. This feature is provided by Kai in kernel side.

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,70 @@
...
+4. Connected monitors should both show display and play audio.
+5. Check dmesg for any unexpected errors.
+* If monitors have USB-C / HDMI input as well, test can also be run with:
+  * DP hub -> DP cables + DP-USB-C adapters -> monitors
+  * DP hub -> DP cables + DP-HDMI adapters -> monitors
+
+# 2) Daisy Chain Test Case:
+## Hardware Requirements
+1. One monitor with DP input **AND** ouput, as well as audio output (monitor1)
+2. An additional monitor with DP input and audio output (monitor2)
```
libinyang: DP output in monitor2 is not a must

test-case/manual/DPMST.md
```diff
@@ -0,0 +1,70 @@
...
+   ```
+4. Connected monitors should both show display and play audio.
+5. Check dmesg for any unexpected errors.
+* If monitors have USB-C / HDMI input as well, test can also be run with:
+  * DP hub -> DP cables + DP-USB-C adapters -> monitors
+  * DP hub -> DP cables + DP-HDMI adapters -> monitors
+
+# 2) Daisy Chain Test Case:
+## Hardware Requirements
+1. One monitor with DP input **AND** ouput, as well as audio output (monitor1)
```
libinyang: It is better to mention to enable DP1.2 in the monitor OSD menu. Many cases failure of DP MST chain test is because of the wrong setting in monitor.

test-case/manual/DPMST.md
stripes416: awesome --- Thanks for checking into that!

test-case/manual/DPMST.md
stripes416: I'll add that comment. Thanks!

test-case/manual/DPMST.md
stripes416: @libinyang do you mean DP input? The statement here just says monitor 2 needs DP input and audio output. (just want to clarify)

libinyang: Actually, for DP MST daisy chain mode, only the first monitor needs output (DP 1.2 supported). The second monitor is just a normal DP monitor.

stripes416: right ... I'm still not seeing what the conflict is. The line says monitor2 just needs DP input and audio output. It doesn't say anything about monitor2 needing DP output.

test-case/manual/DPMST.md
stripes416: added

Issue #37, opened 2020-01-10
Fix exit status of script


Bin-QA:
in  https://github.com/thesofproject/sof-test/pull/11 I find we have some error exit  
```
man bash :
All builtins return an exit status of 2  to  indicate  incorrect usage, 
generally invalid options or missing arguments.
```

We already define  at [exit function](https://github.com/thesofproject/sof-test/blob/master/case-lib/hijack.sh#L51)
```
0: pass
1: fail
2: skip
other: fail
```
but don't document it & fix some exit code in the script

aiChaoSONG:
fixed and close

PR #62, opened 2020-01-31
case-lib/opt.sh: Change exit code when calling func_opt_dump_help

stripes416:
* test will now report as `FAIL` after a problem processing parameters occurs as
  well as when user wants to view help menu via `-h` flag

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -87,7 +87,7 @@ func_opt_parse_option()
             echo -e '    -h |  --help'
             echo -e "\tthis message"
             _func_case_dump_descption
-            exit 0
+            exit 2
         }
 
     # generate the command to load 'getopt'


stripes416:
**Previously:** 
```
ubuntu@jf-apl-up2-pcm512x-64:~/sof-test/test-case$ ./test-speaker.sh -m
: invalid option -- 'm'
Error parsing option
Usage: ./test-speaker.sh [OPTION]

    -l parameter |  --loop parameter
	option of speaker-test
	Default Value: 3
    -s |  --sof-logger
	Open sof-logger trace the data will store at /home/ubuntu/sof-test/logs/test-speaker/last
	Default Value: On
    -t parameter |  --tplg parameter
	tplg file, default value is env TPLG: $TPLG
	Default Value: sof-apl-nocodec.tplg
    -h |  --help
	this message

 Case Name: test speaker test
 Preconditions:
    N/A
 Description:
    using speaker-test to do playback test
 Case step:
    speaker-test on each playback pipelines
 Expect result:
    speaker-test end without error

2020-01-31 00:17:40 UTC [INFO] Test Result: PASS!
```

The reason for this is that when running tests internally -  it is quite easy to misconfigure the suite file and tests will report as `PASS`, leaving users thinking everything is fine and no further investigation is necessary.

Bin-QA:
@stripes416 can you change it to `exit 2`
https://github.com/thesofproject/sof-test/issues/37 this issue already find this problem :)

case-lib/opt.sh
```diff
@@ -87,7 +87,6 @@ func_opt_parse_option()
             echo -e '    -h |  --help'
             echo -e "\tthis message"
             _func_case_dump_descption
-            exit 0
```
Bin-QA: I think you direct `exit 2` and add some comment will more good for this problem  
So you don't need to modify other place

stripes416: Done.

PR #10, opened 2019-12-19
tools: add component searching logic for tplgtool

aiChaoSONG:
This patch adds component searching logic, thus, we can
get some type of components connected to a specified
component.

Based on the searching logic above, add a function to find
componnets connected to PCM, for example, PGA and EQ.

PR #63, opened 2020-01-31
test-case: add equalizer test case

fredoh9:
Tested with sof-apl-eq-pcm512x.tplg on APL_UP2_PCM512X.


fredoh9:
Updated with query amixer id from the system.

In this case, IIR control id=23, FIR control id=24
```
ubuntu@jf-fred-up2-pcm512x:~$ amixer controls | grep EQ
numid=24,iface=MIXER,name='EQFIR1.0 EQFIR'
numid=23,iface=MIXER,name='EQIIR1.0 EQIIR'

```

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+##
+## Case Name: check-audio-equalizer.sh
+## Preconditions:
+##    SOF Topology should have EQ component(s) in any pipeline
+## Description:
+##    check IIR/FIR config working
+## Case step:
+##    1. TODO: Check if the topology has a EQ component in it
```
xiulipan: Could you make this done? I think a pre-check is very import here.
@aiChaoSONG Could you sync with @fredoh9 with your PR https://github.com/thesofproject/sof-test/pull/10 to support IIR/FIR comp

fredoh9: @aiChaoSONG I'd like to add a feature to search specific DAPM name in tplgtool.py. 

1. The tool print all the name in the graph and i can grep it from the output
2. We can add cmdline parameter to search and match the name

I'm fine with either one. I'd like to stay in sync with PGA support. However I'm not getting the idea of your PR #10. How are you going to support PGA name?

Bin-QA: @aiChaoSONG  I think you also need modify sof-tplgreader.py file to add this filed

aiChaoSONG: @fredoh9  each widget is represented by a node structure, the fields of a node are:
- prev: previous node
- next: next node
- name: the name of the node.(in fact, each widget may have two names: the name field and the sname field. both names may be used in the graph from tplg, so the name here is from graph, it may be `name` or `sname` of a widget)
- widget: the widget bind with this node.
- ctrl: kcontrol between two widgets, currently, we don't use this, because all kcontrol is "" in the graph from tplg.

_link_graph function links the graph according to the tplg graph info. it returns two structures:
- node_list: a list of all nodes of this tplg. the node fileds are assigned by traverse the tplg graph, so we have the graph already.
- link_head_list: for each pcm, we have an independent graph, this list only store the head node. (this structure is optional, it is used to separate each pipeline) 

the find_connected_comp is used to find the connected component for a specified component, eg. find PGA component for specified PCM. it use reference node and the component type you want to find for ref node as it's arguments. the second parameter is a string. because you can see that all PGA  names is started with "PGA",and  buffers with "BUF", etc.

to use the feature, first you should have a component name,  and you find the node with `find_node_by_name`, and use `find_connected_comp` to find the connected nodes you want, the 'widget' filed of node is the info you need.

aiChaoSONG: @fredoh9  As I can see, the EQ name is started with "EQ",  and you should have a PCM name, eg PCM0P. with `find_node_by_name`, you got the node for PCM0P, with `find_connected_comp(pcm_node, "EQ")`, you got all the EQ component connected to PCM0P.
if EQIIR and EQFIR matters to you, you can call `find_connected_comp` twice, one with "EQFIR"  and the other with "EQFIR". 
*Note: the return value of `find_connected_comp` is a list.   the "prev" of head node is "None", and the "next" field of tail node is "None", that's how the finding process works.*

aiChaoSONG: @fredoh9 @xiulipan @Bin-QA  after #10 merged, I will modified sof-tplgreader.py, and output EQs and PGAs for each PCM, just like rate and format, I think this is better.

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,87 @@
...
+    local id=$1
+    local conf=$2
+    dlogc "sof-ctl -Dhw:0 -n $id -s $conf"
+    sof-ctl -Dhw:0 -n $id -s $conf
+    if [ $? -ne 0 ]; then
+        dloge "Equalizer setting failure with $conf"
+        return 1
+    fi
+    # test with aplay
+    dlogc "aplay -Dhw:0,0 -f dat -c 2 /dev/zero -d 5"
```
libinyang: It is better not to use the static card and pcm number.

xiulipan: @fredoh9 Could you use some scripts to get the PCM with eq here.

xiulipan:
@fredoh9 Could you check with #10 and add ENV check before start to test with EQ?

fredoh9:
@aiChaoSONG @libinyang @xiulipan I added more checking with tplgtool. I addressed your commments also. can you review again?

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+}
+
+
+eqConfRoot="./eqctl"
+# initialized with negative value. If none of test is performed the test result will be failed.
+result=1
+failed_cnt=0
+
+dlogi "0. Get amixer control id for IIR and FIR"
+IIRid=`amixer -D hw:0 controls | grep EQIIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
```
libinyang: better to use hw:$sofcard ?

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+if [ -z $IIRid ]; then
+        dloge "can't find IIR filter"
+        exit 1
+fi
+
+FIRid=`amixer -D hw:0 controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
```
libinyang: ditto

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+do
+    dlogi "$idx - total pipeline= $PIPELINE_COUNT"
+    dev=$(func_pipeline_parse_value $idx dev)
+    channel=$(func_pipeline_parse_value $idx channel)
+    rate=$(func_pipeline_parse_value $idx rate)
+    fmt=$(func_pipeline_parse_value $idx fmt)
+    eq_support=$(func_pipeline_parse_value $idx eq)
+    if [ -z "$eq_support" ]; then
+        dloge "None of FIR/IIF filter is available in this pipeline, skip"
```
libinyang: this should not be an error. dlogi or dlogw is better

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+        continue
+    fi
+    dlogi "eq_support= $eq_support"
+    is_iir=$(echo $eq_support | grep -i iir | wc -l)
+    is_fir=$(echo $eq_support | grep -i fir | wc -l)
+    if [ $is_iir -ne $is_fir ]; then
+        dloge "no FIR or IIF filter is available in this pipeline, skip"
```
libinyang: ditto

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+fi
+
+FIRid=`amixer -D hw:0 controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+if [ -z $FIRid ]; then
+        dloge "can't find FIR filter"
+        exit 1
+fi
+
+
+for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
```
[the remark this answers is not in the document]
singalsu: Please check this PR for this kind of idea. If you think it's useful to pass the EQ response from top level topology with pipeline.m4 go ahead and copy it https://github.com/thesofproject/sof/pull/2434 . It doesn't make sense to make a new eq pipeline macro for every new needed response. All see how I make the EQ control bytes and tokens (also volume tokens) unique for pipeline instance. Without such it's not possible initalize in topology equalizers to different response because the topology.conf file has "flat" hierarchy.

I would like to do this kind of cleanup anyway but I'm pretty busy with bugs so as my task it would take some time.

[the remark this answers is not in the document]
singalsu: Sorry, I thought this was @fredoh9 's EQ topology proposal. My comment belongs there and to EQ test script.

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,146 @@
...
+    fi
+    dlogi "eq_support= $eq_support"
+    is_iir=$(echo $eq_support | grep -i iir | wc -l)
+    is_fir=$(echo $eq_support | grep -i fir | wc -l)
+    if [ $is_iir -ne $is_fir ]; then
+        dloge "no FIR or IIF filter is available in this pipeline, skip"
+        continue
+    fi
+
+    IIRList=(
```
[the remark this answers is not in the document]
fredoh9: I don't think we plan to add new config files in near future. I will work with seppo if he has any plan to do so.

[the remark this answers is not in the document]
fredoh9: @xiulipan How about merging this test as it is now. And I can make this change with querying EQ pipeline from tplg tool as an enhancement.

test-case/check-audio-equalizer.sh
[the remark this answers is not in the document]
fredoh9: @xiulipan i will work with @aiChaoSONG for this new feature.

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,144 @@
...
+    dlogc "aplay -D$dev -f $fmt -c $channel -r $rate /dev/zero -d $duration"
+    aplay -D$dev -f $fmt -c $channel /dev/zero -d $duration
+    if [ $? -ne 0 ]; then
+        dloge "Equalizer test failure with $conf"
+        return 1
+    fi
+    sleep 1
+}
+
+current_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
xiulipan: What is this? What you want to get here? Won't the following code be enough?
```suggestion
current_dir=$(dirname ${BASH_SOURCE[0]})
```

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,144 @@
...
+    FIRid=`amixer -D hw:$sofcard controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+    if [ -z $FIRid ]; then
+            dloge "can't find FIR filter"
+            exit 1
+    fi
+
+    declare -a IIRList=($(ls -d eqctl/eq_iir_*.txt))
+    nIIRList=${#IIRList[@]}
+    dlogi "IIR list $nIIRList ${IIRList[*]}"
+    declare -a FIRList=($(ls -d eqctl/eq_fir_*.txt))
```
xiulipan: Can you use relative path here? 
```suggestion
    declare -a FIRList=($(ls -d ${current_dir}/eqctl/eq_fir_*.txt))
```

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,144 @@
...
+            exit 1
+    fi
+
+    FIRid=`amixer -D hw:$sofcard controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+    if [ -z $FIRid ]; then
+            dloge "can't find FIR filter"
+            exit 1
+    fi
+
+    declare -a IIRList=($(ls -d eqctl/eq_iir_*.txt))
```
xiulipan:
```suggestion
    declare -a IIRList=($(ls -d ${current_dir}/eqctl/eq_iir_*.txt))
```

test-case/check-audio-equalizer.sh
```diff
@@ -0,0 +1,144 @@
...
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+func_opt_parse_option $*
+tplg=${OPT_VALUE_lst['t']}
+duration=${OPT_VALUE_lst['d']}
+loop_cnt=${OPT_VALUE_lst['l']}
+
+func_pipeline_export $tplg "type:playback"
```
xiulipan: @aiChaoSONG @Bin-QA Can we direct export tplg with EQ here?

test-case/check-audio-equalizer.sh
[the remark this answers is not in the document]
xiulipan: Could you work with @Bin-QA @aiChaoSONG to directly get tplg with EQ here. And skip if the the PCM list is empty.

test-case/check-audio-equalizer.sh
fredoh9: @xiulipan EQ test need eq tplg, this is not absolutely required. I can work with @aiChaoSONG @Bin-QA as an enhancement. if you don't have any other comment, please approve this.

xiulipan: OK, can you add TODO and add issue to track the enhancement.
As you said, we need special TPLG for the test. So check if the TPLG support EQ as early as we can is needed.

aiChaoSONG: @xiulipan @fredoh9 our current pipeline filter function in sof-tplgreader.py can only cover pipeline type filtering, this is not good, I will try to refine it, and add more filtering feature.

PR #64, opened 2020-02-01
case-lib: pkill all aplay / arecord pipelines not closed properly at …

stripes416:
…test end

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -48,6 +48,10 @@ function exit()
     # when exit force check the pulseaudio whether disabled
     func_lib_restore_pulseaudio
 
+    # cleanup any aplay / arecord pipelines not closed properly
+    pkill aplay
+    pkill arecord
+
     case $exit_status in
         0)
             dlogi "Test Result: PASS!"


stripes416:
Some tests are ending without ensuring all aplay / arecord processes are properly closed --- which can cause cascading test failures when successive tests can't take control of the devices needed.

case-lib/hijack.sh
```diff
@@ -48,6 +48,10 @@ function exit()
     # when exit force check the pulseaudio whether disabled
     func_lib_restore_pulseaudio
 
+    # cleanup any aplay / arecord pipelines not closed properly
+    pkill aplay
+    pkill arecord
```
Bin-QA: this is OK for me, but I want to know whether the process is not report by this:
```
    # get ps command result as list
    OLD_IFS="$IFS" IFS=$'\n'
    local -a cmd_lst
    local line
    # $$ as current script pid
    # NOTICE: already test with $BASHPID:
    # it can output the same result of $$
    # but it couldn't be store the result to the array
    for line in $(ps --ppid $$ -o pid,cmd --no-header|grep -vE "ps|${BASH_SOURCE[-1]}");
    do
        cmd_lst=( "${cmd_lst[@]}" "$line")
    done
    IFS="$OLD_IFS"
    # now force kill target process which maybe block the script quit
    if [ ${#cmd_lst[@]} -gt 0 ]; then
        dlogw "${BASH_SOURCE[-1]} load exit still have those process exist:"
        for line in "${cmd_lst[@]}"
        do
            # remove '^[:space:]' because IFS change to keep the '^[:space:]' in variable
            line=$(echo $line|xargs)
            dlogw "Catch pid: $line"
            dlogw "Kill cmd:'${line#* }' by kill -9"
            kill -9 ${line%% *}
        done
    fi
```
if it is reported, so it means your `pkill` will not work
if it is not reported, it means this `aplay/arecord` is not create by our script
it is not work as our expect

xiulipan: @stripes416 I agree with @Bin-QA We do have a clean up in exit hijack. Why need to redo a kill here?
I think if kill is not take effect, pkill will also not have any better result.

xiulipan:
@stripes416 @fredoh9 Please do not leave commit message empty but with a very long title.
The commit title should be simple keywords, you can have more detail in your commit message.

stripes416:
> @stripes416 @fredoh9 Please do not leave commit message empty but with a very long title.
> The commit title should be simple keywords, you can have more detail in your commit message.

Didn't notice when I submitted. Will fix.

case-lib/hijack.sh
stripes416: Yeah ... I guess. Ill close this PR and submit a PR that closes at the end of every test. This issue is causing several cascading CI failures.

PR #65, opened 2020-02-05
Test case/add explicit pkill for aplay & arecord

stripes416:
In check-capture.sh & check-playback.sh to prevent further cascading CI failures due to in-use pcm device from previous tests.

### DIFF
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -93,6 +93,7 @@ do
                 if [[ $? -ne 0 ]]; then
                     dmesg > $LOG_ROOT/arecord_error_${dev}_$i.txt
                     dloge "arecord on PCM $dev failed at $i/$loop_cnt."
+                    pkill -9 arecord
                     exit 1
                 fi
                 dmesg > $LOG_ROOT/arecord_${dev}_$i.txt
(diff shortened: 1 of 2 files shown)


stripes416:
fixed a typo.

>stripes416 force-pushed the stripes416:test-case/add_explicit_pkill_for_aplay_arecord branch from bc2ffd1 to e682ce4 2 minutes ago

test-case/check-playback.sh
```diff
@@ -99,5 +100,6 @@ do
     done
 done
 
+pkill -9 aplay
```
keqiaozhang: If the value of $? is 0, then the previous process did terminate normally. It's overkill.

test-case/check-capture.sh
```diff
@@ -101,5 +102,6 @@ do
     done
 done
 
+pkill -9 arecord
```
keqiaozhang: If the value of $? is 0, then the previous process did terminate normally. It's overkill.

test-case/check-playback.sh
stripes416: If its overkill, how are we still causing all the failures?

PR #66, opened 2020-02-05
utilize bidirectional pipeline for both playback and capture

fredoh9:
Both pipeline was used for only playback or capture in multipipe test
cases. To maximize number of pipelines, use it for both playback and
capture.

### DIFF
--- a/test-case/multiple-pipeline-capture.sh
+++ b/test-case/multiple-pipeline-capture.sh
@@ -48,19 +48,16 @@ tplg=${OPT_VALUE_lst['t']}
 func_pipeline_export $tplg
 [[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
 
-max_count=0
 # get the min value of TPLG:'pipeline count' with Case:'pipeline count'
-[[ $PIPELINE_COUNT -gt ${OPT_VALUE_lst['c']} ]] && max_count=${OPT_VALUE_lst['c']} || max_count=$PIPELINE_COUNT
+max_count=${OPT_VALUE_lst['c']}
 func_lib_setup_kernel_last_line
 
 # now small function define
 declare -A APP_LST DEV_LST
-APP_LST['playback']='aplay'
-DEV_LST['playback']='/dev/zero'
+APP_LST['playback,both']='aplay'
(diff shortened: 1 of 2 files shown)


test-case/multiple-pipeline-capture.sh
```diff
@@ -48,20 +48,17 @@ tplg=${OPT_VALUE_lst['t']}
...
 func_lib_setup_kernel_last_line
 
 # now small function define
 declare -A APP_LST DEV_LST
-APP_LST['playback']='aplay'
-DEV_LST['playback']='/dev/zero'
+APP_LST['playback,both']='aplay'
+DEV_LST['playback,both']='/dev/zero'
 APP_LST['capture,both']='arecord'
 DEV_LST['capture,both']='/dev/null'
```
Bin-QA: @fredoh9 I couldn't understand this logic, it seems conflict with case comments
so you will catch the same pipeline, this behavior is right?

test-case/multiple-pipeline-capture.sh
```diff
@@ -137,7 +138,7 @@ do
     tmp_count=$(expr $tmp_count + $pcount)
     pcount=$(pidof aplay|wc -w)
     tmp_count=$(expr $tmp_count + $pcount)
-    [[ $tmp_count -ne $max_count ]] && func_error_exit "Target pipeline count: $max_count, current process count: $tmp_count"
+    [[ $tmp_count -ne $running_count ]] && func_error_exit "Target pipeline count: $running_count, current process count: $tmp_count"
```
Bin-QA: how to detail the "pipeline count" mismatch with parameter, I think here is conflict with test case comment

test-case/multiple-pipeline-capture.sh
fredoh9: @Bin-QA My idea is both pipeline is counted as 1 but actually it's two pipelines. To maximize utilization, we should use both pipeline for both playback and capture. Same answer for below comment too.

xiulipan:
@fredoh9 I will close this PR as this not on our road map for the next stage.
Please check our plan to remove both type here https://github.com/thesofproject/sof-test/issues/71

PR #72, opened 2020-02-10
bugfix of test-case/check-reboot

Bin-QA:
1. check-reboot have more parameter, but just handle the loop count
2. new feature lock have the warning when load sub-test,

### DIFF
--- a/test-case/check-reboot.sh
+++ b/test-case/check-reboot.sh
@@ -41,7 +41,7 @@ current=$[ $count - $loop_count ]
 cur_kern=$(uname -r)
 
 # compare kernel version
-if [ "$(head -n 1 $status_log|awk '{print $2;}')" != "$(uname -r)" ]; then
+if [ "$orig_kern" != "$cur_kern" ]; then
     sed -i '$s/$/ fail/' $status_log
     dloge "Kernel version already been changed"
     exit 1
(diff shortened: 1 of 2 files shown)


test-case/check-reboot.sh
```diff
@@ -112,17 +114,23 @@ else
...
+# add -d delay
+[[ ! $(echo $full_cmd|grep "'-d $delay'") ]] && full_cmd=$(echo $full_cmd|sed "s:$0:& -d $delay:g")
+# add -t timeout
+[[ ! $(echo $full_cmd|grep "'-t $timeout'") ]] && full_cmd=$(echo $full_cmd|sed "s:$0:& -t $timeout:g")
+# add -l loop_count
+[[ ! $(echo $full_cmd|grep "'-l $loop_count'") ]] && full_cmd=$(echo $full_cmd|sed "s:$0:& -l $loop_count:g")
+# now command will like: $0 -l loop_count -t timeout -d delay
 
 # convert relative path to absolute path
 full_cmd=$(echo $full_cmd|sed "s:$0:$(realpath $0):g")
```
xiulipan: Is this duplicated with 
```
# some load will use '~' which is $HOME, but after system reboot, in rc.local $USER is root
# so the '~' will lead to the error path
full_cmd=$(echo $full_cmd|sed "s:~:$HOME:g")
```

Bin-QA: it is not duplicated, because use different load  
1. load as "./check-reboot.sh" or "xxx/check-reboot.sh"
2. if without this change, it will cause mistake, in bash script, "~" is absolute path ..

Bin-QA: > Most part looks good.
> One mirror question inline and could we just make all cmd parse in one place and use a wrapper function for it

which part? I don't think one line command need to add the function, which will been hard to understand  
on the other hand, just reboot need to convert this information, so let keep it  
if have other case need to fix full command, I will add the new function for it

PR #73, opened 2020-02-10
tools: fix rate error on some platform

aiChaoSONG:
Currently, we use rate_min in tplg as rate for aplay and arecord,
but for some platforms, the value of rate_min is 0, and thus
aplay and arecord will report error.

This patch use rate_min as rate for aplay and arecord if it is not
zero, otherwise, use rate_max

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -12,7 +12,7 @@ class clsTPLGReader:
         self._filed_lst = []
         self._filter_lst = []
         self._block_lst = []
-    
+
     def __comp_pipeline(self, pipeline):
         return pipeline['id']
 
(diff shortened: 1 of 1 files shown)


tools/sof-tplgreader.py
```diff
@@ -51,7 +51,7 @@ def loadFile(self, filename, sofcard=0):
         # format pipeline, this change for script direct access 'rate' 'channel' 'dev'
         for pipeline in self._pipeline_lst:
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
-            pipeline['rate'] = pipeline['rate_min']
+            pipeline['rate'] = pipeline['rate_max']
```
Bin-QA: I'm not sure which is better, rate_min/rate_max  
but from your comment, can you try to judgement the `rate_min` value?
So if it is 0, you setup `rate` as `rate_max`

aiChaoSONG: This is a good idea.

aiChaoSONG:
@Bin-QA that doesn't matter, because rate_min and rate_max is not the reason we got some error.

tools/sof-tplgreader.py
```diff
@@ -51,7 +51,7 @@ def loadFile(self, filename, sofcard=0):
         # format pipeline, this change for script direct access 'rate' 'channel' 'dev'
         for pipeline in self._pipeline_lst:
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
-            pipeline['rate'] = pipeline['rate_min']
+            pipeline['rate'] = pipeline['rate_min'] if pipeline['rate_min'] != 0 else pipeline['rate_max']
```
Bin-QA: I remember pipeline['rate_min'] type is string

PR #75, opened 2020-02-11
test-case: check dsp status before module removal

aiChaoSONG:
On some platforms, after module insertion, it will take
sometime for dsp to become suspended. If sleep time is too
short, we will remove module when dsp is active, this causes
error.

This patch checks dsp status before module removal.

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -50,6 +50,26 @@ do
     ## - 1: remove module section
     func_lib_setup_kernel_last_line
 
+    # 1. check dsp status before removal, only remove module when dsp is suspended.
+    # 2. it will take some time (about 10s) for dsp to become suspended from active
+    # after module insertion on some platforms, here we check dsp status, and only continue
+    # next removal when dsp is suspended.
+    dlogi "wait dsp power status to become suspended"
+    declare I=0 # used to check if for loop is early exit
+    for i in $(seq 1 15)
+    do
+        # Here we pass a hardcoded 0 to python script, and need to ensure
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        let I=i
+        sleep 1
+    done
+    if [ $I -eq 15 ]; then
+        dlogi "dsp is not suspended after 15s, end test"
+        exit 1
+    fi
+
     dlogi "run kmod/sof-kmod-remove.sh"
     sudo sof_remove.sh
     [[ $? -ne 0 ]] && dloge "remove modules error" && exit 1
@@ -79,10 +99,10 @@ do
 
     # successful remove/insert module pass
     dlogi "==== completed boot firmware: $idx of $loop_cnt ===="
+
     # pulseaudio deamon will detect the snd_sof_pci device after 3s
     # so after 2s snd_sof_pci device will in used status which is block current case logic
     # here the logic is to check snd_sof_pci status is not in used status, the max delay is 10s
-    sleep 1
     for i in $(seq 1 10)
     do
         [[ "X$(awk '/^snd_sof_pci/ {print $3;}' /proc/modules)" == "X0" ]] && break
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -303,6 +303,7 @@ if __name__ == "__main__":
     parser.add_argument('-s', '--short', type=int, help='just dump the short name of target id sound card')
     parser.add_argument('-l', '--longname', type=int, help='just dump the longname name of target id sound card')
     parser.add_argument('-P', '--fwpath', action='store_true', help='get firmware path according to DMI info')
+    parser.add_argument('-S', '--dsp_status', type=int, help='get current dsp power status, should specify sof card number')
     parser.add_argument('--version', action='version', version='%(prog)s 1.0')
 
     ret_args = vars(parser.parse_args())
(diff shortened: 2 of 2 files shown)


aiChaoSONG:
@Bin-QA we do have a way to check dsp status, repushed

test-case/check-kmod-load-unload.sh
```diff
@@ -50,6 +50,16 @@ do
     ## - 1: remove module section
     func_lib_setup_kernel_last_line
 
+    # 1. check dsp status before removal, only remove module when dsp is suspended.
+    # 2. it will take some time (about 10s) for dsp to become suspended from active
+    # after module insertion on some platforms, here we check dsp status, and only continue
+    # next removal when dsp is suspended.
+    for i in $(seq 1 15)
+    do
+        [[ $(cat /sys/devices/pci0000:00/0000:00:1f.3/power/runtime_status) == "suspended" ]] && break
```
Bin-QA: here you device id is hard code, i think you can catch it from `sof-dump-status.py`   
https://github.com/thesofproject/sof-test/blob/master/tools/sof-dump-status.py#L168
```
        for run_status in self.sys_power['run_status']:
            exit_code, output=subprocess.getstatusoutput("cat /sys/bus/pci/devices/0000:%s/power/runtime_status" % (run_status['pci_id']))
            if exit_code != 0:
                continue
            run_status['status'] = output
```
so here you can add option for it

aiChaoSONG: @Bin-QA you are right, the pci id is not unique on different platforms. has added an option in sof-dump-status.py, and modified module reload case to call sof-dump-status.py to get current dsp status.

tools/sof-dump-status.py
```diff
@@ -344,6 +345,11 @@ def _covertType(type_lst):
         print(card_info['longname'])
         exit(0)
 
+    if ret_args.get('dsp_status') is True:
+        sysinfo.loadPower()
+        print(sysinfo.sys_power['run_status'][0]['status'])
```
Bin-QA: how about `print(sysinfo.sys_power['run_status'][ret_args['S']]['status'])`
```
parser.add_argument('-S', '--dsp_status', type=int, help='get target id dsp power status')
```

test-case/check-kmod-load-unload.sh
```diff
@@ -50,6 +50,17 @@ do
...
+    # 2. it will take some time (about 10s) for dsp to become suspended from active
+    # after module insertion on some platforms, here we check dsp status, and only continue
+    # next removal when dsp is suspended.
+    dlogi "wait dsp power status to become suspended"
+    for i in $(seq 1 15)
+    do
+        [[ $(sof-dump-status.py --dsp_status) == "suspended" ]] && break
+        sleep 1
+    done
+
```
Bin-QA: so here you couldn't know it is out of define data or it just meet the target  
if out of data means it is not work as our expect, it is the bug?

tools/sof-dump-status.py
aiChaoSONG: sysinfo.sys_power['run_status'] is a list, and I think it always contains one element, is this true?, so i use index 0.
change argument type to int is error prone, because the user never knows which int number should given.

test-case/check-kmod-load-unload.sh
aiChaoSONG: @Bin-QA I understand your concern. as we give a very big number to wait, here 15s, it should be big enough for dsp to be suspended(if it is in normal state, no previous failure). that means, the "out of data" condition never happens.

we can check the exit condition, too, if out of data , tag failed and exit test, but seems it never happens.

tools/sof-dump-status.py
Bin-QA: you can load from case like by this `$SOFCARD`  
because we already meet some patch from kernel change this value from 0 to 1  
you can catch information from [config.sh](https://github.com/thesofproject/sof-test/blob/master/case-lib/config.sh#L38) and [lib.sh](https://github.com/thesofproject/sof-test/blob/master/case-lib/lib.sh)
```
# setup SOFCARD id
if [ ! "$SOFCARD" ];then
    SOFCARD=$(grep '\]: sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
fi
```
On the other hand, we have more than one sound card, so ...

aiChaoSONG: modified with SOFCARD,

test-case/check-kmod-load-unload.sh
aiChaoSONG: check added, if dsp is always active, the script will failed. please review again @xiulipan @Bin-QA

test-case/check-kmod-load-unload.sh
```diff
@@ -50,6 +50,23 @@ do
...
 
+    # 1. check dsp status before removal, only remove module when dsp is suspended.
+    # 2. it will take some time (about 10s) for dsp to become suspended from active
+    # after module insertion on some platforms, here we check dsp status, and only continue
+    # next removal when dsp is suspended.
+    dlogi "wait dsp power status to become suspended"
+    declare I=0 # used to check if for loop is early exit
+    for i in $(seq 1 15)
+    do
+        [[ $(sof-dump-status.py --dsp_status $SOFCARD) == "suspended" ]] && break
```
Bin-QA: I find come problem, here is `lspci` order  
so don't need to use `SOFCARD`

so need to confirm `lspci` refer audio order
here you can direct use 0 instead of `SOFCARD` add the comment at here

aiChaoSONG: done @Bin-QA

xiulipan:
@aiChaoSONG one thing with this patch, will this break the device that do not support runtime pm?
EG. bdw?

aiChaoSONG:
@xiulipan In fact, I am not so sure. it may break the device that do not support runtime pm, because this patch need to wait the dsp status become suspended, then do module unload

aiChaoSONG:
> @aiChaoSONG I will merge this PR then have some test.

@xiulipan  Fine, if we got some problem, I can add a option "-g" for legacy platform that not support runtime pm, then we can skip this code for these platforms. this pr is necessary for module unload and reload on helios, we should keep it.

PR #78, opened 2020-02-14
test-case: check-suspend-resume-with-audio add more dump

Bin-QA:
Current meet some DUT catch lost process the error
Add ps command to dump the process status

This change not refer the code logic

### DIFF
--- a/test-case/check-suspend-resume-with-audio.sh
+++ b/test-case/check-suspend-resume-with-audio.sh
@@ -98,13 +98,27 @@ do
     sleep 1
     # check process status is correct
     sof-process-state.sh $process_id
-    [[ $? -ne 0 ]] && dloge "$cmd error without process" && exit 1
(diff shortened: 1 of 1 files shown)


test-case/check-suspend-resume-with-audio.sh
```diff
@@ -98,13 +98,27 @@ do
     sleep 1
     # check process status is correct
     sof-process-state.sh $process_id
-    [[ $? -ne 0 ]] && dloge "$cmd error without process" && exit 1
+    if [ $? -ne 0 ]; then
+        dloge "$cmd error without process"
```
aiChaoSONG: `$cmd error without process`, this info is misleading, do you mean : we run $cmd, but there is no such process named $cmd? if so, you can change this info to :: `error: no such process named $cmd` or `error: no such process for $cmd` or something like this.

PR #80, opened 2020-02-19
sof-test: split both type pipeline into playback and capture

aiChaoSONG:
1. split both type pipeline into playback and capture pipelines
2.  remove "both" in test cases

@Bin-QA Wubin will modify the simultaneous-playback-capture.sh test case, and add logic to extract "both" type pipeline.

***DO NOT MERGE BEFORE PR #10 ***

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -23,21 +23,33 @@ class clsTPLGReader:
     def loadFile(self, filename, sofcard=0):
         tplg_parser = TplgParser()
         parsed_tplg = tplg_parser.parse(filename)
+        formatter = TplgFormatter(parsed_tplg)
         # ignore the last element, for it is tplg name
         for item in parsed_tplg[:-1]:
             if "pcm" not in item:
                 continue
             for pcm in item['pcm']:
+                pcm_type = TplgFormatter.get_pcm_type(pcm)
+                # if we find None type pcm, there must be errors in topology
+                if pcm_type == "None":
+                    print("type of %s is neither playback nor capture, please check your"
+                        "topology source file" % pcm["pcm_name"])
+                    exit(1)
+                pgas = formatter.find_comp_for_pcm(pcm, 'PGA')
+                eqs = formatter.find_comp_for_pcm(pcm, 'EQ')
                 pipeline_dict = {}
                 pipeline_dict['pcm'] = pcm["pcm_name"]
                 pipeline_dict['id'] = str(pcm["pcm_id"])
                 pipeline_dict['type'] = TplgFormatter.get_pcm_type(pcm)
-                # if we find None type pcm, there must be errors in topology
-                if pipeline_dict['type'] == "None":
-                    print("type of %s is neither playback nor capture, please check your"
-                        "topology source file" % pcm["pcm_name"])
-                    exit(1)
                 cap = pcm["caps"][pcm['capture']]
+                pga = pgas[pcm['capture']]
+                if pga != []:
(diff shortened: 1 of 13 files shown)


tools/sof-tplgreader.py
```diff
@@ -46,6 +59,25 @@ def loadFile(self, filename, sofcard=0):
                 pipeline_dict['fmt'] = pipeline_dict['fmts'].split(' ')[0]
                 pipeline_dict['rate_min'], pipeline_dict['rate_max'] = self._key2str(cap, 'rate')
                 pipeline_dict['ch_min'], pipeline_dict['ch_max'] = self._key2str(cap, 'channels')
+                if pcm_type == "both":
+                    pipeline_dict["type"] = "capture"
+                    # copy pipeline and change values
+                    pb_pipeline_dict = copy.copy(pipeline_dict)
+                    pb_pipeline_dict["type"] = "playback"
+                    cap = pcm["caps"][pcm['playback']]
+                    pga = pgas[pcm['capture']]
```
Bin-QA: Different value between 'playback' and 'capture' ?

tools/sof-tplgreader.py
```diff
@@ -46,6 +59,25 @@ def loadFile(self, filename, sofcard=0):
                 pipeline_dict['fmt'] = pipeline_dict['fmts'].split(' ')[0]
                 pipeline_dict['rate_min'], pipeline_dict['rate_max'] = self._key2str(cap, 'rate')
                 pipeline_dict['ch_min'], pipeline_dict['ch_max'] = self._key2str(cap, 'channels')
+                if pcm_type == "both":
+                    pipeline_dict["type"] = "capture"
+                    # copy pipeline and change values
+                    pb_pipeline_dict = copy.copy(pipeline_dict)
```
Bin-QA: I saw you rebuild the 'pga' and 'eq' don't need deep copy, you can direct to use `pipeline_dict.copy`

tools/sof-tplgreader.py
```diff
@@ -46,6 +59,25 @@ def loadFile(self, filename, sofcard=0):
...
+                        pga_names = [i['name'] for i in pga]
+                        pipeline_dict['pga'] = " ".join(pga_names)
+                    eq = eqs[pcm['capture']]
+                    if eq != []:
+                        eq_names = [i['name'] for i in eq]
+                        pipeline_dict['eq'] = " ".join(eq_names)
+                    pb_pipeline_dict["fmts"] = " ".join(formats[pcm['playback']])
+                    pipeline_dict['fmt'] = pipeline_dict['fmts'].split(' ')[0]
+                    pb_pipeline_dict['rate_min'], pb_pipeline_dict['rate_max'] = self._key2str(cap, 'rate')
+                    pb_pipeline_dict['ch_min'], pb_pipeline_dict['ch_max'] = self._key2str(cap, 'channels')
```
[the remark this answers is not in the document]
aiChaoSONG: @Bin-QA  in fact, these values for playback and capture are from their respectively caps structure under pcm structure. values under pcm structure is the common part, but for values under caps structure, we should use their own values, through they may be the same value currently. I mean, they may not be the same in the future.

tools/sof-tplgreader.py
aiChaoSONG: yes, PGA and EQ is total the different value for playback and capture

tools/sof-tplgreader.py
aiChaoSONG: sure, will refine in next push

PR #81, opened 2020-02-19
test-case: simultaneous-playback-capture.sh to remove 'both' keyword

Bin-QA:
Need to split both type pipeline into playback and capture
Change both keyword into check duplicate pcm id

Wait patch [sof-test: split both type pipeline into playback and capture](https://github.com/thesofproject/sof-test/pull/80 ) can be merge this PR

Because without that change, this pr couldn't worked

### DIFF
--- a/test-case/simultaneous-playback-capture.sh
+++ b/test-case/simultaneous-playback-capture.sh
@@ -37,7 +37,23 @@ tplg=${OPT_VALUE_lst['t']}
 wait_time=${OPT_VALUE_lst['w']}
 loop_cnt=${OPT_VALUE_lst['l']}
 
-func_pipeline_export $tplg "type:both"
+# get 'both' pcm, it means pcm have same id with different type
+declare -A tmp_id_lst
+id_lst_str=""
+[[ -f $TPLG_ROOT/$tplg ]] && tmp_tplg=$TPLG_ROOT/$tplg || tmp_tplg=$tplg
+for i in $(sof-tplgreader.py $tmp_tplg -d id -v)
+do
+    if [ ! "${tmp_id_lst["$i"]}" ]; then  # this id is never used
+        tmp_id_lst["$i"]=0
+    else # this id already used
+        tmp_id_lst["$i"]=1
+        id_lst_str="$id_lst_str,$i"
+    fi
+done
+# now all duplicate ids have already been caught
+unset tmp_id_lst tmp_tplg
+id_lst_str=${id_lst_str/,/} # remove 1st, which is not used
+func_pipeline_export $tplg "id:$id_lst_str"
 [[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
 func_lib_setup_kernel_last_line
 
@@ -54,8 +70,10 @@ do
     dlogi "Testing: (Loop: $i/$loop_cnt)"
     # clean up dmesg
     sudo dmesg -C
-    for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+    # following sof-tplgreader, split 'both' pipelines into separate playback & capture pipelines, with playback occurring first
+    for order in $(seq 0 2 $(expr $PIPELINE_COUNT - 1))
     do
+        idx=$order
         channel=$(func_pipeline_parse_value $idx channel)
         rate=$(func_pipeline_parse_value $idx rate)
         fmt=$(func_pipeline_parse_value $idx fmt)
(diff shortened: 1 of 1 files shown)


stripes416:
Okay, so we need to wait until #80 (which is also dependent on #10) is merged before merging this PR?

test-case/simultaneous-playback-capture.sh
```diff
@@ -37,7 +37,23 @@ tplg=${OPT_VALUE_lst['t']}
 wait_time=${OPT_VALUE_lst['w']}
 loop_cnt=${OPT_VALUE_lst['l']}
 
-func_pipeline_export $tplg "type:both"
+
```
stripes416: lets remove this extra line here

test-case/simultaneous-playback-capture.sh
```diff
@@ -37,7 +37,23 @@ tplg=${OPT_VALUE_lst['t']}
...
+# get 'both' pcm, it means pcm have same id with different type
+declare -A tmp_id_lst
+id_lst_str=""
+for i in $(sof-tplgreader.py -t $TPLG_ROOT $TPLG -d id -v)
+do
+    if [ ! "${tmp_id_lst["$i"]}" ];then  # this id is never used
```
stripes416:
```suggestion
    if [ ! "${tmp_id_lst["$i"]}" ]; then  # this id is never used
```

test-case/simultaneous-playback-capture.sh
```diff
@@ -37,7 +37,23 @@ tplg=${OPT_VALUE_lst['t']}
...
+        tmp_id_lst["$i"]=0
+    else # this id already used
+        tmp_id_lst["$i"]=1
+        id_lst_str="$id_lst_str,$i"
+    fi
+done
+# now all duplicate id already been catched
```
stripes416:
```suggestion
# now all duplicate ids have already been caught
```

test-case/simultaneous-playback-capture.sh
```diff
@@ -54,8 +70,10 @@ do
     dlogi "Testing: (Loop: $i/$loop_cnt)"
     # clean up dmesg
     sudo dmesg -C
-    for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+    # follow sof-tplgreader the capture will after playback
```
stripes416: I'm not sure what you're trying to say here. Can you expand on this a bit?

Bin-QA: @stripes416 check with patch #80 the both will split 2 pipeline the order of those pipeline is 1st is playback, 2nd is capture, so here step change to 2

stripes416: Okay, I think I understand what you're trying to say here now. Maybe change the comment to something like below:

```suggestion
    # following sof-tplgreader, split 'both' pipelines into separate playback & capture pipelines, with playback occurring first
```

xiulipan:
@stripes416 @aiChaoSONG Please help to review and approve as all dependency is merged.

aiChaoSONG:
@Bin-QA still need some fix. hw:0,1, hw:0,2 shouldn't be present
```
2020-02-25 08:26:23 UTC [COMMAND] aplay -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/zero -q &
2020-02-25 08:26:23 UTC [COMMAND] arecord -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/null -q &
2020-02-25 08:26:23 UTC [INFO] Preparing to sleep for 5
2020-02-25 08:26:28 UTC [INFO] check pipeline after 5s
2020-02-25 08:26:28 UTC [COMMAND] killing all pipelines
2020-02-25 08:26:28 UTC [COMMAND] aplay -D hw:0,1 -c 4 -r 48000 -f S32_LE /dev/zero -q &
2020-02-25 08:26:28 UTC [COMMAND] arecord -D hw:0,2 -c 2 -r 48000 -f S16_LE /dev/null -q &
aplay: main:788: audio open error: No such file or directory
2020-02-25 08:26:28 UTC [INFO] Preparing to sleep for 5
arecord: main:788: audio open error: No such file or directory
2020-02-25 08:26:33 UTC [INFO] check pipeline after 5s
...
2020-02-25 08:26:35 UTC [INFO] Test Result: FAIL!
```

Bin-QA:
> @Bin-QA still need some fix. hw:0,1, hw:0,2 shouldn't be present

The root cause is by typo, because we current use `func_pipeline_export`
which will help to check tplg file name, here use the error path & tplg, so the filter is not work  
already update to fix it

PR #82, opened 2020-02-20
test-case: check dsp status before/after module removal/insertion

aiChaoSONG:
On helios, after module removal, `aplay -l` will report no soundcard found, we need to wait about 10s to resume.  This may be an error, and I report it here: https://github.com/thesofproject/linux/issues/1808. 

we should also check dsp status after module insertion and next check-playback for check-kmod-unload-reload-after-playback.sh

for platform support runtime pm, this is a safe and good practice. impact unknown for platform not support runtime pm.

### DIFF
--- a/test-case/check-kmod-load-unload-after-playback.sh
+++ b/test-case/check-kmod-load-unload-after-playback.sh
@@ -77,6 +77,22 @@ do
     ret=$?
     [[ $ret -ne 0 ]] && dloge "kmod reload failed" && exit $ret
 
+    dlogi "wait dsp power status to become suspended"
+    for i in $(seq 1 15)
+    do
+        # Here we pass a hardcoded 0 to python script, and need to ensure
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "unsupported" ]] &&
+            dlogi "platform doesn't support runtime pm, skip waiting" && break
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        sleep 1
+        if [ $i -eq 15 ]; then
+            dlogi "dsp is not suspended after 15s, end test"
+            exit 1
+        fi
+    done
+
     $(dirname ${BASH_SOURCE[0]})/check-playback.sh -l 1 -t $tplg -d $pb_duration
     ret=$?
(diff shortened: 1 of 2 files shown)


test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -77,6 +77,22 @@ do
...
+    for i in $(seq 1 15)
+    do
+        # Here we pass a hardcoded 0 to python script, and need to ensure
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        let I=i
+        sleep 1
+    done
+    if [ $I -eq 15 ]; then
```
Bin-QA: I think you can direct use `i` because you are not release it

aiChaoSONG:
@Bin-QA #75 #72 is the same, but I am not sure when your PR will be merged. without the change, this test cannot run.

stripes416:
Once the change @Bin-QA requested is made, I'll be good to approve.

test-case/check-kmod-load-unload-after-playback.sh
aiChaoSONG: I thought it is just like C language. thanks for the suggestion. I move the if sentence into for loop. this looks better.

xiulipan:
@Bin-QA Please help to change the `sof-dump-status.py` to support the ACPI from BYT and BDW

aiChaoSONG:
@Bin-QA Please make sure sof-dump-status.py can output unsupported on BYT and BDW, than I can add a condition test in our test case like this:
```
[[ $(sof-dump-status.py --dsp_status 0) == "unsupported" ]] && break
```
when unsupported, exit wait immediately.

xiulipan:
@aiChaoSONG Could you test after the https://github.com/thesofproject/sof-test/pull/84 mered
PS: could you also change the commit? This is not workaround at all, this is needed status check for test env.

aiChaoSONG:
> @aiChaoSONG Could you test after the #84 mered
> PS: could you also change the commit? This is not workaround at all, this is needed status check for test env.

Changed the title and commit message, apply patch on master and run check-kmod-load-unload.sh and check-kmod-load-unload-after-playback.sh on helios for 5 times, both Passed without error.
If we run these two case remotely, we need to make timeout larger, because this patch add some wait time.

Issue #83, opened 2020-02-21
[bugs][tools]sof-dump-status do not full support BYT/CHT/BDW


xiulipan:
The sof-dump-status.py is not finding the right DSP device for the BYT, CHT and BDW.
Need to use ACPI device for those platforms.

Bin-QA:
check with source code:  [sof-acpi-dev.c](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-acpi-dev.c#L194)

Catch the keyword, how to deal with the PCI information for 'BYT', 'CHT' and 'BDW'
current template fix for the power status

stripes416:
@xiulipan @Bin-QA can we close this issue now that #84 has been merged?

aiChaoSONG:
close for BYT/CHT/BDW supported

PR #84, opened 2020-02-21
tools: sof-dump-status to add acpi device power detect

Bin-QA:
Some device use acpi to get power status instead of load from
pci device folder

1. add device id for output to know which platform use acpi device
2. query acpi power status & pci power status

fix issue #83


aiChaoSONG:
@Bin-QA I tried this patch on BDW, doesn't work, but on BYT, it worked. Please check the patch again.

xiulipan:
@Bin-QA @aiChaoSONG Please help to update the test result for this PR.

PR #85, opened 2020-02-24
test-case: update comment to match functionality

stripes416:
Update comment in check-pause-resume.sh. Comment now follows program
functionality.

### DIFF
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -91,14 +91,14 @@ set i 1
 expect {
     "*#*+*\%" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
-        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before pause"
+        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms after pause"
         send " "
         after \$sleep_t
         exp_continue
     }
     "*PAUSE*" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
-        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before resume"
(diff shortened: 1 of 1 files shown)


test-case/check-pause-resume.sh
```diff
@@ -91,14 +91,14 @@ set i 1
 expect {
     "*#*+*\%" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
-        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before pause"
+        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms after pause"
```
xiulipan: @stripes416 I think you may confuse with the logic here.
At the moment the log output, we are playing and wait the the time. Then we inject space to enter pause.
So the log here is right to show before pause.

xiulipan:
@stripes416 I will close this PR. You seems to change it wrongly.

test-case/check-pause-resume.sh
stripes416: @xiulipan are you sure? because the `send " "` is what actually pauses the playback and `after `$sleep_t` is what causes the sleep period. We're pausing before we sleep.

test-case/check-pause-resume.sh
```diff
@@ -91,14 +91,14 @@ set i 1
...
         send " "
         after \$sleep_t
         exp_continue
     }
     "*PAUSE*" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
-        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before resume"
+        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms after resume"
```
stripes416: @xiulipan same here --- `send " "` first to unpause then `after \$sleep_t` starts the sleep period

PR #92, opened 2020-02-25
test-case: check aplay/arecord before and after runtime pm

fredoh9:
Before and after runtime pm state transition, check aplay or arecord is
working. The delay is chose from range of 3 to 15 seconds. When the delay
is long enough to have 'suspended' state, then check for the state as well.

Add capture pipeline support for runtime pm test

aplay/arecord can be used by setting -m option. Add scanning dmesg before
switching pipeline.

For reviewers, initially I wanted to merge this two test cases with new input parameter, but it bacame too ugly to read. One more loop and multiple checking mode etc.

### DIFF
--- a/test-case/check-runtime-pm-double-active.sh
+++ b/test-case/check-runtime-pm-double-active.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+
+##
+## Case Name: check-runtime-pm-double-active
+## Preconditions:
+##    N/A
+## Description:
+##    check the audio runtime pm status and make runtime pm status active twice
+##    playback/capture -> stop -> wait till runtime pm get suspended -> playback/capture again
+## Case step:
+##    1. start aplay/arecord
+##    2. stop aplay/arecord
+##    3. keep polling on runtime pm status till 'suspended'
+##    4. when 'suspended', check playback/capture again immediately
+## Expect result:
+##    command line check with $? without error
+##    runtime pm status must be suspended
+##    playback/capture must always work in right after runtime pm is suspended
+##    no error in dmesg
+##
+
+source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=3
+
+OPT_OPT_lst['d']='delay'    OPT_DESC_lst['d']='max delay time for state convert'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=15
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+# param: $1 -> max delay time for dsp pm status switch, unit is second
+func_check_dsp_status()
+{
+    # more frequent checking of the runtime pm status is required. Every loop is 100ms
+    local iloop=$(expr 10 '*' "$1")
+    dlogi "wait dsp power status to become suspended, starting max [100ms x $iloop times] checking"
+    for i in $(seq 1 $iloop)
+    do
+        # Here we pass a hardcoded 0 to python script, and need to ensure
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        sleep 0.1
+    done
+
+    if [ $i -eq $iloop ]; then
+        dlogi "dsp is not suspended after $1s, end test"
+        exit 1
+    else
+        dlogi "dsp suspended in $i try, ${i}00 msec"
+    fi
+}
+
+func_opt_parse_option $*
+tplg=${OPT_VALUE_lst['t']}
+loop_count=${OPT_VALUE_lst['l']}
+[[ -z $tplg ]] && dloge "Miss tplg file to run" && exit 2
+
+[[ $(sof-dump-status.py --dsp_status 0) == "unsupported" ]] &&
+    dlogi "platform doesn't support runtime pm, skip test case" && exit 2
+
+declare -A APP_LST DEV_LST
+APP_LST['playback']='aplay'
+DEV_LST['playback']='/dev/zero'
+APP_LST['capture']='arecord'
+DEV_LST['capture']='/dev/null'
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+func_pipeline_export $tplg
+func_lib_setup_kernel_last_line
+
+for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+do
+    channel=$(func_pipeline_parse_value $idx channel)
+    rate=$(func_pipeline_parse_value $idx rate)
+    fmt=$(func_pipeline_parse_value $idx fmt)
+    dev=$(func_pipeline_parse_value $idx dev)
+    pcm=$(func_pipeline_parse_value $idx pcm)
+    type=$(func_pipeline_parse_value $idx type)
+    snd=$(func_pipeline_parse_value $idx snd)
+
+    cmd="${APP_LST[$type]}"
+    dummy_file="${DEV_LST[$type]}"
+    [[ -z $cmd ]] && dloge "$type is not supported, $cmd, $dummy_file" && exit 2
+
+    # clean up dmesg before the test
+    sudo dmesg -C
+    for i in $(seq 1 $loop_count)
+    do
+        dlogi "Iteration $i of $loop_count for $pcm"
+        # playback or capture device - check status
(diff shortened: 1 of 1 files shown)


xiulipan:
@fredoh9 What is the difference with the `check-runtime-pm-status.sh`. They both try to check if the FW can work normally when runtime pm happen.

fredoh9:
@xiulipan Per pipeline,
`check-runtime-pm-status.sh`
aplay -> atop stop -> sleep for enough time -> check runtime pm status

`check-runtime-pm-status-boundary.sh`
aplay(or arecord) -> atop stop -> sleep range of [3 5 7 10 15] -> check runtime pm status if resired time is passed -> check aplay(or arecord) 

New script will especially check if FW can work at boundary of runtime pm.

test-case/check-runtime-pm-boundary.sh
```diff
@@ -0,0 +1,165 @@
...
+##
+## Case Name: check-runtime-pm-boundary
+## Preconditions:
+##    N/A
+## Description:
+##    check the audio runtime pm status with range of sleep [3 5 7 10 15]
+## Case step:
+##    1. start aplay/arecord
+##    2. stop aplay/arecord
+##    3. sleep of [3 5 7 10 15]
```
stripes416: This is a sleep period in seconds right?

test-case/check-runtime-pm-boundary.sh
```diff
@@ -0,0 +1,165 @@
...
+                $cmd -D $dev -r $rate -c $channel -f $fmt $dummy_file -q &
+                pid=$!
+
+                sleep 1
+
+                kill -0 $pid
+                [[ $? -ne 0 ]] && dloge "$cmd process for pcm $pcm is not alive" && exit 1
+
+                kill -9 $pid && wait $pid 2>/dev/null
+                dlogi "$cmd end"
```
Bin-QA: I don't think this operation is different with case: [check-runtime-pm-status](https://github.com/thesofproject/sof-test/blob/master/test-case/check-runtime-pm-status.sh)  

you can see when you run this logic in the loop, so next round will run the cmd again  
So I will suggest you add the **loop** option for the original case instead of this change  
if `$i` try to run and trigger the error at here, I think original case will trigger at `$i + 1` at begin  
when add **loop** option

test-case/check-runtime-pm-boundary.sh
```diff
@@ -0,0 +1,165 @@
...
+
+    #runtime pm status check at end the end of each pipeline test
+    dlogc "sleep $state_delay at the end"
+    sleep $state_delay
+    result=`cat $runtime_status`
+    dlogi "runtime status: $result"
+    if [[ $result != suspended ]]; then
+        dloge "$cmd process for pcm $pcm runtime status is not suspended as expected"
+        exit 1
+    fi
```
Bin-QA: the original case main logic for the loop
```
loop begin
$cmd # to run cmd
check $runtime_status
is active
    stop $cmd
    check $runtime_status
    is suspended
loop end
```
your new case logic:
```
loop begin
$cmd # to run cmd
check $runtime_status
is active
    stop $cmd
    check $runtime_status
    is suspended
    $cmd # to run cmd
loop end
is suspended
```

So I couldn't see the different between the original 
Because those check will cover by the loop
so if you want check `suspended` you can add at original loop begin  

The different between this case with original, I think this case will catch at `$i`, the original will catch at `$i + 1`

fredoh9: @Bin-QA the main purpose of this test is check aplay/arecord again just before/after runtime pm, having said that probably sleep [3 5 7] are most important delay time to start playback or capture again. There was known issue, it didn't work before.

Adding this logic to known runtime pm test case, it looks to me messy.

Bin-QA: how about give the random time range and increase the loop

xiulipan:
@fredoh9 I agreed with @Bin-QA I did not see any difference of code logic run with https://github.com/thesofproject/sof-test/blob/master/test-case/check-runtime-pm-status.sh
The PM status check also include check DSP is active when pipeline run and check DSP is suspend when pipeline idle for sometime. What will be difference when you wait for [3 5 7] here?

keqiaozhang:
@fredoh9 
This is a corner case, when sof is idle, runtime_status will be changed like this: active->suspending->suspended, so I think "suspending" status is the boundary you want to test.
You can test it like this:
```
while true
do
status=`tail $runtime_status`
echo $status
if [ $status == 'suspending' ]
then
        run aplay/arecord
        check error
fi
done
```

fredoh9:
@keqiaozhang I have a plan to re-visit this test case today. Will update again and remove WIP.

fredoh9:
@xiulipan @keqiaozhang @Bin-QA @libinyang 
I rebased to latest master and updated this again. this is highlighted changes.

1. Need more frequent check of the variable. Updated polling time to 100ms
2. When runtime pm status get suspended, trigger playback/capture again with duration of 1 second
4. commit text changed, file name also is changed

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+            exit 1
+        fi
+    done
+    dlogi "dsp suspended in $i try, ${i}00 msec"
+}
+
+func_opt_parse_option $*
+tplg=${OPT_VALUE_lst['t']}
+loop_count=${OPT_VALUE_lst['l']}
+[[ -z $tplg ]] && dloge "Miss tplg file to run" && exit 1
```
Bin-QA: exit 2

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+    rate=$(func_pipeline_parse_value $idx rate)
+    fmt=$(func_pipeline_parse_value $idx fmt)
+    dev=$(func_pipeline_parse_value $idx dev)
+    pcm=$(func_pipeline_parse_value $idx pcm)
+    type=$(func_pipeline_parse_value $idx type)
+    snd=$(func_pipeline_parse_value $idx snd)
+
+    cmd="${APP_LST[$type]}"
+    dummy_file="${DEV_LST[$type]}"
+    [[ -z $cmd ]] && dloge "$type is not supported, $cmd, $dummy_file" && exit 1
```
Bin-QA: exit 2

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+        dlogc "Rechecking: $cmd -D $dev -r $rate -c $channel -f $fmt $dummy_file -d 1 -q"
+        $cmd -D $dev -r $rate -c $channel -f $fmt $dummy_file -d 1 -q
+        if [[ $? -ne 0 ]]; then
+            func_lib_lsof_error_dump $snd
+            dloge "playback/capture failed on $pcm, $dev at $i/$loop_cnt."
+            exit 1
+        fi
+    done
+
+    sof-kernel-log-check.sh 0
```
Bin-QA: with `sof-kernel-log-check.sh 0` you also need `sudo dmesg -C` before loop

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        sleep 0.1
+        if [ $i -eq $iloop ]; then
+            dlogi "dsp is not suspended after $1s, end test"
+            exit 1
+        fi
+    done
+    dlogi "dsp suspended in $i try, ${i}00 msec"
```
Bin-QA: how about use `return` instead of `break`
then after loop, you can do the error process

libinyang: @Bin-QA Do you mean `exit` or `break`

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+        [[ -d /proc/$pid ]] && result=`sof-dump-status.py --dsp_status 0`
+
+        dlogi "runtime status: $result"
+        if [[ $result == active ]]; then
+            # stop playback or capture device - check status again
+            dlogc "kill process: kill -9 $pid"
+            kill -9 $pid && wait $pid 2>/dev/null
+            dlogi "$cmd killed"
+            func_check_dsp_status ${OPT_VALUE_lst['d']}
+            result=`sof-dump-status.py --dsp_status 0`
```
libinyang: `func_check_dsp_status` has already make sure status is "suspend", otherwise it will exit. So the below `if` will never happen.

test-case/check-runtime-pm-double-active.sh
fredoh9: @Bin-QA once DSP status is suspended, `break ` the loop and measure the time elapsed is the point. Error handling is pretty much timeout only.

test-case/check-runtime-pm-double-active.sh
```diff
@@ -0,0 +1,145 @@
...
+    local iloop=$(expr 10 '*' "$1")
+    dlogi "wait dsp power status to become suspended, starting max [100ms x $iloop times] checking"
+    for i in $(seq 1 $iloop)
+    do
+        # Here we pass a hardcoded 0 to python script, and need to ensure
+        # DSP is the first audio pci device in 'lspci', this is true unless
+        # we have a third-party pci sound card installed.
+        [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
+        sleep 0.1
+        if [ $i -eq $iloop ]; then
```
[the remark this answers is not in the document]
fredoh9: this is good point!

libinyang:
LGTM

PR #93, opened 2020-02-26
test case: simultaneous-playback-capture fix miss 'both'

Bin-QA:
hot fix for when target tplg miss 'both' type pipeline
will load all pipeline in the case execute

### DIFF
--- a/test-case/simultaneous-playback-capture.sh
+++ b/test-case/simultaneous-playback-capture.sh
@@ -53,6 +53,7 @@ done
 # now all duplicate ids have already been caught
 unset tmp_id_lst tmp_tplg
 id_lst_str=${id_lst_str/,/} # remove 1st, which is not used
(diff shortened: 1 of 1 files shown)


test-case/simultaneous-playback-capture.sh
```diff
@@ -53,6 +53,7 @@ done
 # now all duplicate ids have already been caught
 unset tmp_id_lst tmp_tplg
 id_lst_str=${id_lst_str/,/} # remove 1st, which is not used
+[[ ${#id_lst_str} -eq 0 ]] && dlogw "$tplg without 'both' type pipline" && exit 2
```
aiChaoSONG: Since we don't use both any more, let's remove the "both" in the message. maybe change to "no pipeline with both playback and capture capabilities found in $tplg" is better. if you insist to use both, you can change to "no 'both' type pipeline found in $tplg"

xiulipan: @Bin-QA maybe a quick fix for the message can avoid confusing.

Issue #89, opened 2020-02-25
check-kmod-load-unload-after-playback.sh log naming is confusing.


xiulipan:
check-kmod-load-unload-after-playback.sh have pattern like

```
playback 0 // the log named as aplay_hw:0,0_1.tx
kmod reload
playback 1 // the log named looks same as aplay_hw:0,0_1.txt
```

Bin-QA:
this file create by check-playback.sh

stripes416:
@Bin-QA I like the solution youve presented in your branch.

Bin-QA:
@stripes416 thank you, I forget to create the PR

stripes416:
Closing with merge of #94

PR #94, opened 2020-02-28
test case: check-kmod-load-unload-after-playback fix confuse log name

Bin-QA:
fix issue #89
refind the LOG_ROOT to let each sub case write log to the different
log folder

### DIFF
--- a/test-case/check-kmod-load-unload-after-playback.sh
+++ b/test-case/check-kmod-load-unload-after-playback.sh
@@ -57,24 +57,28 @@ func_lib_check_sudo
 # so when load the test-case in current script
 # the test-case will write the log to the store folder LOG_ROOT
 # which is current script log folder
-export LOG_ROOT=$LOG_ROOT
+log_root=$LOG_ROOT
 
 if [ ${OPT_VALUE_lst['p']} -eq 1 ];then
     func_lib_disable_pulseaudio
 fi
 
+export LOG_ROOT=$log_root/round-0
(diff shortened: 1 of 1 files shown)


test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -57,24 +57,28 @@ func_lib_check_sudo
...
+log_root=$LOG_ROOT
 
 if [ ${OPT_VALUE_lst['p']} -eq 1 ];then
     func_lib_disable_pulseaudio
 fi
 
+export LOG_ROOT=$log_root/round-0
 $(dirname ${BASH_SOURCE[0]})/check-playback.sh -l 1 -t $tplg -d $pb_duration
 ret=$?
+export LOG_ROOT=$log_root
```
libinyang: don't need to export here

Bin-QA: turn back the `LOG_ROOT` to current log folder

libinyang: Do we need to turn back here?

Bin-QA: yes, this value common use at [exit function](https://github.com/thesofproject/sof-test/blob/master/case-lib/hijack.sh#L19)   
we define it for when your current case need to write some file will write to `LOG_ROOT`  
here this change, because have sub-case

PR #100, opened 2020-03-04
test-case: add capture pipeline support for runtime pm test

fredoh9:
aplay/arecord can be used by setting -m option. Add scanning dmesg for
error before switching pipelines.

I submitted two commits in in PR#92. This is first part of it.
https://github.com/thesofproject/sof-test/pull/92
For the new test case i will update on PR#92 if I found better solution.

### DIFF
--- a/test-case/check-runtime-pm-status.sh
+++ b/test-case/check-runtime-pm-status.sh
@@ -7,11 +7,14 @@
 ## Description:
 ##    check the audio runtime pm status
 ## Case step:
-##    1. start aplay
-##    2. stop aplay
-##    3. check the runtime pm status
+##    1. start aplay/arecord
+##    2. stop aplay/arecord
+##    3. sleep for runtime pm transition time
+##    4. check the runtime pm status
 ## Expect result:
 ##    command line check with $? without error
+##    runtime pm status must be suspended
+##    no error in dmesg
 ##
 
 source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@fredoh9  Hi, fred, can you refer to https://github.com/thesofproject/sof-test/pull/102. The runtime pm test should be one test case. with your solution, we have to call the script twice, in fact, we can cover both playback and capture with just one call of this scripts.

Bin-QA:
how about go though the TPLG to remove '-m' option?

test-case/check-runtime-pm-status.sh
```diff
@@ -45,16 +52,31 @@ case $platform in
...
+
 runtime_status="/sys/bus/pci/devices/0000:$(lspci |awk '/[Aa]udio/ {print $1;}')/power/runtime_status"
 
 [[ ! -f $runtime_status ]] && dloge "no runtime_status entry: $runtime_status" && exit 1
 
 dlogc "Runtime status check: cat $runtime_status"
 
 [[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
-func_pipeline_export $tplg "type:playback"
+func_pipeline_export $tplg "type:$test_mode"
```
Bin-QA: here you can direct load `func_pipeline_export $tplg` to get full pipeline

test-case/check-runtime-pm-status.sh
```diff
@@ -65,17 +87,17 @@ do
```
Bin-QA: you need to add 'type' refer to decide use which command & dummy_file

stripes416:
Good patch, once @aiChaoSONG & @Bin-QA 's suggestions have been made, this will be a good addition!

wenqingfu:
ping @fredoh9  for follow up

fredoh9:
@aiChaoSONG @Bin-QA I rebased and addressed your comments.

test-case/check-runtime-pm-status.sh
```diff
@@ -49,62 +52,80 @@ func_check_dsp_status()
...
+        ;;
+        "capture")
+            cmd=arecord
+            dummy_file=/dev/null
+        ;;
+        *)
+            dloge "Invalid pipeline: $type (allow value : playback, capture)"
+            exit 1
+        ;;
+    esac
```
Bin-QA: Almost is good to me   
Can you try the bash skill like this usage:  https://github.com/thesofproject/sof-test/blob/master/test-case/multiple-pipeline-playback.sh#L57   

```
help declare
-A        to make NAMEs associative arrays (if supported)
```

so here you can direct to use like: `cmd=${cmd_lst[$type]}`  
if you worry about miss `keyword`, so you can add check for `cmd` is empty  
you can try this skill, I think it is more useful some simple string transform

PR #102, opened 2020-03-04
test-case: add support for runtime pm check with capture

aiChaoSONG:
Before we remove "both" type PCM, we cannot get all pipelines.
So check-runtime-pm-status.sh only do runtime pm check with
playback.

After we split "both" into playback and capture pipelines,
we are able to get a single pipeline.

This patch add support for runtime pm check with capture,
and make the test case integrited


aiChaoSONG:
duplicated

PR #115, opened 2020-03-10
test-case: refine check-runtime-pm-status.sh

aiChaoSONG:
- This patch use sof-dump-status.py to check if platform support runtime pm.
- This patch replaces the fixed delay time with a variable delay time to reduce test time.
- Test info improvement.

### DIFF
--- a/test-case/check-runtime-pm-status.sh
+++ b/test-case/check-runtime-pm-status.sh
@@ -22,12 +22,30 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
 OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
 OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=3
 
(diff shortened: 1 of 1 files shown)


test-case/check-runtime-pm-status.sh
```diff
@@ -22,31 +22,41 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
...
-    ;;
-esac
+[[ $(sof-dump-status.py --dsp_status 0) == "unsupported" ]] &&
+    dlogi "platform doesn't support runtime pm, skip test case" && exit 2
 
 runtime_status="/sys/bus/pci/devices/0000:$(lspci |awk '/[Aa]udio/ {print $1;}')/power/runtime_status"
-
 [[ ! -f $runtime_status ]] && dloge "no runtime_status entry: $runtime_status" && exit 1
 
 dlogc "Runtime status check: cat $runtime_status"
```
Bin-QA: Current you use `sof-dump-status.py` as run time status check, I think this can be removed

aiChaoSONG: deleted

PR #120, opened 2020-03-16
case-lib: refine pulseaudo enable/disable logic

aiChaoSONG:
If pulseaudio not restored within timeout after disabled, information of successful pulseaudio restoration still output to user.

This patch adds more log of pulseaudio manipulation, and tag pulseaudio related test case failed if pulseaudio not restored as expect.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -48,9 +48,18 @@ function exit()
         done
     fi
 
-    # check if function already defined
-    # on exit check whether pulseaudio is disabled
-    [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
+    # check if function already defined.
+    # on exit check whether pulseaudio is disabled.
+    ret=0
+    if [[ $(declare -f func_lib_restore_pulseaudio) ]]; then
+        func_lib_restore_pulseaudio
+        ret=$?
+    fi
+    # if failed to restore pulseaudio, even test caes passed, set exit status to ret
+    # to make test case failed. this helps to dectect pulseaudio failures.
+    if [ $exit_status -eq 0 -a $ret -ne 0 ]; then
+        exit_status=ret
+    fi
 
     if [ -f $SOF_LOCK ]; then
         # use string compare instead of int to confirm file content correct
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -90,6 +90,14 @@ func_lib_disable_pulseaudio()
         sudo "bash -c 'echo \"autospawn = no\" >> $PULSEAUDIO_CONFIG'"
     fi
     sudo pkill -9 pulseaudio
+    sleep 1s # wait pulseaudio to be disabled
+    if [ ! "$(ps -C pulseaudio --no-header)" ]; then
+        dlogi "Pulseaudio disabled"
+    else
+        # if failed to disable pulseaudio before running test case, fail the test case directly.
+        echo "Failed to disable pulseaudio"
+        exit 1
+    fi
 }
 
 func_lib_restore_pulseaudio()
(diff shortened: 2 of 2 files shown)


Bin-QA:
@aiChaoSONG your catch is right  
but normally `func_lib_restore_pulseaudio` load in `exit` process  
so I think you change will cause some other problem

aiChaoSONG:
> @aiChaoSONG your catch is right
> but normally `func_lib_restore_pulseaudio` load in `exit` process
> so I think you change will cause some other problem

redesigned the logic, please review again

case-lib/hijack.sh
```diff
@@ -50,7 +50,12 @@ function exit()
 
     # check if function already defined
     # on exit check whether pulseaudio is disabled
-    [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
+    [[ $(declare -f func_lib_restore_pulseaudio) ]] && ret=func_lib_restore_pulseaudio
+    # if test case passed, but failed to restore pulseaudio, set exit_status to ret,
+    # this helps to dectect pulseaudio failures.
+    if [ $exit_status -eq 0 -a $ret -ne 0]; then
```
Bin-QA: you logic is when case failed, it will continue disable pulseaudio  
but I think it is not refer the test-case  
it should be when case run finished, restore pulseaudio

case-lib/lib.sh
```diff
@@ -86,6 +86,14 @@ func_lib_disable_pulseaudio()
...
     sudo "sed -i '/autospawn/cautospawn = no' $PULSEAUDIO_CONFIG"
     sudo pkill -9 pulseaudio
+    sleep 1 # wait pulseaudio to be disabled
+    if [ ! "$(ps -C pulseaudio --no-header)" ]; then
+        echo "Pulseaudio disabled"
+        return 0
+    else
+        echo "Failed to disable pulseaudio"
+        return 1
+    fi
```
Bin-QA: what try to do?  
when `kill -9 ` failed, it is the big problem,  
but here also do nothing.

aiChaoSONG: just try to output some log for debug. if pulseaudio killed, we will know it, and if pulseaudio kill failed, we also know it. not get it why this code cause big problem.

case-lib/hijack.sh
```diff
@@ -48,9 +48,17 @@ function exit()
...
+    # check if function already defined, this should be always true.
     # on exit check whether pulseaudio is disabled
-    [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
+    if [[ $(declare -f func_lib_restore_pulseaudio) ]]; then
+        func_lib_restore_pulseaudio
+        ret=$?
+    fi
+    # if failed to restore pulseaudio, even test caes passed, set exit status to ret
+    # to make test case failed. this helps to dectect pulseaudio failures.
+    if [ $exit_status -eq 0 -a $ret -ne 0 ]; then
```
Bin-QA: if `$ret` is not setup, here will cause a error

aiChaoSONG: I thought [[ $(declare -f func_lib_restore_pulseaudio) ]] will always return true, seems I am wrong

case-lib/hijack.sh
```diff
@@ -48,9 +48,18 @@ function exit()
...
-    # on exit check whether pulseaudio is disabled
-    [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
+    # check if function already defined.
+    # on exit check whether pulseaudio is disabled.
+    ret=0
+    if [[ $(declare -f func_lib_restore_pulseaudio) ]]; then
+        func_lib_restore_pulseaudio
+        ret=$?
+    fi
+    # if failed to restore pulseaudio, even test caes passed, set exit status to ret
```
xiulipan: I did not get the logic here. can we make it to optional here? As we may have some device use **rename** method to disable pulseaudio. I think it would be better to allow pulseaudio failure as not all test case is depend on it.

But we do need to fail the test case if the test case need pulseaudio and we could not enable it.

case-lib/lib.sh
```diff
@@ -86,6 +86,14 @@ func_lib_disable_pulseaudio()
     # because hijack the sudo command
     sudo "sed -i '/autospawn/cautospawn = no' $PULSEAUDIO_CONFIG"
     sudo pkill -9 pulseaudio
+    sleep 1s # wait pulseaudio to be disabled
+    if [ ! "$(ps -C pulseaudio --no-header)" ]; then
+        dlogi "Pulseaudio disabled"
+    else
+        # if failed to disable pulseaudio, test case related to pulseaudio will be skipped.
```
xiulipan: Let use the N/A to replace skip from now on.

case-lib/hijack.sh
aiChaoSONG: The code behaves like below:
- Cases don't touch PA, or use PA rename method: `[[ "${#PULSECMD_LST[@]}" -ne 0 ]] && return` and `[[ "${#PULSECMD_LST[@]}" -eq 0 ]] && return` will ensure `func_lib_disable_pulseaudio` and `func_lib_restore_pulseaudio` return immediately, `ret` will always be 0, and exit_status is the test case exit_status. Code change has no impact on this.
- Cases need to disable / enable PA: 
    - cases passed, PA disable/enable successfully, exit_status is test case exit status
    - cases failed, PA disable/enable status will be ignored, exit_status is test case exit status
    - cases passed, PA restore failed, exit_status will be replace with PA restore failed status
    - PA disable failed before running case, tag test case skipped(NA)

This is the logic, you can merge the PR into your local repo, and see it. I think there is no problem with the logic. BTW, PA in Ubuntu2004 will be managed by systemd, this allow us to disable PA easier.

case-lib/lib.sh
aiChaoSONG: sure!

aiChaoSONG:
@xiulipan @Bin-QA  rebased and ready to get merged, since you have approved before rebase, please approve again, and get this merged.

PR #122, opened 2020-03-17
add lsof device file check when aplay/arecord/speaker failed

Bin-QA:
this PR include 2 patch

1. add 'snd' which mapping /dev/snd/pcm* in sof-tplgreader
2. add lsof check to confirm resource is not lock by other process

it can catch error like this : `aplay: main:788: audio open error: Device or resource busy`

### DIFF
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -67,6 +67,7 @@ do
         dev=$(func_pipeline_parse_value $idx dev)
         pcm=$(func_pipeline_parse_value $idx pcm)
         type=$(func_pipeline_parse_value $idx type)
+        snd=$(func_pipeline_parse_value $idx snd)
 
         if [ ${OPT_VALUE_lst['F']} = '1' ]; then
             fmt=$(func_pipeline_parse_value $idx fmts)
@@ -91,6 +92,7 @@ do
                 dlogc "arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
                 arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q
                 if [[ $? -ne 0 ]]; then
+                    func_lib_lsof_error_dump $snd
                     dmesg > $LOG_ROOT/arecord_error_${dev}_$i.txt
                     dloge "arecord on PCM $dev failed at $i/$loop_cnt."
                     exit 1
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -79,12 +79,19 @@ class clsTPLGReader:
                     self._pipeline_lst.append(pb_pipeline_dict)
                 self._pipeline_lst.append(pipeline_dict)
 
-        # format pipeline, this change for script direct access 'rate' 'channel' 'dev'
+        # format pipeline, this change for script direct access 'rate' 'channel' 'dev' 'snd'
         for pipeline in self._pipeline_lst:
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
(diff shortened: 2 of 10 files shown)


tools/sof-tplgreader.py
```diff
@@ -79,12 +79,17 @@ def loadFile(self, filename, sofcard=0):
...
                 self._pipeline_lst.append(pipeline_dict)
 
-        # format pipeline, this change for script direct access 'rate' 'channel' 'dev'
+        # format pipeline, this change for script direct access 'rate' 'channel' 'dev' 'snd'
         for pipeline in self._pipeline_lst:
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
             pipeline['rate'] = pipeline['rate_min'] if int(pipeline['rate_min']) != 0 else pipeline['rate_max']
             pipeline['channel'] = pipeline['ch_min']
             pipeline['dev'] = "hw:" + str(sofcard) + ',' + pipeline['id']
+            pipeline['snd'] = "/dev/snd/pcmC" + str(sofcard) + "D" + pipeline['id']
```
aiChaoSONG: Please add a comment upon this line to tell the reader that the string takes this form: /dev/snd/pcmC + card_number + D + device number + capability. so they know why you construct the string like this. the comment can be like this:`the character devices for PCM under /dev/snd take the form of "pcmC + card_number + D + device_number + capability", eg, pcmC0D0p.`

test-case/check-capture.sh
```diff
@@ -91,6 +92,7 @@ do
                 dlogc "arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
                 arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q
                 if [[ $? -ne 0 ]]; then
+                    lsof $snd
```
aiChaoSONG: The user may be confused by the output of `lsof $snd`, because they don't know what the output actually is. my suggestion is: assign the result to a variable, if variable is empty, output nothing, because no device file is in use, and if variable is not empty, we output some hint info and the result. This is user friendly.
```
ret=$(lsof $snd)
[ -n $ret ] && dloge "Sound device file still in use:"
dloge $ret # maybe echo is better here?

```

tools/sof-tplgreader.py
Bin-QA: good idea

test-case/check-capture.sh
Bin-QA: I think if none block it, so `lsof $snd` will dump nothing  
Here don't need to care about who use this device file  
your suggest is good, but I think it maybe complex

aiChaoSONG: My question is: if lsof dump some info, how can the user know what the info is? So I think the hint is important.

aiChaoSONG:
@Bin-QA  Commit message of f07c1bf1a5f023cca98243fd10f3a0020ba0f47e need a little improvement:
```
when case failed, add lsof device file check to ensure
no other process is using the resource

This patch handles: "audio open error: Device or resource busy"
```

Bin-QA:
> @Bin-QA Commit message of [f07c1bf](https://github.com/thesofproject/sof-test/commit/f07c1bf1a5f023cca98243fd10f3a0020ba0f47e) need a little improvement:
> 
> ```
> when case failed, add lsof device file check to ensure
> no other process is using the resource
> 
> This patch handles: "audio open error: Device or resource busy"
> ```

Thank you for your advice

aiChaoSONG:
@Bin-QA commit message title of 21d5a5bb05601c83ebcbaf502453c2499e4c83ad, let't change it to:
`tools: add field 'snd' to pipeline dict in sof-tplgreader.py` , then the PR will be perfect to me.

PR #128, opened 2020-03-18
case-lib: fix pulseaudio disable problem

Bin-QA:
when pulseaudio configure without keyword 'autospawn'
try to disable pulseaudio will failed


libinyang:
LGTM

Issue #105, opened 2020-03-05
[new feature] pipeline filtering enhancement


aiChaoSONG:
- Currently, we globally block DMIC16k (WOV pipeline) to prevent test failure on this pipeline, but this solution will stop us from writing WOV automation test cases in the future.
- We may need to export pipeline with some specific component, such as "all pipeline with EQ", "playback with PGA", "capture with SRC".
So we need to enhance current pipeline filtering feature.

My thought is:
- retain global ignore list, and add a local ignore list. (cases not use DMIC16k ignore it locally)
- add a with list, this help us to get pipeline in "type" with "component"
The final string passed to `func_export_pipeline` should be:
> "type list; with list; ignore list"    For example:
>"type:playback,capture;with:EQ,PGA;ignore:DMIC16k" OR
>"playback,capture;EQ,PGA;DMIC16k"

We also need to make some change to func_export_pipeline and sof-tplgreader.py.

if spilit with bash is tedious, we can provide new python sof-util.py, and use Python to do this. in fact, the sof-combinatoric.py from Wubin, can also be regarded as a util, and can be merged into sof-util.py.

The above solution can also help to get all pipeline without EQ: "type:playback,capture;;ignore:EQ"

aiChaoSONG:
@xiulipan @Bin-QA @fredoh9 @stripes416 @libinyang @keqiaozhang please leave your comments if you have.

Bin-QA:
`TPLG_IGNORE_LST` you can reset it for your test-case

the common array, you also can redefine it
`unset` to cleanup bash variable

So I don't think you need to write a new function 
on the other hand, you means ignore seems: 
 ignore 'pcm:' so this feature already at sof-tplgreader.py  
you can give complex parameter to get it

aiChaoSONG:
> `TPLG_IGNORE_LST` you can reset it for your test-case
> 
> the common array, you also can redefine it
> `unset` to cleanup bash variable
> 
> So I don't think you need to write a new function
> on the other hand, you means ignore seems:
> ignore 'pcm:' so this feature already at sof-tplgreader.py
> you can give complex parameter to get it

@Bin-QA  we can overwrite TPLG_IGNORE_LST, so the ignore part can removed how about the "with part"? is there any way to get all pipeline with EQ?

Bin-QA:
@aiChaoSONG 
```
./sof-tplgreader.py -h

  -f FILTER [FILTER ...], --filter FILTER [FILTER ...]
[...]

PR #129, opened 2020-03-18
tools: pipeline filtering feature enhancement

aiChaoSONG:
This patch fix a pga and eq assignment error, and enhance pipeline filtering feature. Now we can filter pipeline with specific component.

@fredoh9 Some usage listed below:
- all pipelines: `sof-tplgreader.py tplg-path -f "type:playback,capture"`
- capture pipelines with PGA:  `sof-tplgreader.py tplg-path -f "type:capture" "pga:any"`
- playback pipelines with EQ:  `sof-tplgreader.py tplg-path -f "type:playback" "eq:any"`
- all pipelines with PGA:  `sof-tplgreader.py tplg-path -f "pga:any"`
- all pipelines with PGA and EQ:  `sof-tplgreader.py tplg-path -f "pga:any" "eq:any" `

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -10,7 +10,7 @@ class clsTPLGReader:
         self._pipeline_lst = []
         self._output_lst = []
         self._field_lst = []
-        self._filter_lst = []
+        self._filter_dict = {"filter":[], "op":[]}
         self._ignore_lst = []
 
     def __comp_pipeline(self, pipeline):
@@ -87,6 +87,24 @@ class clsTPLGReader:
             pipeline['dev'] = "hw:" + str(sofcard) + ',' + pipeline['id']
         return 0
 
+    @staticmethod
+    def list_and(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = []
+        for elem in lst1:
+            if elem in lst2:
+                output_lst.append(elem)
+        return output_lst
+
+    @staticmethod
+    def list_or(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = lst1.copy()
+        for elem in lst2:
+            if elem not in output_lst:
+                output_lst.append(elem)
+        return output_lst
+
     def _setlist(self, orig_lst):
         tmp_lst = []
         if orig_lst is None:
@@ -98,8 +116,9 @@ class clsTPLGReader:
             tmp_lst.append(orig_lst)
         return tmp_lst
 
-    def setFilter(self, filter_lst=None):
-        self._filter_lst = self._setlist(filter_lst)[:]
+    def setFilter(self, filter_dict=None):
+        self._filter_dict['filter'] = self._setlist(filter_dict['filter'])[:]
+        self._filter_dict['op'] = self._setlist(filter_dict['op'])
 
     def setField(self, field_lst):
         self._field_lst = self._setlist(field_lst)[:]
@@ -110,20 +129,45 @@ class clsTPLGReader:
     def _filterOutput(self, target_lst, filter_dict, bIn):
         self._output_lst.clear()
         for line in target_lst[:]:
+            check = False
             for key, value in filter_dict.items():
-                # match for 'keyword'/'keyword [0-9]' target line
-                check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', line[key], re.I)]) > 0
+                if 'any' in value or value == ['']:
+                    check = True if key in line.keys() else False
+                else:
+                    # match for 'keyword'/'keyword [0-9]' target line
+                    check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', str(line[key]), re.I)]) > 0
                 if check is bIn:
                     break
             else:
                 self._output_lst.append(line)
 
+        return self._output_lst.copy()
+
     def _filterKeyword(self):
-        if len(self._filter_lst) == 0:
+        if len(self._filter_dict['filter']) == 0:
             return
-        for filter_dict in self._filter_lst:
-            tmp_lst = self._output_lst[:]
-            self._filterOutput(tmp_lst, filter_dict, False)
+
+        full_list = self._output_lst[:]
+        filtered = None
+        # pipelines filtered by the first filter item
+        for key, value in self._filter_dict['filter'][0].items():
+            if key.startswith('~'):
+                filtered = self._filterOutput(full_list, {key[1:]:value}, True)
+            else:
+                filtered = self._filterOutput(full_list, {key:value}, False)
+        # do filtering by the rest filter items and logic operations
+        for idx in range(1, len(self._filter_dict['filter'])):
+            new_filtered = None
+            for key, value in self._filter_dict['filter'][idx].items():
+                if key.startswith("~"):
+                    new_filtered = self._filterOutput(full_list, {key[1:]:value} , True)
+                else:
+                    new_filtered = self._filterOutput(full_list, {key:value}, False)
+            if self._filter_dict['op'][idx-1] == '&':
+                filtered = self.list_and(filtered, new_filtered)
+            if self._filter_dict['op'][idx-1] == '|':
+                filtered = self.list_or(filtered, new_filtered)
+        self._output_lst = filtered # the final filtered pipeline list
 
     def _filterField(self):
         if len(self._field_lst) == 0:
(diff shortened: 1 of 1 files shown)


tools/sof-tplgreader.py
```diff
@@ -202,12 +206,12 @@ def func_getPipeline(tplgObj, tplgName, sdcard_id, sort):
     parser.add_argument('filename', type=str, help='tplg file name, multi-tplg file name use "," to split it')
     parser.add_argument('-s', '--sofcard', type=int, help='sofcard id', default=0)
     parser.add_argument('-f', '--filter', type=str, nargs='+',
-        help='''setup filter, this value is format value,
+        help='''setup filter, command line parameter
 string format is 'key':'value','value'
```
Bin-QA: I think you need to add 'any' as 'special value'

tools/sof-tplgreader.py
```diff
@@ -110,9 +110,13 @@ def setIgnore(self, ignore_lst=None):
     def _filterOutput(self, target_lst, filter_dict, bIn):
         self._output_lst.clear()
         for line in target_lst[:]:
+            check = False
             for key, value in filter_dict.items():
-                # match for 'keyword'/'keyword [0-9]' target line
-                check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', line[key], re.I)]) > 0
+                if value == ['any'] or value == ['']:
+                    check = True if key in line.keys() else False
```
Bin-QA: Can receive '' ??

aiChaoSONG: suggestion is good, I will take it.

tools/sof-tplgreader.py
aiChaoSONG: I will add some text to explain 'any' in help info

Bin-QA:
@xiulipan https://github.com/thesofproject/sof-test/blob/master/test-case/simultaneous-playback-capture.sh#L44

1. check pipeline id whether is duplicate, if it is duplicate, it is the 'both' type pipeline
2. just dump pipeline refer this id, will dump 2 pipeline of this id: playback + capture

aiChaoSONG:
filter login need redesign, tag this PR WIP

xiulipan:
@Bin-QA @aiChaoSONG Thanks for the feedback and RFC, let create an issue to discussion.

aiChaoSONG:
@xiulipan https://github.com/thesofproject/sof-test/issues/105 see here

aiChaoSONG:
@xiulipan @Bin-QA redesigned filtering logic, now `&, |, and ~` is supported in filter string. for test case want pipeline with both playback and capture, just need to use "type:dual"
some usage:
- "type:playback & id:3", playback pipeline who's id=3
- "type:capture | pga:any", capture pipelines or pipelines with pga
- "type:playback & ~eq:any", playback pipelines without eq

aiChaoSONG:
@xiulipan @Bin-QA Ping for review

Bin-QA:
@aiChaoSONG I think you miss the filter init status
```
sof-tplgreader.py sof-apl-pcm512x.tplg
  File "/home/user/Templates/sof-test/tools/sof-tplgreader.py", line 342, in <module>
    if len(filter_dict['filter']) > 0:
TypeError: object of type 'NoneType' has no len()

```

aiChaoSONG:
@Bin-QA I tried every combination, except that no -f provided. Now, fix the initial condition.

tools/sof-tplgreader.py
```diff
@@ -231,16 +277,26 @@ def func_getPipeline(tplgObj, tplgName, sdcard_id, sort):
...
+        op_lst = []
+        for filter_elem in ret_args['filter'].split('|'):
+            for item in filter_elem.split('&'):
+                key, _, value = item.partition(':')
+                filter_lst.append({key.strip():value.strip().replace('[','').replace(']','').split(',')})
+        for char in ret_args['filter']:
+            if char == '|' or char == "&":
+                op_lst.append(char)
+        filter_dict['filter'] = filter_lst
+        filter_dict['op'] = op_lst
```
Bin-QA: 1. `[]` value can include this?
2.  you get order `op_lst`, but `filter_lst ` is order by `'|', '&'` , I think this is will cause problem  
maybe you can try `for char in len(ret_args['filter'])`   
here will include those behavior:
```
if '|' in ret_args['filter'] or '&' in ret_args['filter']:
   ...
else:
   loop ret_args['filter']
fi
```

aiChaoSONG: for item 1, delete the replacement of '[' and ']'.

tools/sof-tplgreader.py
```diff
@@ -87,6 +87,24 @@ def loadFile(self, filename, sofcard=0):
...
         return 0
 
+    @staticmethod
+    def list_and(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = []
+        for elem in lst1:
+            if elem in lst2:
+                output_lst.append(elem)
+        return output_lst
```
Bin-QA:
```suggestion
        output_lst = list(set(lst1).intersection(lst2))
        return output_lst
```

tools/sof-tplgreader.py
```diff
@@ -87,6 +87,24 @@ def loadFile(self, filename, sofcard=0):
...
+
+    @staticmethod
+    def list_or(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = lst1.copy()
+        for elem in lst2:
+            if elem not in output_lst:
+                output_lst.append(elem)
+        return output_lst
```
Bin-QA:
```suggestion
        output_lst = list(set(lst1).union(lst2))
        return output_lst
```

tools/sof-tplgreader.py
```diff
@@ -110,20 +129,45 @@ def setIgnore(self, ignore_lst=None):
...
+        # do filtering by the rest filter items and logic operations
+        for idx in range(len(self._filter_dict['filter']) - 1):
+            new_filtered = None
+            for key, value in self._filter_dict['filter'][idx+1].items():
+                if key.startswith("~"):
+                    new_filtered = self._filterOutput(full_list, {key[1:]:value} , True)
+                else:
+                    new_filtered = self._filterOutput(full_list, {key:value}, False)
+            if self._filter_dict['op'][idx] == '&':
+                filtered = self.list_and(filtered, new_filtered)
```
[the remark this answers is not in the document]
aiChaoSONG: they are already static method, just not call them use self rather class name

tools/sof-tplgreader.py
aiChaoSONG: dict is unhashable, unable to use set

tools/sof-tplgreader.py
aiChaoSONG: ditto

tools/sof-tplgreader.py
```diff
@@ -110,20 +129,45 @@ def setIgnore(self, ignore_lst=None):
...
+        full_list = self._output_lst[:]
+        filtered = None
+        # pipelines filtered by the first filter item
+        for key, value in self._filter_dict['filter'][0].items():
+            if key.startswith('~'):
+                filtered = self._filterOutput(full_list, {key[1:]:value}, True)
+            else:
+                filtered = self._filterOutput(full_list, {key:value}, False)
+        # do filtering by the rest filter items and logic operations
+        for idx in range(len(self._filter_dict['filter']) - 1):
```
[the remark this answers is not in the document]
aiChaoSONG: I will pick the first. because the second cannot deal with "type:playback | pga &"

tools/sof-tplgreader.py
```diff
@@ -110,20 +129,45 @@ def setIgnore(self, ignore_lst=None):
...
+            for key, value in self._filter_dict['filter'][idx].items():
+                if key.startswith("~"):
+                    new_filtered = self._filterOutput(full_list, {key[1:]:value} , True)
+                else:
+                    new_filtered = self._filterOutput(full_list, {key:value}, False)
+            if self._filter_dict['op'][idx-1] == '&':
+                filtered = self.list_and(filtered, new_filtered)
+            if self._filter_dict['op'][idx-1] == '|':
+                filtered = self.list_or(filtered, new_filtered)
+        self._output_lst = filtered # the final filterd pipeline list
```
stripes416:
```suggestion
        self._output_lst = filtered # the final filtered pipeline list
```

tools/sof-tplgreader.py
```diff
@@ -201,13 +245,15 @@ def func_getPipeline(tplgObj, tplgName, sdcard_id, sort):
...
-for example: Get "type" is "capture"
-type: capture
-for example: Get "type" is "capture, both" and "fmt" is S16_LE
-type:capture,both fmt:S16_LE
+    parser.add_argument('-f', '--filter', type=str,
+        help='''setup filter, command line parameter
+string format is 'key':'value','value', the filter
+string support & | and ~ logic operation.
+if only care about key, you can use 'key':'any'.
+Example 1: `-f "type:capture & fmt:S16_LE"`
```
stripes416: While these examples are good, its best to be explicit about it. Something more like your usage examples in the description for your first iteration would be great.

> all pipelines: sof-tplgreader.py tplg-path -f "type:playback,capture"
capture pipelines with PGA: sof-tplgreader.py tplg-path -f "type:capture" "pga:any"
playback pipelines with EQ: sof-tplgreader.py tplg-path -f "type:playback" "eq:any"
all pipelines with PGA: sof-tplgreader.py tplg-path -f "pga:any"
all pipelines with PGA and EQ: sof-tplgreader.py tplg-path -f "pga:any" "eq:any"
>

tools/sof-tplgreader.py
aiChaoSONG: good catch, and now fixed

tools/sof-tplgreader.py
aiChaoSONG: I have added more example in the help messages.

aiChaoSONG:
@Bin-QA Got any new ideas? If not, can we make this merged first, and later ideas into new commits later?

xiulipan:
@aiChaoSONG Can you test with this change on some platform? Will the filter change influence any existing test cases?

aiChaoSONG:
> @aiChaoSONG Can you test with this change on some platform? Will the filter change influence any existing test cases?

@xiulipan No any influence on existing test case, because these test cases only use very minor feature, eg, only filter playback or capture.

PR #130, opened 2020-03-19
case-lib: remove SOF-LOCK file

Bin-QA:
Compare the gains with the risks, remove the sof-lock
Now if meet test-case resource problem, #122 will handle this

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -52,16 +52,6 @@ function exit()
     # on exit check whether pulseaudio is disabled
     [[ $(declare -f func_lib_restore_pulseaudio) ]] && func_lib_restore_pulseaudio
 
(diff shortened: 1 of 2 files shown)


libinyang:
There should be a related patch in other places to avoid accessing the same pcm sources. What if 2 test cases are run simultaneously?

Bin-QA:
> There should be a related patch in other places to avoid accessing the same pcm sources. What if 2 test cases are run simultaneously?

the test-case will report failed, because resource is busy, #122 already handle this information  
Current we accept test-case script to load sub test-case to let current case more clear  
In feature if sub-case also load the test-case ...   
So the sof-lock will spend more time to fix issue,  
direct report the error is good enough to know what happend

PR #133, opened 2020-03-19
test-case: add delay for check playback

ClarexZhou:
update to set sleep time array, sleeps time will go through and loop this for each playback. This array contains import timeout and +/- 0.1s. Now timeout is 3s for SOF PCI device, 2s for HDA codec in future (not merged).

### DIFF
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -39,6 +39,9 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
 OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
 
+OPT_OPT_lst['R']='sleep_ctl'   OPT_DESC_lst['R']="set wait/sleep time for runtime PM, sleep time according critical timeout, which is in sleep_lst"
+OPT_PARM_lst['R']=0         OPT_VALUE_lst['R']=0
+
 func_opt_parse_option $*
 
 tplg=${OPT_VALUE_lst['t']}
@@ -46,7 +49,7 @@ round_cnt=${OPT_VALUE_lst['r']}
 duration=${OPT_VALUE_lst['d']}
 loop_cnt=${OPT_VALUE_lst['l']}
 file=${OPT_VALUE_lst['f']}
-
+sleep_ctl=${OPT_VALUE_lst['R']}
 
 [[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
 
@@ -64,6 +67,11 @@ fi
 func_lib_setup_kernel_last_line
 func_lib_check_sudo
 func_pipeline_export $tplg "type:playback"
+
+#Set sleep time array, sleeps time will go through and loop this for each playback. This array contains import timeout and +/- 0.1s. Now timeout is 3s for SOF PCI device, 2s for HDA codec in future (not merged).
+sleep_lst=(0.5 1.9 2 2.1 2.9 3 3.1 5)
+sleep_len=${#sleep_lst[@]}
+j=0
 
 for round in $(seq 1 $round_cnt)
 do
@@ -94,6 +102,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep ${sleep_lst[@]} seconds before running aplay, then aplay can be run under dsp active or suspended status
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ "$sleep_ctl" -eq 1 ]; then
+                        sleep_time=${sleep_lst[$((j++ % sleep_len))]}
+                    else
+                        dlogi "platform doesn't support runtime pm or sleep time is turned off, sleep period keeps ${sleep_lst[0]} second(s)"
+                        sleep_time=${sleep_lst[0]}
+                    fi
+                    sleep "$sleep_time"
+                    pm=$(sof-dump-status.py --dsp_status 0)
+                    dlogi "Sleep: $sleep_time second(s), PM status is $pm"
                 fi
                 dmesg > $LOG_ROOT/aplay_${dev}_$i.txt
             done


test-case/check-playback.sh
```diff
@@ -88,6 +92,10 @@ do
             do
                 dlogi "Testing: (Round: $round/$round_cnt) (PCM: $pcm [$dev]<$type>) (Loop: $i/$loop_cnt)"
                 dlogc "aplay -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
+		#sleep random 0-7 seconds before aplay to cover triggering aplay from both runtime PM suspend and active state.
+		time=$(($RANDOM % $random_range))
+		dlogi "Sleep: $time"
```
aiChaoSONG: dlogi "Sleep $time second(s)"

This makes log user friendly

test-case/check-playback.sh
```diff
@@ -39,13 +39,17 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
 OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
 
+OPT_OPT_lst['p']='rdtime'   OPT_DESC_lst['p']='sleep random duation between 0-7 seconds to cover runtime PM both active and suspend time'
```
aiChaoSONG: In fact, rdtime is a parameter to control random range, right?  if  rdtime is 10, the random range is 0~10, so the explanation here is not accurate.  change to --> `max random sleep time for runtime PM` is enough.

test-case/check-playback.sh
```diff
@@ -88,6 +92,10 @@ do
             do
                 dlogi "Testing: (Round: $round/$round_cnt) (PCM: $pcm [$dev]<$type>) (Loop: $i/$loop_cnt)"
                 dlogc "aplay -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
+		#sleep random 0-7 seconds before aplay to cover triggering aplay from both runtime PM suspend and active state.
```
aiChaoSONG: improve the comment here, may change to `sleep 0~random_range seconds before running aplay/arecord, then aplay/arecord can be run under dsp active or suspended status randomly`

Bin-QA:
I think this case will help you to add the random time [check-suspend-resume.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/check-suspend-resume.sh)

Here setup the the random range and support time value from option

xiulipan:
@ClarexZhou Can you give more detail about why this test case is needed?
Duplicated with https://github.com/thesofproject/sof-test/pull/92 @fredoh9

ClarexZhou:
> @ClarexZhou Can you give more detail about why this test case is needed?
> Duplicated with #92 @fredoh9

Current script, we always trigger aplay when runtime PM is actived. But I think we should cover different scenarios as end user,  trigger aplay in different runtime PM status (Active/Suspend).  
Also to avoid issue when pass if trigger aplay during runtime PM active, but fail when trigger aplay when runtime PM suspend in stress.

test-case/check-playback.sh
```diff
@@ -18,6 +18,9 @@
 
 source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
 
+random_min=0    # wait time should >= 0 for runtime PM
```
xiulipan: Make this into configurable options

test-case/check-playback.sh
```diff
@@ -88,6 +94,14 @@ do
...
                 dlogi "Testing: (Round: $round/$round_cnt) (PCM: $pcm [$dev]<$type>) (Loop: $i/$loop_cnt)"
                 dlogc "aplay -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
+                #sleep $random_min ~$ramdom_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
+                if [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
+                    time=$(func_lib_get_random $random_min $random_max)
+                else
+                    time=$random_min
+                fi
+                dlogi "Sleep: $time second(s)"
+                sleep $time
```
xiulipan: The sleep should go after the aplay finished.
Do not inject any other cmd between dlogc and the real cmd, it will make log more confusing.

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
...
+                    #sleep random_min ~ ramdom_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
+                        time=$(func_lib_get_random $random_min $random_max)
+                    else    
+                        dlogi "platform doesn't support runtime pm or random range is turned off, no need sleep time"
+                        time=$random_min
+                    fi
+                    sleep $time
+                    pm=`sof-dump-status.py --dsp_status 0`
+                    dlogi "Sleep: $time second(s), PM status is $pm"
```
Bin-QA: `no need sleep time`, but you set the time value to the `$random_min`

```suggestion
                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
                        time=$(func_lib_get_random $random_min $random_max)
                        sleep $time
                        dlogi "Sleep: $time second(s)"
                    else    
                        dlogi "platform doesn't support runtime pm or random range is turned off, no need sleep time"
                    fi
                    pm=`sof-dump-status.py --dsp_status 0`
                    dlogi "PM status is $pm"
```

test-case/check-playback.sh
```diff
@@ -39,14 +39,24 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
 OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
 
+OPT_OPT_lst['R']='random'   OPT_DESC_lst['R']="Randomly setup wait/sleep time for rumtime PM, range is [$random_min-$random_max]"
+OPT_PARM_lst['R']=0         OPT_VALUE_lst['R']=1
+
+OPT_OPT_lst['i']='random_min'   OPT_DESC_lst['i']="Minim randomly setup wait/sleep time for rumtime PM"
```
stripes416:
```suggestion
OPT_OPT_lst['i']='random_min'   OPT_DESC_lst['i']="Minimum randomly set wait/sleep time for runtime PM"
```

test-case/check-playback.sh
```diff
@@ -39,14 +39,24 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
 OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
 
+OPT_OPT_lst['R']='random'   OPT_DESC_lst['R']="Randomly setup wait/sleep time for rumtime PM, range is [$random_min-$random_max]"
```
stripes416:
```suggestion
OPT_OPT_lst['R']='random'   OPT_DESC_lst['R']="Randomly set wait/sleep time for runtime PM, range is [$random_min-$random_max]"
```

test-case/check-playback.sh
```diff
@@ -39,14 +39,24 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
...
+OPT_PARM_lst['R']=0         OPT_VALUE_lst['R']=1
+
+OPT_OPT_lst['i']='random_min'   OPT_DESC_lst['i']="Minim randomly setup wait/sleep time for rumtime PM"
+OPT_PARM_lst['i']=1         OPT_VALUE_lst['i']=0
+
+OPT_OPT_lst['x']='random_max'   OPT_DESC_lst['x']="Max randomly setup wait/sleep time for rumtime PM"
```
stripes416:
```suggestion
OPT_OPT_lst['x']='random_max'   OPT_DESC_lst['x']="Max randomly set wait/sleep time for runtime PM"
```

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep random_min ~ ramdom_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
```
stripes416:
```suggestion
                    #sleep random_min ~ random_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
```

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
...
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
+                        time=$(func_lib_get_random $random_min $random_max)
+                    else    
+                        dlogi "platform doesn't support runtime pm or random range is turned off, no need sleep time"
```
stripes416:
```suggestion
                        dlogi "platform doesn't support runtime pm or random range is turned off, sleep period not needed"
```

test-case/check-playback.sh
ClarexZhou: Thanks Bin, and still will keep sleep random_min for no need random delay scenarios.  And i will update the description for this.

xiulipan:
@lgirdwood I think some random test simulate the user behavior is a good idea, but the point of reproducible is also import.
@ClarexZhou Can you do some update about this PR to dump the info when error happen and make it easy to reproduce with the info you dump?

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep random_min ~ random_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
+                        time=$(func_lib_get_random $random_min $random_max)
+                    else
+			    dlogi "platform doesn't support runtime pm or random range is turned off, sleep period keeps $random_min second(s)"
```
Bin-QA, line 112: here have the 'tab' not the 'space'

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep random_min ~ random_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ ${OPT_VALUE_lst['R']} -eq 1 ]; then
+                        time=$(func_lib_get_random $random_min $random_max)
```
Bin-QA, line 110: can you rebase your code: https://github.com/thesofproject/sof-test/blob/master/case-lib/lib.sh#L121

```suggestion
                        time=$(func_lib_get_random $random_max $random_min)
```

test-case/check-playback.sh
```diff
@@ -94,6 +104,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep random_min ~ random_max seconds before running aplay, then aplay can be run under dsp active or suspended status randomly
```
Bin-QA, line 108: here not the randomly, it should be dsp status is 'active' or 'suspended'  
so it means the other status is not our expect

lgirdwood:
> @lgirdwood I think some random test simulate the user behavior is a good idea, but the point of reproducible is also import.

This doesn't buy us anything, we need to deliberately target the timing of the tests so that we cover any areas where timing is used in FW/driver. e.g. runtime PM timeout, DAPM timeout.

xiulipan:
@lgirdwood A good point here. So if a time range in sequence will help here?

ClarexZhou:
@Bin-QA @lgirdwood  I just updated scripts and now this delay parameter default value is changed. Now this delay won't be used in PR verify and daily base test.
This delay will only be used for stress test, to cover more different status for end user test cases.

For bug reproduce: We always using this random delay script for aplay stress test previously. And for the detected bugs,  always able to be reproduced for now. Only the stress times might be different when issue occurred, e.g. issue occurred in 200 interactions  or after 300 interactions...

lgirdwood:
@ClarexZhou the random timing is broken. Please use deterministic values that intersect with driver and FW timings. i.e. HDA link poweroff timeout, DAPM poweroff timout, driver runtime PM entry. i.e. if DAPM power off is at 5 secs then we should do an equal number of stress tests before and after this tmeout (same with the rest). Please align with @xiulipan for all driver and FW timeouts.

Consider we have 3 critical driver and FW events that happen 0.1, 3 and 5 secs after playback. We would then want to stress at  0.09, 0.11, 2.99, 3.01, 4.99 and 5.01 (i.e. X +/- 0.01) and the results here would give us more more information about any failures i.e. failed 3/100 at 0.11

xiulipan:
@lgirdwood I think before we get the right delay or timeout to be deterministic values for test. We may need to tune them. I still think keep a random feature for this test is OK. It will help us to find the  **deterministic values**

@ranj063 @keyonjie @RanderWang Can you help to give some clarify about the timetout value on different type dais? I think we have different driver for I2S/HDA/SDW.

lgirdwood:
The important timeouts are
1) DAPM power off after playback
2) HDA link power off after playback
3) SOF runtime PM entry after playback
4) ???
@kv2019i can you think of any more critical timeouts in the driver and their values.

RanderWang:
> @lgirdwood I think before we get the right delay or timeout to be deterministic values for test. We may need to tune them. I still think keep a random feature for this test is OK. It will help us to find the **deterministic values**
> 
> @ranj063 @keyonjie @RanderWang Can you help to give some clarify about the timetout value on different type dais? I think we have different driver for I2S/HDA/SDW.

There is no special sleep delay for SDW. And it is 3s for SOF PCI device, 2s for HDA codec in future (not merged).

ClarexZhou:
update delay time to array.

test-case/check-playback.sh
```diff
@@ -39,14 +39,17 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
...
+
 func_opt_parse_option $*
 
 tplg=${OPT_VALUE_lst['t']}
 round_cnt=${OPT_VALUE_lst['r']}
 duration=${OPT_VALUE_lst['d']}
 loop_cnt=${OPT_VALUE_lst['l']}
 file=${OPT_VALUE_lst['f']}
-
+sleep=${OPT_VALUE_lst['R']}
```
marc-hb, line 52: `sleep` is also a command name, use something different to avoid confusion.

test-case/check-playback.sh
```diff
@@ -65,6 +68,9 @@ func_lib_setup_kernel_last_line
 func_lib_check_sudo
 func_pipeline_export $tplg "type:playback"
 
+sleep_lst=(0.5 1.9 2 2.1 2.9 3 3.1 5) #And it is 3s for SOF PCI device, 2s for HDA codec in future (not merged).
```
marc-hb, line 71: What is "it" 3s and what is 2s? The timeout before suspend? 

By moving the comment to the previous line they can become shorter than 80 characters. `shellcheck` should show this, please try it.

test-case/check-playback.sh
```diff
@@ -94,6 +100,22 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep ${sleep_lst[@]} seconds before running aplay, then aplay can be run under dsp active or suspended status
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ $sleep -eq 1 ]; then
+                        time=${sleep_lst[${j}]}
+                        j=`expr $j + 1`
```
marc-hb, line 107: `: $((j++))` is shorter, doesn't have deprecated backticks and probably cheaper/faster.

`:` has also  better compatibility with `set -e` compared to the following which aborts when using `set -e` : `j=-1; j=$(expr $j + 1)`

test-case/check-playback.sh
```diff
@@ -94,6 +100,22 @@ do
...
+                        if [ $j -eq ${#sleep_lst[@]} ]; then
+                            j=0
+	                fi
+                    else
+                        dlogi "platform doesn't support runtime pm or sleep time is turned off, sleep period keeps ${sleep_lst[0]} second(s)"
+                        time=${sleep_lst[0]}
```
marc-hb, line 113: `time` is a command, use something different to avoid confusion

test-case/check-playback.sh
```diff
@@ -94,6 +100,22 @@ do
...
+			dlogi "time $time"
+                    fi
+                    sleep $time
+                    pm=`sof-dump-status.py --dsp_status 0`
```
marc-hb, line 117: Backticks https://github.com/koalaman/shellcheck/wiki/SC2006

test-case/check-playback.sh
```diff
@@ -65,6 +68,9 @@ func_lib_setup_kernel_last_line
 func_lib_check_sudo
 func_pipeline_export $tplg "type:playback"
 
+sleep_lst=(0.5 1.9 2 2.1 2.9 3 3.1 5) #And it is 3s for SOF PCI device, 2s for HDA codec in future (not merged).
+j=0
```
marc-hb, line 72: There are 4 nested `for` loops. This new index is used only in the most inner loop but initialized outside the most outer loop, this is unusual. I think I guessed why but a comment explaining how the sleep_lst iteration works would be useful. A longer variable name may help too.

lgirdwood:
@RanderWang NAK to random values. Please fix this PR.

Bin-QA:
> ```shell
> RANDOM=42; echo $RANDOM $RANDOM $RANDOM
> ```
Interesting use, I never noticed, thank you.  
Normally I direct to use the `$RANDOM ` without to setup it,  
because I'm afraid to `unset RANDOM ` when I setup it
```
man bash
RANDOM Each  time  this parameter is referenced, a random integer between 0 and 32767 is generated.  The sequence of random numbers may be initialized by assigning a value to
              RANDOM.  If RANDOM is unset, it loses its special properties, even if it is subsequently reset.
```

RanderWang:
> @RanderWang NAK to random values. Please fix this PR.

@lgirdwood  ? this is PR is not mine.

ClarexZhou:
@lgirdwood Already updated to set delay with loop sleep_lst=(0.5 1.9 2 2.1 2.9 3 3.1 5).

test-case/check-playback.sh
```diff
@@ -94,6 +102,17 @@ do
                     dmesg > $LOG_ROOT/aplay_error_${dev}_$i.txt
                     dloge "aplay on PCM $dev failed at $i/$loop_cnt."
                     exit 1
+                else
+                    #sleep ${sleep_lst[@]} seconds before running aplay, then aplay can be run under dsp active or suspended status
+                    if [[ $(sof-dump-status.py --dsp_status 0) != "unsupported" ]] && [ "$sleep_ctl" -eq 1 ]; then
+                        sleep_time=${sleep_lst[$((j++ % sleep_len))]}
```
ranj063, line 108: @ClarexZhou what are we doing in this part (j++ % sleep_len)? Can we not make it simply use a random number between 1 and sleep_len to make this intuitive?

plbossart: let's be careful with random numbers: https://github.com/thesofproject/sof-test/issues/301

Random with fixed seed: good for CI
Random with variable but assignable seed: good for stress tests, still repeatable
Random without controlled seed: super stress tests, not repeatable.

ranj063: @plbossart we're not randomizing the sleep times themselves but just the index of which sleep time to pick from the pre-defined list. So it should be fine no?

plbossart: The net result is just as non-repeatable.

xiulipan:
close for outdated pr.

PR #135, opened 2020-03-19
test-case: add manual instructions to pause-suspend-resume-release test

stripes416:
The manual instructions for this test should be kept with the test itself for
reference should problems arise and the test needs to be made manual again for a
period of time.

### DIFF
--- a/test-case/check-pause-release-suspend-resume.sh
+++ b/test-case/check-pause-release-suspend-resume.sh
@@ -22,6 +22,32 @@
 ## Expect result:
 ##    no errors occur
 ##
+##
+## To run test manually:
+## Preconditions
+##     1. Device has ability to fully suspend.
+##         - BYTs cannot enter necessary suspend state.
+## Test Description
+##     Check for errors during test cycle of:
+##         playback -> pause -> suspend -> resume -> release cycles
+##     * By changing aplay call to an arecord call, you can manual test capture
+##       via the same method.
+## Test Case:
+##     1. Run in terminal 1:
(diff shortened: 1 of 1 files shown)


Bin-QA:
@stripes416 How about the different between [check-pause-release-suspend-resume.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/check-pause-release-suspend-resume.sh)

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -22,6 +22,30 @@
...
 ##    no errors occur
 ##
+##
+## To run test manually:
+## Preconditions
+##     1. Device has ability to fully suspend.
+##         - BYTs cannot enter necessary suspend state.
+## Test Description
+##     Check for errors during test cycle of:
+##         playback -> pause -> suspend -> resume -> release cycles
```
aiChaoSONG: here, can we change `playback` to `playback/capture`?

aiChaoSONG:
> @stripes416 What is the difference with [check-pause-release-suspend-resume.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/check-pause-release-suspend-resume.sh)?

Not the same test case

test-case/check-pause-release-suspend-resume.sh
stripes416: will do.

stripes416:
Updated manual instructions to include a note saying the method can also be used to test capture via arecord calls.

PR #141, opened 2020-03-20
case-lib: fix error parameter order of func_lib_get_random

Bin-QA:
parameter should format as: max, min; max; empty
older older is: min, max; max; empty


Bin-QA:
@xiulipan please help to review this patch

libinyang:
LGTM

PR #145, opened 2020-03-23
test-case: handle sof-logger errors in check-sof-logger.sh

aiChaoSONG:
Previously, we use the file size generated by sof-logger
to test if we have firmware log to export. But after a
logger change, sof-logger errors will output to generated
file, the size of the file is variant due to sof-logger
errors.
grep "errror|Error|ERROR“ from sof-logger's  stderr redirected file.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -47,23 +47,46 @@ if [[ ! -f $ldcFile ]]; then
 fi
 dlogi "Found file: $(md5sum $ldcFile|awk '{print $2, $1;}')"
 
-tmp_file=/tmp/$RANDOM.sof-logger.log
+rdnum=$RANDOM
+tmp_file=/tmp/$rdnum.logger.log
+err_tmp_file=/tmp/$rdnum.err_logger.log
 
 func_lib_check_sudo
 
 dlogi "Try to dump the dma trace log via sof-logger ..."
-dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2>&1 &"
-sudo $loggerBin -t -l $ldcFile -o $tmp_file 2>&1 &
+# sof-logger errors will output to $err_tmp_file
+dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
+sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &
 sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
 sudo pkill -9 $(basename $loggerBin) 2> /dev/null
 
-# get size of trace log$
-size=`du -k $tmp_file | awk '{print $1}'`
-if [[ $size -lt 1 ]]; then
-    dloge "No available log export."
+# check if we get any sof-logger errors
+logger_err=`grep -i "error" $err_tmp_file`
+if [[ $logger_err ]]; then
+    dloge "No available log to export due to sof-logger errors."
+    dlogi "Logger error BEG>>"
+    cat $err_tmp_file
+    dlogi "<<END Logger error"
     exit 1
 fi
+# get size of trace log$
+size=`du -k $tmp_file | awk '{print $1}'`
+fw_log_err=`grep -i "error" $tmp_file`
+# 100 here is log header size
+# only log header and no fw log
+if [[ $size -le 100 ]]; then
+    dloge "No available log to export."
+    exit 1
+# we catch error from fw log
+elif [[ $fw_log_err ]]; then
+    dloge "Errors in firmware log:"
+    dlogi "Log data BEG>>"
+    cat $tmp_file
+    dlogi "<<END log data"
+    exit 1
+fi
+# no error with sof-logger and no error in fw log
 dlogi "Log data BEG>>"
 cat $tmp_file
 dlogi "<<END log data"


aiChaoSONG:
sof-logger in sof repo seems need fix, because the output is ugly, let's wait that fix.

test-case/check-sof-logger.sh
```diff
@@ -48,23 +48,29 @@ fi
 dlogi "Found file: $(md5sum $ldcFile|awk '{print $2, $1;}')"
 
 tmp_file=/tmp/$RANDOM.sof-logger.log
+err_tmp_file=/tmp/err_logger.log
```
xiulipan: maybe we can also use the $RANDOM prefix to avoid duplicated issue?

test-case/check-sof-logger.sh
```diff
@@ -48,23 +48,29 @@ fi
...
-dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2>&1 &"
-sudo $loggerBin -t -l $ldcFile -o $tmp_file 2>&1 &
+# sof-logger errors will output to $err_tmp_file
+dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
+sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &
 sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
 sudo pkill -9 $(basename $loggerBin) 2> /dev/null
 
-# get size of trace log$
```
xiulipan: can we still keep the trace size? There will also be issue that FW do not send trace to kernel as we have tried to checked in previous. We will need the Error check before this.

test-case/check-sof-logger.sh
```diff
@@ -48,23 +48,29 @@ fi
...
+if [[ $logger_err ]]; then
+    dloge "No available log to export due to sof-logger errors."
+    sudo rm -rf $tmp_file $err_tmp_file
     exit 1
 fi
 dlogi "Log data BEG>>"
 cat $tmp_file
 dlogi "<<END log data"
+
+sudo rm -rf $tmp_file $err_tmp_file
```
xiulipan: Do we need to clean up the tmp files? I think keep them for later debug is better.

test-case/check-sof-logger.sh
```diff
@@ -48,23 +48,29 @@ fi
...
 
-# get size of trace log$
-size=`du -k $tmp_file | awk '{print $1}'`
-if [[ $size -lt 1 ]]; then
-    dloge "No available log export."
+# check if we get any sof-logger errors
+logger_err=`grep -E "Error|error|ERROR" $err_tmp_file`
+if [[ $logger_err ]]; then
+    dloge "No available log to export due to sof-logger errors."
+    sudo rm -rf $tmp_file $err_tmp_file
```
xiulipan: same here.

test-case/check-sof-logger.sh
```diff
@@ -48,23 +48,29 @@ fi
...
     exit 1
 fi
 dlogi "Log data BEG>>"
 cat $tmp_file
 dlogi "<<END log data"
+
+sudo rm -rf $tmp_file $err_tmp_file
+
```
Bin-QA:
```suggestion
dlogi "Log data BEG>>"
cat $tmp_file
dlogi "<<END log data"
# check if we get any sof-logger errors
if [ "$(grep -i 'error' $tmp_file)" ]; then
    dloge "No available log to export due to sof-logger errors."
    sudo rm -rf $tmp_file
    exit 1
fi
sudo rm -rf $tmp_file 
```

aiChaoSONG: let's first check sof-logger error, and then do log file size check. and as xiuli suggested, keep the log files for debug.

test-case/check-sof-logger.sh
```diff
@@ -47,23 +47,34 @@ if [[ ! -f $ldcFile ]]; then
...
-sudo $loggerBin -t -l $ldcFile -o $tmp_file 2>&1 &
+# sof-logger errors will output to $err_tmp_file
+dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
+sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &
 sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
 sudo pkill -9 $(basename $loggerBin) 2> /dev/null
 
+# check if we get any sof-logger errors
+logger_err=`grep -i "error" $err_tmp_file`
```
Bin-QA: if error record at `$tmp_file` ??

here 2 just for the stderr, but if 'error' dump into stdout

`logger_err=`grep -i "error"  $tmp_file $err_tmp_file``

xiulipan: @Bin-QA @aiChaoSONG The logger update is to output error message to both stderr and stdout. So we need to only check stderr to check if error happen.

aiChaoSONG: @Bin-QA @xiulipan if FW got error, the will be "error" keyword in the log,  checking both file, will break the logic.

Bin-QA: if in both, why need `$err_tmp_file` if not in both, miss `$tmp_file` whether will lost something?

aiChaoSONG:
@xiulipan Bin and I checked scripts that check FW and kernel, none of them checked FW log errors after boot. we may add this check in this script

test-case/check-sof-logger.sh
```diff
@@ -47,23 +47,43 @@ if [[ ! -f $ldcFile ]]; then
...
+if [[ $logger_err ]]; then
+    dloge "No available log to export due to sof-logger errors."
+    exit 1
```
Bin-QA: please add `$err_tmp_file` dump

aiChaoSONG: Previously,  logger error info is duplicated in stdout and stderr, that why I don't dump here, but now, only the stderr.  This suggestion is good, and done and repushed

PR #147, opened 2020-03-23
[WIP][DO NOT MERGE] CI Travis: add  shell code style shellcheck

xiulipan:
Add Travis CI to check all shell scripts with shellcheck
Add build status in README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: bash
+
+git:
+  depth: false
+
+before_install:
+  - sudo apt-get update
+  - sudo apt-get -y install shellcheck
+
+stages:
+  - code-check
+
+jobs:
+  include:
+    - stage: code-check
+      name: "bash shellcheck"
+      script: shellcheck */*.sh
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # SOF Test Case repo
 - [SOF Test Case repo](#sof-test-case-repo)
+  - [Status](#status)
   - [Enviroment set up](#enviroment-set-up)
     - [requirments](#requirments)
       - [apt packages](#apt-packages)
@@ -10,6 +11,10 @@
     - [tools](#tools)
   - [Folder description](#folder-description)
   - [Tools list description](#tools-list-description)
+
+## Status
+[![Build Status](https://travis-ci.org/thesofproject/sof-test.svg?branch=master)](https://travis-ci.org/thesofproject/sof-test)
+
 ## Enviroment set up
 ### requirments
 #### apt packages


xiulipan:
@marc-hb @Bin-QA @aiChaoSONG I think we may should have a clean up for the existing warning and errors.

README.md
```diff
@@ -10,6 +10,10 @@
     - [tools](#tools)
   - [Folder description](#folder-description)
   - [Tools list description](#tools-list-description)
+
+## Status
+[![Build Status](https://travis-ci.org/thesofproject/sof-test.svg?branch=master)](https://travis-ci.org/thesofproject/sof-test)
+
```
Bin-QA: can you also update the `Table of contents` at top?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - sudo apt-get update
+  - sudo apt-get -y install shellcheck
+
+stages:
+  - code-check
+
+jobs:
+  include:
+    - stage: code-check
+      name: "bash shellcheck"
```
marc-hb: shellcheck is not just for bash (I think it may even report some bashisms)

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+      script: shellcheck */*.sh
```
marc-hb: Anyone who fixes 1 line and sees 100s of warnings reported on code that has not changed will immediately start ignore this check and will never look back.

At the very least this check should run only on the files changed in the PR. Or even better: https://github.com/AtakamaLLC/lint-diffs (not tested)

xiulipan: @marc-hb so you think this will only test with patches or the whole files?
I can refine this.

xiulipan:
@marc-hb Thanks for the suggestion, I will try to do some update for this scripts to have check for patches.

marc-hb:
> https://github.com/AtakamaLLC/lint-diffs (not tested)

If this doesn't work or takes too long to get working, then as a second best run it on the entire `.sh` script but ONLY for the files that have been modified in the PR. For examples have a look at the `--diff-filter` in the battle-hardened https://github.com/zephyrproject-rtos/ci-tools/blob/master/scripts/check_compliance.py

marc-hb:
@xiulipan can you rebase on top of this commit: https://github.com/marc-hb/sof-test/blob/shellcheck-gitrange/tools/shellcheck-gitrange.bash

and then use in .travis.yml, something like `./tools/shellcheck-gitrange.bash ${TRAVIS_COMMIT_RANGE}`?

marc-hb:
With @xiulipan 's permission I'm taking this task.

marc-hb:
Completed in #183, thanks in advance for reviewing.

PR #148, opened 2020-03-24
tools: force remove snd_usb_audio

keqiaozhang:
snd_usb_audio remove failure on some platfroms, it shows in use
but ‘’by‘’ column is empty. So force remove it to avoid this issue.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -14,7 +14,7 @@ remove_module() {
     fi
 }
 
-remove_module snd_usb_audio
+remove_module -f snd_usb_audio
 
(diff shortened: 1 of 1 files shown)


tools/kmod/sof_remove.sh
```diff
@@ -14,7 +14,7 @@ remove_module() {
     fi
 }
 
-remove_module snd_usb_audio
+remove_module -f snd_usb_audio
```
libinyang: Why do we test snd_usb_audio kernel module remove? I think USB audio driver kernel module remove is out of scope.
BTW: have you checked pulseaudio?

PR #149, opened 2020-03-25
tools: fix direct load sof-kernel-dump.sh miss sof-get-kenrel-line.sh

Bin-QA:
Root cause is reset PATH in lib.sh, direct load sof-kernel-dump
skip reset PATH step, so miss sof-get-kernel-line in the PATH

### DIFF
--- a/tools/sof-kernel-dump.sh
+++ b/tools/sof-kernel-dump.sh
@@ -8,11 +8,20 @@ last_order=$[ $last_order + 1 ]
 log_file=${2:-"/var/log/kern.log"}
 # check whether target file exists
 [[ ! -f "$log_file" ]] && echo "$0 parameter 2 $log_file does not exist" && builtin exit 1
+# load helper script
+which sof-get-kernel-line.sh 2>&1 1>/dev/null
+if [ $? -eq 0 ]; then
+    kernel_line_script="sof-get-kernel-line.sh"
+else
+    cd $(dirname $0)
+    kernel_line_script=$PWD/sof-get-kernel-line.sh
+    cd $OLDPATH
+fi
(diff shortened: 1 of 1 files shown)


Bin-QA:
@aiChaoSONG please help to review this patch

tools/sof-kernel-dump.sh
```diff
@@ -8,11 +8,20 @@ last_order=$[ $last_order + 1 ]
 log_file=${2:-"/var/log/kern.log"}
 # check whether target file exists
 [[ ! -f "$log_file" ]] && echo "$0 parameter 2 $log_file does not exist" && builtin exit 1
+# load help script
```
aiChaoSONG: load helper scirpt

tools/sof-kernel-dump.sh
```diff
@@ -8,11 +8,20 @@ last_order=$[ $last_order + 1 ]
...
+which sof-get-kernel-line.sh 2>&1 1>/dev/null
```
aiChaoSONG: can `which` find the script if it is not in PATH?

Bin-QA: no,  `which - locate a command`

aiChaoSONG: which returns the pathnames of the files (or links) which would be exe‐
       cuted in the current environment, had its arguments been given as  com‐
       mands  in a strictly POSIX-conformant shell.  ***It does this by searching
       the PATH for executable files matching the names of the  arguments.  It
       does not canonicalize path names.***
this is from `man which`, I am afraid you can not locate sof-get-kernel-line.sh, because it only search folders in PATH.

PR #150, opened 2020-03-25
case-lib: opt.sh add sof-test branch and commit info

Bin-QA:
store current sof-test version information


Bin-QA:
@xiulipan please help to review this patch

PR #151, opened 2020-03-25
test-case: fix log redirect issue of check-sof-logger.sh

aiChaoSONG:
Use bash -c to run sof-logger command to avoid redirect
issue of hijacked sudo.

The output size of 'du -k' is block size, and block size
for firmware log header is 4.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -56,7 +56,7 @@ func_lib_check_sudo
 dlogi "Try to dump the dma trace log via sof-logger ..."
 # sof-logger errors will output to $err_tmp_file
 dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
-sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &
+sudo bash -c "'$loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &'"
 sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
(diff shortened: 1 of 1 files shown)


test-case/check-sof-logger.sh
```diff
@@ -56,7 +56,7 @@ func_lib_check_sudo
 dlogi "Try to dump the dma trace log via sof-logger ..."
 # sof-logger errors will output to $err_tmp_file
 dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
-sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &
+bash -c "sudo '$loggerBin' -t -l '$ldcFile' -o '$tmp_file' 2> '$err_tmp_file' &"
```
Bin-QA:
```suggestion
sudo bash -c "'echo $loggerBin -t -l $ldcFile -o $tmp_file 2>$err_tmp_file'"
```

aiChaoSONG: good idea, done

test-case/check-sof-logger.sh
```diff
@@ -75,7 +75,7 @@ size=`du -k $tmp_file | awk '{print $1}'`
 fw_log_err=`grep -i "error" $tmp_file`
 # 100 here is log header size
```
wenqingfu: please update the comment accordingly as well.

aiChaoSONG: done

aiChaoSONG:
@xiulipan Ping for merge

PR #154, opened 2020-03-30
test-case: add kernel log check after module insertion

aiChaoSONG:
Previously, we didn't check dmesg log after module insertion, so test case may passed even there is error in dmesg after module insertion. This patch fixes that condition.

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -59,8 +59,10 @@ do
         # Here we pass a hardcoded 0 to python script, and need to ensure
         # DSP is the first audio pci device in 'lspci', this is true unless
         # we have a third-party pci sound card installed.
-        [[ $(sof-dump-status.py --dsp_status 0) == "unsupported" ]] &&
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@xiulipan Ping for merge

test-case/check-kmod-load-unload.sh
```diff
@@ -89,6 +89,9 @@ do
     ## - 2a: check for errors after insertion
     dlogi "checking for general errors after kmod insert with sof-kernel-log-check tool"
     sof-kernel-log-check.sh $KERNEL_LAST_LINE
+    [[ $? -ne 0 ]] && \
+        dloge "Found error(s) in kernel log after module insertion" && \
```
xiulipan: no need to new line here. This look worse then a long line.

libinyang:
LGTM

test-case/check-kmod-load-unload.sh
aiChaoSONG: replace with if else, and align code style of the whole file

PR #155, opened 2020-03-30
kmod: sof_remove.sh: add snd_soc_max98373 support

keqiaozhang:
Update for machine driver change.


aiChaoSONG:
@keqiaozhang You may need to also add this in sof_insert.sh. because snd_soc_max98373 may not be auto loaded. you can take snd_soc_max98090 as an example.

PR #157, opened 2020-03-30
tools: clean up ignore_str for sof-kernel-log-check.sh

aiChaoSONG:
Some of the items in ignore_str is deprecated, and some new are needed. This patch do a clean up on ignore_str according to current kernel and sof status.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -3,7 +3,7 @@
 begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
-ignore_str="hda_sdw_acpi_scan failed|ACPI scan error|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init|iteration [01]"
+ignore_str="hda_sdw_acpi_scan failed|ACPI scan error|error: debugfs write failed to idle -16|error: status|iteration [01]"
 project_key="sof-audio"
 
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -3,7 +3,7 @@
 begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
-ignore_str="ACPI scan error|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init|iteration [01]"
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
This PR should be merged along side https://github.com/thesofproject/sof-test/pull/154, or the kmod test case will fail due to new ignore item(mainly sdw acpi scan error on none soundwire platforms) not ignored.

tools/sof-kernel-log-check.sh
```diff
@@ -28,8 +28,10 @@ else
 fi
 
 if [ "$err" ]; then
-    echo "Caught error: $err"
-    echo "kernel log has error(s)"
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught dmesg error"
+    echo "===========================>>"
+    echo $err
```
Bin-QA: please use `echo "$err"`, it will keep string format

tools/sof-kernel-log-check.sh
```diff
@@ -3,7 +3,7 @@
 begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
-ignore_str="no reply expected,|error: debugfs write failed to idle -16|error: status |error: cl_dsp_init|iteration [01]"
+ignore_str="hda_sdw_acpi_scan failed|ACPI scan error|error: debugfs write failed to idle -16|error: status|iteration [01]"
```
Bin-QA: we don't need `no reply expected`?  
I remember catch this information with `error` keyword from `dmesg`  

So here you need to confirm which ignore keyword is you want ignored  
when combine `$err_str`, except you modify the `grep` logic to `ignore` as 1st order

tools/sof-kernel-log-check.sh
aiChaoSONG: sure

tools/sof-kernel-log-check.sh
aiChaoSONG: @Bin-QA Oh, for the 'no reply expected' error, it's a long story. Previously, We have ROM message propagate to IRQ thread, this ROM message is not initiated by the kernel, thus not contains any reply, that's why the message is in dmesg. After a kernel fix, ROM message never goes to IRQ thread, so, this message will never appear in dmesg, if we see it, then there must be some error, we should not ignore this message any more.
For the "error: cl_dsp_init", it is fixed in the kernel side, too. If seen later, there should be errors.
For the new added "hda_sdw_acpi_scan failed|ACPI scan error", This error is seen on platform without soundwire, it should not be regarded as an error.
So, the change should be fine.

PR #159, opened 2020-03-31
[IN DISCUSSION, NOT MERGE]case-lib: redirect nohup stdout and stderr to /dev/null

aiChaoSONG:
If only redirect nohup stdout to /dev/null, we will see
'nohup: redirecting stderr to stdout' message from nohup
when manually run './check-kmod-load-unload.sh -l1'.

Which will make the test log looks ugly.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -95,7 +95,7 @@ func_lib_restore_pulseaudio()
     do
         user=${line%% *}
         cmd=${line#* }
-        nohup sudo -u $user $cmd >/dev/null &
+        nohup sudo -u $user $cmd &>/dev/null &
     done
     # now wait for the pulseaudio restore in the ps process
     timeout=10


libinyang:
LGTM

aiChaoSONG:
> I don't this this patch will meet your requirement, because
> when you add '&' for nohup command it will create 'nohup.out' file for stdout & stderr

@Bin-QA Any suggestions？ because “nohup: redirecting stderr to stdout” in the test log， which makes the log ugly， and I test this pr on helios， the output of nohup disappeared.

Bin-QA:
> @Bin-QA Any suggestions？ because “nohup: redirecting stderr to stdout” in the test log， which makes the log ugly， and I test this pr on helios， the output of nohup disappeared.
because you will create 'nohup.out' at your current folder

xiulipan:
@aiChaoSONG will this bring us trouble with debug? As error message is gone

aiChaoSONG:
@Bin-QA When we run sof-test, the repo is cloned to local, so all command is running locally, then why we need a nohup here?

Bin-QA:
@aiChaoSONG because here need recreate pulseaudio command,   
Notice: here recreate pulseaudio is all command , so it is not limited the system server  
So when you  terminal quit, the command attach for your terminal will be killed  
the `nohup` command use to avoid terminal/parent process quit problem

marc-hb:
> Which will make the test log looks ugly.

Test logs are not meant to be beautiful, they have to be informative and not too noisy. That does not seem like a lot of noise, is it?

Logs that are really noisy should be redirected to a file, tests should almost never redirect anything to >/dev/null. In case of a failure any information could be useful.

I have a larger question: why is this script not using `systemctl`? https://docs.ubuntu.com/core/en/stacks/audio/pulseaudio/docs/using-pulseaudio
Is this script not running on Ubuntu?

marc-hb:
> I have a larger question: why is this script not using systemctl? https://docs.ubuntu.com/core/en/stacks/audio/pulseaudio/docs/using-pulseaudio


OK, this doc is obsolete, sorry. I found this more accurate one, tested it and verified it is up to date wrt. Ubuntu 18.04
https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/Running/

Why is the nohup needed at all? I ran `pulseaudio --start` over ssh with no autospawn and it worked without nohup.

Also wondering why use `kill -KILL $(ps)` instead of a simpler `pulseaudio --kill`?

Why use `ps` instead of `pulseaudio --check`?

Why try to extract command line parameters from `ps` as opposed to rely on regular config files?

Also why manually restart pulseaudio after restoring the autospawn which already starts it when needed?

aiChaoSONG:
@Bin-QA Could you please help to answer Marc's questions?

Bin-QA:
@marc-hb because we are not sure target machine how to configure the pulseaudio   
the end user (current just us) maybe not start pulseaudio or something   
so  it detect in `ps`  if target machine configure more `pulseaudio` version,   
how to trace them and restore them also is the problem (normally it will refer configure and test pulseaudio)   
we want the script in here is more common to use

marc-hb:
I'm not sure I follow, are you afraid some operating system(s) hide(s) some pulseaudio --parameters outside the normal pulseaudio configuration files?

How many different operating system versions is `sof-test` regularly used on? I didn't see any mentioned  in https://github.com/thesofproject/sof-test/blob/master/README.md

xiulipan:
@aiChaoSONG @macchian any update? Do we still need this

aiChaoSONG:
No, let's close this, the issue I want to address by this PR is not reproduced. maybe already fixed by other PR

PR #160, opened 2020-04-01
case-lib: rename pulseaudio before disabling it

aiChaoSONG:
Pulseaudio may be started by many other applications,
which will cause some of the test cases failed.

This patch renames pulseaudio in func_lib_disable_pulseaudio
and rename it back in func_lib_restore_pulseaudio

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -57,6 +57,7 @@ func_lib_check_sudo()
 }
 
 declare -ag PULSECMD_LST
+declare -ag PULSE_PATHS
 
 func_lib_disable_pulseaudio()
 {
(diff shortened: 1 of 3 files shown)


aiChaoSONG:
tested on helios, pulseaudio is renamed to pulseaudio.bak before test and restored to pulseaudio after test.

paulstelian97:
What happens if there is a KP or anything else that will cause the rename back to the original name to fail?

aiChaoSONG:
> What happens if there is a KP or anything else that will cause the rename back to the original name to fail?

@paulstelian97 We only disable PA in kmod test cases, failed to restore PA only influence PA related test, and no negative impact on other test cases, even the kmod test cases. And till now, we don't have any PA test cases. If we are going to add some PA test cases in the future, we will firstly check PA and do the rename again.

case-lib/lib.sh
```diff
@@ -90,6 +94,9 @@ func_lib_restore_pulseaudio()
     [[ "${#PULSECMD_LST[@]}" -eq 0 ]] && return
     func_lib_check_sudo
     [[ -f $PULSEAUDIO_CONFIG.bak ]] && sudo mv -f $PULSEAUDIO_CONFIG.bak $PULSEAUDIO_CONFIG
```
Bin-QA, line 96: this line need to remove

aiChaoSONG: Yes, but I am afraid that sometimes if we failed to do the rename, the config can still be a backup solution. Let's  leave it as it is.

case-lib/lib.sh
```diff
@@ -67,13 +68,15 @@ func_lib_disable_pulseaudio()
...
-    [[ ! -f $PULSEAUDIO_CONFIG.bak ]] && sudo cp $PULSEAUDIO_CONFIG $PULSEAUDIO_CONFIG.bak -f
-    # because hijack the sudo command
-    if [ "$(grep 'autospawn' $PULSEAUDIO_CONFIG)" ]; then
-        sudo "sed -i '/autospawn/cautospawn = no' $PULSEAUDIO_CONFIG"
-    else
-        sudo "bash -c 'echo \"autospawn = no\" >> $PULSEAUDIO_CONFIG'"
-    fi
+    # get all running pulseaudio paths
+    PULSE_PATHS=( $(ps -C pulseaudio -o cmd --no-header | awk '{print $1}') )
+    for PA_PATH in $PULSE_PATHS
```
Bin-QA, line 73: `$PULSE_PATHS` is array, so you need `${PULSE_PATHS[@]}` to go though it

case-lib/lib.sh
```diff
@@ -89,7 +92,14 @@ func_lib_restore_pulseaudio()
 {
     [[ "${#PULSECMD_LST[@]}" -eq 0 ]] && return
     func_lib_check_sudo
-    [[ -f $PULSEAUDIO_CONFIG.bak ]] && sudo mv -f $PULSEAUDIO_CONFIG.bak $PULSEAUDIO_CONFIG
+    # restore pulseaudio
+    for PA_PATH in $PULSE_PATHS
```
Bin-QA, line 96: `$PULSE_PATHS` is array, so you need `${PULSE_PATHS[@]}` to go though it

case-lib/lib.sh
aiChaoSONG: thanks, done

PR #161, opened 2020-04-02
test-case: check-sof-logger to fix issue of log file

Bin-QA:
1. fix some test case sequence let sof-logger just dump
HEAD which size is 4K by 'du' command
2. fix detect error keyword overkill to avoid sof-logger
dump really error of sof
3. refine case quit process

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -47,47 +47,53 @@ if [[ ! -f $ldcFile ]]; then
 fi
 dlogi "Found file: $(md5sum $ldcFile|awk '{print $2, $1;}')"
 
-rdnum=$RANDOM
-tmp_file=/tmp/$rdnum.logger.log
-err_tmp_file=/tmp/$rdnum.err_logger.log
+data_file=$LOG_ROOT/logger.data.log
+error_file=$LOG_ROOT/logger.error.log
 
 func_lib_check_sudo
 
 dlogi "Try to dump the dma trace log via sof-logger ..."
-# sof-logger errors will output to $err_tmp_file
-dlogc "sudo $loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &"
-sudo bash -c "'$loggerBin -t -l $ldcFile -o $tmp_file 2> $err_tmp_file &'"
+# sof-logger errors will output to $error_file
+dlogc "sudo $loggerBin -t -l $ldcFile -o $data_file 2> $error_file &"
+sudo bash -c "'$loggerBin -t -l $ldcFile -o $data_file 2> $error_file &'"
 sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
 sudo pkill -9 $(basename $loggerBin) 2> /dev/null
 
+func_logger_exit()
+{
+    local code=$1 type=${2:-data}
+    dlogi "Log $type BEG>>"
+    cat $LOG_ROOT/logger.$type.log
+    dlogi "<<END $type data"
+    exit $code
+}
+
 # check if we get any sof-logger errors
-logger_err=`grep -i "error" $err_tmp_file`
+logger_err=`grep -i "error" $error_file`
 if [[ $logger_err ]]; then
     dloge "No available log to export due to sof-logger errors."
-    dlogi "Logger error BEG>>"
-    cat $err_tmp_file
-    dlogi "<<END Logger error"
-    exit 1
+    func_logger_exit 1 'error'
 fi
 # get size of trace log$
-size=`du -k $tmp_file | awk '{print $1}'`
-fw_log_err=`grep -i "error" $tmp_file`
+size=`du -k $data_file | awk '{print $1}'`
+# '\.c\:[1-9]' to filter like '.c:6' this type keyword like:
+# [3017136.770833]  (11.302083) c0 SA  src/lib/agent.c:65  ERROR validate(), ll drift detected, delta = 25549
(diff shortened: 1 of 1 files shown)


test-case/check-sof-logger.sh
```diff
@@ -61,33 +61,34 @@ sleep 2
 dlogc "sudo pkill -9 $(basename $loggerBin)"
 sudo pkill -9 $(basename $loggerBin) 2> /dev/null
 
+func_logger_exit()
+{
+    local file=$1 code=$2 type=${3:-data}
+    dlogi "Log $type BEG>>"
+    cat $file
+    dlogi "<<END $type data"
+    mv /tmp/$rdnum.*.log $LOG_ROOT
```
xiulipan, line 70: why need to use random name for this logger files? We have a workspace for each test case, no need random name now. Just save it in the LOG_DIR is OK

test-case/check-sof-logger.sh
```diff
@@ -61,33 +61,34 @@ sleep 2
...
-    dlogi "Logger error BEG>>"
-    cat $err_tmp_file
-    dlogi "<<END Logger error"
-    exit 1
+    func_logger_exit $err_tmp_file 1 'error'
 fi
 # get size of trace log$
 size=`du -k $tmp_file | awk '{print $1}'`
-fw_log_err=`grep -i "error" $tmp_file`
+fw_log_err=`grep -i "error" $tmp_file|grep -v '\.c\:[1-9]'`
```
xiulipan, line 82: Can you give a simple comment about why this grep can get the right result?

test-case/check-sof-logger.sh
```diff
@@ -47,47 +47,48 @@ if [[ ! -f $ldcFile ]]; then
...
-    dlogi "<<END Logger error"
-    exit 1
+    func_logger_exit 1 'error'
 fi
 # get size of trace log$
-size=`du -k $tmp_file | awk '{print $1}'`
-fw_log_err=`grep -i "error" $tmp_file`
+size=`du -k $data_file | awk '{print $1}'`
+# '\.c\:[1-9]' to filter like '.c:6' this type keyword like:
+# [  3017136.770833] (       11.302083) c0 SA                          src/lib/agent.c:65    ERROR validate(), ll drift detected, delta = 25549
```
aiChaoSONG, line 81: A direct copy is not good here, this line is too long, can you delete some of the space in the line, make it more compact?
`[3017136.770833]  (11.302083)  c0  SA  src/lib/agent.c:65  ERROR validate(), ll drift detected, delta = 25549`
I recommend two spaces between each item.

test-case/check-sof-logger.sh
```diff
@@ -47,47 +47,48 @@ if [[ ! -f $ldcFile ]]; then
...
+fw_log_err=`grep -i "error" $data_file|grep -v '\.c\:[1-9]'`
```
aiChaoSONG, line 82: for the pipe, I prefer leave some space, this make the code easy to read.
```suggestion
fw_log_err=`grep -i "error" $data_file | grep -v '\.c\:[1-9]'`
```

test-case/check-sof-logger.sh
aiChaoSONG: this is outdated

Bin-QA:
@xiulipan @aiChaoSONG  @fredoh9  please help to review

test-case/check-sof-logger.sh
```diff
@@ -47,47 +47,53 @@ if [[ ! -f $ldcFile ]]; then
...
 elif [[ $fw_log_err ]]; then
     dloge "Errors in firmware log:"
-    dlogi "Log data BEG>>"
-    cat $tmp_file
-    dlogi "<<END log data"
-    exit 1
+    func_logger_exit 1
 fi
+
+if [ "$(grep -i 'error' $data_file)" ]; then
```
libinyang, line 94: Why do we exit 1 when meeting `grep -i "error" $data_file | grep -v '\.c\:[1-9]'` and others are OK?

test-case/check-sof-logger.sh
```diff
@@ -47,47 +47,53 @@ if [[ ! -f $ldcFile ]]; then
...
-size=`du -k $tmp_file | awk '{print $1}'`
-fw_log_err=`grep -i "error" $tmp_file`
+size=`du -k $data_file | awk '{print $1}'`
+# '\.c\:[1-9]' to filter like '.c:6' this type keyword like:
+# [3017136.770833]  (11.302083) c0 SA  src/lib/agent.c:65  ERROR validate(), ll drift detected, delta = 25549
+fw_log_err=`grep -i "error" $data_file | grep -v '\.c\:[1-9]'`
 # 4 here is log header size
 # only log header and no fw log
-if [[ $size -le 4 ]]; then
+if [[ $size -lt 4 ]]; then
```
libinyang, line 85: This should never happen now. If `sof-logger` run successfully, the size always >= 4.

Bin-QA: yes, here fix '=' this problem

test-case/check-sof-logger.sh
Bin-QA: the other means sof-logger work fine

aiChaoSONG: The test case only check if sof-logger function well, don't care if we have error in logger. other test case will do the check.

test-case/check-sof-logger.sh
aiChaoSONG: Previously, if we only catch header, we fail the test case, this is wrong, because we are testing sof-logger function. so now, even if we only catch header, we tag test case passed.  but ,if nothing is catched, that should be wrong.

PR #162, opened 2020-04-02
kmod: fix snd_soc_rt1308 not loaded error

aiChaoSONG:
For platforms with only snd_soc_rt1308_sdw,
the snd_soc_rt1308 module will escape the check
of 'if lsmod | grep "$MODULE" &> /dev/null'.
Removing not existed snd_soc_rt1308 causes error.

This patch does a full match with grep, and thus
avoids the error condition.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -6,7 +6,7 @@ remove_module() {
 
     local MODULE="$1"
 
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@xiulipan @Bin-QA another solution is to do full match with grep, which do you prefer?

PR #166, opened 2020-04-03
case-lib: add topology graph generation support 

aiChaoSONG:
The tplgtool.py for graph generation is done, and now add the graph generation code in exit function.
If you have some other idea on AT WHAT test stage we should generate tplg graph, please add your comments.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -7,6 +7,13 @@ function exit()
 {
     local exit_status=${1:-0}
 
+    # To generate topology graph, test case should export $tplg for us
+    if [[ -n "$tplg" ]]; then
+        tplgtool.py -d graph -D $LOG_ROOT $tplg &> /dev/null
+        [[ "$?" != "0" ]] && dloge "Failed to generate topology graph"
+        unset tplg
+    fi
+
     # when sof logger collect is open
     if [ "X$SOF_LOG_COLLECT" == "X1" ]; then
         # when error occurs, exit and catch etrace log


case-lib/hijack.sh
```diff
@@ -7,6 +7,11 @@ function exit()
 {
     local exit_status=${1:-0}
 
+    # To generate topology graph, test case should export $tplg for us
+    tplgtool.py -d graph -D $LOG_ROOT $tplg > /dev/null
+    [[ "$?" != "0" ]] && dloge "Failed to generate topology graph"
+    unset tplg
+
```
Bin-QA, line 14: this place is good but who pass-down the `$tplg`  
almost our case define the `tplg` but still have case not define it or not use this name  
how about move it into `func_pipeline_export` here you can got the TPLG file name  
almost of our case need TPLG will load this case

aiChaoSONG: I want to add a feature to highlight the pipeline we are testing on when error occurs later, so I think putting the generation of tplg graph at the exit function is a better way. can we modify all the cases to define tplg?
On who pass down $tplg: If case want to exit, the case export $tplg and some other infomation, eg, test error pipeline, to us?

Bin-QA: > I want to add a feature to highlight the pipeline we are testing on when error occurs later, so I think putting the generation of tplg graph at the exit function is a better way. can we modify all the cases to define tplg?  

yes, so you need to check whether each use TPLG case already define tplg   
>On who pass down $tplg: If case want to exit, the case export $tplg and some other infomation, eg, test error pipeline, to us?

you also need to add `tplg` is not define logic  
because not define is allow for some case like 'check-sof-logger.sh' etc.

Issue #138, opened 2020-03-20
[BUG][CML]Test not finished all iterations when test "multiple-pause-resume-25.sh"


Liviali155:
on  CML Mantis 
the daily stress test case  "multiple-pause-resume-25.sh"
test not finished, and the case result is fail. I am not sure if the device has no response or the script error
http://sof-tcr.sh.intel.com/#/tests/0e52af57-0b99-4306-9b62-d139c78eb45a/detail
<img width="669" alt="Capture" src="https://user-images.githubusercontent.com/56707932/77131685-7f812f00-6a97-11ea-84b4-7e2ba2f8f4cf.PNG">

keqiaozhang:
This issue is not related to the timeout. timeout is set as 900s, but this case only run 475s. So maybe other reasons cased this issue.
@Bin-QA  Please have a check.

Bin-QA:
from log:
```
2020-03-19 23:25:05 UTC [REMOTE_INFO] wait for expect process finished

spawn aplay -D hw:0,0 -r 48000 -c 2 spawn-f  S16_LE
aplay  -vv-D  hw:0,5-i  -r /dev/zero48000  -q-c
 2 -f S16_LE -vv -i /dev/zero -q
2020-03-19 23:25:05 UTC [REMOTE_ERROR] Still have expect process not finished after wait for 250
```
mapping script:
```
        for idx in $(echo $pipeline_combine_str)
        do
            func_pause_resume_pipeline $idx
            pid_lst=(${pid_lst[*]} $!)
        done
        # wait for expect script finished
        dlogi "wait for expect process finished"
        i=0
        while [ "$(pidof expect)" ]
        do
            i=$[ $i + 1 ]
            sleep 1s
            [[ $i -ge $max_wait_time ]] && break
        done
        # fix aplay/arecord last output
        echo 
        if [ "$(pidof expect)" ]; then
            dloge "Still have expect process not finished after wait for $max_wait_time"
            # now dump process
            ps -ef |grep -E 'aplay|arecord'
            exit 1
        fi
```
so here block logic is not work, I will check more information for it,   
normally it also means the pidof expect is delay to start  
maybe stress test cause cpu delay  

Current I modify the logic of delay to fix it

PR #169, opened 2020-04-07
test-case: multiple-pause-resume modify wait logic

Bin-QA:
change wait expect finish logic from detect expect at begin
to fix issue #138


libinyang:
Good catch. 1s should be enough now as it is already be tested in other scripts.
LGTM.

Bin-QA:
@stripes416 please help to review this patch

PR #170, opened 2020-04-07
test-case: check-sof-logger use file content instead of size of file

Bin-QA:
use file content keyword instead of file size detect to judgment data
file already dumped

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -75,14 +75,13 @@ if [[ $logger_err ]]; then
     dloge "No available log to export due to sof-logger errors."
     func_logger_exit 1 'error'
 fi
-# get size of trace log$
-size=`du -k $data_file | awk '{print $1}'`
+
 # '\.c\:[1-9]' to filter like '.c:6' this type keyword like:
 # [3017136.770833]  (11.302083) c0 SA  src/lib/agent.c:65  ERROR validate(), ll drift detected, delta = 25549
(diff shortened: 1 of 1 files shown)


test-case/check-sof-logger.sh
```diff
@@ -75,14 +75,13 @@ if [[ $logger_err ]]; then
     dloge "No available log to export due to sof-logger errors."
     func_logger_exit 1 'error'
 fi
-# get size of trace log$
-size=`du -k $data_file | awk '{print $1}'`
+
 # '\.c\:[1-9]' to filter like '.c:6' this type keyword like:
 # [3017136.770833]  (11.302083) c0 SA  src/lib/agent.c:65  ERROR validate(), ll drift detected, delta = 25549
 fw_log_err=`grep -i "error" $data_file | grep -v '\.c\:[1-9]'`
```
wenqingfu, line 81: @xiulipan @Bin-QA , will a normal sof-logger output always contain the pattern of '\.c\:[1-9]'?  Can we just use `grep -v '\.c\:[1-9]'` $data_file` to replace the header field check code below?

Bin-QA: But I meet the file content without header line, it is the root cause why need check whether have content to dump

xiulipan: @Bin-QA I think we may better not have any hard coded style check here. As there is already a DEBUG ABI for logger, I believe the logger style may change soon.

PR #171, opened 2020-04-07
test-case: check-reboot fix sub-case parameter miss

Bin-QA:
sub-case: verify-tplg-binary.sh need TPLG as parameter to pass
test-case


Bin-QA:
@stripes416 @fredoh9 @xiulipan please help to review

PR #172, opened 2020-04-08
 test-case: add firmware load subtest for kmod test

aiChaoSONG:
As the kernel has made its log less verbose, the key
word 'firmware boot complete' we use is not in the
less verbose log now. So remove the check of this
keyword.

Instead, add verify-sof-firmware-load.sh subtest
to check if firmware is loaded successfully after
module insertion.

tested on mantis, glk, PASSED

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -97,15 +97,12 @@ do
         exit 1
     fi
 
-    dlogi "checking for fw_boot success"
-    keyword_info=$(dmesg |grep sof-audio | grep 'boot complete')
-    if [[ ! "$keyword_info" ]]; then
-        dloge "Error: Boot Complete not found in dmesg, fw_boot empty"
-        exit 1
-    fi
+    dlogi "checking if firmware is loaded successfully"
+    $(dirname ${BASH_SOURCE[0]})/verify-sof-firmware-load.sh > /dev/null
+    [[ $? -ne 0 ]] && dloge "Falied to load firmware after module insertion" && exit 1
 
     # successful remove/insert module pass
-    dlogi "==== completed boot firmware: $idx of $loop_cnt ===="
+    dlogi "==== firmware boot complete: $idx of $loop_cnt ===="
 
     # pulseaudio deamon will detect the snd_sof_pci device after 3s
     # so after 2s snd_sof_pci device will in used status which is block current case logic


aiChaoSONG:
check https://github.com/thesofproject/linux/pull/1980 and Kai's email for more info.

test-case/check-kmod-load-unload.sh
```diff
@@ -97,15 +97,15 @@ do
         exit 1
     fi
 
-    dlogi "checking for fw_boot success"
-    keyword_info=$(dmesg |grep sof-audio | grep 'boot complete')
+    dlogi "checking if firmware boot successfully"
+    keyword_info=$(dmesg |grep sof-audio | grep 'Firmware debug build')
```
xiulipan, line 101: I would suggest to use `Firmware info`
```suggestion
    keyword_info=$(dmesg |grep sof-audio | grep 'Firmware info')
```
`Firmware debug build` may not appear if FW is not debug build version.

aiChaoSONG: No, It will always appear. There are still some error checks after 'Firmware info', so it is not as robust as using 'Firmware debug build'. Though there is a 'debug' in the key word, but it has nothing to do with debug. And I have checked other dev_info logs, 'Firmware debug build' is the best.
```
	if (ready->flags & SOF_IPC_INFO_BUILD) {
		dev_info(sdev->dev,
			 "Firmware debug build %d on %s-%s - options:\n"
			 " GDB: %s\n"
			 " lock debug: %s\n"
			 " lock vdebug: %s\n",
			 v->build, v->date, v->time,
			 (ready->flags & SOF_IPC_INFO_GDB) ?
				"enabled" : "disabled",
			 (ready->flags & SOF_IPC_INFO_LOCKS) ?
				"enabled" : "disabled",
			 (ready->flags & SOF_IPC_INFO_LOCKSV) ?
				"enabled" : "disabled");
	}
```

aiChaoSONG:
@xiulipan @Bin-QA We have searched failure logs, another solution is to remove the check of fw boot success status. Vote?

Bin-QA:
@aiChaoSONG I'm not sure the kmod reload logic, how about this part replace by sub-case: verify-sof-firmware-load.sh

aiChaoSONG:
> @aiChaoSONG I'm not sure the kmod reload logic, how about this part replace by sub-case: verify-sof-firmware-load.sh

@Bin-QA Genius, the check I have done in the PR is also in verify-sof-firmware-load.sh, and this script also checked firmware info. this is a perfect solution.

test-case/check-kmod-load-unload.sh
```diff
@@ -97,15 +97,12 @@ do
...
     fi
 
-    dlogi "checking for fw_boot success"
-    keyword_info=$(dmesg |grep sof-audio | grep 'boot complete')
-    if [[ ! "$keyword_info" ]]; then
-        dloge "Error: Boot Complete not found in dmesg, fw_boot empty"
-        exit 1
-    fi
+    dlogi "checking if firmware is loaded successfully"
+    $(dirname ${BASH_SOURCE[0]})/verify-sof-firmware-load.sh > /dev/null
```
libinyang, line 101: verify-sof-firmware-load.sh shouldn't be a separate test case?

test-case/check-kmod-load-unload.sh
```diff
@@ -97,15 +97,12 @@ do
...
+    [[ $? -ne 0 ]] && dloge "Falied to load firmware after module insertion" && exit 1
 
     # successful remove/insert module pass
-    dlogi "==== completed boot firmware: $idx of $loop_cnt ===="
+    dlogi "==== firmware boot complete: $idx of $loop_cnt ===="
```
libinyang, line 105: nitpick: complete => completes

test-case/check-kmod-load-unload.sh
```diff
@@ -97,15 +97,12 @@ do
...
 
-    dlogi "checking for fw_boot success"
-    keyword_info=$(dmesg |grep sof-audio | grep 'boot complete')
-    if [[ ! "$keyword_info" ]]; then
-        dloge "Error: Boot Complete not found in dmesg, fw_boot empty"
-        exit 1
-    fi
+    dlogi "checking if firmware is loaded successfully"
+    $(dirname ${BASH_SOURCE[0]})/verify-sof-firmware-load.sh > /dev/null
+    [[ $? -ne 0 ]] && dloge "Falied to load firmware after module insertion" && exit 1
```
libinyang, line 102: So we don't check the module load successfully or not?

aiChaoSONG: @libinyang  The problem is HERE: the kernel has made the log less verbose, we don't have the keyword "firmware boot complete" to check now, it's disappeared. see  https://github.com/thesofproject/linux/pull/1980. So, we should use a common log keyword both exist in verbose and less verbose log mode. Common keyword candadates:`Firmware debug build | Firmware info`.  They are both checked in `verify-sof-firmware-load.sh`, so, Instead of checking keywords, Use a subtest to do the same thing and ensure firmware is OK after module insertion.

test-case/check-kmod-load-unload.sh
aiChaoSONG: as is merged, may fixed later

test-case/check-kmod-load-unload.sh
aiChaoSONG: before firmware load check, we have grep any errors in dmesg, so a check on module load successfully seems not so necessary.

libinyang: @aiChaoSONG Sorry for my comments after merge. I was reviewing your patch and writing the comments and after I finished the review, I found it's already merged. : )
Yes, I understand that the keyword is removed because of the kernel change and this makes our old method not work. What I mean is even firmware is loaded successfully, kernel module load may still fail.

aiChaoSONG: @libinyang We may add the check later, the possible failure you mentioned is not checked before.

PR #174, opened 2020-04-08
case-lib: misc fix on topology path

aiChaoSONG:
PR test specify topology like this: TPLG=sof-cml-rt1011-rt5682.tplg, under this condition, the tplgtool.py can not find the topology we specified, thus, we will get "Failed to generate topology graph" error.
This patch add a function to get topology path, and use it everywhere it can be used to replace old ugly code.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -8,19 +8,14 @@ function exit()
     local exit_status=${1:-0}
 
     # To generate topology graph, test case should export $tplg for us
-    if [[ -n "$tplg" ]]; then
-        if [[ -f "$tplg" ]]; then
-            tplg_path="$tplg"
-        elif [[ -f "$TPLG_ROOT/$(basename $tplg)" ]]; then
-            tplg_path="$TPLG_ROOT/$(basename $tplg)"
-        else
-            tplg_path=""
-        fi
-        # don't check fw_path here, if fw_path is empty, print debug error.
+    tplg_path=`func_lib_get_tplg_path "$tplg"`
+    if [[ "$?" != "0" ]]; then
+        dlogw "No available topology for graph generation"
+    else
         tplgtool.py -d graph -D $LOG_ROOT $tplg_path &> /dev/null
         [[ "$?" != "0" ]] && dloge "Failed to generate topology graph"
-        unset tplg
     fi
+    unset tplg
 
     # when sof logger collect is open
     if [ "X$SOF_LOG_COLLECT" == "X1" ]; then
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -9,7 +9,15 @@ function exit()
 
     # To generate topology graph, test case should export $tplg for us
     if [[ -n "$tplg" ]]; then
-        tplgtool.py -d graph -D $LOG_ROOT $tplg &> /dev/null
+        if [[ -f "$tplg" ]]; then
+            tplg_path="$tplg"
+        elif [[ -f "$TPLG_ROOT/$(basename $tplg)" ]]; then
+            tplg_path="$TPLG_ROOT/$(basename $tplg)"
+        else
+            tplg_path=""
+        fi
+        # don't check fw_path here, if fw_path is empty, print debug error.
+        tplgtool.py -d graph -D $LOG_ROOT $tplg_path &> /dev/null
         [[ "$?" != "0" ]] && dloge "Failed to generate topology graph"
         unset tplg
     fi
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -155,3 +155,22 @@ func_lib_lsof_error_dump()
         echo "$ret"
     fi
 }
+
+func_lib_get_tplg_path()
+{
+    local tplg=$1
+    # tplg given is empty
+    if [[ -z "$tplg" ]]; then
+        return 1
+    fi
+
+    if [[ -f "$TPLG_ROOT/$(basename "$tplg")" ]]; then
+        echo "$TPLG_ROOT/$(basename "$tplg")"
+    elif [[ -f "$tplg" ]]; then
+        echo $(realpath "$tplg")
+    else
+        return 1
+    fi
+
+    return 0
+}
--- a/test-case/verify-pcm-list.sh
+++ b/test-case/verify-pcm-list.sh
@@ -23,50 +23,31 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
 func_opt_parse_option $*
 tplg=${OPT_VALUE_lst['t']}
 
-[[ ! "$tplg" ]] && dlogw "Missing tplg file for this case" && exit 1
+tplg_path=`func_lib_get_tplg_path "$tplg"`
+[[ "$?" != "0" ]] && dloge "No available topology for this test case" && exit 1
 
 # hijack DMESG_LOG_START_LINE which refer dump kernel log in exit function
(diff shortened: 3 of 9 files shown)


aiChaoSONG:
all commits is tested on helios, PASSED. @Bin-QA @xiulipan Please help to review

case-lib/hijack.sh
```diff
@@ -9,7 +9,15 @@ function exit()
...
     if [[ -n "$tplg" ]]; then
-        tplgtool.py -d graph -D $LOG_ROOT $tplg &> /dev/null
+        if [[ -f "$tplg" ]]; then
+            tplg_path="$tplg"
+        elif [[ -f "$TPLG_ROOT/$(basename $tplg)" ]]; then
+            tplg_path="$TPLG_ROOT/$(basename $tplg)"
+        else
+            tplg_path=""
+        fi
+        # don't check fw_path here, if fw_path is empty, print debug error.
```
libinyang: fw_path => tplg_path

aiChaoSONG: thanks, fixed.

case-lib/lib.sh
```diff
@@ -155,3 +155,22 @@ func_lib_lsof_error_dump()
...
+{
+    local tplg=$1
+    # tplg given is empty
+    if [[ -z "$tplg" ]]; then
+        return 1
+    fi
+
+    if [[ -f "$tplg" ]]; then
+        echo "$tplg"
+    elif [[ -f "$TPLG_ROOT/$(basename $tplg)" ]]; then
```
libinyang, line 169: better to use "$TPLG_ROOT/$(basename "$tplg")"
in case tplg name includes blank space (although this shouldn't happen so far)

aiChaoSONG: the first `"` doesn't match the second `"`, and it matches to the forth `"`,  works like a magic. thanks, fixed.

case-lib/lib.sh
```diff
@@ -155,3 +155,22 @@ func_lib_lsof_error_dump()
...
+        return 1
+    fi
+
+    if [[ -f "$tplg" ]]; then
+        echo "$tplg"
+    elif [[ -f "$TPLG_ROOT/$(basename "$tplg")" ]]; then
+        echo "$TPLG_ROOT/$(basename "$tplg")"
+    else
+        return 1
+    fi
```
Bin-QA, line 173: here will cause a hide bug, so please modify "$TPLG_ROOT/$(basename "$tplg")" as 1st order

test-case/verify-tplg-binary.sh
```diff
@@ -22,29 +22,16 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
...
+dlogi "Checking topology file: $tplg_path"
+dlogi "Found file: $(md5sum $tplg_path|awk '{print $2, $1;}')"
+tplgData=$(sof-tplgreader.py $tplg_path 2>/dev/null)
+[[ -z "$tplgData" ]] && dloge "No valid pipeline(s) found in $tplg_path" && exit 1
+dlogi "Valid pipeline(s) in this topology:"
+echo "===========================>>"
+echo "$tplgData"
+echo "<<==========================="
+
+exit 0
```
libinyang, line 37: You don't need a special update for my below comments, but let's try to follow them later:
some nitpicks (Marc has educated me before):
1. the last line "exit 0" is unnecessary
2. when you are using string, please quote them.
3. from Marc's today email, we should use
```if test1 ; then
  ...
else
  ...
  : $((failures++))
fi
```
instead of 
```
test1
if [ $? = 0] ; then # WRONG!
```

case-lib/lib.sh
Bin-QA: you can catch it with commit: https://github.com/thesofproject/sof-test/commit/dfa4d610a8161218ad6bbf64f947594ae938fec9#diff-c6a24841ec0a3d0dccc065831f99f31a

test-case/verify-tplg-binary.sh
Bin-QA: @libinyang the last line 'exit 0' is necessary, because we hijack the `exit` command,   
so without exit, we couldn't dump the information and do some process for it  

on the other hand it is different requirement for the bash script:
```
# help set
-e  Exit immediately if a command exits with a non-zero status.
```

So for some process, you think it can be ignored you direct use the `-e` option like:  [kmod](https://github.com/thesofproject/sof-test/tree/master/tools/kmod) to use   
but for the full test process,   
you couldn't say direct exit, because you need to do something  
like collect log, like dump the information, etc.   
In fact, test-script just the test process covert to the automation step

test-case/verify-pcm-list.sh
```diff
@@ -23,50 +23,31 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
...
-    fi
-
-    # TODO: Miss multiple TPLG mapping pcm logic
-    break
-done
-
+tplg_str="$(sof-tplgreader.py $tplg_path -d id pcm type -o)"
+pcm_str="$(sof-dump-status.py -i ${SOFCARD:-0})"
+
+dlogc 'sof-tplgreader.py $tplg_file -d id pcm type -o'
```
libinyang, line 35: double quote. And the same to others.

test-case/verify-tplg-binary.sh
libinyang: > @libinyang the last line 'exit 0' is necessary, because we hijack the `exit` command,
> so without exit, we couldn't dump the information and do some process for it
> 
> on the other hand it is different requirement for the bash script:
> 
> ```
> # help set
> -e  Exit immediately if a command exits with a non-zero status.
> ```
> 
> So for some process, you think it can be ignored you direct use the `-e` option like: [kmod](https://github.com/thesofproject/sof-test/tree/master/tools/kmod) to use
> but for the full test process,
> you couldn't say direct exit, because you need to do something
> like collect log, like dump the information, etc.
> In fact, test-script just the test process covert to the automation step

O, right. We do hijack for sof-test. And let's not use exit in the last line for other repos such as sof-sh

case-lib/lib.sh
aiChaoSONG: thanks, fixed.

test-case/verify-pcm-list.sh
aiChaoSONG: @libinyang There are a lot of $var not quoted, let's not do it partially now, and  I will do an overall cleanup on this in a later PR.

libinyang: OK. single quote will show $tplg_file not the real name in the log. It's OK to clean it up later

libinyang:
LGTM

test-case/verify-pcm-list.sh
aiChaoSONG: @libinyang I misunderstand your point, now they are all fixed, and I check dlog* calls, all the arguments to dlog* function are quoted by `""`. Thanks for your reviewing.

case-lib/hijack.sh
```diff
@@ -8,11 +8,14 @@ function exit()
     local exit_status=${1:-0}
 
     # To generate topology graph, test case should export $tplg for us
-    if [[ -n "$tplg" ]]; then
-        tplgtool.py -d graph -D $LOG_ROOT $tplg &> /dev/null
+    tplg_path=`func_lib_get_tplg_path "$tplg"`
+    if [[ "$?" != "0" ]]; then
```
Bin-QA, line 12: normally `!=` use to compare string, you can find it in some environment setup like:  
`  "X$?" != "X0" ` which is more useful

aiChaoSONG: if we quote numbers, "10", it should be string, too

Issue #175, opened 2020-04-08
[BUG]verify-sof-firmware-load.sh as sub-case meet kernel log file is empty


Bin-QA:
1. target file: /var/log/kern.log is empty, sof-kernel-dump.sh will report bash error  (refer sof-kernel-dump.sh and sof-get-kernel-line.sh)
2. verify-sof-firmware as sub-test to reset `DMESG_LOG_START_LINE`, which will be empty

the original case define use as system boot up to check firmware status, but as sub-case this behavior have the problem

solution:
when verify-sof-firmware load as sub-case, force check from dmesg

aiChaoSONG:
issue is found from https://sof-ci.01.org/linuxpr/PR1897/build3515/devicetest/   CML_RVP_HDA

Bin-QA:
patch merged, close issue

PR #178, opened 2020-04-09
case-lib: fix the syntax error

keqiaozhang:
add double quote to prevent globbing and word splitting.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -23,8 +23,8 @@ function exit()
         sleep 1s
     fi
     # when case ends, store kernel log
-    if [ $DMESG_LOG_START_LINE -ne 0 ]; then
-        tail -n +$DMESG_LOG_START_LINE /var/log/kern.log |cut -f5- -d ' ' > $LOG_ROOT/dmesg.txt
+    if [[ -n "$DMESG_LOG_START_LINE" && "$DMESG_LOG_START_LINE" -ne 0 ]]; then
+        tail -n +"$DMESG_LOG_START_LINE" /var/log/kern.log |cut -f5- -d ' ' > $LOG_ROOT/dmesg.txt
     else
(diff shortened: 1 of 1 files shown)


Bin-QA:
how to deal with DMESG_LOG_START_LINE is empty ?
I think #175 is root reason for this

keqiaozhang:
> how to deal with DMESG_LOG_START_LINE is empty ?
> I think #175 is root reason for this

Yes, we need to avoid DMESG_LOG_START_LINE being empty, but better to double quote this variable .

case-lib/hijack.sh
```diff
@@ -23,8 +23,8 @@ function exit()
         sleep 1s
     fi
     # when case ends, store kernel log
-    if [ $DMESG_LOG_START_LINE -ne 0 ]; then
-        tail -n +$DMESG_LOG_START_LINE /var/log/kern.log |cut -f5- -d ' ' > $LOG_ROOT/dmesg.txt
+    if [ "$DMESG_LOG_START_LINE" -ne 0 ]; then
```
Bin-QA, line 26:
```suggestion
    if [[ -n "$DMESG_LOG_START_LINE" && "$DMESG_LOG_START_LINE" -ne 0 ]]; then
```

aiChaoSONG: @Bin-QA can "&&" be used in bash test?  maybe use `-a` for and?

keqiaozhang: > @Bin-QA can "&&" be used in bash test? maybe use `-a` for and?

@aiChaoSONG   In [[..]], we need to use && instead of -a.

aiChaoSONG: @keqiaozhang thanks, Please help to fix this, then we can close #175

keqiaozhang: Updated.

PR #179, opened 2020-04-09
sof-test: misc fix for graph generation

aiChaoSONG:
As we always see "failed to generated graph" error message, this patch does two things:
- add python3-graphviz to env-check.sh
- add debug information for graph generation


aiChaoSONG:
> LGTM
> if you add python-graphviz check in tplgtool is more better

Already there
```
try:
        from graphviz import Digraph
except:
        print("graphviz package not installed, please install with `sudo apt install python3-graphviz`")
        sys.exit(1)
```

PR #181, opened 2020-04-10
case-lib: remove I/O redirection for tplgtool.py

aiChaoSONG:
We may lost debug message if we redirect stdout of tplgtool.py to /dev/null.
```
tplgtool.py -d graph -D $LOG_ROOT $tplg_path > /dev/null
```
As above call of tplgtool.py output nothing but error message, remove the redirection. without the error message, we may get confused in PR test result.

```
2020-04-10 02:37:53 UTC [REMOTE_INFO] Graph of /lib/firmware/intel/sof-tplg/sof-hda-generic-4ch.tplg will be generated to /home/ubuntu/sof-test/logs/check-playback/2020-04-10-10:37:15-29086
2020-04-10 02:37:53 UTC [REMOTE_ERROR] Failed to generate topology graph
2020-04-10 02:37:54 UTC [REMOTE_INFO] Test Result: PASS!
```
This patch helps to tell us why topology graph is failed to generate.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -13,8 +13,8 @@ function exit()
         dlogw "No available topology for graph generation"
     else
         dlogi "Graph of $tplg_path will be generated to $LOG_ROOT"
-        tplgtool.py -d graph -D $LOG_ROOT $tplg_path > /dev/null
-        [[ "$?" != "0" ]] && dloge "Failed to generate topology graph"
+        tplgtool.py -d graph -D $LOG_ROOT $tplg_path
+        [[ "$?" != "0" ]] && dlogw "Failed to generate topology graph"
     fi
     unset tplg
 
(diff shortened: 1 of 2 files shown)


Bin-QA:
1. how about convert `dloge "Failed to generate topology graph"` to `dlogw`
2. not why you need to redirect stdout to /dev/null
2. stderr output is OK for me, but it also means if you are not handle this, it will force output, maybe it also will break your current output

aiChaoSONG:
@Bin-QA  
1. the error is not fatal, change to dlogw is OK
2. in fact, there is no output if we use tplgtool.py like this: `tplgtool.py -d graph -D $LOG_ROOT $tplg_path`,  remove the redirect is OK
3. for tplgtool.py, if failed, program will exit immediately. so the output is invalid, and in most cases, the error check is before output, if failed, no output at all.
so I will do 1 and 2, let's leave 3 as it is(use stderr).

aiChaoSONG:
@Bin-QA Please review again, thanks.

PR #183, opened 2020-04-10
CI: travis: new shellcheck-gitrange.bash

marc-hb:
Travis passes $TRAVIS_COMMIT_RANGE to new shellcheck-gitrange.bash

Runs shellcheck only on files changed by the PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+# bionic = 2018
+dist: bionic
+
+language: minimal
+
+git:
+  depth: false
+  quiet: true
+
+before_install:
+  - sudo apt-get -y install shellcheck
+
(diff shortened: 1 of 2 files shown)


marc-hb:
Shellcheck has a `-f gcc` option that you can pass through `shellcheck-range.bash`. The `-f gcc` option produces "IDE-friendly" output that some editors know how to parse and hyperlink, example running `shellcheck-range.bash HEAD~5.. -f gcc` from Emacs with `Esc-x compile`:

Screenshot 2020-04-10 at 15 36 47

aiChaoSONG:
@marc-hb the check saves us from a lot of shell errors, thanks for the script.

PR #187, opened 2020-04-13
kmod: sof_remove.sh: add snd_sof_ipc_test, snd_sof_probes support

fredoh9:
When multiclient changes are merged, snd_sof_ipc_test is first client
module to test ipc flood test. 2nd client is snf_sof_probes whilch support
Probe feature. The client modules need to be removed before snd_sof.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -52,6 +52,8 @@ remove_module snd_soc_sdw_rt700
 remove_module snd_soc_sdw_rt711_rt1308_rt715
 remove_module snd_soc_sof_sdw
 remove_module snd_soc_ehl_rt5660
+remove_module snd_sof_ipc_test
+remove_module snd_sof_probes
 
 remove_module snd_sof
 remove_module snd_sof_nocodec


fredoh9:
This patch depends on thesofproject/linux#1841 and https://github.com/thesofproject/kconfig/pull/31.

Bin-QA:
@fredoh9 how about add the file system detect for setup default debugfs  
I'm afraid some kernel is `/sys/kernel/debug/sof/ipc_flood_count`, some kernel is `/sys/kernel/debug/sof/ipc-flood-test/ipc_flood_count`

fredoh9:
> @fredoh9 how about add the file system detect for setup default debugfs
> I'm afraid some kernel is `/sys/kernel/debug/sof/ipc_flood_count`, some kernel is `/sys/kernel/debug/sof/ipc-flood-test/ipc_flood_count`

This is good to have. We don't change the root of debugfs often. But to minimize test failures in between transition, this is good idea.
I removed location change and only pushed kmod change. sof_remove.sh change is safe to merge.

PR #189, opened 2020-04-15
test-case: add support for testing on specified pipeline

aiChaoSONG:
with this patch, QA can run test case on specified single pipeline rather than iterate all pipelines

### DIFF
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -42,6 +42,9 @@ OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
 OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
 
+OPT_OPT_lst['S']='filter_string'   OPT_DESC_lst['S']="run this case on specified pipelines"
+OPT_PARM_lst['S']=1             OPT_VALUE_lst['S']="id:any"
+
 func_opt_parse_option $*
 
 tplg=${OPT_VALUE_lst['t']}
@@ -55,7 +58,7 @@ file_prefix=${OPT_VALUE_lst['f']}
 
 func_lib_setup_kernel_last_line
 func_lib_check_sudo
-func_pipeline_export $tplg "type:capture"
+func_pipeline_export $tplg "type:capture & ${OPT_VALUE_lst['S']}"
 
 for round in $(seq 1 $round_cnt)
 do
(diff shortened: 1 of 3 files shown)


Bin-QA:
@aiChaoSONG can you post the test result?
how to deal like: 'id:1,2,3'

aiChaoSONG:
> @aiChaoSONG can you post the test result?
> how to deal like: 'id:1,2,3'

@Bin-QA I know you will ask me this, currently, only one pipeline is supported, if multiple id should be supported, we need to expose the filter string to user, and let user specified filter string

Bin-QA:
> @Bin-QA I know you will ask me this, currently, only one pipeline is supported, if multiple id should be supported, we need to expose the filter string to user, and let user specified filter string

@aiChaoSONG [simultaneous-playback-capture.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/simultaneous-playback-capture.sh#L58) can use like `id:1,2,3`
I think you need post the test result for it  `script -P "1,2,3"`

libinyang:
pipeline ID is not the same of PCM ID. For user, it's hard and complicated to match the PCM id with pipeline id. I suppose @ClarexZhou would like use PCM id, right?

Bin-QA:
> pipeline ID is not the same of PCM ID. For user, it's hard and complicated to match the PCM id with pipeline id. I suppose @ClarexZhou would like use PCM id, right?

the fact they are same value, you can check with [verify-pcm-list.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/verify-pcm-list.sh)

case-lib/pipeline.sh
```diff
@@ -20,7 +20,7 @@ func_pipeline_export()
         opt="-f"
         while [ $# -ne 0 ]
         do
-            opt=$opt" $1"
+            opt=$opt" '$1'"
```
marc-hb, line 23: I don't know this code yet, could you give an example showing why this is required?

libinyang:
> > pipeline ID is not the same of PCM ID. For user, it's hard and complicated to match the PCM id with pipeline id. I suppose @ClarexZhou would like use PCM id, right?
> 
> the fact they are same value, you can check with [verify-pcm-list.sh](https://github.com/thesofproject/sof-test/blob/master/test-case/verify-pcm-list.sh)

OK. Then here the pipeline id is actually the pcm id, not the real "pipeline id" in tplg.

case-lib/pipeline.sh
aiChaoSONG: [cmd_line_log.txt](https://github.com/thesofproject/sof-test/files/4479510/cmd_line_log.txt)
@marc-hb Take a look at the log attached, You should see why I add a `' '`

Bin-QA:
> Is there a plan / github issue to move the `OPT_*` triplication to some common file like `lib.sh` or similar?

`OPT_*` at https://github.com/thesofproject/sof-test/blob/master/case-lib/opt.sh  
`opt.sh` will source at https://github.com/thesofproject/sof-test/blob/master/case-lib/lib.sh#L5

PR #190, opened 2020-04-15
kmod: add snd_soc_rt5660 support for EHL

### LINKED ISSUE #186: [Bug] kmod reload scripts need update for EHL
**Description**
module reload scripts need update for EHL

Need 
remove_module snd_soc_rt5660
remove_module snd_soc_ehl_rt5660
And
insert_module snd_soc_rt5660

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: https://github.com/thesofproject/linux/commits/topic/sof-dev, commit 4b4681e
    * SOF: https://github.com/thesofproject/sof/releases/tag/v1.5-rc1, commit: 5f7d4e2, using sof-tgl.ri rename to sof-ehl.ri
2) Name of the topology file
    * Topology: sof-ehl-rt5660.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: EHL RVP with RT5660 in I2S mode

aiChaoSONG:
fix #186

### DIFF
--- a/tools/kmod/sof_insert.sh
+++ b/tools/kmod/sof_insert.sh
@@ -8,7 +8,7 @@ insert_module() {
 
     if modinfo "$MODULE" &> /dev/null ; then
         echo "Inserting $MODULE"
-        sudo modprobe $MODULE
+        sudo modprobe "$MODULE"
     else
         echo "skipping $MODULE, not in tree"
     fi
@@ -23,6 +23,7 @@ insert_module snd_soc_rt298
 insert_module snd_soc_rt5640
 insert_module snd_soc_rt5645
 insert_module snd_soc_rt5651
+insert_module snd_soc_rt5660
 insert_module snd_soc_rt5670
 insert_module snd_soc_rt5677
 insert_module snd_soc_rt5677_spi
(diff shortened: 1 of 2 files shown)


tools/kmod/sof_insert.sh
```diff
@@ -23,6 +23,7 @@ insert_module snd_soc_rt298
 insert_module snd_soc_rt5640
 insert_module snd_soc_rt5645
 insert_module snd_soc_rt5651
+insert_module snd_soc_rt5660
```
libinyang, line 26: I have an open question unrelate to this PR: Should we manually load the codec driver? Codec driver should be loaded automatically. If we manually load the codec module, will it hide the potential issue that codec driver is not loaded automatically?

aiChaoSONG: @libinyang Before I joined team, there are some codec driver modules exist in the sof_insert.sh. I guess some codec drivers may not be automatically loaded.

aiChaoSONG: @libinyang Pierre once said: we need the I2c devices to be probed before the SOF driver, so we add them all in sof_insert. this should explain why they are in sof_insert.sh

libinyang: Right, without reload i2c driver, the i2c devices will fail to load codec driver. This is an i2c driver scope, not our SOF driver scope. This is different from hdmi codec driver. I saw SOF will try to request HDMI codec driver. ASoC codec driver doesn't support this feature. OK, without this feature, we have to manually load the codec driver.

PR #193, opened 2020-04-16
kmod: add support for snd_sof_ipc_test

aiChaoSONG:
Kernel PR1841 introduces this kernel module,
we need to remove it before snd_sof to resolve
"Module snd_sof is in use by: snd_sof_ipc_test"
error when do module removal.

PR1841 test result here https://sof-ci.01.org/linuxpr/PR1841/build3591/devicetest/


aiChaoSONG:
fixed by fred, close

Issue #177, opened 2020-04-08
EQ test need to support capture pass and IIR or FIR single filter processing


fredoh9:
Playback EQ with IIR + FIR is only tested so far.

Capture EQ also supported, GLK DMIC sof-glk-eq-da7219.tplg. This has only IIR filter. This is new requirement which wasn't the case when the testcase is transferred from sofcitest. Need to grep EQ pipeline and get IIR/FIR, apply proper filter coef to only available filter. Topology util support this already. func_pipeline_export $tplg "eq:any"

And there was warnings due to missing rate option in the command, i.e `-r 48000`
Found from https://github.com/thesofproject/sof/issues/2674
```
...
Applying configuration "././eqctl/eq_fir_pass.txt" into device hw:0 control numid=43.
4607827,0,56,50331648,0,0,0,0,56,65538,0,0,0,0,4294967295,4294901764,0,0,0,0,16384,0
Success.
hdr: magic 0x00464f53
hdr: type 0
hdr: size 56 bytes
hdr: abi 3:0:0
4607827,0,56,50331648,0,0,0,0,56,65538,0,0,0,0,4294967295,4294901764,0,0,0,0,16384,0
2020-04-08 12:53:20 UTC [COMMAND] aplay -Dhw:0,0 -f S16_LE -c 2 -r 48000 /dev/zero -d 5
Playing raw data '/dev/zero' : Signed 16 bit Little Endian, Rate 8000 Hz, Stereo
Warning: rate is not accurate (requested = 8000Hz, got = 48000Hz)
         please, try the plug plugin 
2020-04-08 12:53:21 UTC [INFO] EQ test done: failed_cnt is 0
2020-04-08 12:53:21 UTC [INFO] 1 - total pipeline= 4
2020-04-08 12:53:21 UTC [INFO] None of FIR/IIF filter is available in this pipeline, skip
2020-04-08 12:53:21 UTC [INFO] 2 - total pipeline= 4
2020-04-08 12:53:21 UTC [INFO] None of FIR/IIF filter is available in this pipeline, skip
2020-04-08 12:53:21 UTC [INFO] 3 - total pipeline= 4
2020-04-08 12:53:22 UTC [INFO] None of FIR/IIF filter is available in this pipeline, skip
2020-04-08 12:53:22 UTC [ERROR] Failed to generate topology graph
2020-04-08 12:53:22 UTC [INFO] Test Result: PASS!
```

PR #194, opened 2020-04-16
test-case: test iir and fir filter independently

fredoh9:
Query eq pipelines from topology file. Test iir or fir separately
whichever is available. And base on type of the pipeline, aplay or
arecord is used for the test.

fixes: https://github.com/thesofproject/sof-test/issues/177

### DIFF
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -15,9 +15,10 @@
 ##
 
 # source from the relative path of current folder
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+my_dir=$(dirname "${BASH_SOURCE[0]}")
+source "$my_dir"/../case-lib/lib.sh
 
-OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']="tplg file, default value is env TPLG: $TPLG"
 OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
 
 OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='aplay duration in second'
@@ -34,8 +35,8 @@ tplg=${OPT_VALUE_lst['t']}
 duration=${OPT_VALUE_lst['d']}
 loop_cnt=${OPT_VALUE_lst['l']}
 
-# TODO: direct import only EQ pipeline
-func_pipeline_export $tplg "type:playback"
+# import only EQ pipeline from topology
+func_pipeline_export $tplg "eq:any"
 sofcard=${SOFCARD:-0}
 
 # Test equalizer
@@ -44,103 +45,90 @@ func_test_eq()
     local id=$1
     local conf=$2
 
-    [ ! -f $conf ] && dloge "$conf not exist" && exit 1
-
     dlogc "sof-ctl -Dhw:$sofcard -n $id -s $conf"
-    sof-ctl -Dhw:$sofcard -n $id -s $conf
-    if [ $? -ne 0 ]; then
+    sof-ctl -Dhw:"$sofcard" -n "$id" -s "$conf" || {
         dloge "Equalizer setting failure with $conf"
         return 1
-    fi
-    # test with aplay
-    dlogc "aplay -D$dev -f $fmt -c $channel -r $rate /dev/zero -d $duration"
-    aplay -D$dev -f $fmt -c $channel /dev/zero -d $duration
-    if [ $? -ne 0 ]; then
+    }
+
+    dlogc "$cmd -D $dev -f $fmt -c $channel -r $rate -d $duration $dummy_file"
+    $cmd -D "$dev" -f "$fmt" -c "$channel" -r "$rate" -d "$duration" "$dummy_file" || {
         dloge "Equalizer test failure with $conf"
         return 1
-    fi
+    }
     sleep 1
 }
 
-current_dir=$(dirname ${BASH_SOURCE[0]})
-
-# initialized with negative value. If none of test is performed the test result will be failed.
-result=1
-failed_cnt=0
-
-# TODO: direct import only EQ pipeline and loop for EQ pipeline only
-for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
-do
-    dlogi "$idx - total pipeline= $PIPELINE_COUNT"
-    dev=$(func_pipeline_parse_value $idx dev)
-    channel=$(func_pipeline_parse_value $idx channel)
-    rate=$(func_pipeline_parse_value $idx rate)
-    fmt=$(func_pipeline_parse_value $idx fmt)
-    eq_support=$(func_pipeline_parse_value $idx eq)
-    if [ -z "$eq_support" ]; then
-        dlogi "None of FIR/IIF filter is available in this pipeline, skip"
-        continue
-    fi
-    dlogi "eq_support= $eq_support"
-    is_iir=$(echo $eq_support | grep -i iir | wc -l)
-    is_fir=$(echo $eq_support | grep -i fir | wc -l)
-    if [ $is_iir -ne $is_fir ]; then
-        dlogi "no FIR or IIF filter is available in this pipeline, skip"
-        continue
+# this function performs IIR/FIR filter test
+# param1 must be must be iir or fir
+func_test_filter()
+{
+    local testfilter=$1
+    dlogi "Get amixer control id for $testfilter"
+    # TODO: Need to match alsa control id with the filter in the pipeline,
+    #       currently the test discards EQ pipelines except first one.
+    Filterid=$(amixer -D hw:"$sofcard" controls | sed -n -e "/eq${testfilter}/I "'s/numid=\([0-9]*\),.*/\1/p' | head -1)
+    if [ -z "$Filterid" ]; then
+        dloge "can't find $testfilter"
+        exit 1
     fi
 
-    dlogi "0. Get amixer control id for IIR and FIR"
-    IIRid=`amixer -D hw:$sofcard controls | grep EQIIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
-    if [ -z $IIRid ]; then
-            dloge "can't find IIR filter"
-            exit 1
-    fi
-
-    FIRid=`amixer -D hw:$sofcard controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
-    if [ -z $FIRid ]; then
-            dloge "can't find FIR filter"
-            exit 1
-    fi
-
-    declare -a IIRList=($(ls -d ${current_dir}/eqctl/eq_iir_*.txt))
-    nIIRList=${#IIRList[@]}
-    dlogi "IIR list $nIIRList ${IIRList[*]}"
-    declare -a FIRList=($(ls -d ${current_dir}/eqctl/eq_fir_*.txt))
-    nFIRList=${#FIRList[@]}
-    dlogi "FIR list $nFIRList ${FIRList[*]}"
-    if [ $nIIRList ==  0 ] || [ $nFIRList ==  0 ]; then
-        dloge "IIR or FIR flter coeff list error!"
-	exit 1
+    declare -a FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))
+    nFilterList=${#FilterList[@]}
+    dlogi "$testfilter list, num= $nFilterList, coeff files= ${FilterList[*]}"
+    if [ "$nFilterList" -eq  0 ]; then
+        dloge "$testfilter flter coeff list error!"
+        exit 1
     fi
 
     for i in $(seq 1 $loop_cnt)
     do
-        dlogi "[$i/$loop_cnt] 1. Test IIR config list, IIR amixer control id=$IIRid"
-        for config in ${IIRList[@]}; do
-            func_test_eq $IIRid $current_dir/$config
-            result=$?
-            if [[ $result -ne 0 ]]; then
-                dloge "Failed at $config"
-                let failed_cnt++
-            fi
+        dlogi "[$i/$loop_cnt] Test $testfilter config list, $testfilter amixer control id=$Filterid"
+        for config in "${FilterList[@]}"; do
+            func_test_eq "$Filterid" "$my_dir/$config" || {
+                dloge "EQ test failed with $config"
+                : $((failed_cnt++))
+            }
         done
 
-        dlogi "[$i/$loop_cnt] 2. Test FIR config list, FIR amixer control id=$FIRid"
-        for config in ${FIRList[@]}; do
-            func_test_eq $FIRid $current_dir/$config
-            result=$?
-            if [[ $result -ne 0 ]]; then
-                dloge "Failed at $config"
-                let failed_cnt++
-            fi
-        done
-        dlogi "EQ test done: failed_cnt is $failed_cnt"
+        dlogi "$testfilter test done: failed_cnt is $failed_cnt"
         if [ $failed_cnt -gt 0 ]; then
             exit 1
+        fi
+    done
+}
+
+failed_cnt=0
+
+for idx in $(seq 0 "$((PIPELINE_COUNT-1))")
+do
+    dlogi "$idx - total pipeline= $PIPELINE_COUNT"
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    channel=$(func_pipeline_parse_value "$idx" channel)
+    rate=$(func_pipeline_parse_value "$idx" rate)
+    fmt=$(func_pipeline_parse_value "$idx" fmt)
+    type=$(func_pipeline_parse_value "$idx" type)
+    eq_support=$(func_pipeline_parse_value "$idx" eq)
+
+    case $type in
+        "playback")
+            cmd=aplay
+            dummy_file=/dev/zero
+        ;;
+        "capture")
+            cmd=arecord
+            dummy_file=/dev/null
+        ;;
+    esac
+
+    dlogi "eq_support= $eq_support"
+    # if IIR/FIR filter is avilable, test with coef list
+    for filter_type in iir fir; do
+        if echo "$eq_support" | grep -q -i $filter_type; then
+            func_test_filter $filter_type
         fi
     done
 
 done
 
-exit $result
-
+exit 0


test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
-    fi
 
-    dlogi "0. Get amixer control id for IIR and FIR"
-    IIRid=`amixer -D hw:$sofcard controls | grep EQIIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
-    if [ -z $IIRid ]; then
-            dloge "can't find IIR filter"
-            exit 1
-    fi
+    # if IIR filter is avilable, test wit IIR coef
+    is_iir=$(echo $eq_support | grep -i iir | wc -l)
```
Bin-QA, line 95: IIR is same logic with FIR just some keyword is different  
How about use the function for it

test-case/check-audio-equalizer.sh
```diff
@@ -53,8 +53,8 @@ func_test_eq()
         return 1
     fi
     # test with aplay
-    dlogc "aplay -D$dev -f $fmt -c $channel -r $rate /dev/zero -d $duration"
-    aplay -D$dev -f $fmt -c $channel /dev/zero -d $duration
+    dlogc "$cmd -D$dev -f $fmt -c $channel -r $rate -d $duration $dummy_file"
```
Bin-QA, line 56:
```suggestion
    dlogc "$cmd -D $dev -f $fmt -c $channel -r $rate -d $duration $dummy_file"
```

libinyang, line 56: Please also change the upper description " test with aplay" as we supports arecord now.

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
         fi
-    done
+
+        for i in $(seq 1 $loop_cnt)
+        do
+            dlogi "[$i/$loop_cnt] 2.3 Test FIR config list, FIR amixer control id=$FIRid"
+            for config in ${FIRList[@]}; do
+                func_test_eq $FIRid $current_dir/$config
+                result=$?
+                if [[ $result -ne 0 ]]; then
```
marc-hb, line 157: Shorter and compatible with `set -e`:
```shell
func_test_eq "$FIRid" "$current_dir/$config" || {
    dloge "Failed at $config"
    : $((failed_cnt++))
}
```

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
+        if [ $nFIRList ==  0 ]; then
+            dloge "FIR flter coeff list error!"
             exit 1
         fi
-    done
+
+        for i in $(seq 1 $loop_cnt)
+        do
+            dlogi "[$i/$loop_cnt] 2.3 Test FIR config list, FIR amixer control id=$FIRid"
+            for config in ${FIRList[@]}; do
```
marc-hb, line 154: ` for config in "${FIRList[@]}"; do`

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
+                result=$?
+                if [[ $result -ne 0 ]]; then
+                    dloge "Failed at $config"
+                    let failed_cnt++
+                fi
+            done
+
+            dlogi "FIR test done: failed_cnt is $failed_cnt"
+            if [ $failed_cnt -gt 0 ]; then
+                exit 1
```
marc-hb, line 165: `exit $failed_cnt`, no need for the `if`

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
 
-    dlogi "0. Get amixer control id for IIR and FIR"
-    IIRid=`amixer -D hw:$sofcard controls | grep EQIIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
-    if [ -z $IIRid ]; then
-            dloge "can't find IIR filter"
-            exit 1
-    fi
+    # if IIR filter is avilable, test wit IIR coef
+    is_iir=$(echo $eq_support | grep -i iir | wc -l)
+    if [[ $is_iir -ne "" ]]; then
```
marc-hb, line 96: Do you have an example of some `eq_support` value? The `wc -l` puzzles me.

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
+        # 1. Test IIR
+        dlogi "1.1 Get amixer control id for IIR"
+        IIRid=`amixer -D hw:$sofcard controls | grep EQIIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+        if [ -z $IIRid ]; then
```
marc-hb, line 100: This means `[ -z ]`, -z has no argument  if `$IIRid` is empty. Not good. Just double quote almost everything. https://mywiki.wooledge.org/Quotes

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
+                exit 1
+        fi
 
-    FIRid=`amixer -D hw:$sofcard controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
-    if [ -z $FIRid ]; then
-            dloge "can't find FIR filter"
+        declare -a IIRList=($(ls -d ${current_dir}/eqctl/eq_iir_*.txt))
+        nIIRList=${#IIRList[@]}
+        dlogi "1.2 IIR list $nIIRList ${IIRList[*]}"
+        if [ $nIIRList ==  0 ]; then
```
marc-hb, line 108: `-ne 0` is better style for numbers.

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
-            if [[ $result -ne 0 ]]; then
-                dloge "Failed at $config"
-                let failed_cnt++
+        for i in $(seq 1 $loop_cnt)
+        do
+            dlogi "[$i/$loop_cnt] 1.3 Test IIR config list, IIR amixer control id=$IIRid"
+            for config in ${IIRList[@]}; do
+                func_test_eq $IIRid $current_dir/$config
+                result=$?
+                if [[ $result -ne 0 ]]; then
```
marc-hb, line 119: See below.

test-case/check-audio-equalizer.sh
```diff
@@ -68,77 +68,104 @@ current_dir=$(dirname ${BASH_SOURCE[0]})
...
-        done
-        dlogi "EQ test done: failed_cnt is $failed_cnt"
-        if [ $failed_cnt -gt 0 ]; then
+    # if FIR filter is avilable, test wit FIR coef
+    is_fir=$(echo $eq_support | grep -i fir | wc -l)
+    if [[ $is_fir -ne "" ]]; then
+        # 2. Test FIR
+        dlogi "2.1 Get amixer control id for FIR"
+        FIRid=`amixer -D hw:$sofcard controls | grep EQFIR| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+        if [ -z $FIRid ]; then
```
marc-hb, line 138: See above.

test-case/check-audio-equalizer.sh
fredoh9: good point, that was from very old condition. Now eq_support is oneline.
```
eq_support= EQIIR1.0 EQFIR1.0
```

test-case/check-audio-equalizer.sh
fredoh9: Yes shorter but I like current style better, as it is more intuitive.

fredoh9:
Tested in APL UP2 with sof-apl-pcm512x.tplg and sof-apl-eq-pcm512x.tplg.
Tested in GLK BOBBA with sof-glk-eq-da7219.tplg.

test-case/check-audio-equalizer.sh
```diff
@@ -52,95 +52,95 @@ func_test_eq()
...
+# this function performs IIR/FIR filter test
+# param1 must be must be iir or fir
+func_test_filter()
+{
+    local testfilter=$1
+    dlogi "1. Get amixer control id for $testfilter"
+    Filterid=`amixer -D hw:$sofcard controls | grep -i eq${testfilter}| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
+    if [ -z "$Filterid" ]; then
+        dloge "can't find $testfilter"
+        exit 1
```
Bin-QA, line 74: `fail` / `N/A`, I'm not sure if miss configure the behavior can accept  
couldn't : `failed`
just lost feature: `N/A`

test-case/check-audio-equalizer.sh
```diff
@@ -52,95 +52,95 @@ func_test_eq()
...
+    fi
+
+    declare -a FilterList=($(ls -d ${current_dir}/eqctl/eq_${testfilter}_*.txt))
+    nFilterList=${#FilterList[@]}
+    dlogi "2. $testfilter list $nFilterList ${FilterList[*]}"
+    if [ $nFilterList ==  0 ]; then
+        dloge "$testfilter flter coeff list error!"
+        exit 1
+    fi
```
Bin-QA, line 83: normally lost configure file, it is couldn't to run `exit 2`

test-case/check-audio-equalizer.sh
fredoh9: func_test_filter() will start when the topology support eq filter(s). Hence any failure while running the test should be reported as failure. In this case, we couldn't find alsactl id to test.

test-case/check-audio-equalizer.sh
fredoh9: In this case also same with above. normally missing configuration file can be skipped but in this case, coef files have to be existed before test. This is error case.

test-case/check-audio-equalizer.sh
marc-hb: Sorry but testing `$?` it not acceptable for a number of reasons: https://github.com/koalaman/shellcheck/wiki/SC2181
The main reason is probably compatibility with `set -e`, see other reasons above.

The `do_something || handle_error` style is very common and reads like plain English, you can get used to it quickly. If it's really too hard you can try this instead:

```shell
if ! func_test_eq "$FIRid" "$current_dir/$config"; then
    dloge "Failed at $config"
    : $((failed_cnt++))
fi
```

test-case/check-audio-equalizer.sh
```diff
@@ -52,95 +52,95 @@ func_test_eq()
...
     sleep 1
 }
 
+# this function performs IIR/FIR filter test
+# param1 must be must be iir or fir
+func_test_filter()
+{
+    local testfilter=$1
+    dlogi "1. Get amixer control id for $testfilter"
+    Filterid=`amixer -D hw:$sofcard controls | grep -i eq${testfilter}| head -1| sed 's/numid=\([0-9]*\),.*/\1/'`
```
marc-hb, line 71: Backticks are deprecated https://mywiki.wooledge.org/BashFAQ/082

This was reported by shellcheck, please look at the Travis output. I'll review the rest after you have.

test-case/check-audio-equalizer.sh
marc-hb: In some cases you need to save the error for later; that is the real purpose of `$?`:
```shell
set -e
saved_error=0
do_something || saved_error=$? # saved _for later_
do_something_else1
do_something_else2
if [ 0 -ne $saved_error] then
...
```
But without any `do_something_else` then`$?` is at best pointless and at worst harmful.

test-case/check-audio-equalizer.sh
marc-hb: BTW `sed` has a built-in `grep` feature:
```shell
amixer -D default controls | sed -n -e "/eq${testfilter}/I "'s/numid=\([0-9]*\),.*/\1/p'
```

test-case/check-audio-equalizer.sh
```diff
@@ -52,95 +52,95 @@ func_test_eq()
...
+    if [ -z "$Filterid" ]; then
```
marc-hb, line 72: Can `eq${testfilter}` match multiple `controls` lines?? If yes can you add a comment explaining why the first line is "the best"? If not this should probably error, no?

test-case/check-audio-equalizer.sh
```diff
@@ -52,95 +52,95 @@ func_test_eq()
...
-    dlogi "IIR list $nIIRList ${IIRList[*]}"
-    declare -a FIRList=($(ls -d ${current_dir}/eqctl/eq_fir_*.txt))
-    nFIRList=${#FIRList[@]}
-    dlogi "FIR list $nFIRList ${FIRList[*]}"
-    if [ $nIIRList ==  0 ] || [ $nFIRList ==  0 ]; then
-        dloge "IIR or FIR flter coeff list error!"
-	exit 1
+    # if IIR filter is avilable, test wit IIR coef
+    is_iir=$(echo $eq_support | grep -i iir)
+    if [[ ! -z "$is_iir" ]]; then
```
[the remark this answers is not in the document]
marc-hb: Or:
```shell
if [[ "$eq_support" =~ [iI][iI][rR] ]]; then
...
```

fredoh9:
> There are still some basic shellcheck issues

Most of them are fixed. Remained issues are common for most of test cases. If we need to fix them, we need to apply the fix for all.

test-case/check-audio-equalizer.sh
```diff
@@ -26,16 +26,16 @@ OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=5
 OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
 OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
 
-OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store =at $LOG_ROOT"
 OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
 
-func_opt_parse_option $*
+func_opt_parse_option "$*"
```
marc-hb, line 32: The code in and around this functions seems to be a mess of `$*, "$*", $@,...` but please keep it a _consistent_ mess for now so if something fails it fails the same all across the board :-(

https://mywiki.wooledge.org/Quotes#Expand_argument_lists
https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +44,100 @@ func_test_eq()
...
+
+    # if IIR filter is avilable, test wit IIR coef
+    if echo "$eq_support" | grep -q -i iir; then    
+        func_test_filter iir
+    fi
+
+    # if FIR filter is avilable, test wit FIR coef
+    if echo "$eq_support" | grep -q -i fir; then 
+        func_test_filter fir
+    fi
```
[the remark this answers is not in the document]
fredoh9: thanks, this is better

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +44,100 @@ func_test_eq()
...
-            result=$?
-            if [[ $result -ne 0 ]]; then
-                dloge "Failed at $config"
-                let failed_cnt++
-            fi
-        done
-        dlogi "EQ test done: failed_cnt is $failed_cnt"
+        dlogi "$testfilter test done: failed_cnt is $failed_cnt"
         if [ $failed_cnt -gt 0 ]; then
             exit 1
```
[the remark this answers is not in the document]
fredoh9: all failed case is exit, and if it hit at the end of line, either pass or passthrough(no filter). that's why `result` was initialized with 1.

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +45,100 @@ func_test_eq()
...
+    # if IIR/FIR filter is avilable, test with coef list
+    for filter_type in iir fir; do
+        if echo "$eq_support" | grep -q -i $filter_type; then
+            func_test_filter $filter_type
+        fi
+    done
+
+done
 
+if [ "$has_test_run" == "false" ]; then
```
marc-hb, line 139:
```shell
if $has_test_run; then
  exit 0
else
 dlogi ...
  exit 2
fi
```

test-case/check-audio-equalizer.sh
marc-hb: I still don't understand the `head -1` but if someone else can then fine.

test-case/check-audio-equalizer.sh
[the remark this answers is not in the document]
marc-hb: > all failed case is exit [immediately]

Why not run later tests when one test fail? Is it by design or just because you don't want to add an unrelated change in this PR? Either way is fine, I'm just curious.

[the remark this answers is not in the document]
fredoh9: That's right, failed_cnt was there before. By that logic, it might fail on some coeffs in one filter. I prefer exit immediately at any failure but it's not harm.

test-case/check-audio-equalizer.sh
fredoh9: Even without head -1, it works. but it need to be addressed with `TODO` . I will work on it once this merged.

fredoh9: This is my comment on latest code.
> TODO: Need to match alsa control id with the filter in the pipeline,
>       currently the test discards EQ pipelines except first one.

marc-hb:
The codestyle LGTM. Needs someone else to approve the logic.

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +45,100 @@ func_test_eq()
...
-    dlogi "IIR list $nIIRList ${IIRList[*]}"
-    declare -a FIRList=($(ls -d ${current_dir}/eqctl/eq_fir_*.txt))
-    nFIRList=${#FIRList[@]}
-    dlogi "FIR list $nFIRList ${FIRList[*]}"
-    if [ $nIIRList ==  0 ] || [ $nFIRList ==  0 ]; then
-        dloge "IIR or FIR flter coeff list error!"
-	exit 1
+    declare -a FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))
+    nFilterList=${#FilterList[@]}
+    dlogi "2. $testfilter list $nFilterList ${FilterList[*]}"
```
Bin-QA, line 80: I think "$nFilterList ${FilterList[*]}" when check log maybe confuse, need some flag to split it

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +45,100 @@ func_test_eq()
     local id=$1
     local conf=$2
 
-    [ ! -f $conf ] && dloge "$conf not exist" && exit 1
+    [ ! -f "$conf" ] && dloge "$conf not exist" && exit 1
```
Bin-QA, line 48: I don't think this logic is exist, base on `FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))`

fredoh9:
@Bin-QA I'm busy with my primary task today. I'm planing to address your comments tomorrow.

test-case/check-audio-equalizer.sh
```diff
@@ -26,16 +27,16 @@ OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=5
 OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
 OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
 
-OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store =at $LOG_ROOT"
 OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
```
wenqingfu, line 31: =at?  is this a mis-change accidently?

fredoh9: It was a accident. Will fix this together.

test-case/check-audio-equalizer.sh
```diff
@@ -44,103 +45,90 @@ func_test_eq()
...
-    fi
-    # test with aplay
-    dlogc "aplay -D$dev -f $fmt -c $channel -r $rate /dev/zero -d $duration"
-    aplay -D$dev -f $fmt -c $channel /dev/zero -d $duration
-    if [ $? -ne 0 ]; then
+    }
+
+    dlogc "$cmd -D $dev -f $fmt -c $channel -r $rate -d $duration $dummy_file"
+    $cmd -D "$dev" -f "$fmt" -c "$channel" -r "$rate" -d "$duration" "$dummy_file" || {
         dloge "Equalizer test failure with $conf"
```
Bin-QA, line 56: @aiChaoSONG already update those type code to `die` function, but it is not refer you code

PR #197, opened 2020-04-20
case-lib: remove topology graph generation code

aiChaoSONG:
Generating graph in hijacked exit is not a good
solution to this feature. This patch reverts #166.

As auto revert is not possible after so many PR
merged, this patch remove the graph generation
code in hijacked exit.


wenqingfu:
It's a nice feature. We just need to find a better place to use it.

Issue #188, opened 2020-04-14
Need one parameter to specify pipeline  for playback/capture and pause/resume scripts


ClarexZhou:
**Description**
Will always go through all pipelines for playback/capture,and pause/resume scrips.
Need to specify pipeline ID, and fallback to go through all pipelines w/o this parameter.

Bin-QA:
Update for change of source code:  
https://github.com/thesofproject/linux/commits/topic/sof-dev/sound/soc/sof/sof-pci-dev.c

ClarexZhou:
verified with PR#189, -S id:0  or -S id:1,6,0 can works for playback, capture and pause reaume scripts.

PR #198, opened 2020-04-20
case-lib: refine pipeline filter for pipeline.sh

aiChaoSONG:
Previously, to support "-f filter1 filter2" parameter, we use while to iterate all filter items.
For now, the pipeline filter is only a single string, we don't need a loop here.

### DIFF
--- a/case-lib/pipeline.sh
+++ b/case-lib/pipeline.sh
@@ -7,46 +7,39 @@ func_pipeline_export()
         dlogi "Topology file name is not specified, unable to run command: ${BASH_SOURCE[-1]}" && exit 1
     fi
 
-    local tplg="$1" opt="" sofcard=${SOFCARD:-0} cmd=""
+    local tplg="$1"
+    local sofcard=${SOFCARD:-0}
 
     # got tplg_file, verify file exist
-    tplg_path=`func_lib_get_tplg_path "$tplg"`
+    tplg_path=$(func_lib_get_tplg_path "$tplg")
     [[ "$?" != "0" ]] && dloge "No available topology for pipeline export" && exit 1
     dlogi "${BASH_SOURCE[-1]} will use topology $tplg_path to run the test case"
 
-    shift
-    # create filter option string
-    if [ $# -ne 0 ]; then
-        opt="-f"
-        while [ $# -ne 0 ]
-        do
-            opt=$opt" $1"
-            shift
-        done
-    fi
+    # acquire filter option
+    local filter="$2"
 
     # create block option string
+    local ignore=""
     if [ ${#TPLG_IGNORE_LST[@]} -ne 0 ]; then
-        opt=$opt" -b"
-        for key in ${!TPLG_IGNORE_LST[@]}
+        for key in "${!TPLG_IGNORE_LST[@]}"
         do
             dlogi "Pipeline list to ignore is specified, will ignore '$key=${TPLG_IGNORE_LST[$key]}' in test case"
-            opt=$opt" $key:'${TPLG_IGNORE_LST[$key]}'"
+            ignore=$ignore" $key:${TPLG_IGNORE_LST[$key]}"
         done
     fi
 
-    cmd=$(echo sof-tplgreader.py $tplg_path $opt -s $sofcard -e)
-
-    OLD_IFS="$IFS" IFS=$'\n'
+    opt="-f \"$filter\" -b \"$ignore\""
+    cmd="sof-tplgreader.py $tplg_path $opt -s $sofcard -e"
     dlogi "Run command to get pipeline parameters"
     dlogc "$cmd"
-    for line in $(eval $cmd);
+    OLD_IFS="$IFS" IFS=$'\n'
+    for line in $(eval "$cmd");
     do
-        eval $line
+        eval "$line"
     done
     IFS="$OLD_IFS"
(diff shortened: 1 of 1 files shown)


case-lib/pipeline.sh
```diff
@@ -7,35 +7,26 @@ func_pipeline_export()
...
         for key in ${!TPLG_IGNORE_LST[@]}
         do
             dlogi "Pipeline list to ignore is specified, will ignore '$key=${TPLG_IGNORE_LST[$key]}' in test case"
-            opt=$opt" $key:'${TPLG_IGNORE_LST[$key]}'"
+            ignore=$ignore" $key:'${TPLG_IGNORE_LST[$key]}'"
         done
     fi
 
-    cmd=$(echo sof-tplgreader.py $tplg_path $opt -s $sofcard -e)
+    cmd=$(echo "sof-tplgreader.py $tplg_path -f \"$filter\" -b \"$ignore\" -s $sofcard -e")
```
marc-hb, line 29: Why not just the simpler:
```shell
cmd="sof-tplgreader.py $tplg_path -f \"$filter\" -b \"$ignore\" -s $sofcard -e"
```
Or probably better:
```shell
cmd="sof-tplgreader.py $tplg_path -f '$filter' -b '$ignore' -s $sofcard -e"
```

case-lib/pipeline.sh
```diff
@@ -7,35 +7,26 @@ func_pipeline_export()
...
     # create filter option string
-    if [ $# -ne 0 ]; then
-        opt="-f"
-        while [ $# -ne 0 ]
-        do
-            opt=$opt" $1"
-            shift
-        done
-    fi
+    local filter=$2
```
marc-hb, line 18:
```shell
local filter="$2"
```
https://github.com/koalaman/shellcheck/wiki/SC2046
https://mywiki.wooledge.org/Quotes

case-lib/pipeline.sh
aiChaoSONG: Thanks, I will use the first suggestion, as the single quote in second one is wrong.

```
2020-04-21 05:42:24 UTC [COMMAND] sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-cml-rt1011-rt5682.tplg -f 'type:capture & id:1' -b ' pcm:'HDA Digital,Media Playback,DMIC16k'' -s 0 -e

2020-04-21 05:45:29 UTC [COMMAND] sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-cml-rt1011-rt5682.tplg -f "type:capture & id:0" -b " pcm:'HDA Digital,Media Playback,DMIC16k'" -s 0 -e
```

marc-hb: Embedding quotes inside strings is really complex and nasty and so is `eval` and `ssh $host $cmd`. Complex commands with whitespace like this one should be generated into `mktemp` files and then the main script runs  these files. So much easier to understand and debug.

In this case it's not even clear to me why `$cmd` is needed at all, is this just for logging it? Anyway this is a bit off-topic, I understand this PR has a more limited scope.

case-lib/pipeline.sh
```diff
@@ -7,46 +7,39 @@ func_pipeline_export()
         dlogi "Topology file name is not specified, unable to run command: ${BASH_SOURCE[-1]}" && exit 1
     fi
 
-    local tplg="$1" opt="" sofcard=${SOFCARD:-0} cmd=""
+    local tplg="$1" ignore="" sofcard=${SOFCARD:-0}
```
Bin-QA, line 10: older design `opt` is base on when `$*` have some option like `-s` can also pass-down to here  
so it seems have some confuse  
you also use `opt` and `cmd`

case-lib/pipeline.sh
```diff
@@ -7,46 +7,39 @@ func_pipeline_export()
...
     dlogi "Run command to get pipeline parameters"
     dlogc "$cmd"
-    for line in $(eval $cmd);
+    for line in $(eval "$cmd");
     do
-        eval $line
+        eval "$line"
     done
+
+    opt="-f \"$filter\" -b \"$ignore\""
```
Bin-QA, line 39: move to line 28?

case-lib/pipeline.sh
aiChaoSONG: This function is used to export pipeline, the options we used here is -f -b and -s,  and -s sofcard is here : `sofcard=${SOFCARD:-0}`, so here only process filter string and block option is enough.
Has checked all other option of sof-tplgreader.py, no other option can be used here.

case-lib/pipeline.sh
aiChaoSONG: the opt here is used for output. no other funciton.

Bin-QA: I mean it is the same code when you load it, so here is duplicate with line 29

case-lib/pipeline.sh
Bin-QA: `opt` and `cmd` also been used, so you can add them back

aiChaoSONG:
@Bin-QA The code is more smooth now. The purpose of this PR is that I want the user of func_lib_export_pipeline can directly provide filter string as a parameter,  then https://github.com/thesofproject/sof-test/issues/188 should be easier to fix.

PR #203, opened 2020-04-23
test-case: check-runtime-pm-* miss verify kernel check result

Bin-QA:
when load sof-kernel-log-check it should check the return value
fix case:
check-runtime-pm-status.sh
check-runtime-pm-double-active.sh

### DIFF
--- a/test-case/check-runtime-pm-status.sh
+++ b/test-case/check-runtime-pm-status.sh
@@ -127,7 +127,9 @@ do
         fi
     done
 
-    sof-kernel-log-check.sh 0
+    sof-kernel-log-check.sh 0 || {
+        dloge "Catch error in dmesg" && exit 1
+    }
 done
 
 sof-kernel-log-check.sh $KERNEL_LAST_LINE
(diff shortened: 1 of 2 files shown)


test-case/check-runtime-pm-status.sh
```diff
@@ -128,6 +128,7 @@ do
     done
 
     sof-kernel-log-check.sh 0
+    [[ $? -ne 0 ]] && dloge "Catch error in dmesg" && exit 1
```
marc-hb, line 131: checkpatch warning:
```
[[ $? -ne 0 ]] && dloge "Catch error in dmesg" && exit 1
       ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
```

Fix:
```shell
sof-kernel-log-check.sh 0 || {
 dloge "Catch error in dmesg" && exit 1
}
```
Or if you prefer:
```shell
if ! sof-kernel-log-check.sh 0; then
 dloge "Catch error in dmesg" && exit 1
fi
```

PR #204, opened 2020-04-23
test-case: unify code style for kernel log check function calls

aiChaoSONG:
As Bin add check for kernel log check return value in https://github.com/thesofproject/sof-test/pull/203, and use a new code style, this patch align code style around all of the repo.

### DIFF
--- a/test-case/check-ipc-flood.sh
+++ b/test-case/check-ipc-flood.sh
@@ -47,8 +47,9 @@ do
     dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
     sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-    sof-kernel-log-check.sh 0
-    [[ $? -ne 0 ]] && dloge "Catch error in dmesg" && exit 1
+    sof-kernel-log-check.sh 0 || {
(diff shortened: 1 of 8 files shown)


marc-hb:
```shell
dloge "Catch error in dmesg" && exit 1
```
I don't understand the `&&`. If there's an issue in `dloge` and `dloge` fails then the code will keep running. That seems wrong.

Bin-QA:
https://github.com/thesofproject/sof-test/blob/master/case-lib/logging_ctl.sh#L36

if it is failed, it will meet more serious problem

marc-hb:
Sorry I still don't understand why `&&` instead of `;`

aiChaoSONG:
> Sorry I still don't understand why `&&` instead of `;`

@Bin-QA Marc is right, whatever the return value of `dloge` is, we want the `exit` to be excuted, so need to refine again in a new PR.

Bin-QA:
yes, Marc is right,  
my point is when `echo` failed,  
`exit` seems also couldn't been execute

aiChaoSONG:
@marc-hb I have submit a patch to fix this, https://github.com/thesofproject/sof-test/pull/211. I am graduated from university less than a year, and not as experienced as you, really appreciate your review and suggestions.

PR #206, opened 2020-04-23
lib: opt.sh: remove useless cmd.txt code and shellcheck cleanup

Bin-QA:
1. remove useless code
2. fix shell check error for this file

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -1,19 +1,29 @@
 #!/bin/bash
 
+# get test-case script dirname
+SCRIPT_HOME="$(dirname "$0")"
+# get test-case parent folder name
+SCRIPT_HOME=$(cd "$SCRIPT_HOME/.." && pwd)
+
 # Source from the relative path of current folder
-source $(dirname ${BASH_SOURCE[0]})/config.sh
-source $(dirname ${BASH_SOURCE[0]})/opt.sh
-source $(dirname ${BASH_SOURCE[0]})/logging_ctl.sh
-source $(dirname ${BASH_SOURCE[0]})/pipeline.sh
-source $(dirname ${BASH_SOURCE[0]})/hijack.sh
+# shellcheck disable=SC1091 source=./config.sh
+source "$SCRIPT_HOME/case-lib/config.sh"
+# shellcheck disable=SC1091 source=./opt.sh
+source "$SCRIPT_HOME/case-lib/opt.sh"
+# shellcheck disable=SC1091 source=./logging_ctl.sh
+source "$SCRIPT_HOME/case-lib/logging_ctl.sh"
+# shellcheck disable=SC1091 source=./pipeline.sh
+source "$SCRIPT_HOME/case-lib/pipeline.sh"
+# shellcheck disable=SC1091 source=./hijack.sh
+source "$SCRIPT_HOME/case-lib/hijack.sh"
 
 # force ask buffer data write into file system
 sudo sync -f
 
 # Add tools to command PATH
-if [ ! "$(declare -p TOOL_PATH 2>/dev/null)" ]; then
-    declare -x TOOL_PATH=$(realpath $(dirname ${BASH_SOURCE[1]})/../tools)
-    PATH=$TOOL_PATH:$PATH
+# this line equal `! $(echo $PATH|grep "$SCRIPT_HOME/tools")`
+if [[ ! $PATH =~ $SCRIPT_HOME/tools ]]; then
+    PATH=$SCRIPT_HOME/tools:$PATH
 fi
 
 # setup SOFCARD id
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Before using, please setup these options for the function
-declare -gA OPT_DESC_lst OPT_OPT_lst OPT_PARM_lst OPT_VALUE_lst
+declare -A OPT_DESC_lst OPT_OPT_lst OPT_PARM_lst OPT_VALUE_lst
 
 # option setup && parse function
 func_opt_parse_option()
@@ -32,7 +32,7 @@ func_opt_parse_option()
          local main_script=${BASH_SOURCE[-1]}
          # bash < 4.1
          # local main_script=${BASH_SOURCE[${#BASH_SOURCE[@]} - 1]}
-         cat $main_script |grep '^##'|sed 's/^##//g'
+         grep '^##' "$main_script"|sed 's/^##//g'
     }
 
     _func_create_tmpbash()
@@ -51,13 +51,20 @@ func_opt_parse_option()
             # option will accept parameter
             [ ${OPT_PARM_lst[$i]} -eq 1 ] && _short_opt=$_short_opt':' && _long_opt=$_long_opt':'
         done
-        _op_temp_script=`getopt -o "$_short_opt" --long "$_long_opt" -n "$ST_NAME" -- $@`
-        if [ $? != 0 ] ; then echo "Error parsing option" >&2 ; _func_opt_dump_help ; fi
+        _op_temp_script=$(getopt -o "$_short_opt" --long "$_long_opt" -- "$@")
+        if ! _op_temp_script=$(getopt -o "$_short_opt" --long "$_long_opt" -- "$@") ; then
+            # here output for the End-User who don't care about this function/option couldn't to run
+            echo "Error parsing option" >&2
+            # For debug and fix the option problem, you need to open those code:
+            # printf 'Error parsing %s/%s/%s\n' "$i" "$_short_opt" "$_long_opt" >&2
+            # declare -p |grep 'OPT_[A-Z]*_lst'
+            _func_opt_dump_help
+        fi
     }
 
     _func_opt_dump_help()
     {
-        local i def_value
+        local i
         echo -e "Usage: $0 [OPTION]\n"
         for i in ${!OPT_DESC_lst[*]}
         do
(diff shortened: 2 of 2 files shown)


Bin-QA:
Verify for the CI system, without script break

case-lib/opt.sh
```diff
@@ -115,27 +116,13 @@ func_opt_parse_option()
...
-                if [ $(echo '!'${OPT_VALUE_lst[$opt]}|bc) -eq 1 ]; then
-                    cmd="$cmd -$opt"
-                fi
-            fi
-        done
-        echo $cmd >> $LOG_ROOT/cmd.txt
-        local git_path=$(dirname $0)
-        echo "Branch: "$(git -C $git_path branch |grep '^*') >> $LOG_ROOT/version.txt
-        echo "Commit: "$(git -C $git_path log --oneline -1) >> $LOG_ROOT/version.txt
+    [ "${OPT_VALUE_lst['h']}" -eq 1 ] && _func_opt_dump_help
```
marc-hb, line 119: To be compatible with a future `set -e` and to look more like an assert:
```shell
[ "${OPT_VALUE_lst['h']}" -ne 1 ] || _func_opt_dump_help
```

case-lib/opt.sh
```diff
@@ -115,27 +116,13 @@ func_opt_parse_option()
...
+    # record the git version
+    if [ ! -f "$LOG_ROOT/version.txt" ]; then
+        local git_path
+        git_path=$(dirname "$0")
```
marc-hb, line 123: Can you make `git_path` a global at the top? For consistency with other scripts even if not needed for now.

Bin-QA: good idea, I will update and check it.

case-lib/opt.sh
Bin-QA: do you know which `exit` is load by `set -e`  
we hijack the `exit` function: https://github.com/thesofproject/sof-test/blob/master/case-lib/hijack.sh#L6  
so we can keep test-case more clean and easy to understand  
when case want to finished, it just need to `exit`

marc-hb: > we hijack the exit function

Ouch... :-(

I just did a quick test (`( set -e ; false; exit 2 ; )`) and `set -e` apparently sticks to the native exit(), it ignores the "hijack".

To clean up on exit use `trap sof_test_cleanup EXIT` https://mywiki.wooledge.org/SignalTrap#Examples

This is compatible with `set -e`, demonstration:
```
( exit_sof_test() { printf "MYEXIT"; }; trap exit_sof_test EXIT;    set -e; false )
```
Never redefine built-ins, this is very confusing.

case-lib/opt.sh
```diff
@@ -51,13 +51,14 @@ func_opt_parse_option()
             # option will accept parameter
             [ ${OPT_PARM_lst[$i]} -eq 1 ] && _short_opt=$_short_opt':' && _long_opt=$_long_opt':'
         done
-        _op_temp_script=`getopt -o "$_short_opt" --long "$_long_opt" -n "$ST_NAME" -- $@`
-        if [ $? != 0 ] ; then echo "Error parsing option" >&2 ; _func_opt_dump_help ; fi
+        _op_temp_script=$(getopt -o "$_short_opt" --long "$_long_opt" -- "$@") || {
+            echo "Error parsing option" >&2 && _func_opt_dump_help
+        }
```
marc-hb, line 56: Nit: the `&&` is a bit confusing, it looks like `echo` could fail. Newline or `;` would be clearer.

case-lib/opt.sh
```diff
@@ -115,27 +116,11 @@ func_opt_parse_option()
...
-                fi
-            fi
-        done
-        echo $cmd >> $LOG_ROOT/cmd.txt
-        local git_path=$(dirname $0)
-        echo "Branch: "$(git -C $git_path branch |grep '^*') >> $LOG_ROOT/version.txt
-        echo "Commit: "$(git -C $git_path log --oneline -1) >> $LOG_ROOT/version.txt
+    [ "${OPT_VALUE_lst['h']}" -eq 1 ] && _func_opt_dump_help
+    # record the git version
+    if [ ! -f "$LOG_ROOT/version.txt" -a -f "$SCRIPT_HOME/.git/config" ];then
```
marc-hb, line 121: Shellcheck warning SC2166.

case-lib/opt.sh
```diff
@@ -115,27 +116,11 @@ func_opt_parse_option()
...
+        echo "Branch: $(git -C "$SCRIPT_HOME" branch |grep '^\*')" >> "$LOG_ROOT/version.txt"
+        echo "Commit: $(git -C "$SCRIPT_HOME" log --oneline -1)" >> "$LOG_ROOT/version.txt"
```
marc-hb, line 123:
```shell
 git log -C "$SCRIPT_HOME" --oneline --decorate -n 1 >> "$LOG_ROOT"/version.txt
```
Or maybe `-n 5` for a bit more context.

case-lib/lib.sh
```diff
@@ -1,11 +1,13 @@
 #!/bin/bash
 
+declare -g SCRIPT_HOME=$(realpath "$(dirname $0)/../")
+
```
marc-hb, line 4: 1. You quoted "$( )" which is not necessary. You didn't quote "$0" which _is_ necessary in case of whitespace and reported by shellcheck.

2. I don't think this `realpath` solves any issue (does it?) and it could create some confusion for users with a symbolic link (/home -> /nfs/something or other)

Here's a tested way to make SCRIPT_HOME absolute and whitespace-proof without resolving symbolic links:
```shell
SCRIPT_HOME=$(cd $(dirname "$0")/..; pwd)
```

3. The `-g` has no effect outside functions, did you mean `export`?

case-lib/opt.sh
Bin-QA: when `trap` the `exit` signal will effect `exit` in `tools` ?
I know the `trap` will hijack signal but here also will effect script  
we are not put each `bash function` into `case-lib`  
we also think have some standalone `bash function`  
so we push them into `tools`

case-lib/lib.sh
Bin-QA: yes, you are right  
I try the `dirname $(dirname $0)/../` with `realpath $(dirname $0)../` and `cd $(dirname "$0")/..; pwd`   
the `realpath` maybe confuse when dump the `$SCRIPT_HOME`  
for the option 3, it just my person style, I more like to told the reader to notice this value will be used at other place

case-lib/opt.sh
```diff
@@ -51,13 +51,15 @@ func_opt_parse_option()
             # option will accept parameter
             [ ${OPT_PARM_lst[$i]} -eq 1 ] && _short_opt=$_short_opt':' && _long_opt=$_long_opt':'
         done
-        _op_temp_script=`getopt -o "$_short_opt" --long "$_long_opt" -n "$ST_NAME" -- $@`
-        if [ $? != 0 ] ; then echo "Error parsing option" >&2 ; _func_opt_dump_help ; fi
+        _op_temp_script=$(getopt -o "$_short_opt" --long "$_long_opt" -- "$@") || {
+            echo "Error parsing option" >&2
```
marc-hb, line 55: Would adding some data in the error message make it more useful and help fix mistakes faster?
```shell
>&2 printf 'Error parsing %s/%s/%s\n' "$i" "$short_opt" "$long_opt"# etc.
```

case-lib/opt.sh
```diff
@@ -115,27 +117,11 @@ func_opt_parse_option()
...
-        done
-        echo $cmd >> $LOG_ROOT/cmd.txt
-        local git_path=$(dirname $0)
-        echo "Branch: "$(git -C $git_path branch |grep '^*') >> $LOG_ROOT/version.txt
-        echo "Commit: "$(git -C $git_path log --oneline -1) >> $LOG_ROOT/version.txt
+    [ "${OPT_VALUE_lst['h']}" -eq 1 ] && _func_opt_dump_help
+    # record the git version
+    if [[ ! -f "$LOG_ROOT/version.txt" ]] && [[ -f "$SCRIPT_HOME/.git/config" ]]; then
+        echo -e "Branch:\n$(git -C "$SCRIPT_HOME" branch |sed 's/^/\t/g')" >> "$LOG_ROOT/version.txt"
+        echo -e "Commit:\n$(git -C "$SCRIPT_HOME" log --branches --oneline -n 5)" >> "$LOG_ROOT/version.txt"
```
marc-hb, line 124: See SC2005 and SC2039
```shell
{
  printf 'Branch:'
  git -C "$SCRIPT_HOME" branch |sed 's/^/\t/g'
  printf 'Commit:'
  git -C "$SCRIPT_HOME" log --branches --oneline -n 5

} >> "$LOG_ROOT/version.txt"
```

Bin-QA: sorry, in my environment: shellcheck v0.7.0 without this problem

marc-hb: Sorry that was unclear and inaccurate, I meant have a look at:
https://github.com/koalaman/shellcheck/wiki/SC2005

Bin-QA: got it, how about this style?
```
cat << END >"$LOG_ROOT/version.txt"
output some thing
$(some command)
END
```

marc-hb: This drops the exit status of `some command`.

If you really really want to use `cat <<EOF` for some reason then it should be something like this:
```shell
local val
val=$(some command) # SC2155
cat <<END >"$LOG_ROOT/version.txt"
output something
${val}
END
```
... but that's a lot code more code than simply running `some command`

Bin-QA:
too long time for this PR 
and the target is miss match with original  
I will create the new PR for the code clean and shell check fix

marc-hb:
> I will create the new PR for the code clean and shell check fix

PR #222 correct?

Bin-QA:
> > I will create the new PR for the code clean and shell check fix
> 
> PR #222 correct?

Yes

PR #207, opened 2020-04-24
test-case: change $* to "$@" to pass shellcheck

aiChaoSONG:
this patch does an overall cleanup on all test cases.


aiChaoSONG:
Tested on APL pcm512x with check-capture.sh, PASSED

marc-hb:
> change $* to "$@" to pass shellcheck

change $* to "$@" not to fail in horribly confusing and very time-confusing ways when arguments include stars * or whitespace.

shellcheck is not a purpose in itself (and it's not always right, only most of the time)

Issue #184, opened 2020-04-13
[Bug]check-suspend-resume.sh test failed with non sof issue Call Trace in dmesg


ClarexZhou:
** Description**
Call Trace in dmesg during system suspend and resume. But not sof related. And sof functions still works. But case still failed.
[dmesg_calltrace.txt](https://github.com/thesofproject/sof-test/files/4469391/dmesg_calltrace.txt)

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: https://github.com/thesofproject/linux/commits/topic/sof-dev, commit 4b4681e
    * SOF: https://github.com/thesofproject/sof/releases/tag/v1.5-rc1, commit: 5f7d4e2, using sof-tgl.ri rename to sof-ehl.ri
2) Name of the topology file
    * Topology: sof-ehl-rt5660.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: EHL RVP with RT5660 in I2S mode

dmesg
```
[  826.517489] sof-audio-pci 0000:00:1f.3: Current DSP power state: D3
[  832.067224] ------------[ cut here ]------------
[  832.067229] NETDEV WATCHDOG: enx000ec65b8f43 (asix): transmit queue 0 timed out
[  832.067248] WARNING: CPU: 3 PID: 0 at net/sched/sch_generic.c:443 dev_watchdog+0x23c/0x250
[...]

Issue #167, opened 2020-04-03
Trace test is reporting incorrect test results.


lgirdwood:
The trace test is to test that the trace output is working. i.e. that DMA trace has valid trace data.

The trace test currently parses the trace data for "error" and reports any "error" strings found as a test fail. This is wrong. This test must not parse the trace log for "error" as it reporting failed trace functionality due to "error" strings in the trace output.

Parsing for "error" is valid for other tests but not this one.

aiChaoSONG:
There is already a fix https://github.com/thesofproject/sof-test/pull/161

marc-hb:
> due to "error" strings in the trace output.

Can someone copy an example?

> Parsing for "error" is valid for other tests 

Parsing should always be avoided when possible. I'm not familiar with this test suite yet but in general searching for kernel errors can be done with just one line:
```
sudo journalctl [ --boot ] --dmesg -p 3
```
and for warnings:
```
sudo journalctl [ --boot ] --dmesg -p 4
```

aiChaoSONG:
This issue is already fixed, we are not detecting error from firmware log anymore.
@marc-hb sorry for misleading you here. the firmware log is not in dmesg, it is extract by sof-logger.

marc-hb:
> Parsing for "error" is valid for other tests but not this one.

Too bad no other test does this.

PR #208, opened 2020-04-26
tools: exit successfully if call trace is not from alsa

aiChaoSONG:
Previouly, every call trace will fail the test
case, that's not correct. Only ALSA related
call trace should fail the test case.

This patch checks call trace message, and
only failed test case with ALSA related
call trace.

fix https://github.com/thesofproject/sof-test/issues/184

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -22,16 +22,19 @@ project_key="sof-audio"
 #echo "run $0 with parameter '$*' for check kernel message error"
 
 if [ "$ignore_str" ]; then
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
+    err=$(eval "$cmd"|grep 'Call Trace' -A5 -B3)$(eval "$cmd" | grep "$project_key" | grep -E "$err_str"|grep -vE "$ignore_str")
 else
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str")
+    err=$(eval "$cmd"|grep 'Call Trace' -A5 -B3)$(eval "$cmd" | grep "$project_key" | grep -E "$err_str")
 fi
 
 if [ "$err" ]; then
-    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught dmesg error"
+    echo "$(date -u '+%Y-%m-%d %T %Z') [ERROR] Caught dmesg error"
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
+    # exclude none alsa trace, but still print trace message
+    snd_trace=$(echo "$err" | grep "snd")
+    [ -z "$snd_trace" ] && builtin exit 0
     builtin exit 1
 fi
 


tools/sof-kernel-log-check.sh
```diff
@@ -22,16 +22,19 @@ project_key="sof-audio"
...
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
+    err=$(eval "$cmd"|grep 'Call Trace' -A5 -B3)$(eval "$cmd" | grep "$project_key" | grep -E "$err_str"|grep -vE "$ignore_str")
 else
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str")
+    err=$(eval "$cmd"|grep 'Call Trace' -A5 -B3)$(eval "$cmd" | grep "$project_key" | grep -E "$err_str")
 fi
 
 if [ "$err" ]; then
-    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught dmesg error"
+    echo "$(date -u '+%Y-%m-%d %T %Z')" "[ERROR]" "Caught dmesg error"
```
Bin-QA, line 31: you can include all string into one `"`

tools/sof-kernel-log-check.sh
```diff
@@ -22,16 +22,19 @@ project_key="sof-audio"
...
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
+    # exclude none alsa trace, but still print trace message
+    snd_trace=$(echo "$err" | grep "snd")
```
Bin-QA, line 36: I'm afraid here you will ignore 'sof' refer 'snd'

tools/sof-kernel-log-check.sh
aiChaoSONG: thanks, sure

tools/sof-kernel-log-check.sh
aiChaoSONG: for alsa layer, function name is snd_*, for SOC layer, function name is snd_soc_*, and for SOF layer, function name is snd_sof_*, so only grep "snd" should be enough.

aiChaoSONG:
@marc-hb Sorry for made a mistake. the issue should be #184 , not #167. In #184,  A none SOF call trace caused one test case failed.

aiChaoSONG:
@Bin-QA I think Marc's suggestion is good, is there any possibility to use this: `journalctl --dmesg -p 3 | grep -v -e '$ignore_issue_123" -e "$ignore_dma_trace"`

Bin-QA:
how to clean buffer?  
we force on current, like current case, current loop  
so we need to clean the buffer  
we not just force the kernel boot log

marc-hb:
```shell
test3_start=$(date +%s)
run_test3
test3_end=$(date +%s)
journalctl --boot --since=@"$test3_start" --until=@"$test3_end"
```
`dmesg -c` is more like a hack.

marc-hb:
Longer example
```shell
test3_start=$(date +%s)
run_test3
test3_end=$(date +%s)
# Errors or worse
journalctl --boot --since=@"$test3_start" --until=@"$test3_end" -p 3
# Warnings only
journalctl --boot --since=@"$test3_start" --until=@"$test3_end" -p 4..4
```

PR #210, opened 2020-04-26
test-case: manual: wov trigger test case

YvonneYang2:
Add new wov trigger test case in manual test.

### DIFF
--- a/test-case/manual/check-wov-trigger.md
+++ b/test-case/manual/check-wov-trigger.md
@@ -1,0 +1,63 @@
+
+# Wov pipline trigger test
+Wov pipline trigger testing for now.
+
+# Preconditions
+1. Wov-test.sh in wov-script folder.**Please check the script before run it to make sure use the right wov hw device id which match your platform.**
+2. Test unit support wov test
+
+# Test Description
+* Change to different preamble/history via command worked on wov-test script.
+* Change to fs16/24/32_le format via arecord command 
+
+## Wov Test Case: 
+#####1.check WoV with s16_le format 
+1. Run command to start wov via "arecord -Dhw:0,$device_id -M -N -f s16_le -r 16000 -i -vvv -c2 --buffer-size=68000 ./tmp.wav" 
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####2.Check WoV with s24_le format 
+1. Run command to start wov via "arecord -Dhw:0,$device_id -M -N -f s24_le -r 16000 -i -vvv -c2 --buffer-size=68000 ./tmp.wav" 
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####3.Check WoV with s32_le format 
+1. Run command to start wov via "arecord -Dhw:0,$device_id -M -N -f s32_le -r 16000 -i -vvv -c2 --buffer-size=68000 ./tmp.wav" 
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####4.Check WoV with different buffer sizes(must > 67200) 
+1. Run command to start wov via "arecord -Dhw:0,$device_id -M -N -f s32_le -r 16000 -i -vvv -c2 --buffer-size=80000 ./tmp.wav" ** Note:The buffer size varies with different platform**
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Try different period size. eg.68000/70000/80000...
+4. Check the quality of recorded audio files.
+
+#####5.Check WoV with different period sizes(eg. 400/500/600...) 
+1. Run with command "sudo su"
+2. ./wov-test.sh -p 2100 -h 1800 -f s16_le -s 500 
+3. Try different period sizes, and check the output period sizes if changed.
+4. Clap hands or make some noise to trigger the wov after process starts.
+5. Check the quality of recored audio files. 
+
+#####6.Check WoV with different preamble sizes
+1. Run with command "sudo su"
+2. ./wov-test.sh -p 2100 -h 1800 -f s16_le
+3. Try different preamble sizes, and check the output preamble sizes if changed.
+4. Clap hands or make some noise to trigger the wov after process starts.
+5. Check the quality of recored audio files.
+
+#####7.Check WoV with different history depths
+1. Run with command "sudo su"
+2. ./wov-test.sh -p 2100 -h 1800 -f s16_le
+3. Try different history depths, and check the output history depths if changed.
+4. Clap hands or make some noise to trigger the wov after process starts 
+5. Check the quality of recorded audio files.
+
+#####8.Check WoV with different threadholds
+1. wov test + headphone playback + headset capture.
+2. wov test + headphone playback + Speaker playback + HDMI/DP playback 
+
+## Expected results
+* Wov can be triggerd successfully.
+* Wov audio files quality is well.
+
--- a/test-case/manual/wov-script/wov-test.sh
+++ b/test-case/manual/wov-script/wov-test.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+
+rm -rf blob
+preamble_time=0
+history_depth=0
+format=s16_le
+buffer_size=68000
+max_loop=100
+preamble_time_init=2100
+history_depth_init=1800
+str=$(amixer contents | grep "Detector")
+control=${str#*=}
+control_id=${control%%,*}
+
+
+usage(){
+        echo "Wrong parameters, please specify the test item"
+        echo "Eg: '-p' preamble_time, '-h' history_depth, '-f' format and '-s' period-size\
+        ./$0 -p 2500 -h 3000 -f s16_le -s 500"
+}
+
+# DEC to HEX
+dec2hex(){
+        printf "%x" "$1"
+}
+
+_save_def(){
+	echo "Save WoV as default configure blob:"
+	if [ ! -f detector-default-config ]; then
+		sudo sof-ctl -Dhw:0 -n "$control_id" -br -o detector-default-config
+	fi
+}
+
+_set_default(){
+	echo "Set WoV as default configure blob:"
+	sudo sof-ctl -Dhw:0 -n "$control_id" -br -s detector-default-config
+}
+
+# get the preamble time and history depth
+_get_pt_hd(){
+	_set_default
+        # save current WoV configure blob
+	sof-ctl -Dhw:0 -n "$control_id" -br -o blob
+	def_1st=x08
+	def_snd=x34
+
+}
+
+# set the preamble time & history depth
+_set_pt_hd(){
+        preamble_time_aft=$(dec2hex $preamble_time)
+	length_pt=$(echo "$preamble_time_aft" |wc -L)
+	if [ "$length_pt" -eq 3 ]; then
+		preamble_time_aft=0$preamble_time_aft
+	fi
+	pt_aft_1st=$(echo "$preamble_time_aft" |cut -c 1-2)
+	pt_aft_1st=x$pt_aft_1st
+	pt_aft_snd=$(echo "$preamble_time_aft" |cut -c 3-4)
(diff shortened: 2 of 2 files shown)


aiChaoSONG:
Manual test case is good, I suggest you attach the wov-test.sh script, too.

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,72 @@
...
+2. Test unit support wov test
+
+# Test Description
+* Change to different preamble/history via command worked on wov-test script.
+* Change to fs16/24/32_le format via arecord command 
+
+## Wov Test Case:
+#####1.check WoV with s16_le format 
+
+1)Arecord -Dhw:0,99 -M -N -f s16_le -r 16000 -i -vvv -c2 --period-size=68000 ./tmp.wav
```
Bin-QA, line 17: look like the command, so can you modify it into code style?

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,72 @@
...
+# Test Description
+* Change to different preamble/history via command worked on wov-test script.
+* Change to fs16/24/32_le format via arecord command 
+
+## Wov Test Case:
+#####1.check WoV with s16_le format 
+
+1)Arecord -Dhw:0,99 -M -N -f s16_le -r 16000 -i -vvv -c2 --period-size=68000 ./tmp.wav 
+
+2)if process starts, clap hands or make some noise to trigger the wov.
```
Bin-QA, line 19: 'if' or 'after', because 'if' need failed step

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,72 @@
...
+
+#####2.Check WoV with s24_le format
```
Bin-QA, line 21: need check/verify `tmp.wav` file?

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# DEC to HEX
+dec2hex(){
+        printf "%x" $1
+}
+
+_save_def(){
+	echo "Save WoV as default configure blob:"
+	if [ ! -f detector-default-config ]; then
+		sudo sof-ctl -Dhw:0 -n 163 -br -o detector-default-config
```
keqiaozhang, line 24: Control ID is not fixed and it's different on each platform.
So please don't use hard code. You can use amixer to get the ID number:
eg. 
$ amixer controls |grep "DETECT10.0 Detector Config"
numid=162,iface=MIXER,name='DETECT10.0 Detector Config'

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,112 @@
...
+_set_default(){
+	echo "Set WoV as default configure blob:"
+	sudo sof-ctl -Dhw:0 -n 163 -br -s detector-default-config
+}
+
+# get the preamble time and history depth
+_get_pt_hd(){
+	_set_default
+        # save current WoV configure blob
+	sof-ctl -Dhw:0 -n 163 -br -o blob
```
keqiaozhang, line 37: Same as above.

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,112 @@
...
+_save_def(){
+	echo "Save WoV as default configure blob:"
+	if [ ! -f detector-default-config ]; then
+		sudo sof-ctl -Dhw:0 -n 163 -br -o detector-default-config
+	fi
+}
+
+_set_default(){
+	echo "Set WoV as default configure blob:"
+	sudo sof-ctl -Dhw:0 -n 163 -br -s detector-default-config
```
keqiaozhang, line 30: Same as above.

aiChaoSONG:
@YvonneYang2 
- please remove `.check-wov-trigger.md.swo` and `.check-wov-trigger.md.swp`, these two files are useless.
- please use shellcheck to fix script warnings.

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,64 @@
...
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####3.Check WoV with s32_le format 
+1. Run command"arecord -Dhw:0,99 -M -N -f s24_le -r 16000 -i -vvv -c2 --period-size=68000 ./tmp.wav" 
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####4.Check WoV with different buffer sizes(must > 67200) 
+1. Run command"record -Dhw:0,99 -M -N -f s32_le -r 16000 -i -vvv -c2 --buffer-size=68000 ./tmp.wav"
```
xiulipan, line 30: This size varies with different platform, please add some comments here to mention this.

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,64 @@
...
+* Change to fs16/24/32_le format via arecord command 
+
+## Wov Test Case:
+#####1.check WoV with s16_le format 
+1. Run command via "arecord -Dhw:0,99 -M -N -f s16_le -r 16000 -i -vvv -c2 --period-size=68000 ./tmp.wav" 
+2. Clap hands or make some noise to trigger the wov after process starts.
+3. Check the quality of recorded audio files.
+
+#####2.Check WoV with s24_le format 
+1. Run command"arecord -Dhw:0,99 -M -N -f s24_le -r 16000 -i -vvv -c2 --period-size=68000 ./tmp.wav"
```
xiulipan, line 20: The PCM will be different on different platform, do not hard code to 99 here.

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+	
+# test with arecord
+for (( i=1; i <= "$max_loop"; i++ ));
+do
+	file_name='wov_pt'-"$preamble_time"'_hd'-"$history_depth"'_f-'"$format"'_ps'-"$period_size"'.wav'
+	if arecord -Dhw:0,8 -M -N -c 2 -f "$format" --buffer-size="$buffer_size" -r 16000 -vvv -d 5 "$file_name";then
+		echo "WoV test passed, please check the recorded wav file: wov_pt-""$file_name"
+	else
+		echo "WoV test failed, please have a check"
+		exit 1
```
marc-hb, line 116:
```shell
die()
{
    >&2 printf "ERROR: "
    >&2 printf "$@"
   exit 1
}

die 'WoV test failed, please check %s\n' "$something"

File: test-case/manual/wov-script/wov-test.sh [hunk 10]
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+
+rm -rf blob
```
marc-hb, line 3: This is the only command here, all the rest is just initializing variables so it's out of place. Move it before the first command, I think this means before the `_save_def` invocation

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+        ./$0 -p 2500 -h 3000 -f s16_le -s 500"
+}
+
+# DEC to HEX
+dec2hex(){
+        printf "%x" "$1"
+}
+
+_save_def(){
+	echo "Save WoV as default configure blob:"
```
marc-hb, line 28: I'm confused because there's another file named `blob`.

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+_set_default(){
+	echo "Set WoV as default configure blob:"
+	sudo sof-ctl -Dhw:0 -n "$control_id" -br -s detector-default-config
+}
+
+# get the preamble time and history depth
+_get_pt_hd(){
+	_set_default
+        # save current WoV configure blob
+	sof-ctl -Dhw:0 -n "$control_id" -br -o blob
```
marc-hb, line 43: Use longer, more descriptive names and an indirection variable so they can be changed easily if needed in the future.
```shell
 # just a suggestion, do not use these names, please find better names
wov_config=wov.txt

sof-ctl -Dhw:0 -n "$control_id" -br -o blob "$wov_config"
```

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+	fi
+	pt_aft_1st=$(echo "$preamble_time_aft" |cut -c 1-2)
+	pt_aft_1st=x$pt_aft_1st
+	pt_aft_snd=$(echo "$preamble_time_aft" |cut -c 3-4)
+	pt_aft_snd=x$pt_aft_snd
+        history_depth_aft=$(dec2hex $history_depth)
+	length_hd=$(echo "$history_depth_aft" |wc -L)
+	if [ "$length_hd" -eq 3 ]; then
+		history_depth_aft=0$history_depth_aft
+	fi
```
marc-hb, line 64: I don't understand this, why not just:
```shell
dec2hex () 
{ 
    printf "0x%x" "$1"
}
```
Can you share some examples of actual values?

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+	hd_aft_1st=$(echo "$history_depth_aft" |cut -c 1-2)
+	hd_aft_1st=x$hd_aft_1st
+	hd_aft_snd=$(echo "$history_depth_aft" |cut -c 3-4)
+	hd_aft_snd=x$hd_aft_snd
+}
+
+# update the blob
+_update_blob(){
+	sed -i 's/\'"$def_snd"'\'"$def_1st"'/\'"$hd_aft_snd"'\'"$hd_aft_1st"'/g' blob 
+	sed -i '1s/\'"$hd_aft_snd"'\'"$hd_aft_1st"'/\'"$pt_aft_snd"'\'"$pt_aft_1st"'/1' blob
```
marc-hb, line 74: You can simplify the quoting a great lot (and get rid of the last shellcheck warning) by dropping the outermost quoting. Demonstration:
```shell
str_with_space='i n t e l'
grep intel ~/.gitconfig | sed -e s/intel/"$str_with_space"/ # <- not outer quoting, only one quoting level.

	email = Yvonne@i n t e l.com
```
This works and it is safe because whitespace is dangerous only _before_ variable expansion, whitespace doesn't matter any more after variable expansion.
http://rg1-teaching.mpi-inf.mpg.de/unixffb-ss98/quoting-guide.html

test-case/manual/wov-script/wov-test.sh
```diff
@@ -0,0 +1,118 @@
...
+	fi
+	hd_aft_1st=$(echo "$history_depth_aft" |cut -c 1-2)
+	hd_aft_1st=x$hd_aft_1st
+	hd_aft_snd=$(echo "$history_depth_aft" |cut -c 3-4)
+	hd_aft_snd=x$hd_aft_snd
+}
+
+# update the blob
+_update_blob(){
+	sed -i 's/\'"$def_snd"'\'"$def_1st"'/\'"$hd_aft_snd"'\'"$hd_aft_1st"'/g' blob
```
marc-hb, line 73: This is confusing because `sed` doesn't work on binaries. I suspect what you called "blob" is not actually a blob but a CSV file, correct? If correct then call it something else (and use an indirection variable)

test-case/manual/check-wov-trigger.md
```diff
@@ -0,0 +1,63 @@
...
+#####6.Check WoV with different preamble sizes
+1. Run with command "sudo su"
+2. ./wov-test.sh -p 2100 -h 1800 -f s16_le
+3. Try different preamble sizes, and check the output preamble sizes if changed.
+4. Clap hands or make some noise to trigger the wov after process starts.
+5. Check the quality of recored audio files.
+
+#####7.Check WoV with different history depths
+1. Run with command "sudo su"
+2. ./wov-test.sh -p 2100 -h 1800 -f s16_le
```
ClarexZhou, line 51: -h value should be <= -p value
-h value shouldn't be set bigger than 2100

xiulipan:
close for outdated pr.

PR #211, opened 2020-04-27
sof-test: replace dloge/exit pattern with die

aiChaoSONG:
add die function, and use it to clean up old dloge error message && exit pattern

### DIFF
--- a/case-lib/pipeline.sh
+++ b/case-lib/pipeline.sh
@@ -4,13 +4,11 @@ func_pipeline_export()
 {
     # no parameter input the function
     if [ $# -lt 1 ]; then
-        dlogi "Topology file name is not specified, unable to run command: $SCRIPT_NAME"
-        exit 1
+        die "Topology file name is not specified, unable to run command: $SCRIPT_NAME"
     fi
     # got tplg_file, verify file exist
     tplg_path=$(func_lib_get_tplg_path "$1") || {
-        dloge "No available topology for pipeline export"
-        exit 1
+        die "No available topology for pipeline export"
     }
     dlogi "$SCRIPT_NAME will use topology $tplg_path to run the test case"
 
@@ -39,7 +37,7 @@ func_pipeline_export()
     do
         eval "$line"
     done
-    [[ ! "$PIPELINE_COUNT" ]] && dlogw "Failed to parse $tplg_path, please check topology parsing command" && exit 1
+    [[ ! "$PIPELINE_COUNT" ]] && die "Failed to parse $tplg_path, please check topology parsing command"
     [[ $PIPELINE_COUNT -eq 0 ]] && dlogw "No pipeline found with option: $opt, unable to run $SCRIPT_NAME" && exit 2
     return 0
 }
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -69,16 +69,14 @@ func_test_filter()
     #       currently the test discards EQ pipelines except first one.
     Filterid=$(amixer -D hw:"$sofcard" controls | sed -n -e "/eq${testfilter}/I "'s/numid=\([0-9]*\),.*/\1/p' | head -1)
     if [ -z "$Filterid" ]; then
-        dloge "can't find $testfilter"
-        exit 1
+        die "can't find $testfilter"
     fi
 
     declare -a FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))
     nFilterList=${#FilterList[@]}
     dlogi "$testfilter list, num= $nFilterList, coeff files= ${FilterList[*]}"
     if [ "$nFilterList" -eq  0 ]; then
-        dloge "$testfilter flter coeff list error!"
-        exit 1
+        die "$testfilter flter coeff list error!"
     fi
 
     for i in $(seq 1 $loop_cnt)
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -66,21 +66,19 @@ do
         [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
         sleep 1
         if [ $i -eq 15 ]; then
-            dlogi "dsp is not suspended after 15s, end test"
-            exit 1
+            die "dsp is not suspended after 15s, end test"
         fi
     done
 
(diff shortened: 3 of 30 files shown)


marc-hb:
While doing this, could you go the extra mile and implement the usual `die()` function instead?
```shell
die()
{
  dloge "$@"
  exit 1
}
```

aiChaoSONG:
@marc-hb implemented die function, and do a clean up with it in the whole repo. please help to review again. Currently, please ignore shellcheck, we will do the cleanup for it step by step.

case-lib/pipeline.sh
```diff
@@ -4,7 +4,7 @@ func_pipeline_export()
 {
     # no parameter input the function
     if [ $# -lt 1 ]; then
-        dlogi "Topology file name is not specified, unable to run command: ${BASH_SOURCE[-1]}" && exit 1
+        dloge "Topology file name is not specified, unable to run command: ${BASH_SOURCE[-1]}" && exit 1
```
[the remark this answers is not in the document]
aiChaoSONG: For the topology verification, I think It should be error, and we should make test case failed, rather than NA, NA will make people confuse under this condition.

case-lib/pipeline.sh
```diff
@@ -38,7 +38,7 @@ func_pipeline_export()
         eval "$line"
     done
     IFS="$OLD_IFS"
-    [[ ! "$PIPELINE_COUNT" ]] && dlogw "Failed to parse $tplg_path, please check topology parsing command" && exit 1
+    [[ ! "$PIPELINE_COUNT" ]] && dloge "Failed to parse $tplg_path, please check topology parsing command" && exit 1
```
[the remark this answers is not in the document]
aiChaoSONG: pipeline.sh not source lib.sh, can die be used here?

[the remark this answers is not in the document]
Bin-QA: yes, you source lib.sh,  
which source pipeline.sh, but in pipeline.sh function just define not execute,  
so you can try it in CI system

case-lib/pipeline.sh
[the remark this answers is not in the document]
Bin-QA: for the all case, it don't need to topology verification, it is different scope

case-lib/pipeline.sh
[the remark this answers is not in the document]
aiChaoSONG: @Bin-QA tested on helios, it's ok to call die in pipeline.sh

test-case/check-audio-equalizer.sh
```diff
@@ -44,7 +44,7 @@ func_test_eq()
     local id=$1
     local conf=$2
 
-    [ ! -f $conf ] && dloge "$conf not exist" && exit 1
+    [ ! -f $conf ] && die "$conf not exist"
```
marc-hb, line 47: Not in this PR now (too much change) but in the future  try to stick to the codestyle `test_something || die` and `assert_something || die` codestyle that is standard, reads like plain english and is "more obviously" compatible with `set -e`

For instance here it would be:
```shell
 [ -f $conf ] || die "$conf not exist"
```

test-case/verify-kernel-module-load-probe.sh
```diff
@@ -21,6 +21,6 @@ DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
 dlogi "Checking if sof relative modules loaded"
 dlogc "lsmod | grep \"sof\""
 lsmod | grep "sof"
-[[ $? -ne 0 ]] && dloge "No available sof module found! Dumping lsmod:" && lsmod && exit 1
+[[ $? -ne 0 ]] && die "No available sof module found!"
```
marc-hb, line 24: You dropped the `lsmod`
```shell
lsmod | grep 'sof' || {
 lsmod
 die "No available sof module found!"
}
```

test-case/check-audio-equalizer.sh
aiChaoSONG: thanks, sure. will submit a new PR later to correct this kind of code

test-case/verify-kernel-module-load-probe.sh
aiChaoSONG: done!

marc-hb:
@aiChaoSONG can you resolve the conflicts?

aiChaoSONG:
@Bin-QA @marc-hb conflicts resolved, please review and merge

test-case/verify-kernel-module-load-probe.sh
```diff
@@ -20,7 +20,9 @@ DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
 
 dlogi "Checking if sof relative modules loaded"
 dlogc "lsmod | grep \"sof\""
-lsmod | grep "sof"
-[[ $? -ne 0 ]] && dloge "No available sof module found! Dumping lsmod:" && lsmod && exit 1
+lsmod | grep "sof" || {
+    lsmod
+    die "No available sof module found!"
+}
```
Bin-QA, line 26: already catch 'sof' need `lsmod` again? or need some output

fredoh9: looks `lsmod `is added for more debugging, not very useful but it is paired with original file.

test-case/check-kmod-load-unload.sh
```diff
@@ -93,13 +91,12 @@ do
...
     if [[ $? -ne 0 ]]; then
-        dloge "Found error(s) in kernel log after module insertion"
-        exit 1
+        die "Found error(s) in kernel log after module insertion"
     fi
 
     dlogi "checking if firmware is loaded successfully"
     $(dirname ${BASH_SOURCE[0]})/verify-sof-firmware-load.sh > /dev/null
-    [[ $? -ne 0 ]] && dloge "Falied to load firmware after module insertion" && exit 1
+    [[ $? -ne 0 ]] && die "Falied to load firmware after module insertion"
```
marc-hb, line 99: (older) typo

test-case/verify-kernel-module-load-probe.sh
marc-hb: This PR is a very large search/replace so let's make sure there is no functional changes (hidden) in it.

aiChaoSONG:
@marc-hb all reviews are addressed, please review and merge.

PR #213, opened 2020-04-29
test-case: fix a bash grammer error

aiChaoSONG:
There should be a 'space' between function
and its arguments


marc-hb:
Interesting, did this work before?

Next time use more specific words in the commit message and subject: `multiple-pause-resume.sh`, `func_parse_option`, etc.

aiChaoSONG:
> Interesting, did this work before?
> 
> Next time use more specific words in the commit message and subject: `multiple-pause-resume.sh`, `func_parse_option`, etc.

this function is used to parse cmd-line options, if it fails, we use the default value, so test case still runs.

PR #214, opened 2020-05-07
[WIP]Journalctl instead of query from /var/log/kern.log

Bin-QA:
Use journalctl to instead of query kernel log from /var/log/kern.log  
the query kernel log from /var/log/kern.log design for avoid dmesg buffer flush  
when have too information write into dmesg  
journalctl try to avoid /var/log/kern.log flush delay,  
journalctl also support filter message by level  
if we just want to catch 'warning + error', we can add parameter `-p 3..4` for this command


Bin-QA:
verify patch on CI system

Bin-QA:
Close PR with verify  
because on the DUT have the Clock issue  
when system kernel boot up it will record the time stamp into the journal system,  
we will try to sync time on the DUT, time stamp will been reset,  
so query from journal system with time stamp will catch error information  
Current close this PR when fix the DUT time issue

PR #220, opened 2020-05-12
tools: sof-dump-status add 'know' for pci hw_name

Bin-QA:
when target DUT have multiple PCI device
query from intel id list couldn't find target pci id
this patch to fix out of intel PCI device

### DIFF
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -69,7 +69,7 @@ class clsSYSCardInfo():
 
     def loadPCI(self):
         self.pci_lst.clear()
-        exit_code, output=subprocess.getstatusoutput("lspci |grep audio -i")
+        exit_code, output=subprocess.getstatusoutput("lspci |grep audio -i|grep intel -i")
         # grep exit 1 means nothing matched
         if exit_code != 0:
             return


Bin-QA:
@xiulipan @aiChaoSONG this patch to fix multiple PCI device on the DUT  
whether need to direct filter just intel PCI device with modify at [line 72](https://github.com/thesofproject/sof-test/blob/master/tools/sof-dump-status.py#L72)  
```
-        exit_code, output=subprocess.getstatusoutput("lspci |grep audio -i")
+        exit_code, output=subprocess.getstatusoutput("lspci |grep audio -i|grep intel -i")
```

tools/sof-dump-status.py
```diff
@@ -85,8 +85,12 @@ def loadPCI(self):
...
                     tmp_line = tmp_output[i].split()
                     break
-            pci_info['hw_id']="0x" + tmp_line[2] + tmp_line[1] + " 0x" + tmp_line[4] + tmp_line[3]
-            pci_info['hw_name'] = self._pci_ids["0x" + tmp_line[4] + tmp_line[3]]
+            pci_id = "0x" + tmp_line[4] + tmp_line[3]
+            pci_info['hw_id']="0x" + tmp_line[2] + tmp_line[1] + " " + pci_id
+            if pci_id in self._pci_ids:
+                pci_info['hw_name'] = self._pci_ids[pci_id]
+            else:
+                pci_info['hw_name'] = 'unknow'
```
aiChaoSONG, line 93: please use "unknown“, not unknow

Bin-QA: So you think we need to cover out of intel pci device?

aiChaoSONG: My advice is only an English grammar advice.  In my opinion, all of our DUT will never have a new audio pci device except that USB one. so I think this change is a little redundant. Maybe Xiuli has something to say. @xiulipan

Bin-QA: it is not the USB device, the DUT is notebook,  
so have more than one PCI device of audio is reasonable

PR #222, opened 2020-05-15
fix shell check error for case-lib/*

Bin-QA:
Update for fix shell check error for case-lib/*
expect case-lib/config.sh file

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -88,13 +88,11 @@ func_lib_disable_pulseaudio()
 {
     [[ "${#PULSECMD_LST[@]}" -ne 0 ]] && return
     # store current pulseaudio command
-    OLD_IFS="$IFS" IFS=$'\n'
-    PULSECMD_LST=( $(ps -C pulseaudio -o user,cmd --no-header) )
-    IFS="$OLD_IFS"
+    readarray -t PULSECMD_LST < <(ps -C pulseaudio -o user,cmd --no-header)
     [[ "${#PULSECMD_LST[@]}" -eq 0 ]] && return
     func_lib_check_sudo
     # get all running pulseaudio paths
-    PULSE_PATHS=( $(ps -C pulseaudio -o cmd --no-header | awk '{print $1}') )
+    readarray -t PULSE_PATHS < <(ps -C pulseaudio -o cmd --no-header | awk '{print $1}'|sort -u)
     for PA_PATH in "${PULSE_PATHS[@]}"
     do
         # rename pulseaudio before kill it
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -1,19 +1,36 @@
 #!/bin/bash
 
-# Source from the relative path of current folder
-source $(dirname ${BASH_SOURCE[0]})/config.sh
-source $(dirname ${BASH_SOURCE[0]})/opt.sh
-source $(dirname ${BASH_SOURCE[0]})/logging_ctl.sh
-source $(dirname ${BASH_SOURCE[0]})/pipeline.sh
-source $(dirname ${BASH_SOURCE[0]})/hijack.sh
+# get test-case information
+SCRIPT_HOME="$(dirname "$0")"
+# get test-case parent folder name
+SCRIPT_HOME=$(cd "$SCRIPT_HOME/.." && pwd)
+# shellcheck disable=SC2034 # external script can use it
+SCRIPT_NAME="$0"  # get test-case script load name
+# shellcheck disable=SC2034 # external script can use it
+SCRIPT_PRAM="$*"  # get test-case parameter
 
-# force ask buffer data write into file system
-sudo sync -f
+# Source from the relative path of current folder
+# shellcheck disable=SC1091 source=./config.sh
+source "$SCRIPT_HOME/case-lib/config.sh"
+# shellcheck disable=SC1091 source=./opt.sh
+source "$SCRIPT_HOME/case-lib/opt.sh"
+# shellcheck disable=SC1091 source=./logging_ctl.sh
+source "$SCRIPT_HOME/case-lib/logging_ctl.sh"
+# shellcheck disable=SC1091 source=./pipeline.sh
+source "$SCRIPT_HOME/case-lib/pipeline.sh"
+# shellcheck disable=SC1091 source=./hijack.sh
+source "$SCRIPT_HOME/case-lib/hijack.sh"
+
+# restrict bash version for some bash feature
+[[ $(echo -e "$BASH_VERSION\n4.1"|sort -V|head -n 1) == '4.1' ]] || {
+    dlogw "Bash version: ${BASH_VERSINFO[0]}.${BASH_VERSINFO[1]} should > 4.1"
+    exit 2
+}
 
 # Add tools to command PATH
-if [ ! "$(declare -p TOOL_PATH 2>/dev/null)" ]; then
-    declare -x TOOL_PATH=$(realpath $(dirname ${BASH_SOURCE[1]})/../tools)
-    PATH=$TOOL_PATH:$PATH
+# this line equal `! $(echo $PATH|grep "$SCRIPT_HOME/tools")`
+if [[ ! $PATH =~ $SCRIPT_HOME/tools ]]; then
+    export PATH=$SCRIPT_HOME/tools:$PATH
 fi
 
 # setup SOFCARD id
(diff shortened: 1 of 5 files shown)


Bin-QA:
Wait for CI system detect error

case-lib/lib.sh
```diff
@@ -1,43 +1,70 @@
...
+export DMESG_LOG_START_LINE
 
-declare -g SOF_LOG_COLLECT=0
+export SOF_LOG_COLLECT=0
 
 func_lib_setup_kernel_last_line()
 {
-    declare -g KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    export KERNEL_LAST_LINE
```
marc-hb, line 56: See above. `declare -g` is a no-op, drop it (`export` is `declare -x`, not `declare -g`)

case-lib/lib.sh
```diff
@@ -1,43 +1,70 @@
...
 
 func_lib_start_log_collect()
 {
-    local is_etrace=${1:-0}
-    local ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc
+    local is_etrace=${1:-0} ldcFile
+    ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+        dlogw "sof-dump-status.py -p to query platformat failed"
+        return
+    }
```
marc-hb, line 65: This is useful and generally not visible externally but it's unfortunate to hide this in a commit with a placeholder commit message.

case-lib/lib.sh
```diff
@@ -46,39 +73,38 @@ func_lib_start_log_collect()
...
 
 func_lib_check_sudo()
 {
-    func_hijack_setup_sudo_level
-    [[ $? -ne 0 ]] && \
-        dlogw "Command needs root privilege to run, please configure SUDO_PASSWD in case-lib/config.sh" && \
+    func_hijack_setup_sudo_level || {
+        dlogw "Command needs root privilege to run, please configure SUDO_PASSWD in case-lib/config.sh"
         exit 2
+    }
```
marc-hb, line 88: Use a `die()` function

case-lib/logging_ctl.sh
```diff
@@ -27,52 +26,41 @@ _func_log_cmd()
...
+    ppcmd=$(ps -p $PPID -o cmd --noheader|awk '{print $2;}')
+    local ext_message="" cmd
+    # confirm this script is loaded by other script, and add the flag for it
+    file "$ppcmd" 2>/dev/null |grep -q 'shell script' && ext_message=" Sub-Test:"
+    for key in "${!LOG_LIST[@]}";
+    do
+        # dymaic alias the command with different content
+        cmd="alias $key='echo \$(date -u \"+%F %T %Z\")$ext_message ${LOG_LIST[$key]}'"
+        eval "$cmd"
+    done
```
marc-hb, line 41: Please explain this change in a different commit.

case-lib/logging_ctl.sh
```diff
@@ -27,52 +26,41 @@ _func_log_cmd()
...
-    local log_dir=$(dirname ${BASH_SOURCE[0]})/../logs/
-    log_dir=$(realpath $log_dir)
-    local timetag=$(date +%F"-"%T)"-"$RANDOM
-    local cur_pwd=$PWD
+    local case_name log_dir timetag
+    case_name=$(basename "$SCRIPT_NAME")
     case_name=${case_name%.*}
-    local case_folder=$log_dir/$case_name
-    mkdir -p $case_folder/$timetag
+    log_dir=$SCRIPT_HOME/logs/$case_name
```
marc-hb, line 55: While quoting is not necessary here it would be good for consistency.

case-lib/opt.sh
```diff
@@ -77,9 +79,10 @@ func_opt_parse_option()
                 if [ "${OPT_VALUE_lst[$i]}" ]; then
                     if [ "${OPT_PARM_lst[$i]}" -eq 1 ]; then
                         echo -e "\t""Default Value: ${OPT_VALUE_lst[$i]}"
+                    elif [ "${OPT_VALUE_lst[$i]}" -eq 0 ]; then
+                        echo -e "\t""Default Value: Off"
                     else
-                        [ "${OPT_VALUE_lst[$i]}" -eq 0 ] && echo -e "\t""Default Value: Off" \
-                            || echo -e "\t""Default Value: On"
+                        echo -e "\t""Default Value: On"
```
marc-hb, line 85: Avoid `echo`, use `printf`
```shell
printf '\tDefault Value: '
if [ "${OPT_PARM_lst[$i]}" -eq 1 ]; then
   printf '%s\n' "${OPT_VALUE_lst[$i]}"
...
   printf 'On\n'
```

case-lib/pipeline.sh
```diff
@@ -4,19 +4,15 @@ func_pipeline_export()
 {
     # no parameter input the function
     if [ $# -lt 1 ]; then
-        dlogi "Topology file name is not specified, unable to run command: ${BASH_SOURCE[-1]}" && exit 1
+        dlogi "Topology file name is not specified, unable to run command: $SCRIPT_NAME"
+        exit 1
```
marc-hb, line 8: `die()`

case-lib/pipeline.sh
```diff
@@ -4,19 +4,15 @@ func_pipeline_export()
...
-    tplg_path=$(func_lib_get_tplg_path "$tplg")
-    [[ "$?" != "0" ]] && dloge "No available topology for pipeline export" && exit 1
-    dlogi "${BASH_SOURCE[-1]} will use topology $tplg_path to run the test case"
-
-    # acquire filter option
-    local filter="$2"
+    tplg_path=$(func_lib_get_tplg_path "$1") || {
+        dloge "No available topology for pipeline export"
+        exit 1
+    }
```
marc-hb, line 14: `die()`

case-lib/pipeline.sh
```diff
@@ -28,18 +24,23 @@ func_pipeline_export()
...
-    local opt="-f \"$filter\" -b \"$ignore\""
-    local cmd="sof-tplgreader.py $tplg_path $opt -s $sofcard -e"
+    local opt=""
+    # acquire filter option
+    [[ "$2" ]] && opt="-f '$2' "
+    [[ "$ignore" ]] && opt="-b '$ignore' "
+    [[ "$SOFCARD" ]] && opt="-s $SOFCARD"
+
+    local -a pipeline_lst
+    local cmd="sof-tplgreader.py $tplg_path $opt -e" line=""
```
marc-hb, line 34: This looks like real change, really not just fixing a warning. Explain this in a different commit or even better in a different PR.

case-lib/pipeline.sh
```diff
@@ -48,5 +49,6 @@ func_pipeline_parse_value()
     local idx=$1
     local key=$2
     [[ $idx -ge $PIPELINE_COUNT ]] && echo "" && return
-    eval echo "\${PIPELINE_$idx['$key']}"
+    # shellcheck disable=SC1087
+    eval echo "\${PIPELINE_$idx['$key']}" # dynmaic echo the target value of the PIPELINE
```
marc-hb, line 53: Using an intermediate variable (with a good name) should get rid of SC1087 while having simpler quoting.

marc-hb:
About the [WIP] : if you submit the PR as a "github draft" then no reviewer will be automatically added.
https://github.blog/2019-02-14-introducing-draft-pull-requests/

paulstelian97:
> It's too late now, you can't go back to Draft. Next time.

Not too late apparently. It can always be converted back and review requests re-sent.

Bin-QA:
@marc-hb  
1. `die()` function to fix I want to add into other PR,  
it also refer the problem we have different `exit code`  
so here `die()` just focus on `dloge + exit 1`? 
2. `echo` replace with `print` just for the `opt.sh` or other file

Already verify on CI system

marc-hb:
1. I keep losing track of where `die()` is available versus not. If it's not available here yet then forget it for now.

2. For very simple strings with no control characters `echo` is IMHO not the best but acceptable. For anything more advanced like for instance control characters please use `printf`.

Bin-QA:
> 2. For very simple strings with no control characters `echo` is IMHO not the best but acceptable. For anything more advanced like for instance control characters please use `printf`.

@marc-hb I already try with `printf`
but I find, if I want use this style string `"description need two\nlines"` will not output
like my design, so it need to parse the string, how to deal those string for output

paulstelian97:
> but I find, if I want use this style string `"description need two\nlines"` will not output
> like my design, so it need to parse the string, how to deal those string for output

Use single quotes on that, otherwise the shell will eat the `\`. Or do double escaping.

marc-hb:
Exactly: always single quote the format string:
```shell
printf 'format: \t%s\nstring: \t%d' "$arg1" "$arg2"
```

Bin-QA:
@marc-hb Can you help me to review this patch, or you think what I need to update

marc-hb:
> Next time please submit several smaller PRs, one at a time thanks!

It's not only impossible to review changes that large, it's also impossible to test them in a reasonable time. See for instance regression #266

PR #229, opened 2020-05-29
case-lib: hijack.sh use trap to hijack exit command

Bin-QA:
use 'trap' command hijack exit command instead of function
'exit' which perpare for 'set -e'

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -2,8 +2,9 @@
 
 SUDO_CMD=$(command -v sudo)
 
+trap 'func_exit_handler' EXIT
 # Overwrite other functions' exit to perform environment cleanup
-function exit()
+function func_exit_handler()
 {
     local exit_status=${1:-0}
 


aiChaoSONG:
@Bin-QA Once we use this version of exit, we need to clean up old exit calls in test case, right?

Bin-QA:
@aiChaoSONG current without more old exit

aiChaoSONG:
@Bin-QA , Yes, you are right. when use trap, the func_hijack_exit will be automatically called. Now got the idea

case-lib/hijack.sh
```diff
@@ -77,6 +78,7 @@ function exit()
         ;;
     esac
 
+    trap '' EXIT
     builtin exit $exit_status
```
aiChaoSONG, line 82: We trapped exit, I think the `builtin exit $exit_status` may cause recursive call of exit.

Bin-QA: here trap is store the `exit`

case-lib/hijack.sh
```diff
@@ -77,6 +78,7 @@ function exit()
         ;;
     esac
 
+    trap '' EXIT
```
marc-hb, line 81: I don't think you need this.

case-lib/hijack.sh
```diff
@@ -2,8 +2,9 @@
 
 SUDO_CMD=$(command -v sudo)
 
+trap 'func_hijack_exit' EXIT
 # Overwrite other functions' exit to perform environment cleanup
-function exit()
+function func_hijack_exit()
```
marc-hb, line 7: You can remove "hijack" from the name, this is not hijacking anything any more. Suggestions: `cleanup_on_exit()` or `exit_handler()`.

case-lib/hijack.sh
marc-hb: I missed all the processing done on the exit code, sorry. I don't think you don't have access to the exit code in `trap EXIT`, so I don't think this can work. How did you test it?

Bin-QA: yes, you are right, I already test it without this line, it can be worked without recursive call the trap function

case-lib/hijack.sh
Bin-QA: in my test, it is not the recursive call, so this line removed

case-lib/hijack.sh
marc-hb: @Bin-QA I don't understand your last comment, I don't understand how this can work and I would like to know with what failure and  non-zero exit you tested this change because I really don't understand how this can report failures. Try this:

```shell
trap 'func_exit_handler' EXIT

func_exit_handler()
{
    printf '$1 = |%s|\n' "$1"
    local exit_status=${1:-0}
    printf '(wrong) exit status  = |%s|\n' "${exit_status}"
}
exit 5
```
This prints:
```
$1 = ||
(wrong) exit status  = |0|
```
Cc: @xiulipan

Bin-QA: Sorry for miss your point

Bin-QA: @marc-hb let's track it at gitlab

marc-hb:
Fixed in PR #249 / commit fd5a73ef9bbab6f0

So this #229 was never tested, was it? Testing test code requires making it (manually) FAIL. Test code that has never been red is not tested.

Bin-QA:
> Fixed in PR #249 / commit [fd5a73e](https://github.com/thesofproject/sof-test/commit/fd5a73ef9bbab6f0a3feaabc0d164c3dc06e6062)
> 
> So this #229 was never tested, was it? Testing test code requires making it (manually) FAIL. Test code that has never been red is not tested.

No, it tested, but it is not display the hide issue from CI system, just confirm the change without effect current logic

marc-hb:
I still don't understand sorry. Which failure and non-zero exit_status was #229  tested with?

Bin-QA:
> builtin exit

In case test verify and simple step, it is not effect in current script,   
so I push this PR,  
but the case result confuse me for the `exit_code`  and I miss understand your point  
it is the reason why have PR #249

PR #251, opened 2020-06-11
tools: sof-kernel-dump: use journalctl command to query kernel log

Bin-QA:
use journalctl command instead of this step:
1. get kernel bootup line number by sof-get-kernel-line.sh
2. user step 1 information to dump the kernel log from /var/log/kern.log file

now direct use journalctl to dump the kernel log after system bootup

### DIFF
--- a/tools/sof-kernel-dump.sh
+++ b/tools/sof-kernel-dump.sh
@@ -1,35 +1,8 @@
 #!/bin/bash
 
 # dump the kernel information from target
-last_order=${1:-0}
-# check whether parameter 1 has the Non-Number
-[[ "${last_order//[0-9]/}" ]] && echo "$0 parameter 1 just support +N number, N start from 0" && builtin exit 1
-last_order=$[ $last_order + 1 ]
-log_file=${2:-"/var/log/kern.log"}
-# check whether target file exists
-[[ ! -f "$log_file" ]] && echo "$0 parameter 2 $log_file does not exist" && builtin exit 1
-# load helper script
-which sof-get-kernel-line.sh 2>&1 1>/dev/null
-if [ $? -eq 0 ]; then
-    kernel_line_script="sof-get-kernel-line.sh"
-else
-    cd $(dirname $0)
-    kernel_line_script=$PWD/sof-get-kernel-line.sh
-    cd $OLDPWD
-fi
-declare -a line_lst
-# here line lst to keep 2 value, [0] is target line, [1] is end line
-# if target is last one, so will cut content form [0] -> file end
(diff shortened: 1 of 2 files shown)


tools/sof-kernel-dump.sh
```diff
@@ -1,35 +1,9 @@
...
-# check target is last one
-if [ ! "${line_lst[1]}" ];then
-    # tr '\0' ' ' to fix warning message: "command substitution: ignored null byte in input"
-    tail -n +${line_lst[0]} $log_file|tr '\0' ' '|cut -f5- -d ' '
-else
-    line_count=$[ ${line_lst[1]} - ${line_lst[0]} ]
-    # tr '\0' ' ' to fix warning message: "command substitution: ignored null byte in input"
-    tail -n +${line_lst[0]} $log_file|tr '\0' ' '|head -n $line_count|cut -f5- -d ' '
-fi
+[[ "${boot_number//[0-9]/}" ]] && echo "$0 parameter 1 just support +N number, N start from 0" && builtin exit 1
```
marc-hb, line 6: What does `builtin` do?

marc-hb: You can remove this sanity check, `journalctl` has a good error message already.

tools/sof-kernel-dump.sh
```diff
@@ -1,35 +1,8 @@
...
-    tail -n +${line_lst[0]} $log_file|tr '\0' ' '|cut -f5- -d ' '
-else
-    line_count=$[ ${line_lst[1]} - ${line_lst[0]} ]
-    # tr '\0' ' ' to fix warning message: "command substitution: ignored null byte in input"
-    tail -n +${line_lst[0]} $log_file|tr '\0' ' '|head -n $line_count|cut -f5- -d ' '
-fi
+boot_number=${1:-0}
+
+# format time stamp output for journalctl command
+LC_TIME='en_US.UTF-8'
```
marc-hb, line 7: This should really be the default locale on all test systems. `update-locale` seems to be the CLI tool for the job (not tested)

Bin-QA: so force update for DUT?

Current with the check and diff older style, so add this for format output

Bin-QA:
> LGTM except I still didn't check the deleted code, sorry.
> 
> In the future maybe this entire script could become just a small function in `lib.sh`? (Not now)

yes, it become more easy to understand with older one

xiulipan:
@marc-hb I will close this one and please help to take over this PR later as @Bin-QA no longer work on this.

PR #254, opened 2020-06-12
tools/shellcheck-gitrange.bash to add '-x' option

Bin-QA:
Add '-x' option for shellcheck to fix SC1090 error


marc-hb:
https://travis-ci.org/github/thesofproject/sof-test/builds/697773603 passed for throw-away test PR #254

Issue #173, opened 2020-04-08
[new case] check dynamic debug logs


xiulipan:
With update from https://github.com/thesofproject/linux/pull/1980, we will have dynamic debug feature.
Need a new test case for it.

Example for /etc/modprobe.d/sof-dyndgb.conf is here
https://github.com/thesofproject/linux/pull/1980/commits/bcf28058ef97264a774ec8dc4a3a48bc0a7193da

We need to compare there is indeed difference w/o /etc/modprobe.d/sof-dyndgb.conf

fredoh9:
@xiulipan This is matter of updating sof-dyndbg.conf for all test devices. What test case is required for this. 

Updating /etc/modprobe.d/sof-dyndbg.con will be done when deploying build.

In addition, probably we need to udpate env-check.sh to warn if sof-dyndbg.conf is not available in the system.

plbossart:
@fredoh9 we should check which logs CI depends on, the file I provided initially was more for developer debug and may contains too many traces.

fredoh9:
@plbossart Sure, i will discuss this.
In my understanding CI didn't depends on any debug log, but only error logs. This is more likely to provide more verbose logs for developers. Having said that, your initial version is good to go.

fredoh9:
sof-dyndbg.conf will be downloaded to /etc/modprobe.d/ in every deploy of daily/pr build.
I don't think any new test case is required for this. closing this issue.

PR #260, opened 2020-06-24
env-check: add sof-dyndbg.conf file check

fredoh9:
Check sof-dyndbg.conf available in the system, give optional warning and
short usage of dyndbg for SOF modules.

---------
Output file:
```
Optional: Enable dynamic debug logs on  /etc/modprobe.d/sof-dyndbg.conf  file
	For example,
	options snd_sof dyndbg=+p
	options snd_sof_pci dyndbg=+p
```

deploy.sh change will be merged soon. This can fix
https://github.com/thesofproject/sof-test/issues/173

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# enable dynamic debug logs for SOF modules
+DYNDBG="/etc/modprobe.d/sof-dyndbg.conf"
+
 # check for the system package
 out_str=""
 func_check_pkg(){
@@ -24,14 +27,25 @@ func_check_python_pkg(){
     fi
 }
 
+func_check_file(){
+    if [ -e "$1" ]; then
+        return
+    fi
+    out_str=$out_str"\nOptional: Enable dynamic debug logs on \e[31m $1 \e[0m file\n\tFor example,\n\toptions snd_sof dyndbg=+p\n\toptions snd_sof_pci dyndbg=+p"
+    check_res=1
+}
+
 check_res=0
 echo -ne "Check for the package:\t\t"
 func_check_pkg expect
 func_check_pkg aplay
 func_check_pkg python3
 func_check_python_pkg graphviz
+func_check_file "$DYNDBG"
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    # We know exactly what $out_str is, it's not some random and unsanitized input
+    # shellcheck disable=SC2059
+    printf '\e[31mWarning\e[0m\nSolution:'"$out_str"'\n'
 
 # check for the tools folder
 old_path=$PWD


fredoh9:
Debug log verbosity was discussed in https://github.com/thesofproject/linux/issues/2204 as well. Same sof-dyndbg.conf to all test devices are very much required.

env-check.sh
```diff
@@ -24,12 +27,22 @@ func_check_python_pkg(){
     fi
 }
 
+func_check_file(){
+    if [ -e $1 ]; then
+        return
+    else
+        out_str=$out_str"\nOptional: Enable dynamic debug logs on \e[31m $1 \e[0m file\n\tFor example,\n\toptions snd_sof dyndbg=+p\n\toptions snd_sof_pci dyndbg=+p"
```
plbossart, line 34: it's not optional for CI? Do we have a mechanism to be strict with env-check.sh and make it really picky when running from a CI environment (maybe extra parameter?)

plbossart:
> Debug log verbosity was discussed in [thesofproject/linux#2204](https://github.com/thesofproject/linux/issues/2204) as well. Same sof-dyndbg.conf to all test devices are very much required.

right, but the contents of sof-dyndbg.conf should be discussed.

env-check.sh
fredoh9: yes, this is not optional for CI. We will mandate same sof-dyndbg.conf to all CI devices. When a build is deployed to CI device, it will be downloaded.
Separate discussion is started optimal verbosity for SOF CI.

env-check.sh
```diff
@@ -24,12 +27,22 @@ func_check_python_pkg(){
     fi
 }
 
+func_check_file(){
+    if [ -e $1 ]; then
+        return
+    else
```
marc-hb, line 33: No need for the `else` and extra indentation

env-check.sh
```diff
@@ -24,12 +27,22 @@ func_check_python_pkg(){
     fi
 }
 
+func_check_file(){
+    if [ -e $1 ]; then
```
marc-hb, line 31: quotes, run `shellcheck`

env-check.sh
```diff
@@ -24,12 +27,22 @@ func_check_python_pkg(){
...
+    fi
+}
+
 check_res=0
 echo -ne "Check for the package:\t\t"
 func_check_pkg expect
 func_check_pkg aplay
 func_check_pkg python3
 func_check_python_pkg graphviz
+func_check_file $DYNDBG
```
marc-hb, line 45: quotes, run `shellcheck`

env-check.sh
```diff
@@ -16,22 +19,31 @@ func_check_python_pkg(){
...
 check_res=0
 echo -ne "Check for the package:\t\t"
 func_check_pkg expect
 func_check_pkg aplay
 func_check_pkg python3
 func_check_python_pkg graphviz
+func_check_file "$DYNDBG"
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    echo -e "\e[31mWarning\e[0m\nSolution:""$out_str"
```
marc-hb, line 46: Thanks for fixing these quoting issues but:
- in a different and earlier commit
- use `printf '\e[31mWarning\e[0m\nSolution: %s\n' "$out_str"`
https://www.google.com/search?q=echo+printf

fredoh9: Thank you for reviewing this. If I just replace printf with echo -e, special characters are just came out. And line change also messed up.
```
Solution:\nPlease install \e[31m expect \e[0m package\nOptional: Enable dynamic debug logs on \e[31m /etc/modprobe.d/sof-dyndbg.conf \e[0m file\n\tFor example,\n\toptions snd_sof dyndbg=+p\n\toptions snd_sof_pci dyndbg=+p
Check for tools folder:         Warning
Solution:\n
\tMissing execution permission of script/binary in tools folder\n
\tWarning: you need to make sure the current user has execution permssion\n
\tPlease use the following command to give execution permission:\n
\t\e[31mcd .\n
\tchmod a+x tools/*\e[0m
Checking for case folder:               pass
Checking the permission:                Warning
Solution:\n
\tMissing permission to access sound card\n
\t\tPlease use the following command to add current user to the audio group:\n
\t\e[31msed -i '/^audio:/s:$:,fredoh:g' /etc/group\e[0m
Checking the config setup:              pass
```

Came up with "%b" trick. This is the output from my build machine.
```
$ ./env-check.sh
Check for the package:          Warning
Solution:
Please install  expect  package
Optional: Enable dynamic debug logs on  /etc/modprobe.d/sof-dyndbg.conf  file
        For example,
        options snd_sof dyndbg=+p
        options snd_sof_pci dyndbg=+p
Check for tools folder:         Warning
Solution:
        Missing execution permission of script/binary in tools folder
        Warning: you need to make sure the current user has execution permssion
        Please use the following command to give execution permission:
        cd .
        chmod a+x tools/*
Checking for case folder:               pass
Checking the permission:                Warning
Solution:
        Missing permission to access sound card
                Please use the following command to add current user to the audio group:
        sed -i '/^audio:/s:$:,fredoh:g' /etc/group
Checking the config setup:              pass
```

marc-hb: Sorry I missed there are \e escape codes in the `$out_str` argument too.

`%b` works, otherwise this should also work and is easier to read?
```shell
printf '\e[31mWarning\e[0m\nSolution:'"$out_str"'\n' # no argument, everything in the format string
# OK as long as we control what goes into `$out_str`
```

marc-hb: > Thanks for fixing these quoting issues but: in a different and earlier commit

Please?

env-check.sh
marc-hb: I mean:
```shell
if [ -e "$1" ]; then
   return 0
fi
out_str=...
check_res=1

Comment (plbossart) [reply env-check.sh]: agree with @marc-hb, it's the same style in kernel-land. When you do a return the following else is not useful.

Comment (marc-hb) [reply env-check.sh]: > When you do a return the following else is not useful.

https://www.google.com/search?q=chad+versace+early+return+perfect+ice+cream

https://en.wikipedia.org/wiki/Cyclomatic_complexity

File: env-check.sh [hunk 6]
```diff
@@ -24,14 +27,25 @@ func_check_python_pkg(){
...
 func_check_pkg expect
 func_check_pkg aplay
 func_check_pkg python3
 func_check_python_pkg graphviz
+func_check_file "$DYNDBG"
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    # We know exactly what $out_str is, it's not some random and unsanitized input
+    # shellcheck disable=SC2059
+    printf '\e[31mWarning\e[0m\nSolution:'"$out_str"'\n'
```
marc-hb, line 48: This doesn't work, it prints the warning every time because the backslash applies only to the comment line. This  totally confuses shellcheck BTW, did you run shellcheck?

```shell
false || \
# only the comment is "backslashed"
   printf 'RUNS ANYWAY'
```

marc-hb:
re-submitted without the echo/printf change at #309

PR #262, opened 2020-06-28
tools: sof-kernel-log-check: use journalctl command to query kernel log

Bin-QA:
use journalctl command instead of query the begin line from /var/log/kern.log file

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -41,7 +41,9 @@ fi
 func_lib_setup_kernel_last_line()
 {
     # shellcheck disable=SC2034 # external script will use it
-    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_LINE=$(journalctl --dmesg --no-pager -n 1 -o short-iso-precise|awk '/kernel/ {print $1;}')
+    KERNEL_LAST_LINE=${KERNEL_LAST_LINE:0:-5}
+    KERNEL_LAST_LINE=${KERNEL_LAST_LINE/T/ }
 }
 
 SOF_LOG_COLLECT=0
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -1,38 +1,39 @@
 #!/bin/bash
 
-begin_line=${1:-1}
+begin_time="${@:-0}"
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
 ignore_str="error: debugfs write failed to idle -16|error: status|iteration [01]"
 project_key="sof-audio"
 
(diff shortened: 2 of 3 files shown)


case-lib/lib.sh
```diff
@@ -41,7 +41,9 @@ fi
 func_lib_setup_kernel_last_line()
 {
     # shellcheck disable=SC2034 # external script will use it
-    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_LINE=$(journalctl --dmesg --no-pager -n 1 -o short-iso-precise|awk '/kernel/ {print $1;}')
```
marc-hb, line 44: On my system this prints the boot time (uptime -s) which doesn't seem to match what was before.

Should this simply be `KERNEL_LAST_LINE=$(date)`?

I like that you're trying to save a lot of changed lines by keeping the KERNEL_LAST_LINE name the same, however there should be a big fat warning comment saying "This is not a line anymore"

tools/sof-kernel-log-check.sh
```diff
@@ -1,38 +1,39 @@
...
+        echo "Error parameter for date: $begin_time"
+        echo "Support date format: date +'%F %T'"
+        exit 0
+    }
+    cmd="journalctl --dmesg --no-pager --no-hostname -o short-precise --since='$begin_time'"
+fi
 
 if [ "$ignore_str" ]; then
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
+    err=$("$cmd" |grep 'Call Trace' -A5 -B3)$("$cmd" | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
```
marc-hb, line 26: Please submit these cleanups first and separately so they don't noise the functional changes later.

case-lib/lib.sh
```diff
@@ -41,7 +41,9 @@ fi
...
+    KERNEL_LAST_LINE=${KERNEL_LAST_LINE:0:-5}
+    KERNEL_LAST_LINE=${KERNEL_LAST_LINE/T/ }
```
marc-hb, line 46: There should be a comment explaining what this removes and why. It's surprising to drop such a critical piece of information.

case-lib/lib.sh
plbossart: I rename the whole thing as KERNEL_LAST_TIMESTAMP in PR #354

xiulipan:
I will close this one as @Bin-QA no longer work on this.
@aiChaoSONG @marc-hb do we still need some similar issues? I will leave all journalctl related issue to you @marc-hb

marc-hb:
@plbossart is undeterred in #354

plbossart:
This PR would not work anyways because there are still tons of 'sudo dmesg -C' everywhere.

Issue #261, opened 2020-06-25
ignore_str to have root cause or known issue description


fredoh9:
There is array of ignoring error message.
https://github.com/thesofproject/sof-test/blob/master/tools/sof-kernel-log-check.sh#L6

This error message was captured but it is known problem. But there is no description why this is safely ignored.
`ignore_str="error: debugfs write failed to idle -16|error: status|iteration [01]"`

@marc-hb added comments, https://github.com/thesofproject/sof-test/pull/258

Ignoring errors is very dangerous and should be carefully reviewed and extensively documented. At the very least this code should be changed to something like:
```shell
# because blabla, see bug 1234
ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
# because blabla, see bug 5678
ignore_str="$ignore_str"'|error: status'
```

marc-hb:
I tig blamed the history of that `ignorestr` list:

- The `iteration [01]` exclusion was added by d238c48cf which mentions issue #43
- `error: status` was added by commit 50f6ccf895c5 . It mentions https://github.com/thesofproject/linux/pull/1676
- `debugfs write failed...` was added by initial commit b3b7d8df9c84 which is unfortunately not specific.

@aiChaoSONG could you implement and test this line split? For the `debugfs write...` line leave a `# TODO` for now.

@xiulipan, @plbossart  do you know why this this `debugfs write...` error must be ignored?

aiChaoSONG:
@marc-hb I have no idea why `debugfs write..` is added, it is far before I joined CI, but some error string in ignore_str is deprecated. I will try to clean up and split ignore_str after Dragon Boat Holiday.

aiChaoSONG:
@Bin-QA Could you please help to find the reason for each ignore_str? though Marc gives some hint, but from the releated commit's message, It's really hard to know the reason, because the message is not clear.

aiChaoSONG:
WIP PR: https://github.com/thesofproject/sof-test/pull/268

Issue #43, opened 2020-01-15
dmesg check still influenced by DSP init fail


xiulipan:
From https://sof-ci.01.org/linuxpr/PR1706/build2893/devicetest/ICL_RVP_HDA/check-runtime-pm-status.sh/check-runtime-pm-status.sh.txt

The DSP init fail in retry is still caught by our dmesg check by keyword "failed"

xiulipan:
Fixed by #45

PR #268, opened 2020-06-29
[Wait Test Result]tools: split ignore_str and give reason for each ignore item

aiChaoSONG:
Each ignore item should have a reason. This helps us
to trace every ignore item.

This patch split ignore_str into ignore items, and provide
descirption on why each item is ignored.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -3,8 +3,21 @@
 begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
-ignore_str="error: debugfs write failed to idle -16|error: status|iteration [01]"
 project_key="sof-audio"
+
+# There will be debug logs at each failed initializaiton of DSP:
+#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
+#   sof-audio-pci 0000:00:1f.3: error: Error code=0xffffffff: FW status=0xffffffff
+#   sof-audio-pci 0000:00:1f.3: error: iteration 0 of Core En/ROM load failed: -110
+# We will reinit DSP if it is failed to init, and retry 3 times, so the errors in
+# debug logs at the frist and second retry can be ignored.
+# Check https://github.com/thesofproject/linux/pull/1676 for more information.
+# TODO explain for "error: Error ..."
+ignore_str='error: iteration [01]'
+ignore_str="$ignore_str"'|error: status'
+
+# TODO explain
+ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
 
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 # dmesg KB size buffer size


marc-hb:
Nit: you can add
```
Fixes: #261 
```
... in the commit message.

tools/sof-kernel-log-check.sh
```diff
@@ -3,9 +3,22 @@
...
 
+# There will be debug logs at each failed initializaiton of DSP:
+#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
+#   sof-audio-pci 0000:00:1f.3: error: Error code=0xffffffff: FW status=0xffffffff
+#   sof-audio-pci 0000:00:1f.3: error: iteration 0 of Core En/ROM load failed: -110
+# We will reinit DSP when it is failed to init. The errors in debug logs can be
+# ignored due to retry mechanism.
+# Check https://github.com/thesofproject/linux/pull/1676 for more information.
+ignore_str="$ignore_str|iteration [01]"
+ignore_str="$ignore_str|error: status"
```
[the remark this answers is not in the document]
aiChaoSONG: @marc-hb  The `space` between `error:` and `iteration [01]` will always be a space, I think we don't need a RE here, let use `error: iteration [01]`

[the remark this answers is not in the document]
aiChaoSONG: @marc-hb all reviews are addressed.

tools/sof-kernel-log-check.sh
```diff
@@ -3,9 +3,23 @@
...
+# There will be debug logs at each failed initializaiton of DSP:
+#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
+#   sof-audio-pci 0000:00:1f.3: error: Error code=0xffffffff: FW status=0xffffffff
+#   sof-audio-pci 0000:00:1f.3: error: iteration 0 of Core En/ROM load failed: -110
+# We will reinit DSP if it is failed to init, and retry times is 3, so the errors in
+# debug logs at the frist and second retry can be ignored.
+# Check https://github.com/thesofproject/linux/pull/1676 for more information.
+# TODO explain for "error: Error ..."
+ignore_str="$ignore_str"|'error: iteration [01]'
+ignore_str="$ignore_str"|'error: status'
```
Bin-QA, line 18: I think you can add for ignore `error: status`, which is not really error, the error will report by other content

marc-hb: @Bin-QA can you point at the code that prints `error: status`? Thanks!

aiChaoSONG: As @marc-hb pointed here: https://github.com/thesofproject/sof-test/issues/261#issuecomment-649837495:
- `error: status` added by https://github.com/thesofproject/sof-test/commit/50f6ccf895c5a70ba071cd21b4fe15ce3090d74a, mentioned https://github.com/thesofproject/linux/pull/1676
- `error: iteration [01]` added by https://github.com/thesofproject/sof-test/commit/50f6ccf895c5a70ba071cd21b4fe15ce3090d74a mentioned https://github.com/thesofproject/sof-test/issues/43

Take a look at https://github.com/thesofproject/sof-test/issues/43, you should see it also points to https://github.com/thesofproject/linux/pull/1676

tools/sof-kernel-log-check.sh
```diff
@@ -3,9 +3,23 @@
...
 
+# There will be debug logs at each failed initializaiton of DSP:
+#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
+#   sof-audio-pci 0000:00:1f.3: error: Error code=0xffffffff: FW status=0xffffffff
+#   sof-audio-pci 0000:00:1f.3: error: iteration 0 of Core En/ROM load failed: -110
+# We will reinit DSP if it is failed to init, and retry 3 times, so the errors in
+# debug logs at the frist and second retry can be ignored.
+# Check https://github.com/thesofproject/linux/pull/1676 for more information.
+# TODO explain for "error: Error ..."
+ignore_str="$ignore_str"'|error: iteration [01]'
```
marc-hb, line 17: Wait, I just realized the string starts with a vertical bar. That means `ignore_str`probably matches anything.

How did you test this?

aiChaoSONG: @marc-hb the ignored item here is not 100% reproduced, and some may need a lot of stress test. No good way to test but just run script without error. 

Maybe initialize ignore_str with the first ignore item? This will make code style not unified.

aiChaoSONG: Is this OK for you?
```
ignore_str='error: iteration [01]'
ignore_str="$ignore_str"'|error: status'
```

marc-hb: This version looks  better but do test the code change locally with a couple errors that are easy to reproduce: at lest one ignored error and another  not ignored. Then change the ignored message(s) and only the message(s) just before submitting.

This part of the code is very critical because a single, one character bug like this vertical bar can turn the entire results green. So it must be very thoroughly tested.

aiChaoSONG:
Fake some errors in dmesg to test the ignore filter, and ignored strings are filtered out, and other errors strings are still there.

aiChaoSONG:
Fake some errors in and not in the ignore_str to test the filter, errors in ignore_str are filtered out, other errors keep unchanged. tested and merge

aiChaoSONG:
Merge,super thanks @marc-hb

PR #274, opened 2020-07-02
tools: add support for drawing loop in topology graph

aiChaoSONG:
Previously, we use the processed graph link to
draw topology graph. If there is loop in topology
graph, we will be traped into an endless recursion
when traverse the graph.

This patch adds check for loop, and use the raw
widget-contorl-widget denoted link from topology
to draw the graph.


aiChaoSONG:
verified with all topologies currently we have.

aiChaoSONG:
@xiulipan Ping for review

PR #275, opened 2020-07-03
case-lib/lib.sh: remove disable=SC1091 thanks to shellcheck -x

marc-hb:
commit b8d6a278d1f0 added -x to the shellcheck call, use it.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -10,15 +10,15 @@ SCRIPT_NAME="$0"  # get test-case script load name
 SCRIPT_PRAM="$*"  # get test-case parameter
 
 # Source from the relative path of current folder
-# shellcheck disable=SC1091 source=./config.sh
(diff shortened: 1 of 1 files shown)


marc-hb:
> Patch works fine on my side with/without -x option.

Can you elaborate what you mean by "works"? After this commit I need -x, otherwise shellcheck prints warnings. Did you find a way to enable -x by default?

aiChaoSONG:
> > Patch works fine on my side with/without -x option.
> 
> Can you elaborate what you mean by "works"? After this commit I need -x, otherwise shellcheck prints warnings. Did you find a way to enable -x by default?

with -x, no warnings; without -x, there are warnings like "SC1091: Not Following: some file was not specified as input". Yes, we need -x,  or there will be warnings. I don't think there is a way to enable -x by default, so need to add it explicitly.

PR #277, opened 2020-07-03
Errexit fixes

marc-hb:
See commit messages


marc-hb:
Thanks @aiChaoSONG . `sudo` is an extremely well known tool, I don't think we need to duplicate its documentation.

Issue #282, opened 2020-07-09
tplgtool.py duplicate links


plbossart:
This could be a topology or a tool issue, but basically when we use the new baytrail/cherrytrail topologies one can see duplicate arrows after the mixer

![sof-byt-rt5640](https://user-images.githubusercontent.com/1149035/87091880-0d785580-c200-11ea-9704-8a42319fb97b.png)

Topology file included for convenience:
[sof-byt-rt5640.tplg.gz](https://github.com/thesofproject/sof-test/files/4899585/sof-byt-rt5640.tplg.gz)

PR #283, opened 2020-07-10
tools: fix duplicated link in topology graph

aiChaoSONG:
Previously, topology graph is drawn per pipeline,
so there will be duplicated links on component like
mixer.

This patch records components' links, and only links
conponents not previouly linked.

Fixes: #282

### DIFF
--- a/tools/tplgtool.py
+++ b/tools/tplgtool.py
@@ -854,11 +854,15 @@ if __name__ == "__main__":
             tplg_paths.append(f)
         return tplg_paths
 
-    def connect(graph, raw_graph_list, nodes_added):
+    def connect(graph, raw_graph_list, nodes_added, edges):
         for raw_graph in raw_graph_list:
             for link in raw_graph:
                 if link[0] in nodes_added and link[2] in nodes_added:
-                    graph.edge(link[0], link[2])
+                    new_edge = (link[0], link[2])
+                    if new_edge not in edges:
+                        graph.edge(*new_edge)
+                        edges.append(new_edge)
+
 
     # traverse all nodes, and add them to graph
     def init_node(graph, head, nodes_added):
@@ -911,12 +915,15 @@ if __name__ == "__main__":
             raw_graph_list = formatter.get_tplg_raw_graph_list()
 
             graph = Digraph("Topology Graph", format=format)
+            # There is no method in python-graphviz to detect if there is already an edge
+            # between two nodes, we need to record it manually
+            edges = []
             # Here we make every pipeline as a subgraph, this gives us more precise control
             for head in head_list:
                 subgraph = Digraph('Pipeline' + head['name'])
                 nodes_added = []
                 init_node(subgraph, head, nodes_added)
-                connect(subgraph, raw_graph_list, nodes_added)
+                connect(subgraph, raw_graph_list, nodes_added, edges)
                 # add subgraph to the graph
                 graph.subgraph(graph=subgraph)
             # Developers may want to view graph without saving it.


aiChaoSONG:
Tested on all topology currently we have, all passed.

tools/tplgtool.py
```diff
@@ -854,11 +854,15 @@ def get_tplg_paths(cmd_args):
...
+    def connect(graph, raw_graph_list, nodes_added, connnected_link):
         for raw_graph in raw_graph_list:
             for link in raw_graph:
                 if link[0] in nodes_added and link[2] in nodes_added:
-                    graph.edge(link[0], link[2])
+                    link_str = link[0] + '->' + link[2]
+                    if link_str not in connnected_link:
+                        graph.edge(link[0], link[2])
+                        connnected_link.append(link_str)
+
```
marc-hb, line 865: You don't need to use strings, they use more CPU and memory. This should work:
```python
new_edge = (link[0], link[2])
if new_edge not in edges:
    graph.edge(*new_edge)
    edges.append(new_edge)
```
PS: stick to the graphviz terminology, this avoids wondering whether it's something different.

tools/tplgtool.py
```diff
@@ -911,12 +915,15 @@ def dump_graph(parsed_tplgs, cmd_args):
             raw_graph_list = formatter.get_tplg_raw_graph_list()
 
             graph = Digraph("Topology Graph", format=format)
+            # There is no method in python-graphviz to dectect if there is already a link
+            # between two nodes, we need to record it manually
+            connnected_link = []
```
marc-hb, line 920: Very useful comment.

marc-hb:
> Previously, topology graph is drawn per pipeline, so there will be duplicated links on component like mixer.

Wait, isn't the number of pipelines useful information that should be kept? Naive question sorry.

aiChaoSONG:
@Marc  Thanks for the review. updated.
Q:  isn't the number of pipelines useful information that should be kept? 
A:  if a case need this number, here is how to get the number: `sof-tplgreader.py tplg-path | wc -l`, we can also add one line in sof-tplgreader.py or tplgtool.py `print(len(some_list_already_there))` to print this number. but currently, No test case uses this number. 

Q:  Don't we have tests with expected .gv output?
A: Here's how topology graph is rendered:  generate dot file(.gv) -> render png -> delete dot file. the `cleanup` in `graph.render(filename=outfile, directory=dir, cleanup=True)` contorls if we should delete .gv dot file. if you see the PNG, the .gv should good too. no test case need the .gv, so it is deleted.

marc-hb:
So the tests don't check the output, they just make sure the software doesn't crash?

aiChaoSONG:
> So the tests don't check the output, they just make sure the software doesn't crash?

I have used this tool to generate all topology drawing, and check each drawing, duplicated links are moved now.

tools/tplgtool.py
```diff
@@ -911,12 +915,15 @@ def dump_graph(parsed_tplgs, cmd_args):
             raw_graph_list = formatter.get_tplg_raw_graph_list()
 
             graph = Digraph("Topology Graph", format=format)
+            # There is no method in python-graphviz to dectect if there is already an edge
```
plbossart, line 918: typo: detect

marc-hb:
Adding some `test3_expected.gv` files in git would make testing one order of magnitude better for a reasonably small effort. This PR is a great opportunity to add a couple `_expected.gv` files. 1. Add a couple _before_ this PR, including one with double edges. 2 Rebase this PR on top and make sure the new versions of the `_expected.gv` files submitted have all their double edges removed (test driven development)

I think Travis could easily run such tests.

tools/tplgtool.py
aiChaoSONG: Thanks, updated.

PR #290, opened 2020-07-16
Remove sof-audio before scanning dmesg for error and Ignore known issues

fredoh9:
When scan errors in kernel log in sof-kernel-log-check.sh, current code filter "sof-audio" first then check for error. This will limit scanning logs and not scalable. Below $project_key is nothing but "sof-audio".

```
err_str="error|failed|timed out|panic|oops"
project_key="sof-audio"

if [ "$ignore_str" ]; then
    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
else
    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str")
fi
```

We should remove this limitation. If we see unrelated error, we can add it to ignoring error list. This is first step toward this approach. I ran first test on APL, CML Helios, ICL RVP. If no objection is raised, I will scan logs for all supported platforms and update ignore_str.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -22,6 +22,10 @@ ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
 # Possible fix - https://github.com/thesofproject/linux/pull/1984
 ignore_str="$ignore_str"'|Parity error detected'
 
+# CML Helios known issue related with xhci_hcd
+# https://bugzilla.kernel.org/show_bug.cgi?id=202541
+ignore_str="$ignore_str"'|xhci_hcd 0000:00:14.0: WARN Set TR Deq Ptr cmd failed due to incorrect slot or ep state'
+
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 # dmesg KB size buffer size
 #dmesg_config_define=$(awk -F '=' '/CONFIG_LOG_BUF_SHIFT/ {print $2;}' /boot/config-$(uname -r))
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -18,6 +18,10 @@ ignore_str="$ignore_str"'|error: status'
 # TODO explain
 ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
 
+# Realtek codecs thrown an error on startup, checking with Realtek
+# Possible fix - https://github.com/thesofproject/linux/pull/1984
+ignore_str="$ignore_str"'|Parity error detected'
+
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 # dmesg KB size buffer size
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -19,6 +18,12 @@ ignore_str="$ignore_str"'|error: status'
 # TODO explain
 ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
 
+# Realtek codecs thrown an error on startup, checking with Realtek
+ignore_str="$ignore_str"'|Parity error detected'
+
+# CML Helios known issue related with xhci_hcd
+ignore_str="$ignore_str"'|xhci_hcd 0000:00:14.0: WARN Set TR Deq Ptr cmd failed due to incorrect slot or ep state'
```
marc-hb, line 25: Please one context URL (bug, mailing-list discussion, anything...) per ignored error. File a new bug/task if really needed.

tools/sof-kernel-log-check.sh
```diff
@@ -35,9 +40,9 @@ ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
 #echo "run $0 with parameter '$*' for check kernel message error"
 
 if [ "$ignore_str" ]; then
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str"|grep -vE "$ignore_str")
+    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep -E "$err_str"|grep -vE "$ignore_str")
 else
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep $project_key | grep -E "$err_str")
+    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep -E "$err_str")
```
marc-hb, line 45: The `eval` seems unnecessary, can you try to remove it?

fredoh9: Without it, I'm seeing errors.  How about handling this when we switch to journalctl
```
sed: -e expression #1, char 1: unknown command: `''
sed: -e expression #1, char 1: unknown command: `''

```

fredoh9:
> > err_str="error|failed|timed out|panic|oops"
> 
> Later we should drop such "screen scraping" and use something more robust like `journalctl --dmesg --priority err` with no false positive (for instance a debug message with one of these keywords) and no false negative (e.g. `ERR:`). I'm not suggesting to do that right now but probably good to keep in mind while you're doing this PR.

This is good idea, I will create an issue to follow up.

tools/sof-kernel-log-check.sh
marc-hb: OK forget it for now; one change at a time.

Issue #285, opened 2020-07-13
[Bug] Need skip echo reference capture pipline


ClarexZhou:
Need skip echo reference capture pipline.  Since this pipline only works during max98357a playback has data output.

```
2020-06-30 03:53:52 UTC Sub-Test: [COMMAND] sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-tgl-max98357a-rt5682.tplg -f 'type:capture & id:any' -b ' pcm:HDA Digital,Media Playback,DMIC16k' -s 0 -e
2020-06-30 03:53:52 UTC Sub-Test: [INFO] ===== Testing: (Round: 1/1) (PCM: DMIC48kHz [hw:0,99]<capture>) (Loop: 1/1) =====
2020-06-30 03:53:52 UTC Sub-Test: [INFO] no file prefix, use /dev/null as dummy capture output
2020-06-30 03:53:52 UTC Sub-Test: [COMMAND] arecord -Dhw:0,99 -r 48000 -c 4 -f S32_LE -d 100 /dev/null -v -q
2020-06-30 03:55:32 UTC Sub-Test: [INFO] ===== Testing: (Round: 1/1) (PCM: Headset [hw:0,1]<capture>) (Loop: 1/1) =====
2020-06-30 03:55:32 UTC Sub-Test: [INFO] no file prefix, use /dev/null as dummy capture output
2020-06-30 03:55:32 UTC Sub-Test: [COMMAND] arecord -Dhw:0,1 -r 48000 -c 2 -f S16_LE -d 100 /dev/null -v -q
2020-06-30 03:57:13 UTC Sub-Test: [INFO] ===== Testing: (Round: 1/1) (PCM: EchoRef [hw:0,6]<capture>) (Loop: 1/1) =====
2020-06-30 03:57:13 UTC Sub-Test: [INFO] no file prefix, use /dev/null as dummy capture output
2020-06-30 03:57:13 UTC Sub-Test: [COMMAND] arecord -Dhw:0,6 -r 48000 -c 2 -f S16_LE -d 100 /dev/null -v -q
2020-06-30 03:57:13 UTC Sub-Test: [ERROR] arecord on PCM hw:0,6 failed at 1/1.
2020-06-30 03:57:13 UTC Sub-Test: [INFO] Test Result: FAIL!
```

PR #293, opened 2020-07-21
tools: enhance pipeline filtering with block feature

aiChaoSONG:
If run general test on echo reference capture pipeline,
there will be errors due to no feed data. So we need
to block this pipeline in general test while keep
echo reference playback pipeline.

This patch enhances the pipeline filtering with block
feature, and makes it possible to block specified pipeline
by keyword.

Usage:
1. default behavior is to dump all pipelines, for example:
    sof-tplgreader.py sof-tgl-max98357a-rt5682.tplg

2. specify -f, WOV and ECHO REF capture pipeline will be blocked,
and specified filter keyword still works, for example, below line
will filter out all capture pipelines except WOV and ECHO REF capture:
    sof-tplgreader.py -f "type:capture" sof-cml-rt1011-rt5682.tplg

3. specify -b, specified pipelines will be blocked, for example,
below line will block all playback pipelines, WOV and ECHO REF
capture pipeline:
    sof-tplgreader.py -b "type:playback" sof-cml-rt1011-rt5682.tplg

Fixes: #285

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -10,8 +10,8 @@ class clsTPLGReader:
         self._pipeline_lst = []
         self._output_lst = []
         self._field_lst = []
-        self._filter_dict = {"filter":[], "op":[]}
-        self._ignore_lst = []
+        self._filter_dict = {}
+        self._block_lst = []
 
     def __comp_pipeline(self, pipeline):
         return int(pipeline['id'])
@@ -125,6 +125,15 @@ class clsTPLGReader:
                 output_lst.append(elem)
         return output_lst
 
+    @staticmethod
+    def list_diff(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = []
+        for elem in lst1:
+            if elem not in lst2:
+                output_lst.append(elem)
+        return output_lst
+
     def _setlist(self, orig_lst):
         tmp_lst = []
         if orig_lst is None:
@@ -137,17 +146,16 @@ class clsTPLGReader:
         return tmp_lst
 
     def setFilter(self, filter_dict=None):
-        self._filter_dict['filter'] = self._setlist(filter_dict['filter'])[:]
-        self._filter_dict['op'] = self._setlist(filter_dict['op'])
+        self._filter_dict = filter_dict
 
     def setField(self, field_lst):
         self._field_lst = self._setlist(field_lst)[:]
 
-    def setIgnore(self, ignore_lst=None):
-        self._ignore_lst = self._setlist(ignore_lst)[:]
+    def setBlock(self, block_lst=None):
+        self._block_lst = block_lst
 
     def _filterOutput(self, target_lst, filter_dict, bIn):
-        self._output_lst.clear()
+        filtered = []
         for line in target_lst[:]:
             check = False
             for key, value in filter_dict.items():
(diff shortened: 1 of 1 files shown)


tools/sof-tplgreader.py
```diff
@@ -125,6 +125,15 @@ def list_or(lst1, lst2):
...
         return output_lst
 
+    @staticmethod
+    def list_diff(lst1, lst2):
+        assert(lst1 is not None and lst2 is not None)
+        output_lst = []
+        for elem in lst1:
+            if elem not in lst2:
+                output_lst.append(elem)
+        return output_lst
```
marc-hb, line 135: Can input lists have duplicates? Does their order matter? If both answers are "no" then:
```python
return list(set(list1) - set(list2))
```

case-lib/pipeline.sh
```diff
@@ -18,7 +18,7 @@ func_pipeline_export()
         for key in "${!TPLG_IGNORE_LST[@]}"
         do
             dlogi "Pipeline list to ignore is specified, will ignore '$key=${TPLG_IGNORE_LST[$key]}' in test case"
-            ignore=$ignore" $key:${TPLG_IGNORE_LST[$key]}"
+            ignore=$ignore"$key:${TPLG_IGNORE_LST[$key]};"
```
marc-hb, line 21: The commit message states "use semicolon to comply with semantics..." Are these semantics new and if yes then since when? If there are not new then was this code buggy and  not working? Is this a fix? Maybe some ignore example(s) in the commit message would help.

I can't understand the new location of the delimiter either.

Please quote `"$ignore"` too, at least for consistency.

tools/sof-tplgreader.py
aiChaoSONG: @marc-hb Both answers are no, but the two lists used here are lists of dicts, they are not hashable, Python will report error on `set(list1)` . "unhahable type: dict"

case-lib/pipeline.sh
aiChaoSONG: Semantics is new, and since this PR, commit id is 9c2750405873dc8c178c91201168b89919f3a6da. I will explain more.
> The filtering feature now allows us to acquire specific pipelines we want,  for example, pipelines with/without PGA/EQ/ASRC/KPBM/ECHO REF/SMART_AMP component, pipelines of specific IDs and pipelines of playback/capture. 

>  Recently, Echo Reference feature is introduced, this feature requires two pipelines, the capture pipeline will capture the data of the playback pipeline, you can regard this as loop back. with the filtering feature, `-f "echo:any" ` will help to acquire the two pipelines as they both contain ECHO REF compnent, and it will be easy to put this case to CI.

>  Our capture test case use -f "type:capture" to acquire all capture pipelines, and the capture from Echo Reference is also here. Here comes the problem, Echo Reference capture cannot work alone, all of our capture case will fail due to this pipeline.

> This PR is intended to solve the above problem. The block feature allow us to block a single pipeline. `-b type:capture & echo:any` will do. We also need to block WOV pipeline in capture test, that's where the `delimiter ";"` works. `-b type:capture & echo:any; kpbm:any`

aiChaoSONG: @marc-hb As the block feature is enhanced in this PR, the test script do need a enhance to use the new feature. Let's do this  in another PR to make changes consistent. So this PR, only the enhancement on sof-tplgreader.py. this enhancement is required to add echo reference case to CI.

PR #296, opened 2020-07-27
test-case: add case for DSM feature test

aiChaoSONG:
- add a wavetool for DSM/WOV/EchoReference feature test (used to analyze wave)
- add DSM test case.

### DIFF
--- a/test-case/check-smart-amplifier.sh
+++ b/test-case/check-smart-amplifier.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+set -e
+
+##
+## Case Name: check-smart-amplifier
+## Preconditions:
+##    require python3-numpy, python3-scipy and python3-matplotlib
+##    to be installed, or wavetool.py will not work
+## Description:
+##    test smart amplifier with customized wavetool.py,
+##    this tool will do binary comparison of reference wave file
+##    and recorded wave file.
+## Case step:
+##    1. acquire playback & capture pipelines with smart_amp component
+##    2. generate playback wave file (default 997Hz sine wave)
+##    3. play reference wave file through smart_amp playback pipeline
+##       and record it back through smart_amp capture pipeline
+##    4. compare channel 0/1, 2/3 of recorded wave with reference wave
+## Expect result:
+##    1. reference wave and recorded wave are binary same
+##    2. delay in recorded wave less than 5ms
+##
+
+# source from the relative path of current folder
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# What we want here is the "$TPLG" string
+# shellcheck disable=SC2016
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+# $TPLG is assigned outside this script as env variable
+# shellcheck disable=SC2153
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
+
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='playback/capture duration in second'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=6
+
+# We need OPT_OPT_lst to tell what the command option is, and OPT_PARM_lst to tell
+# how many arguments this option required, though they are not used.
+# shellcheck disable=SC2034
+OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
+# shellcheck disable=SC2034
+OPT_PARM_lst['F']=0         OPT_VALUE_lst['F']=0
+
+func_opt_parse_option "$@"
+
+duration=${OPT_VALUE_lst['d']}
+loop_cnt=${OPT_VALUE_lst['l']}
+tplg=${OPT_VALUE_lst['t']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+func_pipeline_export $tplg "smart_amp:any"
+func_lib_setup_kernel_last_line
+
+[ "$PIPELINE_COUNT" == "2" ] || die "Only detect $PIPELINE_COUNT pipeline(s) from topology, but two are needed"
+
+for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
+do
+    type=$(func_pipeline_parse_value "$idx" type)
+    if [ "$type" == "playback" ]; then
+        pb_chan=$(func_pipeline_parse_value "$idx" ch_max)
+        pb_rate=$(func_pipeline_parse_value "$idx" rate)
+        pb_dev=$(func_pipeline_parse_value "$idx" dev)
+    else
+        cp_chan=$(func_pipeline_parse_value "$idx" ch_max)
+        cp_rate=$(func_pipeline_parse_value "$idx" rate)
+        cp_fmt=$(func_pipeline_parse_value "$idx" fmt)
+        cp_fmts=$(func_pipeline_parse_value "$idx" fmts)
+        cp_dev=$(func_pipeline_parse_value "$idx" dev)
+    fi
+done
+
+fmts="$cp_fmt"
+if [ ${OPT_VALUE_lst['F']} = '1' ]; then
+    fmts="$cp_fmts"
+fi
+
+for i in $(seq 1 $loop_cnt)
+do
+    for fmt in $fmts
+    do
+        dlogi "Testing: iteration $i of $loop_cnt with $fmt format"
+        # S24_LE format is not supported
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # generate wave file
+        tmp_dir="/tmp"
+        file="$tmp_dir/tmp_wave_${fmt%_*}.wav"
+        recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
+        wavetool.py -gsine -A0.8 -B"${fmt%_*}" -o"$file"
+        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file &"
+        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file"
+        dlogi "Comparing recorded wave and reference wave"
+        wavetool.py -a"smart_amp" -R"$recorded_file" -r"$file" || die "wavetool.py exit with $?"
+        # clean up generated wave files
+        rm -rf "$file" "$recorded_file"
+    done
+done
+
+sof-kernel-log-check.sh "$KERNEL_LAST_LINE"
--- a/tools/wavetool.py
+++ b/tools/wavetool.py
@@ -1,0 +1,160 @@
+#! /usr/bin/python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+
+"""
+A Tool to Generate and Manipulate Wave Files
+
+Supported features:
+- generate sinusoids
+- give verdict in check-smart-amplifier test case by binay wave comparison
+"""
+
+import os
+import sys
+import argparse
+import numpy as np
+import scipy.signal as signal
+import scipy.io.wavfile as wavefile
+
+# The acceptable threshold of smart amplifier delay, unit: ms.
+# If the dealy is longer, then DSP is overloaded or something
+# is wrong with firmware scheduler.
+SMART_AMP_DELAY_THRESHOLD = 5
+
+def generate_sinusoids(**p):
+    """
+    Generate sine or cosine wave.
+
+    ``y(t) = A * sin(2 * pi * f * t + phi)``
+
+    ``y(t) = A * cos(2 * pi * f * t + phi)``
+
+    Parameters
+    ----------
+    dict of sinusiod parameters:
+    p['type']: Wave type, ``sine`` or ``cosine``
+    p['amp']: Amplitude, range: 0.0 ~ 1.0
+    p['freq']: Frequency, unit: Herz
+    p['phase']: Phase, unit: Radian
+    p['chan']: Channels
+    p['sample_rate']: Sample rate, unit: Herz
+    p['duration']: Duration, unit: Second
+
+    Returns
+    ----------
+    Real n-D Array with shape ``(duration * sample_rate, channel)``
+    """
+    func = np.sin if p['type'] == 'sine' else np.cos
+    time = np.arange(0, p['duration'], 1.0 / p['sample_rate'], dtype=np.float32)
+    data = p['amp'] * func(2 * np.pi * p['freq'] * time + p['phase'])
+    return np.reshape(np.repeat(data, p['chan']),[len(data), p['chan']])
+
+def generate_wav(**cmd):
+    if cmd['generate'] in ['sine', 'cosine']:
+        wave_param = {
+            'type': cmd['generate'], 'amp': cmd['amp'][0], 'freq': cmd['freq'][0],
+            'phase': cmd['phase'][0], 'chan': cmd['channel'], 'sample_rate': cmd['sample_rate'],
+            'duration': cmd['duration'][0]
+        }
+        wave_data = generate_sinusoids(**wave_param)
+    else:
+        raise Exception('invalid generate function, will generate nothing')
+    return wave_data
+
+def save_wave(wave_data, **cmd):
+    wave_path = cmd['output']
+    if wave_path == '.' or not wave_path.endswith('wav'):
+        wave_path = wave_path + '/tmp.wav'
+    sample_bits = cmd['bits']
+    np_types = {'S8': np.int8, 'S16': np.int16, 'S32': np.int32}
+    if sample_bits in np_types.keys():
+        # range of wave_data is floating point [-1.0, 1.0]. When saving to any integer format use its full range.
+        wave_data = (np.iinfo(np_types[sample_bits]).max * wave_data).astype(np_types[sample_bits])
+    wavefile.write(wave_path, cmd['sample_rate'], wave_data)
+
+def do_wave_analysis(**cmd):
+    fs_wav, wave = wavefile.read(cmd['recorded_wave'])
+    fs_ref, ref_wave = wavefile.read(cmd['reference_wave'])
+    if fs_wav != fs_ref:
+        print('Can not compare wave with different sample rate')
+        sys.exit(1)
+    if cmd['analyze'] == 'smart_amp':
+        analyze_wav_smart_amp(wave, ref_wave, fs_wav, cmd['zero_threshold'])
+
+# remove digital zeros in two sides
+def trim_wave(wave, zero_threshold):
+    # once waves go through DAC/ADC, zero will become small value close to zero,
+    # here we set the digital zero threshold to 100, and cut samples below 100
+    # of two sides, this has no negative impact to processing.
+    ZERO_THRESHOLD_LEVEL = np.power(10, zero_threshold / 20.) * np.iinfo(wave.dtype).max
+    wave_mono = wave[:,0]
+    left_idx = 0
+    right_idx = wave_mono.shape[0] - 1
+    while True:
+        if abs(wave_mono[left_idx]) > ZERO_THRESHOLD_LEVEL:
+            break
+        left_idx = left_idx + 1
+    while True:
+        if abs(wave_mono[right_idx]) > ZERO_THRESHOLD_LEVEL:
+            break
+        right_idx = right_idx - 1
+    return wave[left_idx:right_idx,:], left_idx
+
+# float point binary comparison is not supported, and will not be supported
+# check recorded wave through smart amplifier component
+def analyze_wav_smart_amp(wave, ref_wave, fs, zero_threshold):
+    trimed_ref_wave, _ = trim_wave(ref_wave, zero_threshold)
+    # compare the first two channel
+    trimed_wave_ch_0_1, delay1 = trim_wave(wave[:,0:2], zero_threshold)
+    trimed_ref_wave = trimed_ref_wave[0:trimed_wave_ch_0_1.shape[0],:]
+    compare_result0_1 = np.array_equal(trimed_ref_wave, trimed_wave_ch_0_1)
+    # compare the second two channel
+    trimed_wave_ch_2_3, delay2 = trim_wave(wave[:,2:4], zero_threshold)
+    trimed_ref_wave = trimed_ref_wave[0:trimed_wave_ch_2_3.shape[0],:]
+    compare_result2_3 = np.array_equal(trimed_ref_wave, trimed_wave_ch_2_3)
+
+    smart_amp_delay = ((delay2 - delay1) / fs * 1000)
+    print('Delay of smart amplifier is %0.3fms' % smart_amp_delay)
+    if compare_result0_1 and compare_result2_3 and smart_amp_delay < SMART_AMP_DELAY_THRESHOLD:
+        print('Recorded wave is binary same as reference wave')
+        print('Wave comparison result: PASSED')
+    else:
+        print('Recorded wave is not binary same as reference wave')
+        print('Wave comparison result: FAILED')
+        sys.exit(1001)
+
+def parse_cmdline():
+    parser = argparse.ArgumentParser(add_help=True, formatter_class=argparse.RawTextHelpFormatter,
+        description='A Tool to Generate and Manipulate Wave Files.')
+    parser.add_argument('-v', '--version', action='version', version='%(prog)s 1.0')
+    # wave parameters
+    parser.add_argument('-g', '--generate', type=str, choices=['sine','cosine'], help='generate specified types of wave')
+    parser.add_argument('-A', '--amp', type=float, nargs='+', default=[1.0], help='amplitude of generated wave')
+    parser.add_argument('-F', '--freq', type=float, nargs='+', default=[997.0], help='frequency of generated wave')
+    parser.add_argument('-P', '--phase', type=float, nargs='+', default=[0.0], help='phase of generated wave')
+    parser.add_argument('-D', '--duration', type=float, nargs='+', default=[10.], help='duration of generated wave')
+    parser.add_argument('-S', '--sample_rate', type=int, default=48000, help='sample rate of generated wave')
+    parser.add_argument('-C', '--channel', type=int, default=2, help='channels of generated wave')
+    parser.add_argument('-B', '--bits', type=str, choices=['S8', 'S16', 'S32', 'F32'], default='S16',
+    help='sample bits of generated wave')
+    parser.add_argument('-o', '--output', type=str, help='path to store generated files', default='.')
+    # wave comparison arguments
+    parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp'], help='analyze reocrded wave to give case verdict')
+    parser.add_argument('-R', '--recorded_wave', type=str, help='path of recorded wave')
+    parser.add_argument('-r', '--reference_wave', type=str, help='path of reference wave')
+    parser.add_argument('-Z', '--zero_threshold', type=float, default=-50.3, help='zero threshold in dBFS')
+    return vars(parser.parse_args())
+
+def main():
+    cmd_args = parse_cmdline()
+
+    if cmd_args['generate'] is not None:
+        wave_data = generate_wav(**cmd_args)
+        save_wave(wave_data, **cmd_args)
+
+    if cmd_args['analyze'] is not None:
+        do_wave_analysis(**cmd_args)
+
+if __name__ == '__main__':
+    main()


test-case/check-dsm.sh
```diff
@@ -0,0 +1,113 @@
...
+## Case step:
+##    1. acquire DSM playback & capture pipelines
+##    2. generate playback wave file (1000Hz sine wave)
+##    3. play wave file and recorded it back through smart amp
+##    4. compare channel 0/1, 2/3 of recorded wave with reference wave
+## Expect result:
+##    1. reference wave and recorded wave are binary same
+##    2. delay in recorded wave less than 5ms
+##
+
```
marc-hb, line 21: As this a new script add `set -e` and test it with `set -e`.

test-case/check-dsm.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+
```
marc-hb, line 2: SPDX license

tools/wavetool.py
```diff
@@ -0,0 +1,279 @@
...
+    def generateRandom(param_dict):
+        wave_size = int(float(param_dict['duration']) * float(param_dict['sample_rate']))
+        data = np.random.normal(float(param_dict['freq']), float(param_dict['phase']), wave_size)
+        maxnum = abs(np.max(data))
+        if abs(np.max(data)) < abs(np.min(data)):
+            maxnum = abs(np.min(data))
+        data = float(param_dict['amp']) / maxnum * data
+        return np.reshape(np.repeat(data, int(param_dict['channel'])),[len(data), int(param_dict['channel'])])
+
+if __name__ == "__main__":
```
marc-hb, line 36: Please create a (not too big) `main` function: https://docs.python.org/3/library/__main__.html

tools/wavetool.py
```diff
@@ -0,0 +1,279 @@
...
+    def parse_cmdline():
```
marc-hb, line 37: There's no reason not to have these functions at the top-level. In fact they must be at the top-level for interactive debug or importing https://docs.python.org/3/library/__main__.html

tools/wavetool.py
```diff
@@ -0,0 +1,279 @@
...
+import sys
+import argparse
+import numpy as np
+import scipy.signal as signal
+import scipy.io.wavfile as wavefile
+import scipy.fftpack as fftpack
+
+eps = 0.0000000001 # used to avoid zero division error
+
+class WavGenerator:
```
marc-hb, line 13: I don't see the point of a class with only @staticmethods. No need to indent these functions in a class, just move them to the top-level.

tools/wavetool.py
```diff
@@ -0,0 +1,279 @@
...
+        time = np.arange(0, float(param_dict['duration']), 1.0/float(param_dict['sample_rate']), dtype=np.float32)
+        data = float(param_dict['amp']) * np.sin(2 * np.pi * float(param_dict['freq']) * time + float(param_dict['phase']))
+        return np.reshape(np.repeat(data, int(param_dict['channel'])),[len(data), int(param_dict['channel'])])
+
+    @staticmethod
+    def generateCosine(param_dict):
+        time = np.arange(0, float(param_dict['duration']), 1.0/float(param_dict['sample_rate']), dtype=np.float32)
+        data = float(param_dict['amp']) * np.cos(2 * np.pi * float(param_dict['freq']) * time + float(param_dict['phase']))
+        return np.reshape(np.repeat(data, int(param_dict['channel'])),[len(data), int(param_dict['channel'])])
+
```
marc-hb, line 25: These functions are nearly identical to each other which makes it very difficult to see the differences between them. Make "sin" and "cos" an  "f" parameter and you can have only one function, no duplication, less code and less maintenance.

aiChaoSONG:
> Before we spend too much time reviewing it, what does this new Python tool do that ALSA BAT doesn't already do?
> https://github.com/alsa-project/alsa-utils/blob/master/bat/alsabat.1
> https://www.alsa-project.org/wiki/BAT

This tool is intended for DSM/WOV test. 
- For DSM test, we need to binary compare recorded wave and reference wave, before comparison, we need to trim zeros of recorded wave. 
- For WOV test, we use this tool to generate special designed wave, and analyze recorded wave to ensure data integrity in history buffer.
- For future use, I decide to add more complex wave analysis for EQ test.

marc-hb:
I'm still a bit surprised that tool(s) to perform these sorts of analysis don't already exist but I'm new to this project.

It's nice to split the new tool and the test case in two different commits but I think this doesn't go far enough. Creating a generic tool like this from scratch is a pretty big, long term commitment so it should be more visible and deserves its own PR, not semi-hidden behind the addition of a new test case. Maybe it should be located in `sof/tools/` to be even more visible?

aiChaoSONG:
> I'm still a bit surprised that tool(s) to perform these sorts of analysis don't already exist but I'm new to this project.

I think audio analysis tools already exist in sof repo, see `sof/tools/test/audio`. but tools are written in octave, which is not suitable for test case to use.

test-case/check-hw-echo-reference.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+
+##
+## Case Name: check-hw-echo-reference
+## Preconditions:
+##    require python3-numpy, python3-scipy and python3-matplotlib
```
fredoh9, line 8: Let's add this to env-check.sh

test-case/check-hw-echo-reference.sh
```diff
@@ -0,0 +1,117 @@
...
+do
+    for fmt in $fmts
+    do
+        printf "Testing: iteration %d of %d with %s format\n" "$i" "$loop_cnt" "$fmt"
+        # S24_LE format is not supported
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # generate wave file
+        file="tmp_wave_${fmt%_*}.wav"
```
fredoh9, line 98: Are we creating temp files in current directory? We should create file in /tmp

test-case/check-hw-echo-reference.sh
```diff
@@ -0,0 +1,117 @@
...
+fmts="$cp_fmt"
+if [ ${OPT_VALUE_lst['F']} = '1' ]; then
+    fmts="$cp_fmts"
+fi
+
+for i in $(seq 1 $loop_cnt)
+do
+    for fmt in $fmts
+    do
+        printf "Testing: iteration %d of %d with %s format\n" "$i" "$loop_cnt" "$fmt"
```
fredoh9, line 92: use dlogX api to log. below also same

test-case/check-hw-echo-reference.sh
```diff
@@ -0,0 +1,117 @@
...
+        # S24_LE format is not supported
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # generate wave file
+        file="tmp_wave_${fmt%_*}.wav"
+        recorded_file="recorded_$file"
+        wavetool.py -g "sine;1.0;1000;0.;48000;2;${fmt%_*};10" -o "$file"
+        dlogi "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
```
fredoh9, line 101: dlogc?

test-case/check-hw-echo-reference.sh
```diff
@@ -0,0 +1,117 @@
...
+    do
+        printf "Testing: iteration %d of %d with %s format\n" "$i" "$loop_cnt" "$fmt"
+        # S24_LE format is not supported
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # generate wave file
+        file="tmp_wave_${fmt%_*}.wav"
+        recorded_file="recorded_$file"
+        wavetool.py -g "sine;1.0;1000;0.;48000;2;${fmt%_*};10" -o "$file"
```
fredoh9, line 100: For amplitude, i think 0.8 is better idea to avoid possible saturation problem. (give 20% buffer) I haven't check about `smart_amp` but always good idea.

aiChaoSONG:
@fredoh9 This case is used to check the dummy-DSM feature, I say `dummy` because there is no codec. you can run this case on any platform that with smart_amp component in the topology, `func_pipeline_export $tplg "smart_amp:any"`. Check echo reference case here: https://github.com/thesofproject/sof-test/pull/299
There is some difference between echo reference and this case:
- Channel of recorded wave: 2ch in echo reference, and 4 in dsm
- frequency analysis should work for both case, but in this case we do more strict check: binary comparison after trimming two sides' zeros.

BTW, I don't think alsabat can handle the condition which playback and capture channel is different. That's why I write this tool. I also consider to add more feature to test EQ, for example, we can check if a 800Hz low-cut filter really works with freq analysis, which alsabat cannot do.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    func = np.sin
+    if params['type'] == 'cosine':
+        func = np.cos
+    time = np.arange(0, float(params['duration']), 1.0/float(params['sample_rate']), dtype=np.float32)
+    data = float(params['amp']) * func(2 * np.pi * float(params['freq']) * time + float(params['phase']))
+    return np.reshape(np.repeat(data, int(params['channel'])),[len(data), int(params['channel'])])
+
+def parse_cmdline():
+    parser = argparse.ArgumentParser(add_help=True, formatter_class=argparse.RawTextHelpFormatter,
+        description='A Tool to Generate and Manipulate Wave Files.')
```
marc-hb, line 25: If you put a  string at the top of thee file it automatically gets added to the help:
```python
#! /usr/bin/python3
# SPDX-License-Identifier: BSD-3-Clause
# Copyright(c) 2020 Intel Corporation. All rights reserved.

"""
A Tool to Generate and Manipulate Wave Files
"""
```

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,118 @@
...
+##    1. reference wave and recorded wave are binary same
+##    2. delay in recorded wave less than 5ms
+##
+
+set -e
+# source from the relative path of current folder
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# we do not want the $TPLG to expand
```
marc-hb, line 30: ... because `eval` will expand it later? If correct add to comment (or explain other reason)

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+        wave_data = (np.iinfo(np.int16).max * wave_data).astype(np.int16)
+    if wave_format == 'S32':
+        wave_data = (np.iinfo(np.int32).max * wave_data).astype(np.int32)
+    try:
+        wavefile.write(wave_path, int(wave_params['sample_rate']), wave_data)
+    except:
+        print("Path specified not valid: %s" % wave_path)
+
+# remove zeros in two sides
+def trim_wave(wave):
```
marc-hb, line 94: It looks like `numpy` already has a `trim_zeroes()` function?

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    'parameter format: "mode;recorded_wave_path; reference_wave_path"\n'
+    'Supported comparison modes are: HWEchoRef')
+    return vars(parser.parse_args())
+
+def parse_params_and_gen_wav(cmd_param):
+    supported_type = ['sine', 'cosine']
+    param_keys = ['type', 'amp', 'freq', 'phase', 'sample_rate', 'channel', 'format', 'duration']
+    split_cmd_params = cmd_param.split(';')
+    wave_params = [param.strip() for param in split_cmd_params]
+    param_dict = dict(zip(param_keys, wave_params))
```
marc-hb, line 45: You're re-implementing a mini option parser here. Please use `argparse` for these too. It can provide default values and https://docs.python.org/3/library/argparse.html#choices

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    if param_dict['type'] not in supported_type:
+        print("Unsupported wave type")
+        sys.exit(1)
```
marc-hb, line 48: You're re-implementing a mini option parser here. Please use `argparse` for each `-g` param. It can provide default values and https://docs.python.org/3/library/argparse.html#choices

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+        return wave_data, param_dict
+
+def do_wave_compare(cmd_args):
+    supported_mode = ['HWEchoRef']
+    cmd_line = cmd_args['compare'].strip().split(';')
+    comparison_mode = cmd_line[0].strip()
+    if comparison_mode not in supported_mode:
+        print("Unsupported comparison mode: %s" % comparison_mode)
+        print('Supported comparison modes are: ' + str(supported_mode))
+        sys.exit(1)
```
marc-hb, line 60: Again this is duplicating `argparse` and https://docs.python.org/3/library/argparse.html#choices

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    wav_path, ref_path = [path.strip() for path in cmd_line[1:]]
+    if not os.path.exists(wav_path):
+        print('Recorded wave path does not exist: %s' % wav_path)
+        sys.exit(1)
+    if not os.path.exists(ref_path):
+        print('Reference wave path does not exist: %s' % ref_path)
+        sys.exit(1)
```
marc-hb, line 67: One big advantage of a high level language like Python over something like C is that you don't need checks like these which means one order of magnitude less code. You're losing this advantage. Just open the file and if not found then the built-in Python exception is perfectly readable (try it)

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    supported_format = ['S8', 'S16', 'S32', 'F32']
+    wave_path = path
+    if path == '.' or not path.endswith('wav'):
+        wave_name = wave_params['type'] + '_' + wave_params['channel'] + 'ch_' + wave_params['freq'] + 'Hz_' \
+        + wave_params['sample_rate'] + '_' + wave_params['format'] + '.wav'
+        wave_path = path + '/'  + wave_name
+    wave_format = wave_params['format'].upper()
+    if wave_format not in supported_format:
+        print('Unsupported wave format: %s' % wave_format)
+        sys.exit(1)
```
marc-hb, line 81: https://docs.python.org/3/library/argparse.html#choices does this check for free.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    if wave_format == 'S8':
+        wave_data = (np.iinfo(np.int8).max * wave_data).astype(np.int8)
+    if wave_format == 'S16':
+        wave_data = (np.iinfo(np.int16).max * wave_data).astype(np.int16)
+    if wave_format == 'S32':
+        wave_data = (np.iinfo(np.int32).max * wave_data).astype(np.int32)
+    try:
+        wavefile.write(wave_path, int(wave_params['sample_rate']), wave_data)
+    except:
+        print("Path specified not valid: %s" % wave_path)
```
marc-hb, line 91: I prefer the built-in python exception and less code.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    wave_format = wave_params['format'].upper()
+    if wave_format not in supported_format:
+        print('Unsupported wave format: %s' % wave_format)
+        sys.exit(1)
+    if wave_format == 'S8':
+        wave_data = (np.iinfo(np.int8).max * wave_data).astype(np.int8)
+    if wave_format == 'S16':
+        wave_data = (np.iinfo(np.int16).max * wave_data).astype(np.int16)
+    if wave_format == 'S32':
+        wave_data = (np.iinfo(np.int32).max * wave_data).astype(np.int32)
```
marc-hb, line 87:
```python
np_types = { 'S8' : np.int8, ... }
wave_data = (np.iinfo(np.int32).max * wave_data).astype(np_types[wave_format])
```

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    return wave[left_idx:right_idx,:], left_idx
+
+# float point binary comparison is not supported, and will not be supported
+# check HW echo referece wave
+def compare_wav_hw_echo_ref(wave_path, ref_path):
+    fs_wav, wave = wavefile.read(wave_path)
+    fs_ref, ref_wav = wavefile.read(ref_path)
+    if fs_wav != fs_ref:
+        print('Can not compare wave with different sample rate')
+        sys.exit(1)
```
marc-hb, line 115: Does this just duplicate a built-in exception and if yes which one?

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,118 @@
...
+        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file"
+        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file"
+        dlogi "Comparing recorded wave and reference wave"
+        output=$(wavetool.py -c "HWEchoRef;$recorded_file;$file")
+        printf "%s\n" "$output"
+        # clean up generated wave files
+        rm -rf "$file" "$recorded_file"
+        compare_result=$(echo "$output" | grep "PASSED")
+        if [ -z "$compare_result" ]; then
+            exit 1
```
marc-hb, line 113: Change `wavetool.py` to exit 0 for PASS, so much less code.
```shell
wavetool.py -c ... || die "test FAILED..."

File: test-case/check-smart-amplifier.sh [hunk 24]
```diff
@@ -0,0 +1,118 @@
...
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # generate wave file
+        tmp_dir="/tmp"
+        file="$tmp_dir/tmp_wave_${fmt%_*}.wav"
+        recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
+        wavetool.py -g "sine;0.8;1000;0.;48000;2;${fmt%_*};10" -o "$file"
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file &"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file" &
```
marc-hb, line 103:
```shell
arecord -D"$cp_dev" -r ... & arecordPID=$!
dlogc ..
...


# Now make sure arecord didn't fail
timeout 3 wait $arecordPID 
```

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+        if wave_mono[right_idx] != 0:
+            break
+        right_idx = right_idx - 1
+    return wave[left_idx:right_idx,:], left_idx
+
+# float point binary comparison is not supported, and will not be supported
+# check HW echo referece wave
+def compare_wav_hw_echo_ref(wave_path, ref_path):
+    fs_wav, wave = wavefile.read(wave_path)
+    fs_ref, ref_wav = wavefile.read(ref_path)
```
marc-hb, line 112: reading the files outside this function would probably make it easier to re-use.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    compare_result2_3 = np.array_equal(trimed_ref_wave, trimed_wave_ch_2_3)
+
+    hw_echo_delay = ((delay2 - delay1) / 48000 * 1000)
+    print("delay of hardware echo reference is %0.3fms" % hw_echo_delay)
+    if compare_result0_1 and compare_result2_3 and hw_echo_delay < 5.:
+        print('Recorded wave is binary same as reference wave')
+        print("Wave comparison result: PASSED")
+    else:
+        print('Recorded wave is not binary same as reference wave')
+        print("Wave comparison result: FAILED")
```
marc-hb, line 133: Use `logging.info()` so this can easily be silenced or redirect later (don't add any silencing or redirection feature yet)

Use an exit code so the caller does not have to grep

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+import scipy.signal as signal
+import scipy.io.wavfile as wavefile
+import scipy.fftpack as fftpack
+
+eps = 0.0000000001 # used to avoid zero division error
+
+def generateSinusoid(**params):
+    func = np.sin
+    if params['type'] == 'cosine':
+        func = np.cos
```
marc-hb, line 18: On one line:
```python
func = np.cos if params['type'] == 'cosine' else np.sin
```

marc-hb:
> > I'm still a bit surprised that tool(s) to perform these sorts of analysis don't already exist but I'm new to this project.

After a bit more research it looks like `scipy` is relatively popular for audio analysis and there seems to be a fair number of other Python libraries for sound processing so maybe Python is not a bad long  term investment choice after all. I have to admit I was initially deceived by the relatively large amount of unnecessary code. Once this script is trimmed down to a small fraction of what it would be in `C` for instance, the value of using Python, `scipy`, `wave` etc. should be more obvious.

Besides Octave/matlab/scipy and python, another "obvious" option is R which is the most popular language for data analysis. For audio I found this http://rug.mnhn.fr/seewave/exam.html

@singalsu ?

> I think audio analysis tools already exist in sof repo, see sof/tools/test/audio. but tools are written in octave, which is not suitable for test case to use.

Why not?

> This tool is specific to DSM/WOV test case. we need to analyze recorded wave to know if case passed or failed.

Comparing two wave files really doesn't seem specific to any test case so I still suspect this tool is being submitted to the wrong repo. Anyway let's finish the first round of review in this repo as it's already started.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+import sys
+import argparse
+import numpy as np
+import scipy.signal as signal
+import scipy.io.wavfile as wavefile
+import scipy.fftpack as fftpack
+
+eps = 0.0000000001 # used to avoid zero division error
+
+def generateSinusoid(**params):
```
marc-hb, line 15: In a function so short you can rename `params` to just `p`, it will fix the problem of too long lines and focus the reader's attention on what matters: parameter names like 'freq'.
You can add a pydoc if you want to explain what `p` is.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+
+##
+## Case Name: check-smart-amplifier
+## Preconditions:
```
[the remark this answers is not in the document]
aiChaoSONG: Currently, help information of test case is processed in opt.sh:  `func_opt_parse_option`. With unified grammar: "##" to specified case description, and `OPT_OPT_lst\OPT_DESC_lst\OPT_PARAM_lst/OPT_VALUE_lst` to give case option. The `usage` you recommended here will break current architecture. If you think we need to change this, I suggest to do overall change over all test case.

tools/wavetool.py
aiChaoSONG: Thanks.

tools/wavetool.py
aiChaoSONG: thanks.

tools/wavetool.py
aiChaoSONG: No, this is not. for DSM comparison, we do bit-by-bit comparison, and we need to ensure the waves have the same sample rate.

tools/wavetool.py
aiChaoSONG: This is good, thanks.

tools/wavetool.py
aiChaoSONG: thanks, will do

tools/wavetool.py
aiChaoSONG: thanks, will do

tools/wavetool.py
aiChaoSONG: 1. trim_zeros() in numpy only applies to 1-D arrary. and our audio is 2-D array.
2. Once wave goes through DAC/ADC, the zero may become a small value close to zero, eg, 15, -48.  (This is the situation we met in WOV test)
I need to keep self-implemented function.

tools/wavetool.py
aiChaoSONG: For Wake on Voice feature test, We need to generate a wave like this:
we need to give params of two sine wave, it's a little bit hard to specify with argparse to achieve this.
with the mini parser, we can specify:  -g type;amp1;freq1;duration1;amp2;freq2;duration2; sample_rate;channels;format

image

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,118 @@
...
+##    4. compare channel 0/1, 2/3 of recorded wave with reference wave
+## Expect result:
+##    1. reference wave and recorded wave are binary same
+##    2. delay in recorded wave less than 5ms
+##
+
+set -e
+# source from the relative path of current folder
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
```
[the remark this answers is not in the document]
aiChaoSONG: Currently, all our case use this form, but I think your idea is good. I will do this to all of our test case later.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,118 @@
...
+##    2. generate playback wave file (1000Hz sine wave)
+##    3. play reference wave file through smart_amp playback pipeline
+##       and record it back through smart_amp capture pipeline
+##    4. compare channel 0/1, 2/3 of recorded wave with reference wave
+## Expect result:
+##    1. reference wave and recorded wave are binary same
+##    2. delay in recorded wave less than 5ms
+##
+
+set -e
```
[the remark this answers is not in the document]
aiChaoSONG: I am afraid we cannot add `set -e` to this script, this option will end the process once there is any error. in the smart amplifier(dsm) test, there is a bug with firmware, the FW will hang, and sof-logger will exit with error, which will kill the test script process, and cause test incomplete, (can not report failure).

test-case/check-smart-amplifier.sh
aiChaoSONG: this explain is ambiguous,  here we just want the string "$TPLG",  fixed

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
marc-hb: > in the smart amplifier(dsm) test, there is a bug with firmware, the FW will hang, and sof-logger will exit with error

You can easily ignore any error with `sof-logger ... || true` but why would a test script silently ignore FW hangs?!? That makes no sense to me. At the very least it should print a warning `sof-logger .... || dlog ...`

singalsu:
I think Python makes the most efficient validation environment but there's more coding effort vs. higher level R or Matlab languages. Personally I think R looks gibberish when doing signal processing (it's for statistics originally) but AFAIK it's equally capable as Matlab. I think everyone developing tests could use what they are familiar to get the testing needs covered and the test eventually should run in default Python version available in current Ubuntu distro.

plbossart:
> I think Python makes the most efficient validation environment but there's more coding effort vs. higher level R or Matlab languages. Personally I think R looks gibberish when doing signal processing (it's for statistics originally) but AFAIK it's equally capable as Matlab (for scripts & graphs output, when no application GUI needed). I think everyone developing tests could use what they are familiar to get the testing needs covered and the test eventually should run in default Python version available in current Ubuntu distro.

I don't fully agree @singalsu. The tests need to rely on specific algorithms and 'not everyone' should be writing theses tests. I would be a lot more comfortable if we took your quality tests and implemented them both in Python and Matlab, so that the *same criteria* are used in different environments. If we have different test methodologies because different people wrote their own tests in their preferred language we are going to see divergence between simulation and validation.

juimonen:
@singalsu @plbossart just my 2 cents here:
I would also like to restrict the languages as few as possible. So if we would use only 1, I would go for python. But I would still suggest to use combination of python and matlab/octave: The python/scipy syntax is close to matlab, but if you are doing anything little bit more elaborate (more complex signal analysis with graphs) matlab/octave is still 1 notch ahead of python in easiness and function offering. Also matlab/octave is still de facto design tool for algorithm designers, we have also some filter designs in that language -> a lot of folks in dsp community use it.

tools/wavetool.py
```diff
@@ -0,0 +1,157 @@
...
+    'Support format: S8, S16, S32, F32\n')
+    parser.add_argument('-o', '--output', type=str, help='Path to store generated files', default='.')
+    parser.add_argument('-p', '--preset', type=str, help='Commonly used wave presets\n'
+    'Supported presets are: sine_1k_16bit_48k_10s')
+    parser.add_argument('-c', '--compare', type=str, help='compare recorded wave and reference wave\n'
+    'parameter format: "mode;recorded_wave_path; reference_wave_path"\n'
+    'Supported comparison modes are: HWEchoRef')
+    return vars(parser.parse_args())
+
+def parse_params_and_gen_wav(cmd_param):
```
marc-hb, line 40: Considering everything is in dictionaries, it should be easy to replace this single function with two names by two functions with a single name.

tools/wavetool.py
marc-hb: 1. `wave[:,0]` is a 1-D array, you can use `trim_zeros()` on it.
2. Where is this rounding? I don't this it in this function. 2. bis On line 115 you mentioned a bit-for-bit comparison. I'm confused and also a bit worried because rounding if any should be one of the most visible things  / main parameter in this code and I haven't spotted it yet. It's also a sure way to get different results with duplicated test code, see @plbossart's comment.

tools/wavetool.py
marc-hb: If you compare two diffferent rates bit by bit they won't be equal for sure and the test will fail which seems good.

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
aiChaoSONG: we will start sof-logger at the beginning of out test cases with `[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
`, currently, there is no `set -e` in every case. once with `set -e`, our test case cannot run to complete if sof-logger exit too early due to FW issue. I can fix this issue later with `sof-logger ... || true`, but not in this PR

test-case/check-hw-echo-reference.sh
aiChaoSONG: Here is the dummy-dsm test, signal will not goes to codec, only digital transmission, so amp here have no positive/negtive impact

aiChaoSONG:
@marc-hb Most of the reviews are addressed, no mini parser anymore, redundant codes are removed.  BTW, WOV test is based on this tool, too. without this merged, it's impossible to start WOV test case. we need both case in TGL release.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
...
+tplg=${OPT_VALUE_lst['t']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+func_pipeline_export $tplg "smart_amp:any"
+func_lib_setup_kernel_last_line
+
+[ "$PIPELINE_COUNT" == "2" ] ||{
+    die "Only detect $PIPELINE_COUNT pipeline(s) from topology, but two are needed"
+}
```
marc-hb, line 64: Nit: you don't need the braces `{ }`

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+def trim_wave(wave):
+    wave_mono = wave[:,0]
+    left_idx = 0
+    right_idx = wave_mono.shape[0] - 1
+    while True:
+        if abs(wave_mono[left_idx]) > 100:
+            break
+        left_idx = left_idx + 1
+    while True:
+        if abs(wave_mono[right_idx]) > 100:
```
marc-hb, line 86: I guess this 100 value should be at least a named constant. Maybe even a script parameter?

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
marc-hb: Maybe I'm missing something here but it still feels very wrong to add a new test that PASSes when the firmware actually hangs.

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+
+def generate_wav(**cmd):
+    if cmd['generate'] in ['sine', 'cosine']:
+        wave_param = {
+            'type': cmd['generate'], 'amp': cmd['amp'][0], 'freq': cmd['freq'][0],
+            'phase': cmd['phase'][0], 'chan': cmd['channel'], 'sample_rate': cmd['sample_rate'],
+            'duration': cmd['duration'][0]
+        }
+        wave_data = generate_sinusoids(**wave_param)
+        return wave_data
```
marc-hb, line 54: It looks like you want `generate_wav()` to be ready for other types of signals, correct? If correct then make it completely ready and clearer by:

1. adding the missing `default case`
```python
    else:
        raise Exception("invalid generate function")
```
... you never want this function to silently return `None` in case of a future bug.

2. Indenting `return wave_data` one notch left

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+    parser.add_argument('-S', '--sample_rate', type=int, default=48000, help='sample rate of generated wave')
+    parser.add_argument('-C', '--channel', type=int, default=2, help='channels of generated wave')
+    parser.add_argument('-B', '--bits', type=str, choices=['S8', 'S16', 'S32', 'F32'], default='S16',
+    help='sample bits of generated wave')
+    parser.add_argument('-o', '--output', type=str, help='path to store generated files', default='.')
+    # wave comparison arguments
+    parser.add_argument('-c', '--compare', type=str, choices=['smart_amp'], help='compare recorded wave and reference wave')
+    parser.add_argument('-R', '--recorded_wave', type=str, help='path of recorded wave')
+    parser.add_argument('-r', '--reference_wave', type=str, help='path of reference wave')
+    return vars(parser.parse_args())
```
marc-hb, line 133: What's wrong with `cmd_args.generate`, `cmd_args.compare`, etc.?

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+        return wave_data
+
+def save_wave(wave_data, **cmd):
+    wave_path = cmd['output'].strip()
+    if wave_path == '.' or not wave_path.endswith('wav'):
+        wave_path = wave_path + '/tmp.wav'
+    sample_bits = cmd['bits']
+    np_types = {'S8': np.int8, 'S16': np.int16, 'S32': np.int32}
+    if sample_bits in np_types.keys():
+        wave_data = (np.iinfo(np_types[sample_bits]).max * wave_data).astype(np_types[sample_bits])
```
marc-hb, line 63: This multiplies the values, is this really what you want to do? Please add a comment.

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+    wavefile.write(wave_path, cmd['sample_rate'], wave_data)
+
+def do_wave_compare(**cmd):
+    fs_wav, wave = wavefile.read(cmd['recorded_wave'].strip())
+    fs_ref, ref_wave = wavefile.read(cmd['reference_wave'].strip())
```
marc-hb, line 68: argparse doesn't add whitespace. If there is whitespace it's because the user took the very unlikely pain to add whitespace so don't remove it.

tools/wavetool.py
marc-hb: Some rounding just got added, see below.

tools/wavetool.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+
+import os
+import sys
+import argparse
+import numpy as np
+import scipy.signal as signal
+import scipy.io.wavfile as wavefile
+
+eps = 0.0000000001 # used to avoid zero division error
```
marc-hb, line 16: On which line(s) is this used?

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
...
+        fi
+        # generate wave file
+        tmp_dir="/tmp"
+        file="$tmp_dir/tmp_wave_${fmt%_*}.wav"
+        recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
+        wavetool.py -gsine -A0.8 -B${fmt%_*} -o"$file"
+        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file &"
+        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file"
```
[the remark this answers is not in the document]
aiChaoSONG: Yes, but before we add `set -e`, we need to another fix: https://github.com/thesofproject/sof-test/pull/313. I will add set -e, please help to review #313

tools/wavetool.py
aiChaoSONG: yes, to same data to sign 8 bit, 16 bit or 32 bit, need to multiply value range from [-1.0, 1.0] to [-127, 127]... will fix

tools/wavetool.py
aiChaoSONG: not currently used, removed

tools/wavetool.py
aiChaoSONG: didn't totally get the point here, sorry.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
...
+
+[ "$PIPELINE_COUNT" == "2" ] ||{
+    die "Only detect $PIPELINE_COUNT pipeline(s) from topology, but two are needed"
+}
+
+for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
+do
+    type=$(func_pipeline_parse_value "$idx" type)
+    if [ "$type" == "playback" ]; then
+        pb_chan=$(func_pipeline_parse_value "$idx" ch_max)
```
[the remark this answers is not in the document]
aiChaoSONG: we have two pipelines here, one is playback pipeline, one is capture pipeline, but we cannot assume which goes first.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
...
+
+OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
+
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='playback/capture duration in second'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=6
+
+# we do know this is not used, but we need it.
+# shellcheck disable=SC2034
+OPT_OPT_lst['F']='fmts'   OPT_DESC_lst['F']='Iterate all supported formats'
```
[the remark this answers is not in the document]
aiChaoSONG: there should be option description already, see `OPT_DESC_lst['F']='Iterate all supported formats'`

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
...
+    type=$(func_pipeline_parse_value "$idx" type)
+    if [ "$type" == "playback" ]; then
+        pb_chan=$(func_pipeline_parse_value "$idx" ch_max)
+        pb_rate=$(func_pipeline_parse_value "$idx" rate)
+        pb_dev=$(func_pipeline_parse_value "$idx" dev)
+    else
+        cp_chan=$(func_pipeline_parse_value "$idx" ch_max)
+        cp_rate=$(func_pipeline_parse_value "$idx" rate)
+        cp_fmt=$(func_pipeline_parse_value "$idx" fmt)
+        cp_fmts=$(func_pipeline_parse_value "$idx" fmts)
```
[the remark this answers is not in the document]
aiChaoSONG: with a variable `get_pipeline_param='func_pipeline_parse_value "$idx"'`,  we still need `cp_chan=$(get_pipeline_param ch_max)`, duplication is not eliminated. don't see improvement here, or I get the wrong point?

aiChaoSONG:
> My sof-logger ... || true suggestion was a bad suggestion, forget it. It assumed the sof-logger issue was harmless. A FW hangs does really not seem harmless.
> I'm suggesting to add set -e only to this test to make sure this new test does its job of finding bugs. We can fix green failures in other tests later.

I think it is not that bad, once dsp hang, we will have a lot of errors in dmesg, we can still catch the error. without this suggestion,  we are unable to add `set -e` to any test case. because if sof-logger exit early, the kill in case-lib/hijack.sh line 19 fails, then test case exit without complete. maybe add some error print?

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
marc-hb: > this is not used, but we need it

I think there is not enough information to understand this contradiction.

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
marc-hb: Thanks! I think your explanation deserves to be a comment.

test-case/check-smart-amplifier.sh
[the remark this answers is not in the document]
marc-hb: Sorry I was currying without realizing I was. Here's how actual currying can make the code significantly shorter and a bit clearer:
```shell
for idx in $(seq 0 $((PIPELINE_COUNT - 1))); do

    parse_val() { func_pipeline_parse_value "$idx" "$1" ; }
    ... 
    pb_rate=$(parse_val rate)
    pb_dev=$(parse_val dev)
    ...

done
```
bash is also a functional language.

tools/wavetool.py
marc-hb: What is the purpose of `vars()`? Why not `return parse.parse_args()`?

tools/wavetool.py
```diff
@@ -0,0 +1,151 @@
...
+    parser.add_argument('-v', '--version', action='version', version='%(prog)s 1.0')
+    # wave parameters
+    parser.add_argument('-g', '--generate', type=str, choices=['sine','cosine'], help='generate specified types of wave')
+    parser.add_argument('-A', '--amp', type=float, nargs='+', default=[1.0], help='amplitude of generated wave')
+    parser.add_argument('-F', '--freq', type=float, nargs='+', default=[997.0], help='frequency of generated wave')
+    parser.add_argument('-P', '--phase', type=float, nargs='+', default=[0.0], help='phase of generated wave')
+    parser.add_argument('-D', '--duration', type=float, nargs='+', default=[10.], help='duration of generated wave')
+    parser.add_argument('-S', '--sample_rate', type=int, default=48000, help='sample rate of generated wave')
+    parser.add_argument('-C', '--channel', type=int, default=2, help='channels of generated wave')
+    parser.add_argument('-B', '--bits', type=str, choices=['S8', 'S16', 'S32', 'F32'], default='S16',
```
marc-hb, line 131: Could we re-use the numpy names here? I mean `int8`, `int16`,...

tools/wavetool.py
```diff
@@ -0,0 +1,151 @@
...
+    return wave_data
+
+def save_wave(wave_data, **cmd):
+    wave_path = cmd['output']
+    if wave_path == '.' or not wave_path.endswith('wav'):
+        wave_path = wave_path + '/tmp.wav'
+    sample_bits = cmd['bits']
+    np_types = {'S8': np.int8, 'S16': np.int16, 'S32': np.int32}
+    if sample_bits in np_types.keys():
+        # range of wave_data is [-1.0, 1.0], to save to S8 format, need to multiply the maximum of S8
```
marc-hb, line 63: Thanks for the comment, now I finally understand this code :-)
Suggestion:
```
# range of wave_data is floating point [-1.0, 1.0]. When saving to any integer format use its full range.
```

aiChaoSONG:
> What is the purpose of vars()? Why not return parse.parse_args()?

@marc-hb the `vars()` here is to transform return value of `parse.parse_args()` into a python dict,  without `vars`, the returned value looks like this:
```
Namespace(amp=[1.0], bits='S16', channel=2, compare=None, duration=[10.0], freq=[997.0], generate=None, output='.', phase=[0.0], recorded_wave=None, reference_wave=None, sample_rate=48000)
``` 
this value is not as easy to use as python dict

aiChaoSONG:
@marc-hb The only logic in this PR is to do bit-by-bit comparison in function `compare_wav_smart_amp`: `np.array_equal(trimed_ref_wave, trimed_wave_ch_0_1)`. The concept here is that if 1 is a 32 bit value, then 1 == 1 also means that 1 and 1 are binary same, right?

tools/wavetool.py
aiChaoSONG: If we want to reuse these value, we need a meta class, seems there is not such a class.

marc-hb: I meant this:
```python
parser.add_argument('-B', '--bits', type=str, choices=['int8', 'int16', 'in32', ...], default='S16',
```
No real code change, only a string change for consistency with numpy.

aiChaoSONG: The `parse_val()` do make code shorter in this case, but since all of our case using this style, we may consider to do overall clean up, rather than only do it in this case.  A `parse_val` in this case may cause this function duplicated in other cases.   I think we may rename "func_pipeline_parse_value" to be shorter.

aiChaoSONG:
@marc-hb I'd like to remove `vars` and use 'args.amp', but for the int8, int16, I think S8, S16 and S32 is better, because aplay and arecord use S16_LE, S32_LE, etc. It's better to add a layer in wavetool (S32_LE->int32) than add a layer in test case

aiChaoSONG:
@marc-hb Almost every function in wavetool use command line parameters, pass a argparse Namespace class is no better than pass a python dict. at least **cmd denote multiple parameters. I suggest we keep the dict implemetation.

marc-hb:
>  pass a argparse Namespace class is no better than pass a python dict. 

I agree there's probably not much technical difference (the dict implementation is just a bit more verbose); the problem is serious reader confusion. It took me a very long time and multiple rounds of reviews before I realized all these dictionaries were actually just `args`, same thing as in every Python script but in disguise here.

> Almost every function in wavetool use command line parameters,

That's exactly what took me a very long time to realize.

In fact it would be clearer if these functions didn't even take `args` parameters at all but use a global instead. The idea of function parameters is that they change between invocations whereas `args` are constants while the script runs.

> at least **cmd denote multiple parameters

`args` or `cmd_args` also denotes that and much more: exactly where it comes from.

> I suggest we keep the dict implemetation.

I suggest a large but very simple and mechanical search/replace requiring very little testing.

tools/wavetool.py
marc-hb: Reposting a comment lost by github?

> Thanks for the comment, now I finally understand this code :-)
Suggestion:
```
# range of wave_data is floating point [-1.0, 1.0]. When saving to any integer format use its full range.
```

marc-hb: It seems `numpy` uses standard 2's complement so I think there's an underflow bug for values close to `-1.0`. This should fix it (untested)
```python
# range of wave_data is floating point [-1.0, 1.0]. When saving to any integer format use its full range.
wave_data = (- np.iinfo(np_types[sample_bits]).min * wave_data).astype(np_types[sample_bits])
```
```shell
for i in 8 16 32; do python3 -c "import numpy as np; print(np.iinfo(np.int$i))"; done
Machine parameters for int8
---------------------------------------------------------------
min = -128
max = 127
---------------------------------------------------------------

Machine parameters for int16
---------------------------------------------------------------
min = -32768
max = 32767
---------------------------------------------------------------

Machine parameters for int32
---------------------------------------------------------------
min = -2147483648
max = 2147483647
---------------------------------------------------------------
```

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
+#! /usr/bin/python3
```
singalsu, line 1: There's no description in commit, please add.

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
...
+    Parameters
+    ----------
+    dict of sinusiod parameters:
+    p['type']: Wave type, ``sine`` or ``cosine``
+    p['amp']: Amplitude
+    p['freq']: Frequency
+    p['phase']: Phase
+    p['chan']: Channels
+    p['sample_rate']: Sample rate
+    p['duration']: Duration
```
singalsu, line 33: Would be good to add units to parameters descriptions (linear, Hz, rad, Hz, s, etc.).

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
...
+        sys.exit(1)
+    if cmd['analyze'] == 'smart_amp':
+        analyze_wav_smart_amp(wave, ref_wave, fs_wav)
+
+# remove digital zeros in two sides
+def trim_wave(wave):
+    # once waves go through DAC/ADC, zero will become small value close to zero,
+    # here we set the digital zero threshold to 100, and cut samples below 100
+    # of two sides, this has no negative impact to processing.
+    DIGITAL_ZERO_THRESHOLD = 100
```
singalsu, line 81: Not a good idea to refer directly to integer values of s16/s24/s32 format. The code will fail with different word length used. Such threshold should be a float or decibel number relative to full scale.

But the most important question - why do you need to trim the samples based on this kind of silence detection?

aiChaoSONG: For DSM, there are only digital process, a comparison with zero is OK to trim zeros in two sides. but for WOV, to trim zero, we need to find a zero threshold because of ADC/DAC.

I have done some experiment with the current mounted DMIC injector under S16_LE,  and found 100 is a proper value to discard silence. I understand that under S24/S32 some more valid values may be trimmed, but it doesn't impact later data process that much.
Without much knowledge in audio processing senario, I found this strategy worked. But as you have stated, dBFS will provide the same criteria under s16/s24/s32, and is better to be used here. 
Will change to dBFS implementation in next push.

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
...
+    trimed_ref_wave = trimed_ref_wave[0:trimed_wave_ch_0_1.shape[0],:]
+    compare_result0_1 = np.array_equal(trimed_ref_wave, trimed_wave_ch_0_1)
+    # compare the second two channel
+    trimed_wave_ch_2_3, delay2 = trim_wave(wave[:,2:4])
+    trimed_ref_wave = trimed_ref_wave[0:trimed_wave_ch_2_3.shape[0],:]
+    compare_result2_3 = np.array_equal(trimed_ref_wave, trimed_wave_ch_2_3)
+
+    smart_amp_delay = ((delay2 - delay1) / fs * 1000)
+    print('Delay of smart amplifier is %0.3fms' % smart_amp_delay)
+    if compare_result0_1 and compare_result2_3 and smart_amp_delay < 5.:
```
singalsu, line 110: Magic number 5 (ms), since it's acceptance criteria could it be defined in the beginning of module and described?

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    Generate sine or cosine wave.
+
+    ``y(t) = A * sin(2 * pi * f * t + phi)``
+
+    ``y(t) = A * cos(2 * pi * f * t + phi)``
+
+    Parameters
+    ----------
+    dict of sinusiod parameters:
```
singalsu, line 26: Just a generic question. Can this function use different sine wave parameters for each channel (amp, frequency, phase)?

aiChaoSONG: Currently, no, but if we need this kind of sine wave, I can extend this feature.

tools/wavetool.py
```diff
@@ -0,0 +1,150 @@
+#! /usr/bin/python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+
+"""
+A Tool to Generate and Manipulate Wave Files
```
singalsu, line 6: Also this is rather brief description. Is there somewhere a graph of test system? If not, could even add an ASCII graph here for description what the test does.

test-case/check-smart-amplifier.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
singalsu, line 1: No commit message, just title.

aiChaoSONG:
@singalsu Three procedures in the DSM test case: 
- generate playback wave, which is a sine wave.
- play sine wave through DSM playback pipeline and recorded it back through DSM capture pipeline
- do binary comparison between playback wave and captured wave.

There is no processing in the DSM component, just copy the playback data back to the capture pipeline.

What the wavetool.py did is just to do binary comparison, to be short, compare S16/S24/S32 value sample by sample.
As the recorded wave is not aligned, so need to trim zero of two sides to make it aligned with playback wave.
In the DSM case,  comparison to zero should define if a value is zero,  But in WOV(which will be implement next) test, because ADC/DAC is present, we need a zero threshold.

tools/wavetool.py
singalsu: It's useful to have different signal in channels to check that one channel is not duplicated to all channels. If such is a possible fault for DSM the change should be done.

tools/wavetool.py
singalsu: If 100 is suitable for s16, then please change it to 100.0/INT16_MAX that is about 0.003. Or -50.3 decibels. I think that would be the minimum fix for this.

It's not possible to test e.g. volume control over it's range with such simple test signal find. It gets unreliable with low levels and noise. That's why I use in SOF/tools/test/audio stuff the chirp markers and code to detect them. It works with noisy ADC/DAC too and even with acoustics and some office noise. Lack of doing that is not blocker for smart amp test apparently but once implemented as library it's easy to use in all tests and make them robust.

tools/wavetool.py
aiChaoSONG: For DSM case, the tool only do binary comparison. There is sure case description in check-smart-amplifier.sh

tools/wavetool.py
aiChaoSONG: Currently for the DSM case, we do binary comparison, so wave type/parameter don't matter. I will add this feature and chirp sound generation in my TODO list.

tools/wavetool.py
aiChaoSONG: current implementation is OK to cope with the TGL release, later we can add more feature to this tool, for example, I'd like to add EQ test by checking if some frequency component is decreased.

aiChaoSONG:
Thanks Marc, Seppo for very useful reviews, educated me a lot.  I am going to merge this as WOV test case is en route. Any further review will be applied as bug fix of improvement. Thank you all.

marc-hb:
I just tried to generate a signal and compare it with itself and it fails like this. @aiChaoSONG how did you test this?

```
$ ./tools/wavetool.py -gsine && ./tools/wavetool.py -a smart_amp -r tmp.wav -R tmp.wav 

  File "./tools/wavetool.py", line 160, in <module>
    main()
  File "./tools/wavetool.py", line 157, in main
    do_wave_analysis(**cmd_args)
  File "./tools/wavetool.py", line 83, in do_wave_analysis
    analyze_wav_smart_amp(wave, ref_wave, fs_wav, cmd['zero_threshold'])
  File "./tools/wavetool.py", line 113, in analyze_wav_smart_amp
    trimed_wave_ch_2_3, delay2 = trim_wave(wave[:,2:4], zero_threshold)
  File "./tools/wavetool.py", line 91, in trim_wave
    wave_mono = wave[:,0]
IndexError: index 0 is out of bounds for axis 1 with size 0
```

aiChaoSONG:
@marc-hb I guess your waves are both 2 channels, right? For the DSM, the recorded wave is 4 channels,  so you need `-R recorded-four-channel-wave -r two-channel-reference-wave`.

marc-hb:
> @marc-hb I guess your waves are both 2 channels, right? 

You don't need to guess, you can just run the command I gave.

> so you need -R recorded-four-channel-wave -r two-channel-reference-wave.

Indeed `sof-test/test-case/check-smart-amplifier.sh` seems to generate a reference with two channels. 

That's quite confusing,  I thought this wavetool was meant to become less specialized.

Now I understand the duplication in `analyze_wav_smart_amp()` better. Can you at least create a function that compares two channels and removes this duplication? (after reviewing and merging this cleanup #334). I bet it will make the backtrace above a bit clearer.

PR #299, opened 2020-07-29
test-case: add echo reference test case

keqiaozhang:
add one new test case to check the echo reference on
on supported platforms.

### DIFF
--- a/test-case/check-fw-echo-reference.sh
+++ b/test-case/check-fw-echo-reference.sh
@@ -1,0 +1,116 @@
+#!/bin/bash
+
+##
+## Case Name: check_echo_reference
+## Preconditions:
+##    N/A
+## Description:
+##    using alsabat to check the echo reference
+## Case step:
+##    1. play sine wave on echo reference playback pipeline
+##    2. capture the data through the internel loopback
+##    3. use alsabat to analyze the recorded data
+## Expect result:
+##    no errors and the captured data should match with the orignal one
+##
+
+# remove the existing alsabat wav files
+rm -f /tmp/bat.wav.*
+
+# shellcheck source=case-lib/lib.sh
+# shellcheck disable=SC2016
+# shellcheck disable=SC1091
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# TPLG is assigned as env variable
+# shellcheck disable=SC2154
+# shellcheck disable=SC2016
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+# shellcheck disable=SC2153
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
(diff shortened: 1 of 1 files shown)


xiulipan:
@keqiaozhang Do we really need a new test scripts for the echo reference test? I think this is almost same like `check-alsabat.sh`  
Also please name this as FW echo reference to make difference with the DSM hw echo reference

keqiaozhang:
> @keqiaozhang Do we really need a new test scripts for the echo reference test? I think this is almost same like `check-alsabat.sh`

> Also please name this as FW echo reference to make difference with the DSM hw echo reference

Yes, I considered to reuse alsabat test script before, but we better to use tplgreader to get the PCM IDs of echo reference pipelines instead of the hard code and also in alsabat test script, we use plugin, but we want to test echo reference with different formats.
Case name is updated to check-fw-echo-reference.sh

marc-hb:
13 `shellcheck disable` (with barely any comment) in a single small file like this really doesn't really make sense, fixed in PR #315 .

test-case/check-fw-echo-reference.sh
```diff
@@ -0,0 +1,116 @@
...
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # run echo reference test
+        dlogc "alsabat -P $pb_dev --standalone -c $channel -f $fmt -r $rate -n $frames -F $frequency"
+        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" &
+        # playback may have low latency, add 0.5 second delay to aviod recording zero at beginning.
+        sleep 1
+        dlogc "alsabat -C $cp_dev -c $channel -f $fmt -r $rate -F $frequency"
+        alsabat -C "$cp_dev" -c "$channel" -f "$fmt" -r "$rate" -F "$frequency"
```
marc-hb, line 104: Please help review #313 and then test this script with `set -e`, see #312 

I think we don't want to risk missing any `alsabat` failure.

test-case/check-fw-echo-reference.sh
```diff
@@ -0,0 +1,116 @@
...
+    for fmt in $fmts
+    do
+    printf "Testing: iteration %d of %d with %s format\n" "$i" "$loop_cnt" "$fmt"
+        # S24_LE format is not supported
+        if [ "$fmt" == "S24_LE" ]; then
+            continue
+        fi
+        # run echo reference test
+        dlogc "alsabat -P $pb_dev --standalone -c $channel -f $fmt -r $rate -n $frames -F $frequency"
+        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" &
```
marc-hb, line 100: Make sure we don't miss any `alsabat` failure, this is a test script:
```shell
alsabat ... & alsabatPID=$!
...
...
timeout 5 wait $alsabatPID
```

PR #309, opened 2020-08-01
env-check: add sof-dyndbg.conf file check

marc-hb:
Check sof-dyndbg.conf available in the system, give optional warning and
print short usage of dyndbg for SOF modules.


marc-hb:
This is just:
- a shorter version of #260 that has been approved (and in review for 1+ week)
- trivial double quotes

=> fast tracking

PR #310, opened 2020-08-01
env-check.sh: fix 8/22 shellcheck warnings

marc-hb:
See commit messages.

Only the easiest fixes for now.

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -111,6 +111,7 @@ check_res=1 && out_str=$out_str"\n
 
 out_str="" check_res=0
 echo -ne "Checking the config setup:\t\t"
+# shellcheck source=case-lib/config.sh
 source  "$(dirname "$0")/case-lib/config.sh"
 # effect check
 case "$SUDO_LEVEL" in
--- a/env-check.sh
+++ b/env-check.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+mydir=$(cd "$(dirname "$0")"; pwd)
+
 # enable dynamic debug logs for SOF modules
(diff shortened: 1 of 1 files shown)


fredoh9:
Thanks! we are getting less shellcheck warnings. But I'm seeing shellcheck warnings on the lines in you are changing. Are you going to fix? Otherwises, should we put some comments to ignore these warnings?

```
In env-check.sh line 3:
        ^------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
		
In env-check.sh line 54:
cd "$mydir"
^---------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

In env-check.sh line 56:
[[ "$(stat -c "%A" ./tools/* |grep -v 'x')" ]] && check_res=1 && out_str=$out_str"\n
   ^-- SC2143: Use grep -q instead of comparing output with [ -n .. ].
   
In env-check.sh line 67:
[[ "$(stat -c "%A" ./test-case/* |grep -v 'x')" ]] && check_res=1 && out_str="\n
   ^-- SC2143: Use grep -q instead of comparing output with [ -n .. ].
```

marc-hb:
> But I'm seeing shellcheck warnings on the lines in you are changing.

I didn't add these warnings, I just moved several of them into a single place to reduce their number.

> Are you going to fix? Otherwises, should we put some comments to ignore these warnings?

I've been trying to keep this PR reasonably small to minimize conflicts with other work in progress and speed up review. Looks like the last part isn't working :-)

One way I'm keeping it small and very low risk is by keeping the number of logical changes very small.

marc-hb:
This script is not in any critical path so one approval will be enough...

PR #311, opened 2020-08-01
 env-check.sh: replace sed /etc/group hack with usermod command 

marc-hb:
2 commits. The main change is replacing the `sed /etc/group` suggested hack with a `usermod` command. The other commit is just a minor cleanup.

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -9,7 +9,7 @@ func_check_pkg(){
     if command -v "$1" >/dev/null; then
         return
     else
-        out_str=$out_str"\nPlease install \e[31m $1 \e[0m package"
+        out_str="$out_str""\tPlease install the \e[31m $1 \e[0m package\n"
         check_res=1
     fi
 }
@@ -19,7 +19,7 @@ func_check_python_pkg(){
         if python3 -c "import $1" &> /dev/null; then
             return
         else
-            out_str=$out_str"\nPlease install \e[31m python3-$1 \e[0m package"
+            out_str="$out_str""\tPlease install the \e[31m python3-$1 \e[0m package\n"
             check_res=1
         fi
     else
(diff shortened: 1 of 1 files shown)


marc-hb:
All the shellcheck warnings were already there, this PR doesn't add any (it removes one or two).

One third of these warnings are fixed by unrelated PR #310

env-check.sh
```diff
@@ -31,19 +31,26 @@ func_check_file(){
...
-[[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+if [ $check_res -eq 0 ]; then
+    printf "pass\n"
+else
+    printf '\e[31mWarning\e[0m\n'
+# Need ANSI color characters to be the format string. This is not
+# unsanitized input.
+# shellcheck disable=SC2059
+    printf "$out_str"
```
lyakh, line 52: what are the advantages of using `printf` over `echo`?

marc-hb: `echo` is a mess https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819

lyakh: ok, good to know, thanks. Still, I think as long as we don't use sufficiently advanced "echo" features and our shell is fixed to `#!/bin/bash` in most cases `echo` should work just fine, so hopefully we aren't going to start replacing it everywhere with `printf`

lyakh: @marc-hb btw, while at it, do we really need those escape sequences for colour coding etc.? They are fun and (arguably) useful when working with a terminal, but they aren't helpful when redirecting output to a file. Ideally we'd want to check whether the output is redirected like git and others do...

paulstelian97: Preferably variables for the escape sequences. Yeah.

marc-hb: Thanks for the reviews!

I'm not aware of any plan to mass search/replace of `echo` but I generally replace it when I'm modifying the code anyway for some other reason. It's rare to print a pure string without any variable and then it's easier to just `printf '%s' "$VAR"` rather than think about what could be in the string; and then these assumptions to become wrong in the future. While shell is admittedly safer than C, you would never `printf(some_string_variable);` in C without a format either. This would be flagged by most compilers and analyzers anyway.

I'm not a fan of the escape sequences either but I'm trying to focus on more important stuff right now :-) In this case I'm not aware of `env-check.sh` being used programmatically.

BTW you can use `less -R` to view files with escape sequences.

> Preferably variables for the escape sequences. Yeah.

I don't understand this sorry.

paulstelian97: I mean have variables like $RED, $RESET etc which could either contain the escape sequences or empty strings.

marc-hb: There are many tasks where no other language can beat a short shell script (< few hundred lines). But for building a good, interactive user interface I would probably looks elsewhere.
tl;dr: I think the colors are overkill here, some basic "ASCII art" would be simpler and good enough.

PR #313, opened 2020-08-05
case-lib: fix case exit early caused by sof-logger

aiChaoSONG:
Once DSP hangs, sof-logger will exit early before
we explicitly kill it. Test case will exit due to
the failure of killing sof-logger if we have 'set -e'
in the case.

This patch makes the kill of sof-logger always return
true.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -16,7 +16,10 @@ function func_exit_handler()
             sleep 1s
         }
         local loggerBin; loggerBin=$(basename "$SOFLOGGER")
-        sudo pkill -9 "$loggerBin" 2>/dev/null
+        sudo pkill -9 "$loggerBin" 2>/dev/null || {
+            dloge "sof-logger was already dead"
+            exit_status=1
+        }
         sleep 1s
     fi
     # when case ends, store kernel log


case-lib/hijack.sh
```diff
@@ -16,7 +16,7 @@ function func_exit_handler()
             sleep 1s
         }
         local loggerBin; loggerBin=$(basename "$SOFLOGGER")
-        sudo pkill -9 "$loggerBin" 2>/dev/null
+        sudo pkill -9 "$loggerBin" 2>/dev/null || true && dloge "sof-logger exit too early"
```
marc-hb, line 19: Can we have this?
```shell
 sudo pkill -9 "$loggerBin" 2>/dev/null ||  {
  dloge "sof-logger was already dead"
  exit_status=1
}
```

aiChaoSONG: Yeah, this is very good idea

marc-hb:
As we have no CI for sof-test yet, please submit a "dummy PR" to sof immediately after this get merged.

aiChaoSONG:
test PR: https://github.com/thesofproject/sof/pull/3277, result: https://sof-ci.01.org/sofpr/PR3277/build6817/devicetest

PR #314, opened 2020-08-05
Ignore non-audio errors on CML Mantis

fredoh9:
First commit is for Intel(R) Management Engine Interface(mei) errors
2nd commit is for DELL touchpad i2c errors

Both are not 100% reproducible but fairly easy to reproduce.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -26,6 +26,11 @@ ignore_str="$ignore_str"'|Parity error detected'
 # https://bugzilla.kernel.org/show_bug.cgi?id=202541
 ignore_str="$ignore_str"'|xhci_hcd 0000:00:14.0: WARN Set TR Deq Ptr cmd failed due to incorrect slot or ep state'
 
+# CML Mantis occasionally throws Intel(R) Management Engine Interface(mei) errors
+# https://unix.stackexchange.com/questions/109294/mei-00000016-0-init-hw-failure
+ignore_str="$ignore_str"'|mei_me 0000:00:16.0: wait hw ready failed'
+ignore_str="$ignore_str"'|mei_me 0000:00:16.0: hw_start failed ret = -62'
+
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 # dmesg KB size buffer size
 #dmesg_config_define=$(awk -F '=' '/CONFIG_LOG_BUF_SHIFT/ {print $2;}' /boot/config-$(uname -r))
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -26,10 +26,17 @@ ignore_str="$ignore_str"'|Parity error detected'
...
+# CML Mantis occasionally throws Intel(R) Management Engine Interface(mei) errors 
+# https://unix.stackexchange.com/questions/109294/mei-00000016-0-init-hw-failure
 ignore_str="$ignore_str"'|mei_me 0000:00:16.0: wait hw ready failed'
 ignore_str="$ignore_str"'|mei_me 0000:00:16.0: hw_start failed ret = -62'
 
+# CML Mantis has DELL touchpad i2c error on suspend/resume
+ignore_str="$ignore_str"'|i2c_designware i2c_designware.0: controller timed out'
+ignore_str="$ignore_str"'|i2c_hid i2c-DELL0955:00: failed to change power setting'
+ignore_str="$ignore_str"'|PM: dpm_run_callback(): i2c_hid_resume+0x0/0xf0 [i2c_hid] returns -110'
+ignore_str="$ignore_str"'|PM: Device i2c-DELL0955:00 failed to resume async: error -110'
```
ranj063, line 38: @fredoh9 this change looks good. But I wonder if it would be possible to make it a bit more generic may be? Say for example, if a log reports a failed/error, can we look at the device name and ignore it if it is not in the list of devices we care about?

fredoh9:
Thank you for reviewing and feedback.

> Are you sure the ME failure is totally unrelevant to SOF?
I can't imagine any relation between MEI and sound. If I miss any thing, please let me know.

fredoh9:
@marc-hb 
Tested again with/without SOF modules. I can reproduce same mei errors. So we can ignore this errors safely.

tools/sof-kernel-log-check.sh
fredoh9: > `ignore_str` is an Extended regular expression. Can you replace all the special characters (`[ ]`, `( )`, `+,`... with dots? (I'm surprised the one with brackets worked...)

@marc-hb , regarding extended regalar expression, that was good point. Special characters should be replaced in this mode. I was wondering how my test went through. It turned out that line is not required. It didn't hit out error logs hence nothing to ignore.
`
ignore_str="$ignore_str"'|PM: dpm_run_callback(): i2c_hid_resume+0x0/0xf0 [i2c_hid] returns -110'
`

fredoh9:
> `ignore_str` is an Extended regular expression. Can you replace all the special characters (`[ ]`, `( )`, `+,`... with dots? (I'm surprised the one with brackets worked...)

@marc-hb , regarding extended regalar expression, that was good point. Special characters should be replaced in this mode. I was wondering how my test went through. It turned out that line is not required. It didn't hit out error logs hence nothing to ignore.
`
ignore_str="$ignore_str"'|PM: dpm_run_callback(): i2c_hid_resume+0x0/0xf0 [i2c_hid] returns -110'
`

fredoh9:
> @fredoh9 Please correct the typo in commit title, ingore -> ignore

Thank you for catching typo. Just updated.

PR #315, opened 2020-08-05
check-fw-echo-reference.sh: remove 13 abusive 'shellcheck disable'

marc-hb:
There's no point running shellcheck and disable every warning without
any attempt to fix them or explain why they're disabled.

Most of the warnings that were disabled don't even exist!

Let's keep the 2 warnings that actually exist as reminders of real bugs
that should be fixed eventually.

This brings down the number of "shellcheck disable" for the _entire
sof-test repo_ from 17 down to 4.

### DIFF
--- a/test-case/check-fw-echo-reference.sh
+++ b/test-case/check-fw-echo-reference.sh
@@ -18,29 +18,17 @@
 rm -f /tmp/bat.wav.*
 
 # shellcheck source=case-lib/lib.sh
-# shellcheck disable=SC2016
-# shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


keqiaozhang:
Thanks @marc-hb  for your review and comments, I will address all shellcheck errors next.

marc-hb:
@keqiaozhang  the last two warnings do not seem urgent to me.

marc-hb:
Forgot this sorry: on the other hand, please add `set -e` to this new test ASAP  (#313 has been merged)

PR #316, opened 2020-08-06
test-case: refine the test script and fix the shellcheck warnings

keqiaozhang:
Refine the check-fw-echo-reference.sh script and fix some shellcheck warnings.

### DIFF
--- a/test-case/check-fw-echo-reference.sh
+++ b/test-case/check-fw-echo-reference.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
 ## Case Name: check_echo_reference
 ## Preconditions:
@@ -18,15 +20,16 @@
 rm -f /tmp/bat.wav.*
 
 # shellcheck source=case-lib/lib.sh
-source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+libdir=$(dirname "${BASH_SOURCE[0]}")
+source "$libdir"/../case-lib/lib.sh
 
-OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
-OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+OPT_OPT_lst['t']='tplg'         OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
+OPT_PARM_lst['t']=1             OPT_VALUE_lst['t']="$TPLG"
 OPT_OPT_lst['s']='sof-logger'   OPT_DESC_lst['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_PARM_lst['s']=0             OPT_VALUE_lst['s']=1
-OPT_OPT_lst['l']='loop'     OPT_DESC_lst['l']='loop count'
-OPT_PARM_lst['l']=1         OPT_VALUE_lst['l']=1
-OPT_OPT_lst['n']='frames'     OPT_DESC_lst['n']='test frames'
+OPT_OPT_lst['l']='loop'         OPT_DESC_lst['l']='loop count'
+OPT_PARM_lst['l']=1             OPT_VALUE_lst['l']=1
+OPT_OPT_lst['n']='frames'       OPT_DESC_lst['n']='test frames'
 OPT_PARM_lst['n']=1             OPT_VALUE_lst['n']=240000
 OPT_OPT_lst['f']='frequency'    OPT_DESC_lst['f']='target frequency'
 OPT_PARM_lst['f']=1             OPT_VALUE_lst['f']=997
@@ -43,23 +46,11 @@ frequency=${OPT_VALUE_lst['f']}
 func_pipeline_export $tplg "echo:any"
 func_lib_setup_kernel_last_line
 
-function __upload_wav_file
-{
-    # upload the alsabat wav file
-    for file in /tmp/bat.wav.*
-    do
-        size=$(ls -l "$file" | awk '{print $5}')
-        if [[ $size -gt 0 ]]; then
-            cp "$file" "$LOG_ROOT/"
-        fi
-    done
-}
-
 if [ "$PIPELINE_COUNT" != "2" ]; then
     die "Only detect $PIPELINE_COUNT pipeline(s) from topology, but two are needed"
 fi
 
-for idx in $(seq 0 $(("$PIPELINE_COUNT" - 1)))
+for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
 do
     type=$(func_pipeline_parse_value "$idx" type)
     if [ "$type" == "playback" ]; then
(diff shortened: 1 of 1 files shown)


test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
         fi
         # run echo reference test
         dlogc "alsabat -P $pb_dev --standalone -c $channel -f $fmt -r $rate -n $frames -F $frequency"
-        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" &
+        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" & alsabatPID=$!
```
lyakh, line 78: I'd rather have the alsabatPID assignment on a separate line

test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
...
         # playback may have low latency, add 0.5 second delay to aviod recording zero at beginning.
         sleep 1
         dlogc "alsabat -C $cp_dev -c $channel -f $fmt -r $rate -F $frequency"
-        alsabat -C "$cp_dev" -c "$channel" -f "$fmt" -r "$rate" -F "$frequency"
+        if ! alsabat -C "$cp_dev" -c "$channel" -f "$fmt" -r "$rate" -F "$frequency"; then
+            # upload failed wav files
+            find /tmp -maxdepth 1 -type f -name "bat.wav.*" -size +0 -exec cp {} "$LOG_ROOT/" \;
```
lyakh, line 84: ouch... Taking all *.wav from /tmp (up to 1 level deep) seems a bit too broad to me. Can we have a specific directory under /tmp?

test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
         fi
         # run echo reference test
         dlogc "alsabat -P $pb_dev --standalone -c $channel -f $fmt -r $rate -n $frames -F $frequency"
-        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" &
+        alsabat -P "$pb_dev" --standalone -c "$channel" -f "$fmt" -r "$rate" -n "$frames" -F "$frequency" & alsabatPID=$!
         # playback may have low latency, add 0.5 second delay to aviod recording zero at beginning.
         sleep 1
```
marc-hb, line 80: The comment says 0.5 but the code says 1.
Did you mean "high" latency?

test-case/check-fw-echo-reference.sh
marc-hb: `mktemp` maybe?

test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
         fi
```
marc-hb, line 75: Shouldn't this print a warning?

test-case/check-fw-echo-reference.sh
keqiaozhang: The directory that used to store alsabat wav files is defined by the alsabat tool,  all alsabat generated wav files are stored in /tmp with format as bat.wav.xxxxxx, "“xxxxxx”  are random codes.

test-case/check-fw-echo-reference.sh
keqiaozhang: it should be 1 second delay.

test-case/check-fw-echo-reference.sh
keqiaozhang: fixed.

test-case/check-fw-echo-reference.sh
keqiaozhang: updated.

test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
...
+            # upload failed wav files
+            find /tmp -maxdepth 1 -type f -name "bat.wav.*" -size +0 -exec cp {} "$LOG_ROOT/" \;
+            exit 1
+        fi
 
-        # upload failed wav file
-        if [[ "$?" != "0" ]]; then
-            __upload_wav_file
-        exit 1
+        if ! timeout 2 bash -c wait $alsabatPID; then
```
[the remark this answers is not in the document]
keqiaozhang: I tried timeout 2 wait $pid directly, but failed with "timeout: failed to run command ‘wait’: No such file or directory"
but looks like with bash -c, it can only fix the error, but it cannot return error code when $pid is timed out.
so I changed the way to check it.

test-case/check-fw-echo-reference.sh
```diff
@@ -84,17 +75,20 @@ do
...
+            kill -9 $alsabatPID
```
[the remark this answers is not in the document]
keqiaozhang: updated.

test-case/check-fw-echo-reference.sh
lyakh: @marc-hb no, just because it's a new command, don't remember ever seeing `cmd1 & cmd2` in one line in shell. I understand trying to keep `$!` directly attached to the previous command to avoid them being ever separated by accident, so, I'm not strongly against this, just looks weird to me.

test-case/check-fw-echo-reference.sh
lyakh: @keqiaozhang oh, have I misread the name pattern? If it's `bat.wav.*` then it isn't that bad any more... But we could also use `--saveplay=` to specify a filename and maybe a directory.

keqiaozhang: @lyakh , right, we can use --saveplay to specify a filename even with a directly, but this is only for playback, so in order to unify the processing of playback and recorded files, I think the default name pattern is easy to handle.

test-case/check-fw-echo-reference.sh
marc-hb: BTW I just checked the documentation about this (because of some unrelated review): `&` and `;` have the same precedence.

test-case/check-fw-echo-reference.sh
```diff
@@ -80,21 +71,24 @@ do
...
         sleep 1
         dlogc "alsabat -C $cp_dev -c $channel -f $fmt -r $rate -F $frequency"
-        alsabat -C "$cp_dev" -c "$channel" -f "$fmt" -r "$rate" -F "$frequency"
+        alsabat -C "$cp_dev" -c "$channel" -f "$fmt" -r "$rate" -F "$frequency" || {
+            # upload failed wav files
+            find /tmp -maxdepth 1 -type f -name "bat.wav.*" -size +0 -exec cp {} "$LOG_ROOT/" \;
+            die "alsabat test failed on pcm: $cp_dev"
+        }
+
+        timeout 5 tail --pid=$alsabatPID -f /dev/null || die "Failed to stop alsabat playback"
```
marc-hb, line 90: Sorry I gave you some wrong advice earlier: `timeout wait $pid` does not work because `wait` is a built-in that  works only for direct child processes. `tail --pid` is an interesting idea but it doesn't not collect the exit status/error code of the alsabat command which is the goal here so `tail --pid` doesn't not solve the problem. This works:

```shell
timeout -k 10 5 alsabat -P "$pb_dev" -... -F "$frequency" & timedAlsabatPID=$!
...
...
wait $timedAlsabatPID || die "alsabat failed..."
```
Unlike my previous advice I tested this logic and it does fail when alsabat fail.

keqiaozhang: Thanks @marc-hb  for your suggestion, I have updated the code and did some fine-tuning of the time.

Issue #308, opened 2020-07-31
TPLG file not well handled


plbossart:
unclear to me why some tests fail because of topology:

````
root@plb-olympic:~/sof-test/test-case# export TPLG=/lib/firmware/intel/sof-tplg/sof-cml-rt711-rt1308-mono-rt715.tplg 
root@plb-olympic:~/sof-test/test-case# ./run-all-tests.sh 
 ---------- 
 ---------- 
 starting test_firmware-presence 
2020-07-31 21:41:05 UTC Sub-Test: [INFO] Checking file: /lib/firmware/intel/sof/sof-cml.ri
2020-07-31 21:41:05 UTC Sub-Test: [INFO] Found file: /lib/firmware/intel/sof/sof-cml.ri 45239d45304afe791de8075a48feb95c
2020-07-31 21:41:05 UTC Sub-Test: [INFO] Test Result: PASS!
 ---------- 
 ---------- 
 starting test_firmware-load 
2020-07-31 21:41:08 UTC Sub-Test: [INFO] Checking SOF Firmware load info in kernel log
[   21.035180] sof-audio-pci 0000:00:1f.3: Firmware info: version 0:0:0-69d92
[   21.035184] sof-audio-pci 0000:00:1f.3: Firmware: ABI 3:15:0 Kernel ABI 3:17:0
--
[   21.035478] sof-audio-pci 0000:00:1f.3: Firmware info: used compiler XCC 12:0:8 <RG-2017.8-linux> used optimization flags O2
[   21.035532] sof-audio-pci 0000:00:1f.3: found ext header type 3 size 0x1c
2020-07-31 21:41:08 UTC Sub-Test: [INFO] Test Result: PASS!
 ---------- 
 ---------- 
 starting test_tplg-binary 
2020-07-31 21:41:12 UTC Sub-Test: [ERROR] No available topology for this test case
2020-07-31 21:41:12 UTC Sub-Test: [INFO] Test Result: FAIL!
 ---------- 
 ---------- 
 starting test_pcm_list 
2020-07-31 21:41:15 UTC Sub-Test: [ERROR] No available topology for this test case
2020-07-31 21:41:15 UTC Sub-Test: [INFO] Test Result: FAIL!
 ---------- 
 ---------- 
 [...]
 ---------- 
 starting test_ipc-flood 
2020-07-31 21:41:23 UTC Sub-Test: [WARNING] ./check-ipc-flood.sh need debug version firmware
[...]

PR #322, opened 2020-08-07
test-case: fix TPLG path issue and refine run-all-tests

aiChaoSONG:
If user specifies TPLG in env, we will get a wrong
topology path in run-all-tests.sh. So test case which
requires topology path will fail.

This patch adds usage and command option parser, and
fix TPLG path issue by checking and reexporting TPLG env.

Fixes: #308

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -1,8 +1,4 @@
 #!/bin/bash
-TPLG=$1
-Time=3
-export TPLG=/lib/firmware/intel/sof-tplg/$TPLG
-
 set -e
 
 testlist="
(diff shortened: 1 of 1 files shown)


test-case/run-all-tests.sh
```diff
@@ -39,15 +35,27 @@ suspend-resume-with-capture"
...
+	while getopts "ht:T:" OPTION; do
+	case "$OPTION" in
+		t) tplg_name="$OPTARG" ;;
+		T) time_delay="$OPTARG" ;;
+		*) usage; exit 1 ;;
+		esac
+	done
+	if [ -z "$TPLG" ]; then
+		export TPLG="/lib/firmware/intel/sof-tplg/$tplg_name"
+	fi
```
marc-hb, line 50: I think this gives the environment precedence over the (more specific) command line parameter, so it's not OK.

Do we really need this `-t` option? I mean do we really need to do anything with TPLG in this file? I think all other tests use the $TPLG environment variable, correct? Why not just do nothing here and stay consistent with all other tests?

BTW why not pass-through "$@" to every test?

aiChaoSONG: OK, I can remove the -t option, and just use TPLG to pass through topology path to all test case. We can not use  "$@" here, because every single test case takes different parameters.

marc-hb: > We can not use "$@" here, because every single test case takes different parameters.

I've seen some options that _seem_ to be present in every test (`git grep OPT_`). I admittedly haven't checked whether some options actually exist in every single test.

PR #323, opened 2020-08-07
hijack.sh: print a call trace on every failure

marc-hb:
Example:
```
06:03:34 UTC [WARNING] No sof-logger found
06:03:34 UTC [INFO] no source file, use /dev/zero as dummy playback source
06:03:34 UTC [ERROR] No available topology for pipeline export
06:03:34 UTC [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
06:03:34 UTC [ERROR]    die()  @  ./test-case/../case-lib/lib.sh
06:03:34 UTC [ERROR]    func_pipeline_export()  @  sof-test/case-lib/pipeline.sh:11
06:03:34 UTC [ERROR]    main()  @  ./test-case/check-playback.sh:71
06:03:34 UTC [WARNING] No sof-logger found
06:03:35 UTC [ERROR] sof-logger was already dead
06:03:36 UTC [INFO] Test Result: FAIL!
```

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -8,6 +8,21 @@ function func_exit_handler()
 {
     local exit_status=${1:-0}
 
+    # call trace
+    if [ "$exit_status" -ne 0 ] ; then
+        dloge "Starting ${FUNCNAME[0]}(), exit status=$exit_status, FUNCNAME stack:"
+        local i line_no
+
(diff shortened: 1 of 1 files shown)


marc-hb:
Thanks @ranj063 . I would normally wait longer and for more than just one review but this is very low-risk, short and easy to revert if needed, heavily tested locally and very high value.

PR #327, opened 2020-08-08
verify-sof-firmware-load: stop using dmesg; ring buffers can expire

marc-hb:
Switch to journalctl --dmesg which is slower because it gets the log
from boot which means it works.

Also remove the useless and confusing eval and the useless [ ]
test. Switch one grep to grep -q to avoid generating a massive string
and cap the other two to just 12 lines (enough to see some repetition if
any)

This was tricky to reproduce because it required some indirection script
like run-all-tests.sh.

Good reminder to gradually get rid of dmesg everywhere.

### DIFF
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -3,9 +3,9 @@
 ##
 ## Case Name: verify-sof-firmware-load
 ## Description:
-##    Check if the SOF fw loaded successfully in dmesg
+##    Check if the SOF fw loaded successfully in the kernel logs
 ## Case step:
-##    Check dmesg to search fw load info
+##    Check kernel logs to search fw load info
 ## Expect result:
 ##    Get fw version info in dmesg
 ##    sof-audio-pci 0000:00:0e.0: Firmware info: version 1:1:0-e5fe2
@@ -22,15 +22,15 @@ if [ ! "$(alias |grep 'Sub-Test')" ];then
     DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
     cmd="sof-kernel-dump.sh"
 else
-    cmd="dmesg"
+    cmd="journalctl --dmesg"
 fi
 
 dlogi "Checking SOF Firmware load info in kernel log"
-if [[ $(eval $cmd | grep "] sof-audio.*version") ]]; then
+if $cmd | grep -q " sof-audio.*version"; then
     # dump the version info and ABI info
-    eval $cmd | grep "Firmware info" -A1
+    $cmd | grep "Firmware info" -A1 | head -n 12
     # dump the debug info
-    eval $cmd | grep "Firmware debug build" -A3
+    $cmd | grep "Firmware debug build" -A3 | head -n 12
     exit 0
 else
     die "Cannot find the sof audio version"


test-case/verify-sof-firmware-load.sh
```diff
@@ -22,15 +22,15 @@ if [ ! "$(alias |grep 'Sub-Test')" ];then
...
 
 dlogi "Checking SOF Firmware load info in kernel log"
-if [[ $(eval $cmd | grep "] sof-audio.*version") ]]; then
+if $cmd | grep -q " sof-audio.*version"; then
     # dump the version info and ABI info
-    eval $cmd | grep "Firmware info" -A1
+    $cmd | grep "Firmware info" -A1 | head -n 12
     # dump the debug info
-    eval $cmd | grep "Firmware debug build" -A3
+    $cmd | grep "Firmware debug build" -A3 | head -n 12
```
lyakh, line 33: if we're already fixing this and since you say yourself, that `journalctl` is slower, maybe we could change this to call it only once?

marc-hb: I considered this. It would mean using `grep -A3` for `Firmware info` too which seems OK after all, here's a sample output with `-A3`:
```
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: Firmware info: version 1:5:0-d9431
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: Firmware: ABI 3:16:0 Kernel ABI 3:16:0
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: found ext header type 1 size 0xb8
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: found ext header type 2 size 0x5c
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: Firmware info: used compiler XCC 12:0:8 xtensa-apl-elf used optimization flags O2
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: found ext header type 3 size 0x1c
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: warning: unknown ext header type 3 size 0x1c
Aug 07 22:43:25 up2 kernel: sof-audio-pci 0000:00:0e.0: found ext header type 4 size 0x10
```
> you say yourself, that journalctl is slower

On my Up2 board `journalctl --dmesg` extracts approximately 60,000 `--dmesg` kernel lines per second so it would likely not affect CI for instance because the CI systems reboot frequently.

marc-hb: I just tested and verified that `grep -q` is smart enough to exit as soon as possible; so the first `$cmd` invocation is now much faster, even when calling `journalctl` it is very fast.

I will regroup the last two `$cmd` calls  into one.

marc-hb: > I will regroup the last two $cmd calls into one.

Actually I can't because I want the first `head -n 12` lines of `Firmware info` on one hand and the first 12 lines of `Firmware debug build` on the other hand and they could be in different places so I really need two independent `grep` commands.

Now I could extract the entire `--dmesg` to a temporary file and then `grep` that temporary file twice, however:
- the code complexity would increase significantly
- in my (extreme) example with `240000` lines of debug a temporary file is actually a bit _slower_ because it loses the benefit of `head -n12` that interrupts `journalctl` before it finishes and saves time.

PR #328, opened 2020-08-10
run-all-tests.sh: add $mydir to allow invoking from anywhere

marc-hb:
Use pwd to resolve relative paths that can be broken by symbolic links.

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -34,6 +34,9 @@ suspend-resume-with-capture"
 
 main()
 {
+	local mydir
+	mydir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
+
 	local failures=()
 	local passed=()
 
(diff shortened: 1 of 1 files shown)


marc-hb:
>  is it better to use name like: cwd, work_dir, current_dir?

It's the location of the script so it's neither of those.

plbossart:
see alternate solution at https://github.com/plbossart/sof-test/tree/fix/run-all-test-path

marc-hb:
> see alternate solution at https://github.com/plbossart/sof-test/tree/fix/run-all-test-path

More specifically https://github.com/plbossart/sof-test/commit/61a50a5d73b42745b

Thanks Pierre, do you want me to resubmit with some of your code or are you going to resubmit yourself?

[ ] I can change `$0` to `BASH_SOURCE[0]`. I don't expect this script itself to be included any time soon but `BASH_SOURCE[0]` can't hurt. The `pwd` resolves relative paths, this helps when using some symbolic links. I'll keep it.
[ ] I can rename $mydir to $TEST_DIR (with quotes)

aiChaoSONG:
@marc-hb I don't think Pierre will send his patch, could you please continue this?

marc-hb:
> @marc-hb I don't think Pierre will send his patch, could you please continue this?

I changed `$0` to `${BASH_SOURCE[0]}`.

I didn't rename `$mydir` to `$TEST_DIR` because the name `$TEST_DIR` is better but IMHO not so much better that it's worth my time rebasing my other, local and conflicting `run-all-tests.sh` work in progress.

aiChaoSONG (approved):
Thanks, LGTM

PR #337, opened 2020-08-13
[DO NOT MERGE]test-case: smart amplifier case workaround to run on master FW

aiChaoSONG:
The smart amplifier case can only run under some
special sequence of aplay/arecord with delay.

This workaround is used to cope with the latest
release, after the release, this workaround should
be reverted or the delay between aplay and arecord
should be removed.

### DIFF
--- a/test-case/check-smart-amplifier.sh
+++ b/test-case/check-smart-amplifier.sh
@@ -98,14 +98,16 @@ do
         file="$tmp_dir/tmp_wave_${fmt%_*}.wav"
         recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
         wavetool.py -gsine -A0.8 -B"${fmt%_*}" -o"$file"
-        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file &"
-        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
-        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
-        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file"
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -v -q $recorded_file &"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file" &
+        sleep 0.5
+        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -v -q $file"
+        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file"
         dlogi "Comparing recorded wave and reference wave"
         wavetool.py -a"smart_amp" -R"$recorded_file" -r"$file" || die "wavetool.py exit with $?"
         # clean up generated wave files
         rm -rf "$file" "$recorded_file"
+        sleep 2
     done
 done
 


test-case/check-smart-amplifier.sh
```diff
@@ -98,14 +98,16 @@ do
...
         recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
         wavetool.py -gsine -A0.8 -B"${fmt%_*}" -o"$file"
-        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file &"
-        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
-        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
-        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file"
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -v -q $recorded_file &"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file" &
+        sleep 0.5
+        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -v -q $file"
```
dbaluta, line 104: @aiChaoSONG any idea why the 0.5 sleep? What happens without it?

aiChaoSONG: There is issue with DSM feature on master firmware, without the delay, we will get IPC timeout, and the firmware will not response anymore, unless reboot. In fact, 0.5 is an experiment value, 1 second should also work.

paulstelian97: Hm so similar issue with us except in our case it's just regular playback, two aplay calls in a row. Any half-idea on what could cause that?

aiChaoSONG: @paulstelian97  add comment here may get more response, https://github.com/thesofproject/sof/issues/3249

paulstelian97: Hm issue does seem to differ then. In our case it actually doesn't IPC timeout but timeout in the actual processing flow. Guess I'll have to look more into the issue on i.MX separately.

aiChaoSONG: @paulstelian97 If corresponding SSP is not enabled in BIOS, aplay/arecord will fail either. I am not familiar with ARM, but you should have the same setting in some place.

paulstelian97: Yeah again, that's not the nature of the issue (first aplay succeeds, the second one times out). I'll check it out because we suspect the DMA driver isn't perfect.

test-case/check-smart-amplifier.sh
```diff
@@ -98,14 +98,16 @@ do
         file="$tmp_dir/tmp_wave_${fmt%_*}.wav"
         recorded_file="$tmp_dir/recorded_tmp_wave_${fmt%_*}.wav"
         wavetool.py -gsine -A0.8 -B"${fmt%_*}" -o"$file"
-        dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -q $file &"
-        aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
-        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -q $recorded_file"
-        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file"
+        dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -v -q $recorded_file &"
+        arecord -D"$cp_dev" -r "$cp_rate" -c "$cp_chan" -f "$fmt" -d "$duration" -v -q "$recorded_file" &
```
marc-hb, line 102: To make sure no error is left behind ("green failures"), `wait` for all background commands at the end which collects their status. Especially commands like `arecord` or `aplay`; we _really_ don't want to miss any error from ALSA commands because these are audio tests! See example in recent commit fd545aefac3a:
```shell
timeout -k 3 10 some_backgrounded_command & somePID=$!
... test ... test .... test
wait $somePID || die "alsa_blabla failed"
```

aiChaoSONG: @marc-hb Yes, We are hiding the bug for release test with this case next week. But seems FW team find the root cause of the bug, so we don't need the workaround. I will come back later to apply your suggestion, as WOV case is first priority.

PR #339, opened 2020-08-13
wavetool: don't pass cmd line parameters to function

aiChaoSONG:
The command line variable is made global in 4f989221,
we don't need to pass any sub-parameters from command
line to functions, we can use them directly.


aiChaoSONG:
@marc-hb we may need this to be merged ASAP to continue WOV test case, or if you think this is bad idea, we can close this PR.

PR #341, opened 2020-08-17
test-case: add keyword detection test support

keqiaozhang:
add one new test case to check the keyword detection and use
the wavetool to do the analysis.

Signed-off-by: Zhang Keqiao <keqiao.zhang@linux.intel.com>

PR #350, opened 2020-08-25
test-case: fix the missing separator in config blob and remove the unused parameter

keqiaozhang:
The comma separator is missing when updating the config
blob, which will cause errors when writing it back.

### DIFF
--- a/test-case/check-keyword-detection.sh
+++ b/test-case/check-keyword-detection.sh
@@ -113,7 +113,8 @@ _update_blob(){
     [ $preamble_time -ge $history_depth ] || {
         die "Warning: invalid arguments, preamble_time must be greater than or equal to history_depth"
     }
-    awk -F "," '{if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; $7='"$history_depth"'}1' "$def_blob" > "$new_blob"
+    awk -F, -v OFS=, '{if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; $7='"$history_depth"'}1' \
+        "$def_blob" > "$new_blob"
     dlogi "kwd config blob is updated to:"
     cat "$new_blob"
     dlogi "write back the new kwd config blob"
--- a/test-case/check-keyword-detection.sh
+++ b/test-case/check-keyword-detection.sh
@@ -37,9 +37,6 @@ OPT_PARM_lst['s']=1                  OPT_VALUE_lst['s']=2100
 OPT_OPT_lst['b']='buffer'            OPT_DESC_lst['b']='buffer size'
 OPT_PARM_lst['b']=1                  OPT_VALUE_lst['b']=67200
 
-OPT_OPT_lst['f']='frequency'         OPT_DESC_lst['f']='target frequency of sine wav'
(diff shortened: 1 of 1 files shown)


test-case/check-keyword-detection.sh
```diff
@@ -112,7 +112,8 @@ _update_blob(){
     [ $preamble_time -ge $history_depth ] || {
         die "Warning: invalid arguments, preamble_time must be greater than or equal to history_depth"
     }
-    awk -F "," '{if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; $7='"$history_depth"'}1' "$def_blob" > "$new_blob"
+    awk 'BEGIN{FS=OFS=","} {if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; \
+        $7='"$history_depth"'}1' "$def_blob" > "$new_blob"
```
marc-hb, line 116: To reduce the amount of quotation needed and make the code slightly easier to read:
```shell
awk -F, -v OFS=, '{ if ...
```

marc-hb:
All but one shellcheck warning are fixed by trivial two-lines commit #352, please help review and merge #352 and rebase on top of it.

test-case/check-keyword-detection.sh
keqiaozhang: Updated.

marc-hb:
> Indeed it looks like this cannot work without OFS. How was #341 tested?

The shellcheck warning at line 115 seems valid https://travis-ci.org/github/thesofproject/sof-test/builds/721549726

How was this awk script tested now?

keqiaozhang:
> > Indeed it looks like this cannot work without OFS. How was #341 tested?
> 
> The shellcheck warning at line 115 seems valid https://travis-ci.org/github/thesofproject/sof-test/builds/721549726
> 
> How was this awk script tested now?

shellcheck warning is fixed and tested this script on TGL HDA.
I missed this fix in the first version of https://github.com/thesofproject/sof-test/pull/341, then I took 4 days leave, After I came back,  PR341 has been merged.

marc-hb:
There is still a "frequency unused" warning. Is it really unused?

> tested this script on TGL HDA.
> I missed this fix in the first version of #341,

How is your latest testing of this different from the previous testing of #341? Your previous testing did apparently not run this awk code. How/why does the new testing run this awk code now?

Untested awk code was already merged once. Let's not merge untested awk code again.

keqiaozhang:
> How is your latest testing of this different from the previous testing of #341? Your previous testing did apparently not run this awk code. How/why does the new testing run this awk code now?
> 
> Untested awk code was already merged once. Let's not merge untested awk code again.
> 
> I'm concerned this awk code was still not tested because the backslash warning at line 115 reported by https://travis-ci.org/github/thesofproject/sof-test/builds/721549726 looks like it is still not tested.

This awk code is used to replace the history_depth and preamble_time with different values in wov config blob.  In this script, we can specify these two options by using -p and -s  parameters, during the test, this script will use awk code to replace these 2 options and print out the updated config blob in line 119 and write it back.

I tuned wov case on TGL-HDA and submitted https://github.com/thesofproject/sof-test/pull/341 on my development machine,  but I forgot to synchronize the code before I submitted PR341. That's why this PR is here.
Now I have verified this script on both TGL-HDA and JSL-HDA and this awk code is in effect.

PR #352, opened 2020-08-26
check-keyword-detection.sh: fix 5 trivial shellcheck warnings

marc-hb:
The only warning left 'frequency appears unused' seems legit.


marc-hb:
There's no actual failure from jenkins, only NA (Not Available or Applicable) and SKIPs.

aiChaoSONG (approved):
Looks good

PR #354, opened 2020-08-27
Fix error filters and use journaltcl

plbossart:
Remove useless filters, add new ones for Dell CML-U

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -38,10 +38,9 @@ if [ ! "$SOFCARD" ]; then
     SOFCARD=$(grep '\]: sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
 fi
 
-func_lib_setup_kernel_last_line()
+func_lib_setup_kernel_last_timestamp()
 {
-    # shellcheck disable=SC2034 # external script will use it
-    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S.%6N")
 }
 
 SOF_LOG_COLLECT=0
--- a/test-case/check-ipc-flood.sh
+++ b/test-case/check-ipc-flood.sh
@@ -32,15 +32,15 @@ loop_cnt=${OPT_VALUE_lst['l']}
 
 [[ ! "$(sof-kernel-dump.sh|grep 'sof-audio'|grep 'Firmware debug build')" ]] && dlogw "${BASH_SOURCE[0]} need debug version firmware" && exit 2
 
-func_lib_setup_kernel_last_line
+func_lib_setup_kernel_last_timestamp
 func_lib_check_sudo
 
 dlogi "Check sof debug fs environment"
 [[ "$(sudo file $ipc_flood_dfs|grep 'No such file')" ]] && dlogw "${BASH_SOURCE[0]} need $ipc_flood_dfs to run the test case" && exit 2
 dlogi "Checking ipc flood test!"
 
-# cleanup dmesg buffer before test
-sudo dmesg -c > /dev/null
+# discard old kernel logs
+func_lib_setup_kernel_last_timestamp
 
 for i in $(seq 1 $loop_cnt)
 do
@@ -48,11 +48,12 @@ do
     dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
     sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-    sof-kernel-log-check.sh 0 || die "Catched error in dmesg"
+    sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP" || die "Catched error in journalctl"
 
     dlogi "Dumping test logs!"
-    dmesg | grep "IPC Flood count" -A 2
+    journalctl -k --since="$KERNEL_LAST_TIMESTAMP" | grep "IPC Flood count" -A 2
 done
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE
+sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP"
+
 exit 0
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -67,7 +67,7 @@ else
     dlogi "using $file as playback source"
 fi
 
-func_lib_setup_kernel_last_line
+func_lib_setup_kernel_last_timestamp
 func_lib_check_sudo
 func_pipeline_export $tplg "type:playback & ${OPT_VALUE_lst['S']}"
 
@@ -86,8 +86,9 @@ do
         if [ ${OPT_VALUE_lst['F']} = '1' ]; then
             fmt=$(func_pipeline_parse_value $idx fmts)
         fi
-        # clean up dmesg
-        sudo dmesg -C
+        # discard old kernel logs
+	func_lib_setup_kernel_last_timestamp
+
         for fmt_elem in $(echo $fmt)
         do
             for i in $(seq 1 $loop_cnt)
@@ -104,5 +105,5 @@ do
     done
 done
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE
+sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP"
 exit $?
--- a/test-case/multiple-pipeline-capture.sh
+++ b/test-case/multiple-pipeline-capture.sh
@@ -50,7 +50,7 @@ tplg=${OPT_VALUE_lst['t']}
 max_count=0
 # get the min value of TPLG:'pipeline count' with Case:'pipeline count'
 [[ $PIPELINE_COUNT -gt ${OPT_VALUE_lst['c']} ]] && max_count=${OPT_VALUE_lst['c']} || max_count=$PIPELINE_COUNT
-func_lib_setup_kernel_last_line
+func_lib_setup_kernel_last_timestamp
 
 # now small function define
 declare -A APP_LST DEV_LST
@@ -102,8 +102,8 @@ func_error_exit()
 for i in $(seq 1 $loop_cnt)
 do
     dlogi "===== Testing: (Loop: $i/$loop_cnt) ====="
-    # clean up dmesg
-    sudo dmesg -C
+    # discard old kernel logs
+    func_lib_setup_kernel_last_timestamp
 
     # start capture:
     func_run_pipeline_with_type "capture"
@@ -150,8 +150,8 @@ do
     pkill -9 arecord
     pkill -9 aplay
 
-    sof-kernel-log-check.sh 0 || die "Catch error in dmesg"
+    sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP" || die "Catch error in journalctl -k"
 done
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE >/dev/null
+sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP" >/dev/null
 exit $?
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -51,13 +51,6 @@ ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. R
 ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
 
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
-# dmesg KB size buffer size
-#dmesg_config_define=$(awk -F '=' '/CONFIG_LOG_BUF_SHIFT/ {print $2;}' /boot/config-$(uname -r))
-#dmesg_buffer_size=$( echo $(( (1<<$dmesg_config_define) / 1024 )) )
-# kernel file log buffer size
-#kernel_buffer_size=$(du -k /var/log/kern.log |awk '{print $1;}')
-# now decide using which to catch the kernel log
-#[[ $kernel_buffer_size -lt $dmesg ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
 
 # confirm begin_line is number, if it is not the number, direct using dmesg
 [[ "${begin_line//[0-9]/}" ]] && begin_line=0
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -4,19 +4,9 @@ begin_line=${1:-1}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
 
-# There will be debug logs at each failed initializaiton of DSP:
-#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
-#   sof-audio-pci 0000:00:1f.3: error: Error code=0xffffffff: FW status=0xffffffff
-#   sof-audio-pci 0000:00:1f.3: error: iteration 0 of Core En/ROM load failed: -110
-# We will reinit DSP if it is failed to init, and retry 3 times, so the errors in
-# debug logs at the frist and second retry can be ignored.
-# Check https://github.com/thesofproject/linux/pull/1676 for more information.
-# TODO explain for "error: Error ..."
-ignore_str='error: iteration [01]'
-ignore_str="$ignore_str"'|error: status'
-
 # TODO explain
-ignore_str="$ignore_str"'|error: debugfs write failed to idle -16'
+# The first string cannot start by |
+ignore_str='error: debugfs write failed to idle -16'
 
 # Realtek codecs thrown an error on startup, checking with Realtek
 # Possible fix - https://github.com/thesofproject/linux/pull/1984
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-begin_line=${1:-1}
+begin_timestamp=${1:-0}
 declare err_str ignore_str project_key
 err_str="error|failed|timed out|panic|oops"
 
@@ -52,22 +52,34 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
 
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 
-# confirm begin_line is number, if it is not the number, direct using dmesg
-[[ "${begin_line//[0-9]/}" ]] && begin_line=0
-[[ "$begin_line" -eq 0 ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
-
-#echo "run $0 with parameter '$*' for check kernel message error"
-
-if [ "$ignore_str" ]; then
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep -E "$err_str"|grep -vE "$ignore_str")
+# confirm begin_timestamp is valid, if it is not the number, search full log
+re='^[0-9]+$'
+if [[ $begin_timestamp =~ $re ]] ; then
+    cmd="journalctl -k"
 else
-    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep -E "$err_str")
+    cmd="journalctl -k --since='$begin_timestamp'"
 fi
 
+# Handle warnings first
+if [ "$ignore_str" ]; then
+    err=$(eval $cmd --priority=warning | grep -E "$err_str" | grep -vE "$ignore_str")
+else
+    err=$(eval $cmd --priority=warning | grep -E "$err_str")
+fi
 if [ "$err" ]; then
-    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught dmesg error"
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught journalctl -k error"
     echo "===========================>>"
     echo "$err"
+    echo "<<==========================="
+    builtin exit 1
+fi
+
+# Handle Call Trace separately
+err_trace=$(eval $cmd | grep 'Call Trace' -A5 -B3)
+if [ "$err_trace" ]; then
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught trace call"
+    echo "===========================>>"
+    echo "$err_trace"
     echo "<<==========================="
     builtin exit 1
 fi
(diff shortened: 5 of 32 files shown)


plbossart:
> Can you please add the bug numbers in the code itself?

Done

tools/sof-kernel-log-check.sh
```diff
@@ -40,6 +25,30 @@ ignore_str="$ignore_str"'|PM: Device i2c-DELL0955:00 failed to resume async: err
 # https://github.com/thesofproject/sof-test/issues/348
 ignore_str="$ignore_str"'|da7219 i2c-DLGS7219:00: SRM failed to lock'
 
+# Dell CML-U laptop with SoundWire, issues reported by sof-test
+# https://github.com/thesofproject/sof-test/issues/343
+ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send(): error -5'
+ignore_str="$ignore_str"'|platform regulatory.0: Direct firmware load for regulatory.db failed with error -2'
+ignore_str="$ignore_str"'|cfg80211: failed to load regulatory.db'
+ignore_str="$ignore_str"'|EXT4-fs (nvme0n1p6): re-mounted. Opts: errors=remount-ro'
```
marc-hb, line 33: Are you sure this works? Parentheses `( )` are special characters in Extended regular expressions:

```shell
printf '%s' 'h(ell)o' | grep   'h(ell)o'
h(ell)o # match OK
printf '%s' 'h(ell)o' | grep -E 'h(ell)o'
# nothing, no match!
```
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04

If lines with parentheses fail to filter as I'm afraid they do, then you can simply replace them with dots `.`

case-lib/lib.sh
```diff
@@ -38,10 +38,9 @@ if [ ! "$SOFCARD" ]; then
     SOFCARD=$(grep '\]: sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
 fi
 
-func_lib_setup_kernel_last_line()
+func_lib_setup_kernel_last_timestamp()
 {
-    # shellcheck disable=SC2034 # external script will use it
-    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_TIMESTAMP=date "+%Y-%m-%d %H:%M:%S.%6N"
```
aiChaoSONG, line 43: I think you may want `KERNEL_LAST_TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S.%6N")`

tools/sof-kernel-log-check.sh
plbossart: it doesn't, we need to use escape characters. fixed now.

case-lib/lib.sh
plbossart: if you say so, fixed now.

plbossart:
I will need some help here, I am completely out of my depth and I don't have a clue how these scripts are supposed to work. It's just horrible to even look at this code, sorry. 
@marc-hb @aiChaoSONG @xiulipan please be mindful of folks like me who are users and don't have any plans to learn advanced bash. CI tests work and run-all-tests.sh doesn't.

````
/root/sof-test/tools/sof-kernel-log-check.sh: line 3: 1: "NONE": syntax error: operand expected (error token is ""NONE"")
Failed to parse timestamp: 
Failed to parse timestamp: 
````

**Comment by marc-hb** [comment]
Not many people have a clue how these possibly buggy scripts work, and some of these people are gone. Switching to journalctl and error levels is awesome but it's not so easy to implement and very difficult to test because it requires injecting some errors manually. This (past) mistake caught by @aiChaoSONG  (thx!) tells me that testing didn't happen yet:
```shell
KERNEL_LAST_TIMESTAMP=date "+%Y-%m-%d %H:%M:%S.%6N"
```
So I don't think the switch to journalctl and error levels should be combined with unrelated changes and refactorings in the same PR and - hopefully no offence - I think it should be performed by someone who is (unfortunately) willing to "learn some advanced bash" or has already done so.

Fred said he already started to work on this, can we please wait until he's back from vacation next week and keep filtering "as usual" in the mean time? I already offered to help him any way I can.

A middle-ground is maybe to simply append a `--level` option to the existing `dmesg` and not perform any major surgery.

tools/sof-kernel-log-check.sh
```diff
@@ -1,64 +1,70 @@
 #!/bin/bash
 
-begin_line=${1:-1}
+begin_timestamp=${1:"NONE"}
```
xiulipan, line 3: @plbossart I think this is some bash issue. It should be used as `${[number]:-default}`
Reference https://unix.stackexchange.com/questions/338146/bash-defining-variables-with-var-number-default
```suggestion
begin_timestamp=${1:-NONE}
```

tools/sof-kernel-log-check.sh
```diff
@@ -1,64 +1,70 @@
...
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-56\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-55\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-54\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-53\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-52\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-51\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-50\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-49\.ucode failed with error -2'
+ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwl-debug-yoyo\.bin failed with error -2'
```
xiulipan, line 45: Do we really need `\.` as we already use '' to quote the string?

tools/sof-kernel-log-check.sh
```diff
@@ -1,64 +1,70 @@
...
-#dmesg_buffer_size=$( echo $(( (1<<$dmesg_config_define) / 1024 )) )
-# kernel file log buffer size
-#kernel_buffer_size=$(du -k /var/log/kern.log |awk '{print $1;}')
-# now decide using which to catch the kernel log
-#[[ $kernel_buffer_size -lt $dmesg ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
 
-# confirm begin_line is number, if it is not the number, direct using dmesg
-[[ "${begin_line//[0-9]/}" ]] && begin_line=0
-[[ "$begin_line" -eq 0 ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
+# confirm begin_timestamp is number, if it is not the number, direct using dmesg
```
xiulipan, line 55: The comment should be update
```suggestion
# confirm begin_timestamp is valid, if it is not, search full log
```

tools/sof-kernel-log-check.sh
```diff
@@ -1,64 +1,70 @@
...
+if [[ "$begin_timestamp" == "NONE" ]]; then
+    cmd="journalctl -k "
```
xiulipan, line 57:
```suggestion
    cmd="journalctl -k"
```

xiulipan (commented):
Thanks Pierre to implement the journalctl feature. Comments for fix inline.

tools/sof-kernel-log-check.sh
plbossart: ok, thanks for the suggestion, will give it a try.

tools/sof-kernel-log-check.sh
plbossart: yes. the strings are used as regular expressions, so we cannot have things like ., ), (

tools/sof-kernel-log-check.sh
plbossart: applied as is.

tools/sof-kernel-log-check.sh
plbossart: applied as suggested.

tools/sof-kernel-log-check.sh
plbossart: I missed the fact that the code already uses 0 as a possible parameter, so the default should also be zero here and we use the entire log.

tools/sof-kernel-log-check.sh
```diff
@@ -1,72 +1,120 @@
...
-[[ "${begin_line//[0-9]/}" ]] && begin_line=0
-[[ "$begin_line" -eq 0 ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
+
+# confirm begin_timestamp is valid, if it is not the number, search full log
+re='^[0-9]+$'
+if [[ $begin_timestamp =~ $re ]] ; then
+    echo "using all journal"
+    cmd='journalctl -k'
+else
+    cmd='journalctl -k --since=$begin_timestamp'
```
xiulipan, line 90: I got the issue now. begin_timestamp has space inside. So we need to use quote for this args variable
```suggestion
    cmd="journalctl -k --since='$begin_timestamp'"
```

some simple test shared.
```
LASTTIME="2020-09-02 11:46:57.075394"
pxl@pxl-UP2:~$ cmd="journalctl -k --since='$LASTTIME'"
pxl@pxl-UP2:~$ eval $cmd
-- Logs begin at Wed 2019-12-11 11:26:27 CST, end at Wed 2020-09-02 11:51:20 CST. --
-- No entries --
pxl@pxl-UP2:~$ cmd="journalctl -k --since=$LASTTIME"
pxl@pxl-UP2:~$ eval $cmd
Failed to add match '11:46:57.075394': Invalid argument
```

test-case/check-ipc-flood.sh
```diff
@@ -32,27 +32,30 @@ loop_cnt=${OPT_VALUE_lst['l']}
...
     dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
     sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-    sof-kernel-log-check.sh 0 || die "Catched error in dmesg"
+    dlogc "last timestamp check: $KERNEL_LAST_TIMESTAMP"
+    sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP || die "Catched error in journalctl"
 
     dlogi "Dumping test logs!"
-    dmesg | grep "IPC Flood count" -A 2
+    journalctl -k --since=$KERNEL_LAST_TIMESTAMP | grep "IPC Flood count" -A 2
```
xiulipan, line 55:
```suggestion
    journalctl -k --since="$KERNEL_LAST_TIMESTAMP" | grep "IPC Flood count" -A 2
```
KERNEL_LAST_TIMESTAMP has space inside, need quote

xiulipan (changes requested):
Some bash string issue with quote.

plbossart:
@xiulipan your suggestions don't work

````
2020-09-02 16:12:21 UTC [REMOTE_COMMAND] last timestamp check: 2020-09-02 16:12:20.480743
check begin_timestamp: 2020-09-02
check cmd: journalctl -k --since='2020-09-02'
````

I really hate this bash thing, it's just horrible.

**Comment by plbossart** [comment]
> @xiulipan your suggestions don't work
> 
> ```
> 2020-09-02 16:12:21 UTC [REMOTE_COMMAND] last timestamp check: 2020-09-02 16:12:20.480743
> check begin_timestamp: 2020-09-02
> check cmd: journalctl -k --since='2020-09-02'
> ```
> 
> I really hate this bash thing, it's just horrible.

It turns out the parameter also needed to be in double-quotes. Not all scripts did it. Fixed now.

**Comment by plbossart** [comment]
@marc-hb @xiulipan @aiChaoSONG this should be ready to review now, I cleaned-up the commits. Phew.

**Comment by plbossart** [comment]
SOFCI TEST

File: tools/sof-kernel-log-check.sh [hunk 9]
```diff
@@ -52,24 +52,36 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
...
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught journalctl -k error"
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
     builtin exit 1
 fi
 
+# Handle Call Trace separately
+err_trace=$(eval $cmd | grep 'Call Trace' -A5 -B3)
+if [ "$err_trace" ]; then
```
marc-hb, line 108: I think
```shell
if [ -n "$err_trace" ]; then
```
... is not only clearer but also "safer": less likely to break if `$err_trace` starts with some unusual characters.

case-lib/lib.sh
```diff
@@ -38,10 +38,9 @@ if [ ! "$SOFCARD" ]; then
     SOFCARD=$(grep '\]: sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
 fi
 
-func_lib_setup_kernel_last_line()
+func_lib_setup_kernel_last_timestamp()
```
marc-hb, line 41: I think you could have kept this function name for now, it would have made this (risky) refactoring so much smaller. Too late.

tools/sof-kernel-log-check.sh
```diff
@@ -52,24 +52,36 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
...
 if [ "$err" ]; then
-    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught dmesg error"
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught journalctl -k error"
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
     builtin exit 1
 fi
 
+# Handle Call Trace separately
```
marc-hb, line 106: Can you help me understand how this was done before? (if at all)

marc-hb (commented):
I'm afraid the `--since` logic doesn't work and falls back on scanning the entire log in every test.

How did you test this locally? To inject errors easily I would locally replace the `-k` option in `journalctl -k` with `--boot` and add `logger hello $0` statements in tests.

case-lib/lib.sh
```diff
@@ -38,10 +38,9 @@ if [ ! "$SOFCARD" ]; then
...
 {
-    # shellcheck disable=SC2034 # external script will use it
-    KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+    KERNEL_LAST_TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S.%6N")
```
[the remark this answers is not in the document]
plbossart: I couldn't find a way to make journalctl understand a date in UTC time. If you find a solution this can be a follow-up PR.

tools/sof-kernel-log-check.sh
```diff
@@ -52,24 +52,36 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
...
 
-# confirm begin_line is number, if it is not the number, direct using dmesg
-[[ "${begin_line//[0-9]/}" ]] && begin_line=0
-[[ "$begin_line" -eq 0 ]] && cmd="dmesg" || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"
-
-#echo "run $0 with parameter '$*' for check kernel message error"
+# confirm begin_timestamp is valid, if it is not the number, search full log
+re='^[0-9]+$'
+if [[ $begin_timestamp =~ $re ]] ; then
+    cmd="journalctl -k"
```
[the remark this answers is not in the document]
plbossart: this is for the case where no argument is passed. But in practice we always pass one, so that branch is never taken.

tools/sof-kernel-log-check.sh
```diff
@@ -52,24 +52,36 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
...
 fi
 
+# Handle Call Trace separately
+err_trace=$(eval $cmd | grep 'Call Trace' -A5 -B3)
+if [ "$err_trace" ]; then
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught trace call"
+    echo "===========================>>"
+    echo "$err_trace"
+    echo "<<==========================="
+    builtin exit 1
```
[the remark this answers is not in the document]
plbossart: follow-up cleanup, there's just too many of those.

tools/sof-kernel-log-check.sh
```diff
@@ -52,24 +52,36 @@ ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
...
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
     builtin exit 1
 fi
 
+# Handle Call Trace separately
+err_trace=$(eval $cmd | grep 'Call Trace' -A5 -B3)
+if [ "$err_trace" ]; then
+    echo `date -u '+%Y-%m-%d %T %Z'` "[ERROR]" "Caught trace call"
```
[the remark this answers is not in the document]
plbossart: follow-up cleanup.

tools/sof-kernel-log-check.sh
plbossart: follow-up cleanup. I didn't add this way of doing things.

case-lib/lib.sh
plbossart: it's just a rename, not a refactoring.

tools/sof-kernel-log-check.sh
plbossart: I have no idea what the previous code did:

````bash
    err=$(eval $cmd|grep 'Call Trace' -A5 -B3)$(eval $cmd | grep -E "$err_str"|grep -vE "$ignore_str")	
````

File: test-case/check-ipc-flood.sh (hunk 8 above)
Comment (marc-hb) [reply test-case/check-ipc-flood.sh]: > KERNEL_LAST_TIMESTAMP has space inside, need quote

@plbossart shell variable expansion needs to  be double-quoted 95% of the time and is harmless in 4.5% of the time.

Double quote first, think later. 
https://mywiki.wooledge.org/Quotes#I.27m_Too_Lazy_to_Read.2C_Just_Tell_Me_What_to_Do

Shellcheck will warn every time https://github.com/koalaman/shellcheck/wiki/SC2046

**Review by marc-hb — CHANGES_REQUESTED** [review]
tl;dr: I finally tested this PR for real and it introduces at least 1 or 2 regressions.

Based on some previous issues (now fixed) and discussions I think it's clear you haven't been testing this PR by making some tests FAIL. Tests that don't fail are not tested, so for such a risky change of our core (and messy) error detection "framework", green results in CI don't prove anything.

So I bit the bullet, implemented and tested a (small but very useful) function that injects fake kernel errors in the logs to properly test this part of the code at last. It's in PR #361, everyone please help review. It's small.

Now of course I used this new `fake_kern_error()` function to manually and locally add one-line errors in a few tests. Then I used run-all-tests.sh to locally check that all the tests where I injected errors were failing; and only those.

So far I had time to try only one combination of artificially failing tests yet this very small scale testing was enough to immediately find that you're introducing a "green failure" regression in `check-playback.sh`, see more details below.

I also observed a "red success" that is also introduced by this PR in another test but I didn't have time to analyze it yet. Hopefully tomorrow.

Please use PR #361 (or better) too to make some tests fail and actually test this PR. Draft PRs with fake errors should help too, I will likely launch a few tomorrow too.

I really do appreciate the huge effort to attempt to clean up this mess but we really can't afford to make it worse than it already is and the only way to avoid making it worse is to test any change very thoroughly. The worse the code is, the more testing any change requires.

File: test-case/check-ipc-flood.sh (hunk 8 above)
Comment (plbossart) [reply test-case/check-ipc-flood.sh]: I don't know why you are arguing, this was fixed
https://github.com/thesofproject/sof-test/pull/354/commits/8cb97f7d42e253ba179d1ad529d0d5a419fb097d#diff-7467b2c8abcf990e16a6e4f76a7a1211R54

File: test-case/check-playback.sh [hunk 16]
```diff
@@ -86,8 +86,9 @@ do
         if [ ${OPT_VALUE_lst['F']} = '1' ]; then
             fmt=$(func_pipeline_parse_value $idx fmts)
         fi
-        # clean up dmesg
-        sudo dmesg -C
+        # discard old kernel logs
+	func_lib_setup_kernel_last_timestamp
+
```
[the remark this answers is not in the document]
plbossart: I don't understand the comment, sorry. what is the issue?

test-case/multiple-pipeline-capture.sh
```diff
@@ -150,8 +150,8 @@ do
     pkill -9 arecord
     pkill -9 aplay
 
-    sof-kernel-log-check.sh 0 || die "Catch error in dmesg"
+    sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP" || die "Catch error in journalctl -k"
 done
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE >/dev/null
+sof-kernel-log-check.sh "$KERNEL_LAST_TIMESTAMP" >/dev/null
```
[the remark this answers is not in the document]
plbossart: I just performed a dmesg->journactl change. You can't possibly expect me to transform this bag of idiotic scripts into something perfect in one shot.

tools/sof-kernel-log-check.sh
plbossart: yes, I didn't understand it, and that's a good enough reason, no?

tools/sof-kernel-log-check.sh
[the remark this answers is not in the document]
plbossart: the test means that if there's anything other than numbers, it's not a timestamp.

I don't know why we are arguing about this.

plbossart:
I AM DONE.

marc-hb:
> You can't possibly expect me to transform this bag of idiotic scripts into something perfect in one shot.

Not "perfect", certainly not. But not giving more misleading results than it gives now, which is unfortunately a huge challenge that I'm afraid you underestimated despite my "don't go there" warnings and my various, less ambitious suggestions. BTW I will resubmit now the "simpler" yet very useful commits of this PR in a smaller PR.

"Not giving more misleading results" = not adding any new green failure which unfortunately requires a significant test effort involving some fake errors (see small #361)

marc-hb:
> BTW I will resubmit now the "simpler" yet very useful commits of this PR in a smaller PR.

Done in #362

Issue #340, opened 2020-08-13
Ignore firmware loading failure if retrial is successful


fredoh9:
FW load may fail at any time, we have up to 3 time retrials. If FW loading file fails, below errors cause test case failure.

```
kernel: [  636.613727] sof-audio-pci 0000:00:1f.3: unknown ROM status value 06000010
kernel: [  636.613734] sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
kernel: [  636.613768] sof-audio-pci 0000:00:1f.3: error: extended rom status:0x6000010 0x0 0x0 0x0 0x0 0x0 0x2001180 0x0 
kernel: [  636.613773] sof-audio-pci 0000:00:1f.3: unknown ROM status value 06000010
```
We can ignore, but I'm afraid status and panic codes are NOT specific enough.

- "0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000"
- "0000:00:1f.3: error: extended rom status:0x6000010"

Logically we should check iteration, `Attempting iteration 1 of Core En/ROM load...` and firmware boot fine `firmware boot complete` then ignore previous above errors. This is some work to do in current error checking.

https://sof-ci.01.org/linuxpr/PR2359/build4303/devicetest/

```
kernel: [  636.311495] sof-audio-pci 0000:00:1f.3: loading firmware
kernel: [  636.311518] sof-audio-pci 0000:00:1f.3: Debug PCIR: 00000002 at  00000048
kernel: [  636.311539] sof-audio-pci 0000:00:1f.3: Debug PCIW: 00000000 at  00000048
kernel: [  636.311545] sof-audio-pci 0000:00:1f.3: Debug PCIR: 00000000 at  00000044
kernel: [  636.311551] sof-audio-pci 0000:00:1f.3: Debug PCIW: 00000004 at  00000044
kernel: [  636.311552] sof-audio-pci 0000:00:1f.3: booting DSP firmware
kernel: [  636.311638] sof-audio-pci 0000:00:1f.3: FW Poll Status: reg=0x40000 successful
kernel: [  636.311689] sof-audio-pci 0000:00:1f.3: FW Poll Status: reg=0x40000 successful
kernel: [  636.311695] sof-audio-pci 0000:00:1f.3: period_bytes:0x0
kernel: [  636.311697] sof-audio-pci 0000:00:1f.3: periods:1
[...]

PR #355, opened 2020-08-28
tools: update the checklist and ignore list with boot time

xiulipan:
With https://github.com/thesofproject/linux/pull/2382
no error logs will generate in boot time. Udpate the err list
and ignore list with this update.

Fix https://github.com/thesofproject/sof-test/issues/340

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -2,18 +2,7 @@
 
 begin_line=${1:-1}
 declare err_str ignore_str project_key
-err_str="error|failed|timed out|panic|oops"
-
-# There will be debug logs at each failed initializaiton of DSP:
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
Pierre has done this in https://github.com/thesofproject/sof-test/pull/354/commits/2feec89a55999876b0a0c83a2b574c1d8d6afc6e, close this and review https://github.com/thesofproject/sof-test/pull/354?

lyakh (commented):
you removed "panic" because it also triggers a different keyword, maybe "error"?

plbossart:
> you removed "panic" because it also triggers a different keyword, maybe "error"?

we need to keep panic, but only when it's a dev_err/warn level. We have a couple of dev_dbg() logs which use 'panic', but they are harmless.

plbossart:
yes this seems redundant with #354, let's close this one.

marc-hb:
See also #362 (and #340)

PR #356, opened 2020-08-31
env-check: add check for more required files and python packages

aiChaoSONG:
This PR improve env-check.sh:
- add check for sof-logger, sof-ctl and sof-*.ldc
- add check for python3-numpy and python3-scipy
- fix some shellcheck warnings

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 mydir=$(cd "$(dirname "$0")"; pwd)
 
 # enable dynamic debug logs for SOF modules
@@ -78,7 +80,7 @@ cd "$mydir"
 \t\e[31mcd ${mydir}\n
 \tchmod a+x tools/*\e[0m"
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    echo -e "\e[31mWarning\e[0m\nSolution:$out_str"
 
 out_str="" check_res=0
 echo -ne "Checking for case folder:\t\t"
@@ -89,7 +91,7 @@ echo -ne "Checking for case folder:\t\t"
 \t\e[31mcd ${mydir}\n
 \tchmod a+x test-case/*\e[0m"
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    echo -e "\e[31mWarning\e[0m\nSolution:$out_str"
 
 out_str="" check_res=0
 echo -ne "Checking the permission:\t\t"
@@ -125,7 +127,7 @@ check_res=1 && out_str=$out_str"\n
 \t\tPlease create the \e[31mlink\e[0m of your distribution kernel log file at \e[31m/var/log/kern.log\e[0m"
 
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    echo -e "\e[31mWarning\e[0m\nSolution:$out_str"
 
 out_str="" check_res=0
 echo -ne "Checking the config setup:\t\t"
@@ -159,4 +161,4 @@ esac
 \t\tthe permissions are properly set up according to instructions."
 
 [[ $check_res -eq 0 ]] && echo "pass" || \
-    echo -e "\e[31mWarning\e[0m\nSolution:"$out_str
+    echo -e "\e[31mWarning\e[0m\nSolution:$out_str"
--- a/env-check.sh
+++ b/env-check.sh
@@ -43,6 +43,8 @@ func_check_pkg expect
 func_check_pkg aplay
 func_check_pkg python3
 func_check_python_pkg graphviz
+func_check_python_pkg numpy
+func_check_python_pkg scipy
 func_check_file "$DYNDBG"
 if [ $check_res -eq 0 ]; then
     printf "pass\n"
--- a/env-check.sh
+++ b/env-check.sh
@@ -37,6 +37,13 @@ func_check_file(){
     check_res=1
 }
 
+func_check_exec_binary() {
+    if ! type "$1" &> /dev/null; then
+        out_str="$out_str""\tExecutable \e[31m $1 \e[0m not found, please put $1 in PATH\n"
+        check_res=1
+    fi
+}
+
 check_res=0
 printf "Checking for some OS packages:\t\t"
 func_check_pkg expect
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@marc-hb I have no idea how to fix some of the shellcheck warnings here: https://travis-ci.org/github/thesofproject/sof-test/builds/722676292?utm_source=github_status&utm_medium=notification, maybe you can help with the 
`^-- SC2143: Use grep -q instead of comparing output with [ -n .. ].` ?

env-check.sh
```diff
@@ -1,12 +1,14 @@
 #!/bin/bash
 
-mydir=$(cd "$(dirname "$0")"; pwd)
+mydir=$(cd "$(dirname "$0")" || exit; pwd)
```
marc-hb, line 3: The better way to fix this warning is `set -e`.

This is a bit more work though because `set -e` probably requires something like this too:
```shell
func_check_pkg expect || : $((check_res++))
func_check_pkg aplay || : $((check_res++))
func_check_pkg python3 || : $((check_res++))
func_check_python_pkg graphviz || : $((check_res++))
```

env-check.sh
```diff
@@ -88,7 +88,7 @@ func_check_return_val "$check_res"
 out_str="" check_res=0
 echo -ne "Check for tools folder:\t\t"
 
-cd "$mydir"
+cd "$mydir" || exit
```
marc-hb, line 91: see above

env-check.sh
```diff
@@ -31,34 +33,62 @@ func_check_python_pkg(){
...
+    case $1 in
+        "$dyn_dbg_conf")
+            out_str="$out_str""\tptional: Enable dynamic debug logs in \e[31m $1 \e[0m file\n"
+            out_str="$out_str""\tFor example,\n\toptions snd_sof dyndbg=+p\n\toptions snd_sof_pci dyndbg=+p\n"
+            ;;
+        "$sof_logger")
+            out_str="$out_str""\tsof-logger with excute permisson should be installed to $sof_logger\n"
+            ;;
+        "$sof_ctl")
+            out_str="$out_str""\tsof-ctl with excute permission should be installed to $sof_ctl\n"
```
marc-hb, line 47: Please use a spellchecker.

env-check.sh
aiChaoSONG: Only `SC2143: Use grep -q instead of comparing output with [ -n .. ]` left, I think this is releated to code login improvement, let's do this in a new PR.

env-check.sh
```diff
@@ -1,17 +1,20 @@
 #!/bin/bash
+set -e
 
 mydir=$(cd "$(dirname "$0")"; pwd)
 
 # enable dynamic debug logs for SOF modules
-DYNDBG="/etc/modprobe.d/sof-dyndbg.conf"
+dyn_dbg_conf="/etc/modprobe.d/sof-dyndbg.conf"
+sof_logger="/usr/bin/sof-logger"
```
xiulipan, line 8: better to use `which `or `type` to find if we have sof-logger or other executable binary in our PATH
user may not put sof-logger in this DIR

env-check.sh
```diff
@@ -1,17 +1,20 @@
...
+sof_ctl="/usr/bin/sof-ctl"
+sof_ldc="/etc/sof/sof-$("$mydir"/tools/sof-dump-status.py -p).ldc"
```
xiulipan, line 10: we'd better provide a method to have some ENV override for ldc path. This is all for CI environment, we should make it more generic to normal user.

env-check.sh
```diff
@@ -31,28 +34,55 @@ func_check_python_pkg(){
...
 func_check_file(){
     if [ -e "$1" ]; then
-        return
+        return 0
     fi
-    out_str="$out_str""Optional: Enable dynamic debug logs in \e[31m $1 \e[0m file\n\tFor example,\n\toptions snd_sof dyndbg=+p\n\toptions snd_sof_pci dyndbg=+p\n"
+    case $1 in
+        "$dyn_dbg_conf")
+            out_str="$out_str""\tOptional: Enable dynamic debug logs in \e[31m$1\e[0m file\n"
+            out_str="$out_str""\tFor example:\n\t\toptions snd_sof dyndbg=+p\n\t\toptions snd_sof_pci dyndbg=+p\n"
```
xiulipan, line 42: I think we can provide a copy of sof-dyndbg.conf in our repo for user reference.
PS: user can put this option in any conf files. So better also provide a simple cmd line here with the files provide.

env-check.sh
```diff
@@ -31,28 +34,55 @@ func_check_python_pkg(){
...
+            ;;
+        "$sof_logger")
+            out_str="$out_str""\tExecutable sof-logger should be installed to \e[31m$1\e[0m\n"
```
xiulipan, line 45: suggest to instal in PATH

env-check.sh
```diff
@@ -31,28 +34,55 @@ func_check_python_pkg(){
...
+        printf '\e[31mWarning\e[0m\n'
+        # Need ANSI color characters.
+        # shellcheck disable=SC2059
+        printf "$out_str"
+    fi
+}
+
+out_str="" check_res=0
+printf "Checking for required files\t\t"
+func_check_file "$sof_logger"
```
xiulipan, line 71: For executable binary check, a new func_check_path would be good. Please see above comments

xiulipan (changes requested):
Some comments about executable binary

marc-hb:
I agree with @xiulipan that a new `type  $cmd` test would be better. However if this is urgently needed for CI or some  other reason then this can be merged now and enhanced later.

`type` is for Bourne shells (including) bash whereas `which` is for C-shells like csh, tcsh etc. Compare:

```shell
$ which type
/usr/bin/type 
$ type type
type is a shell builtin
```

xiulipan:
@marc-hb @aiChaoSONG I think this is some enhancement for user experience. No influence for CI environment.

aiChaoSONG:
@xiulipan @marc-hb It is not urgent, before I continue with this PR, please help to review https://github.com/thesofproject/sof-test/pull/358, which makes sof-*.ldc configurable.

xiulipan:
@aiChaoSONG Ping?

aiChaoSONG:
@xiulipan I will pick this soon

xiulipan:
Ping?

aiChaoSONG:
@xiulipan Updated

env-check.sh
```diff
@@ -37,13 +39,24 @@ func_check_file(){
     check_res=1
 }
 
+func_check_exec_binary() {
+    if ! type "$1" &> /dev/null; then
```
marc-hb, line 43: If you redirect only stderr and not stdout then it will show which binary comes first in the PATH which is useful.

You can avoid the negation with `type ... || {`

marc-hb (approved):

xiulipan (approved):

Issue #351, opened 2020-08-26
kmod-load-unload shouldn't exit with modules unloaded


lyakh:
If an error is detected when unloading kernel modules (see e.g. https://github.com/thesofproject/linux/issues/2334) the check-kmod-load-unload.sh script exits immediately with modules unloaded, so no further testing is possible. If a proposed policy is implemented where the tested system is rebooted after a failure, this would be fixed. Otherwise an attempt should be made to re-load the modules after the error.

marc-hb:
In https://github.com/thesofproject/linux/issues/2334 @plbossart suggested we try multiple times and count the number of errors.

I looked at the code it should be doable but please first please help review the simple shellcheck cleanup in #368

aiChaoSONG:
It is not easy to fix it in the script, because kmod test touched kernel, and it is hard to make sure the kernel module can be reloaded properly, and the HW status is status, so current solution is 'when kmod test falied, we will reboot and continue test'

PR #357, opened 2020-09-01
kmod-case: fix case exit with module unloaded

aiChaoSONG:
Fixes: #351

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -72,7 +72,7 @@ do
 
     dlogi "run kmod/sof-kmod-remove.sh"
     sudo sof_remove.sh
-    [[ $? -ne 0 ]] && die "remove modules error"
+    [[ $? -ne 0 ]] && sudo sof_insert.sh && die "remove modules error"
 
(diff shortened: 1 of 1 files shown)


xiulipan:
@aiChaoSONG What I am think here is could module reload successfully when error already happen? It may break the device directly.
what @lyakh wants in https://github.com/thesofproject/sof-test/issues/351 is that we should reset the DUT if error happen in those test case.

I think we should handle this in our framework.

xiulipan:
@marc-hb we already put all suspend/resume and kmod test in last.
I am also working on the framework in CI to reset the device.

aiChaoSONG:
@marc-hb @xiulipan maybe we should  depend on the framework to resolve this. close ?

marc-hb:
> @marc-hb @xiulipan maybe we should depend on the framework to resolve this. close ?

When you write "depend on" I read "wait for". Why not but for how long? Do we need another workaround in the mean time?

Ask  the submitter in #351

aiChaoSONG:
This is a bad try to solve #351,  I agree that if issue happens in kmod test, reset DUT is a good way to restore. let's close this.

PR #358, opened 2020-09-02
case-lib: make sof-*.ldc file path configurable

aiChaoSONG:

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -47,11 +47,20 @@ func_lib_setup_kernel_last_line()
 SOF_LOG_COLLECT=0
 func_lib_start_log_collect()
 {
-    local is_etrace=${1:-0} ldcFile
-    ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
-        >&2 dlogw "sof-dump-status.py -p to query platform failed"
-        return
+    local is_etrace=${1:-0} ldcFile="$SOFLDC"
+    # if user doesn't specify file path of sof-*.ldc, fall back to
+    # /etc/sof/sof-PLATFORM.ldc, which is the default path used by CI.
+    [[ -n "$ldcFile" ]] || {
+        ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+            >&2 dloge "Failed to query platform with sof-dump-status.py"
+            return 1
+        }
     }
+    [[ -e "$ldcFile" ]] || {
+        dloge "LDC file $ldcFile not found, check the SOFLDC environment variable or put your sof-*.ldc to /etc/sof"
+        return 1
+    }
+
     local loggerBin="" logfile="" logopt="-t"
     [[ "$SOFLOGGER" ]] && loggerBin=$SOFLOGGER || loggerBin=$(command -v sof-logger)
     if [ "X$is_etrace" == "X0" ];then


xiulipan (dismissed):

case-lib/config.sh
```diff
@@ -34,6 +34,10 @@ LOG_ROOT=${LOG_ROOT:-}
 # EXAMPLE: the sof-logger will collect the data into this folder
 SOFLOGGER=${SOFLOGGER:-}
 
+# User may not put sof-*.ldc into the default folder "/etc/sof", this env allows
+# user to specify the path of their own sof-*.ldc file.
+SOFLDC=${SOFLDC:-}
```
marc-hb, line 39: I don't understand the purpose of these lines. The only thing they do is changing a missing variable to an empty variable. As we don't use `set -o nounset` this doesn't seem to make any difference.

case-lib/lib.sh
```diff
@@ -47,11 +47,13 @@ func_lib_setup_kernel_last_line()
 SOF_LOG_COLLECT=0
 func_lib_start_log_collect()
 {
-    local is_etrace=${1:-0} ldcFile
-    ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
-        >&2 dlogw "sof-dump-status.py -p to query platform failed"
-        return
-    }
+    local is_etrace=${1:-0} ldcFile="$SOFLDC"
+    if [[ ! -e "$SOFLDC" ]]; then
```
marc-hb, line 51:
```shell
if [ -z "$SOFLDC" ]; then
```

If I misconfigure SOFLDC by mistake, then I want the logger to fail. I don't want something else to try to correct my mistake without telling me.

case-lib/config.sh
aiChaoSONG: I didn't realize this before, if you say so, I think most content of config.sh can be dropped, but for now, can we just align with  others, eg: LOG_ROOT, SOFLOGGER, etc? I can drop all of them in a new PR.

case-lib/lib.sh
aiChaoSONG: updated

case-lib/lib.sh
```diff
@@ -47,11 +47,13 @@ func_lib_setup_kernel_last_line()
...
 {
-    local is_etrace=${1:-0} ldcFile
-    ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
-        >&2 dlogw "sof-dump-status.py -p to query platform failed"
-        return
-    }
+    local is_etrace=${1:-0} ldcFile="$SOFLDC"
+    if [[ -z "$SOFLDC" ]]; then
+        ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+            >&2 dlogw "sof-dump-status.py -p to query platform failed"
```
marc-hb, line 53: This error message doesn't say what's missing (the ldcfile) and it doesn't suggest any solution.

case-lib/config.sh
marc-hb: If I install something in the default location and the scripts find it in the default location then everything just works and I don't want to hear about implementation details, it's just noise.

If I forget to install something or put in the wrong location, _then_ I want the error message or `env-check.sh` to help me fix it.

Either way I don't think such a list is useful. I didn't know about this list until a couple days ago and I never missed it.

A list of _required_ options that don't have defaults all gathered in a single place would be useful but fortunately we require almost nothing so we still don't need such a list.

This is a bit off-topic, for now in this PR I recommend to leave `config.sh` as it is.

marc-hb (changes requested):

case-lib/lib.sh
aiChaoSONG: @marc-hb  If $SOFLDC is empty,  we will try to use /etc/sof/sof-\*.ldc, (default sof-\*.ldc folder in CI), I think we should add error message/solution after `ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || { >&2 dlogw "sof-dump-status.py -p to query platform failed" `}, before I make any change, I want to get input from you.

I will remove SOFLDC in config.sh.

`$(sof-dump-status.py -p)` will give the platform name, eg, cml, apl, etc.

xiulipan (approved):

case-lib/lib.sh
```diff
@@ -47,11 +47,15 @@ func_lib_setup_kernel_last_line()
 SOF_LOG_COLLECT=0
 func_lib_start_log_collect()
 {
-    local is_etrace=${1:-0} ldcFile
-    ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
-        >&2 dlogw "sof-dump-status.py -p to query platform failed"
-        return
-    }
+    local is_etrace=${1:-0} ldcFile="$SOFLDC"
+    if [[ -z "$SOFLDC" ]]; then
```
marc-hb, line 51: It's easier to read if you just use `$ldcFile` everywhere, I mean:
```shell
if [[ -z "$ldcFile" ]]; then
```

case-lib/lib.sh
```diff
@@ -47,11 +47,15 @@ func_lib_setup_kernel_last_line()
...
+        # Default sof-*.ldc path used by CI is /etc/sof/sof-PLATFORM.ldc
+        ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+            >&2 dlogw "Failed to query platform with sof-dump-status.py"
+            return 1
+        }
+        [[ ! -e "$ldcFile" ]] && die "LDC file not found, check SOFLDC environment variable or put your sof-*.ldc to /etc/sof"
```
marc-hb, line 57: Shouldn't this test be after and outside the `if` statement?

If `sof-dump-status.py -p` fails right now you `die`, on the other hand if the user misconfigures `$SOFLDC` then you let the `$loggerBin` fail later. This is inconsistent, I think both cases should fail the same way. I'm not sure which way though they should both fail though: how does `$loggerBin` fail when there is a typo in`$SOFLDC`? If there's already good error message (I think a good one was added recently) then we don't need this other assert here; "less is more"

Also, I'm not sure this sub-function should `die`, in fact the previous failure above does not `die` it just `returns 1`. A `return 1` lets the caller decide (with `set -e` or anything else), `die` gives the caller no choice.

Finally, I prefer this code style because there's no negation and it reads like English:
```shell
assert_something || return 1
```

marc-hb (changes requested):

case-lib/lib.sh
aiChaoSONG: @marc-hb If there is a typo in $SOFLDC, sof-logger will exit immediately with error message; if there is no typo, sof-logger will runs during the whole test case, we have to manually kill it after test case finished. Currently, error message of sof-logger is redirected to /dev/null, because when we kill it, we get noise in the log like this
```
/home/ubuntu/sof-test/case-lib/hijack.sh: line 118: 120976 Killed 
/usr/bin/sudo env 'PATH=/home/ubuntu/sof-test/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin' /usr/bin/sof-logger -t -l /etc/sof/sof-cml.ldcbak -o /home/ubuntu/sof-test/logs/check-playback/2020-09-09-13:52:57-6300/slogger.txt
```
according to you comment, let's do this:
- use `ldcFile` every where
- don't die, return 1 if `sof-dump-status.py` failed or user misconfigures `ldcFile`
- move [[ ! -e "$ldcFile" ]] out of the first -z test.

I test this on my Helios, this way makes it quite user friendly.

marc-hb: > Currently, error message of sof-logger is redirected to /dev/null,

This is really, really bad because it hides _all_ errors for sof-logger. Test code should never redirect anything to /dev/null.

> because when we kill it, we get noise in the log like this

Test logs are not meant to be beautiful, they're meant to be useful and readable. Confusion if any can be avoided with a simple "Killing sof-logger" message.

case-lib/lib.sh
```diff
@@ -47,11 +47,20 @@ func_lib_setup_kernel_last_line()
...
+    # if user doesn't specify ldcFile, fall back to /etc/sof/sof-PLATFORM.ldc,
+    # which is used by CI.
+    [[ -n "$ldcFile" ]] || {
+        ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+            >&2 dlogw "Failed to query platform with sof-dump-status.py"
+            return 1
+        }
+    }
+    [[ -e "$ldcFile" ]] || {
+        dlogw "LDC file not found, check ldcFile environment variable or put your sof-*.ldc to /etc/sof"
```
marc-hb, line 60: `$ldcFile` is the main information, it must be in the error message.

aiChaoSONG: fixed.  I agree with your point, the log should be readable rather than  beautiful, but I also think the original author may make a lot of compromise to redirect the sof-logger's stderr. The purpose of this PR is making the log more readable.

case-lib/lib.sh
marc-hb: I meant "everywhere in this function", sorry I wasn't clear. I meant this:
```shell
    local is_etrace=${1:-0} ldcFile="$SOFLDC"
    if [[ -z "$ldcFile ]]; then
```
Anyway I'm fine with `$ldcFile` in the interface too.

case-lib/lib.sh
marc-hb: I still don't see `$ldcFile` in the message.

case-lib/lib.sh
aiChaoSONG: thanks, your review helps a lot. fixed

marc-hb (approved):

marc-hb:
Merging with just one approval because it's small and low risk, it has been under review for 9 days and I need to resolve local git conflicts with it.

Issue #232, opened 2020-06-04
misleading error report of 'no available topology'


plbossart:
I have an UpExtreme board with SoundWire. headset playback works fine.

sof-test provides very misleading error reports

````
root@upExtreme2:~/sof-test/test-case# ./check-playback.sh -d 4
2020-06-04 19:28:59 UTC [REMOTE_INFO] no source file, use /dev/zero as dummy playback source
2020-06-04 19:28:59 UTC [REMOTE_ERROR] No available topology for pipeline export
2020-06-04 19:28:59 UTC [REMOTE_INFO] Test Result: FAIL!
root@upExtreme2:~/sof-test/test-case# test-speaker.sh
-bash: test-speaker.sh: command not found
root@upExtreme2:~/sof-test/test-case# ./test-speaker.sh
2020-06-04 19:29:56 UTC [REMOTE_ERROR] No available topology for pipeline export
2020-06-04 19:29:56 UTC [REMOTE_INFO] Test Result: FAIL!
root@upExtreme2:~/sof-test/test-case# speaker-test -Dhw:0,0 -c2 -r48000

speaker-test 1.2.2

Playback device is hw:0,0
Stream parameters are 48000Hz, S16_LE, 2 channels
Using 16 octaves of pink noise
Rate set to 48000Hz (requested 48000Hz)
Buffer size range from 96 to 16368
Period size range from 48 to 4080
Using max buffer size 16368
Periods = 4
was set period_size = 3312
was set buffer_size = 16368
 0 - Front Left
 1 - Front Right
^CTime per period = 3.095371
````

WTH is this 'no available topology' error?

@keqiaozhang @marc-hb FYI

plbossart:
there is actually a topology check
````
root@upExtreme2:~/sof-test/test-case# ./verify-tplg-binary.sh
2020-06-04 19:43:44 UTC [REMOTE_ERROR] No available topology for this test case
2020-06-04 19:43:45 UTC [REMOTE_INFO] Test Result: FAIL!
root@upExtreme2:~/sof-test/test-case# ls -l /lib/firmware/intel/sof-tplg/
total 48
-rw------- 1 root root 25451 Jun  3 19:23 sof-cnl-rt711vd_sdw0.tplg
-rw------- 1 root root 16411 May 12 17:35 sof-hda-generic-idisp.tplg
````

but it's just wrong, there is a topology for this device. The requirements are completely unclear here.

plbossart:
````
[...]

PR #359, opened 2020-09-02
pipeline: fix misleading pipeline exporting info

aiChaoSONG:
The "No available topology ..." info is misleading,
this patch makes it clearer.

Fixes: #232

### DIFF
--- a/case-lib/pipeline.sh
+++ b/case-lib/pipeline.sh
@@ -8,7 +8,7 @@ func_pipeline_export()
     fi
     # got tplg_file, verify file exist
     tplg_path=$(func_lib_get_tplg_path "$1") || {
-        die "No available topology for pipeline export"
+        die "Topology $1 not found, check the TPLG environment variable or specify topology path with -t"
     }
     dlogi "$SCRIPT_NAME will use topology $tplg_path to run the test case"
 


case-lib/pipeline.sh
```diff
@@ -8,7 +8,7 @@ func_pipeline_export()
     fi
     # got tplg_file, verify file exist
     tplg_path=$(func_lib_get_tplg_path "$1") || {
-        die "No available topology for pipeline export"
+        die "No topology found! Please configure TPLG env to point to you topology path."
```
marc-hb, line 11: Many tests seem to have a `-t` option so setting TPLG is not the only possible solution.

It looks like this new error message will ask me set TPLG even when I make a typo in TPLG  that's confusing. It will also not show my typo.

How about:
```shell
die "Topology $1 not found, check the TPLG environment variable" 
```

aiChaoSONG: @marc-hb The default value for -t option is $TPLG. if No TPLG is set, and -t not specified, the "$1" will be empty.  How about this: `Topology not found, check TPLG environment variable or specify topology path with -t`

marc-hb: - if `$1` is empty: `Topology not found, check TPLG`
- if `$1` is wrong: `Topology wrong_file not found, check TPLG`.

I don't mind mentioning `-t` in the message but it should also print `$1`.

aiChaoSONG: I'd like the -t in the message, updated

marc-hb (approved):

PR #252, opened 2020-06-11
test-case: verify-sof-firmware-load: update filter of firmware keyword

Bin-QA:
with replace kern.log file the keyword filter already changed from
'] sof-audio' to 'sof-audio' so need refine the case filter

Signed-off-by: Wu, BinX <binx.wu@intel.com>

PR #360, opened 2020-09-02
verify-sof-firmware-load: fix the last shellcheck warnings and more specific grep

marc-hb:
Supersedes #252


marc-hb:
The modified test is passing everywhere except on BSW_CYN_MAX98090 which doesn't boot at all.

xiulipan (approved):

PR #361, opened 2020-09-03
case-lib/lib.sh: add new fake_kern_error()

marc-hb:
Needed to (manually and locally) test our overly complex error handling framework, see for instance PR #354
    
### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -44,6 +44,37 @@ func_lib_setup_kernel_last_line()
     KERNEL_LAST_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
 }
 
+# This function adds a fake error to dmesg (which is always saved by
+# journald). It also adds it to kern.log, see why in comment below.
+#
+# It is surprisingly easy to write a test that always succeeds,
+# especially in shell script and it has already happened a number of
+# times. Temporarily add this function to a test under development to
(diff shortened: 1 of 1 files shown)


xiulipan (commented):
@marc-hb why this fake_kern_error function is needed? And where we will use this function?
Please also add some explanation in the commit.

marc-hb:
@xiulipan what part of the commit message needs more detail?

xiulipan:
@marc-hb For the commit `case-lib/lib.sh: add new fake_kern_error()`. Can you explain why we need to add such a function?
I do not think some discussion on GitHub can answer the question when people looking for the source code. (Please also add some inline comments if this function is for test only)

PS: will this function take effect also with journalctl?

marc-hb:
Thanks @xiulipan  for the review, I added more details.

> PS: will this function take effect also with journalctl?

Yes. Unlike `kern.log`, `journalctl` saves everything from `dmesg`.

marc-hb:
Besides the regression mentioned in #354, here's another example of bugs found thanks to this function: #366

xiulipan:
@marc-hb I got the idea now. Can you also add some DEBUG prefix in commit and inline comments to show this function should only be used for debug.

PS: I think it would be better if you can have some unit test for our lib with this debug or test function.

xiulipan (approved):

Issue #312, opened 2020-08-04
[sof-test] add set -e to all test cases


aiChaoSONG:
add set -e to all of our test case for better error handling.

marc-hb:
This should be done very gradually on a test by test basis, starting of course with all new test cases like #296

marc-hb:
I think there is still a fair number of tests without `set -e`.

mengdonglin:
@marc-hb @aiChaoSONG @keqiaozhang Is this feature done?

marc-hb:
No but I'm going to reduce the priority because a good number of tests have `set -e` now.

libinyang:
@marc-hb I found some cases are hard to use `set -e`, such as cases with `grep`. Do you have some good suggestion? My current user space cases met such issue.

marc-hb:
Sorry for the delay, too much email.

For `grep` it's usually as easy as:
```shell
something=$( ... | grep something || true)
```
Please provide more details/sample code if this does not seem possible.

marc-hb:
>  Do you have some good suggestion? 

I just added a very long description with rationale, guidelines, workarounds and a good reference.

marc-hb:
https://sof-ci.01.org/softestpr/PR663/build673/devicetest/ is an interesting list of tests that don't have `set -e`. Almost all tests should have failed on BYT but some did not because they don't have `set -e`.

greg-intel:
Reviewing the tests here: https://github.com/thesofproject/sof-test/tree/main/test-case

The remaining tests without `set -e`:
* check-alsabat.sh
* check-runtime-pm-double-active.sh
* check-runtime-pm-status.sh
* check-xrun-injection.sh
* simultaneous-playback-capture.sh
* test-speaker.sh
* verify-ucm-config.sh
* volume-basic-test.sh

Also, the location of `set -e` bounces between different spots in the files. sometimes before the comment block, sometimes after...

PR #363, opened 2020-09-07
add User space test cases

libinyang:
This patches add the test cases for user space.
It includes:
1. verify-ucm-config (already merged)
2. check-user-packages: test installing user space packages, restart pulseaudio 
3. check-userspace-pa-status: check the pulseaudio can be run successfully
4. check-userspace-cardinfo: check pulseaudio can parse the ucm correctly and exports the correct card
5. check-userspace-paplay: go through all the active sinks and play on it if port is available (almost finished, not included in this patchset)
6. check-userspace-parecord: go through all the active sources and record on it if port is available (almost finished, not included in this patchset)

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -252,6 +252,14 @@ func_lib_get_tplg_path()
     return 0
 }
 
+func_lib_check_pa()
+{
+    pactl stat &>/dev/null || {
+        dloge "pactl stat failed"
+        return 1
+    }
+}
+
 die()
 {
     dloge "$@"
--- a/test-case/check-userspace-cardinfo.sh
+++ b/test-case/check-userspace-cardinfo.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+##
+## Case Name: check-userspace-cardinfo
+## Preconditions:
+##    N/A
+## Description:
+##    Get the card name from pactl info
+##    check if there is an available card to do the following tests
+## Case step:
+##    1. run pactl to get the cards info
+##    2. check the card name
+## Expect result:
+##    There is at least one available card for test
+##
+
+set -e
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# check pulseaudio runs properly or not
+func_lib_check_pa || die "Please check whether pulseaudio runs correctly or not"
+
+func_opt_parse_option "$@"
+
+OLDIFS=$IFS
+dlogc "pactl list cards short"
+cardlist=$(pactl list cards short)
+: $((available_card=0))
+IFS=$'\n'
+for card in $cardlist; do
+    # pactl list cards short format should be like:
+    # 0	alsa_card.pci-0000_00_1f.3-platform-sof_sdw	module-alsa-card.c
+    dlogi "found card: $(echo "$card" | awk '{print $2}')"
+    echo "$card" |grep -i "usb" &>/dev/null || : $((available_card++))
+done
+IFS=$OLDIFS
+
+if [ "$available_card" == "0" ]; then
+    # TODO: do more check to give hint why there is no available card
+    die "no available card for test"
+fi
--- a/test-case/check-userspace-paplay.sh
+++ b/test-case/check-userspace-paplay.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+##
+## Case Name: check-userspace-paplay
+## Preconditions:
+##    N/A
+## Description:
+##    Go through all the sinks and paplay on the sinks whose
+##    active port is available (jack connected) or unknown (speaker)
+## Case step:
+##    1. go through all the sinks
+##    2. check the sink's active port is available or not
+##    3. paplay on the active port which is available or unknown
+## Expect result:
+##    paplay on the sinks successfully
+##
+
+set -e
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# check pulseaudio runs properly or not
+func_lib_check_pa || die "Please check whether pulseaudio runs correctly or not"
+
+OPT_OPT_lst['r']='round'     OPT_DESC_lst['r']='round count'
+OPT_PARM_lst['r']=1         OPT_VALUE_lst['r']=3
+
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='paplay duration in second'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=8
+
+OPT_OPT_lst['f']='file'   OPT_DESC_lst['f']='source file path'
+OPT_PARM_lst['f']=1         OPT_VALUE_lst['f']='/dev/zero'
+
+OPT_OPT_lst['F']='format'   OPT_DESC_lst['F']='sample format'
+OPT_PARM_lst['F']=1         OPT_VALUE_lst['F']=s16le
+
+OPT_OPT_lst['R']='rate'   OPT_DESC_lst['R']='sample rate'
+OPT_PARM_lst['R']=1         OPT_VALUE_lst['R']=44100
+
+OPT_OPT_lst['C']='channels'   OPT_DESC_lst['C']='channels'
+OPT_PARM_lst['C']=1         OPT_VALUE_lst['C']=2
+
+func_opt_parse_option "$@"
+
+round_cnt=${OPT_VALUE_lst['r']}
+duration=${OPT_VALUE_lst['d']}
+file=${OPT_VALUE_lst['f']}
+format=${OPT_VALUE_lst['F']}
+rate=${OPT_VALUE_lst['R']}
+channel=${OPT_VALUE_lst['C']}
+
+[[ -e $file ]] || { dlogw "$file does not exist, use /dev/zero as dummy playback source" && file=/dev/zero; }
+
+# TODO: check the parameter is valid or not
+
+# go through all the sinks
+# get all the sinks name
+sinkkeys=$(pactlinfo.py --showsinks)
+for round in $(seq 1 $round_cnt); do
+    for i in $sinkkeys; do
+        sinkcard=$(pactlinfo.py --getsinkcardname "$i") || {
+            dlogw "failed to get sink $i card_name"
+            continue
+        }
+
+        # Let's skip testing on USB card
+        # TODO: add a list for other skipped cards such as HDA
+        if echo "$sinkcard" |grep -i "usb" &>/dev/null; then
+            continue
+        fi
+
+        # get the sink's active port
+        actport=$(pactlinfo.py --getsinkactport "$i") || {
+            dlogw "failed to get sink $i active port"
+            continue
+        }
+
+        # get the active port's information
+        actportvalue=$(pactlinfo.py --getsinkportinfo "$actport") || {
+            dlogw "failed to get sink $i active port $actport info"
+            continue
+        }
+
+        # check the active port is available or not from the port's information
+        portavailable=$(echo "$actportvalue" |grep "not available") || true
+        if [ -z "$portavailable" ]; then
+            # now prepare to paplay on this sink as the active port is not "not available"
+            # get the sink's name
+            sinkname=$(pactlinfo.py --getsinkname "$i")
+            sinkname=$(eval echo "$sinkname") #formalize the sink name
+            dlogi "===== Testing: (Round: $round/$round_cnt) (sink: $sinkname.$actport) ====="
+            dlogc "paplay -v --device=$sinkname --raw --rate=$rate --format=$format --channels=$channel $file"
+            paplay -v --device="$sinkname" --raw --rate=$rate --format=$format --channels=$channel $file &
+            pid=$!
+            sleep $duration
+            # check whether process is still running
+            count=$(ps -A| grep -c $pid) || true
+            if [[ $count -eq 0 ]]; then
+                if wait $pid; then #checks if process executed successfully or not
+                    dlogi "paplay has stopped successfully"
+                else
+                    die "paplay on $sinkname failed (returned $?)"
+                fi
+            else
+                dlogi "paplay runs successfully"
+                # kill all paplay process
+                pkill -9 paplay
+                sleep 0.5
+            fi
+        fi
+    done
+done
--- a/test-case/check-userspace-parecord.sh
+++ b/test-case/check-userspace-parecord.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+##
+## Case Name: check-userspace-parecord
+## Preconditions:
+##    N/A
+## Description:
+##    Go through all the sources and parecord on the sources whose
+##    active port is available (jack connected) or unknown (DMIC)
+## Case step:
+##    1. go through all the sources
+##    2. check the source's active port is available or not
+##    3. parecord on the active port which is available or unknown
+## Expect result:
+##    parecord on the source successfully
+##
+
+set -e
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# check pulseaudio runs properly or not
+func_lib_check_pa || die "Please check whether pulseaudio runs correctly or not"
+
+OPT_OPT_lst['r']='round'     OPT_DESC_lst['r']='round count'
+OPT_PARM_lst['r']=1         OPT_VALUE_lst['r']=3
+
+OPT_OPT_lst['d']='duration' OPT_DESC_lst['d']='parecord duration in second'
+OPT_PARM_lst['d']=1         OPT_VALUE_lst['d']=8
+
+OPT_OPT_lst['f']='file'   OPT_DESC_lst['f']='source file path'
+OPT_PARM_lst['f']=1         OPT_VALUE_lst['f']='/dev/zero'
+
+OPT_OPT_lst['F']='format'   OPT_DESC_lst['F']='sample format'
+OPT_PARM_lst['F']=1         OPT_VALUE_lst['F']=s16le
+
+OPT_OPT_lst['R']='rate'   OPT_DESC_lst['R']='sample rate'
+OPT_PARM_lst['R']=1         OPT_VALUE_lst['R']=44100
+
+OPT_OPT_lst['C']='channels'   OPT_DESC_lst['C']='channels'
+OPT_PARM_lst['C']=1         OPT_VALUE_lst['C']=2
+
+func_opt_parse_option "$@"
+
+round_cnt=${OPT_VALUE_lst['r']}
+duration=${OPT_VALUE_lst['d']}
+file=${OPT_VALUE_lst['f']}
+format=${OPT_VALUE_lst['F']}
+rate=${OPT_VALUE_lst['R']}
+channel=${OPT_VALUE_lst['C']}
+
+[[ -e $file ]] || { dlogw "$file does not exist, use /dev/zero as dummy playback source" && file=/dev/null; }
+
+# TODO: check the parameter is valid or not
+
+# go through all the sources
+# get all the sources name
+sourcekeys=$(pactlinfo.py --showsources)
+for round in $(seq 1 $round_cnt); do
+    for i in $sourcekeys; do
+        sourcecard=$(pactlinfo.py --getsourcecardname "$i") || {
+            dlogw "failed to get source $i card_name"
+            continue
+        }
+
+        # Let's skip testing on USB card
+        # TODO: add a list for other skipped cards such as HDA
+        if echo "$sourcecard" |grep -i "usb" &>/dev/null; then
+            continue
+        fi
+
+        sourceclass=$(pactlinfo.py --getsourcedeviceclass "$i") || {
+            dlogw "failed to get source $i device class"
+            continue
+        }
+        # Let's skip the monitor sources
+        if echo "$sourceclass" |grep "monitor" &>/dev/null; then
+            continue
+        fi
+
+        # get the source's active port
+        actport=$(pactlinfo.py --getsourceactport "$i") || {
+            dlogw "failed to get source $i active port"
+            continue
+        }
+        # get the active port's information
+        actportvalue=$(pactlinfo.py --getsourceportinfo "$actport") || {
+            dlogw "failed to get source $i active port $actport info"
+            continue
+        }
+        # check the active port is available or not from the port's information
+        portavailable=$(echo "$actportvalue" |grep "not available") || true
+        if [ -z "$portavailable" ]; then
+            # now prepare to parecord on this source as the active port is not "not available"
+            # get the source's name
+            sourcename=$(pactlinfo.py --getsourcename "$i")
+            sourcename=$(eval echo "$sourcename")
+            dlogi "===== Testing: (Round: $round/$round_cnt) (source: $sourcename.$actport) ====="
+            dlogc "parecord -v --device=$sourcename --raw --rate=$rate --format=$format --channels=$channel $file"
+            parecord -v --device="$sourcename" --raw --rate=$rate --format=$format --channels=$channel $file &
+            pid=$!
+            sleep $duration
+            # check whether process is still running
+            count=$(ps -A| grep -c $pid) || true
+            if [[ $count -eq 0 ]]; then
+                if wait $pid; then #checks if process executed successfully or not
+                    die "parecord has stopped successfully, which is abnormal"
+                else
+                    die "parecord on $sourcename failed (returned $?)"
+                fi
+            else
+                dlogi "parecord runs successfully"
+                # kill all parecord processes
+                pkill -9 parecord >/dev/null
+                sleep 0.5
+            fi
+        fi
+
+    done
+done
--- a/tools/pactlinfo.py
+++ b/tools/pactlinfo.py
@@ -1,0 +1,153 @@
+#!/usr/bin/python3
+
+import re
+import os
+import argparse
+
+# get the specified sink information from "pactl list sink"
+def get_sink(sinkname):
+    sinks = os.popen("pactl list sinks").read()
+    sink=""
+    start = 0
+    for line in sinks.splitlines():
+        # sink information starts with "Sink #n"
+        m = re.match('^Sink #(\d+)', line)
+        if m:
+            m1 = re.match('^Sink #'+sinkname, line)
+            if m1:
+                start = 1
+            else:
+                start = 0
+        else:
+            if start == 1:
+                sink += line+'\n'
+    return sink
+
+# get the specified source information from "pactl list source"
+def get_source(sourcename):
+    sources = os.popen("pactl list sources").read()
+    source=""
+    start = 0
+    for line in sources.splitlines():
+        # source information starts with "Source #n"
+        m = re.match('^Source #(\d+)', line)
+        if m:
+            m1 = re.match('^Source #'+sourcename, line)
+            if m1:
+                start = 1
+            else:
+                start = 0
+        else:
+            if start == 1:
+                source += line+'\n'
+    return source
+
+# get the value of the key in the bulk
+# For example, the bulk is the information of a sink, which
+# includes "State", "Name", "Driver" and etc information.
+# get_value(bulk, "Name: ") will get the name information.
+# The "Name: " is the key word format, please refer 'pactl list' for
+# more keys format.
+# TODO: some informations are splitted into multiple lines. Need to handle
+#       this situation if necessary.
+def get_value(bulk, key):
+    # refine the key to handle the special characters
+    schar='\[@_!#$%^&*()<>?/|}{~]'
+    for char in schar:
+        key=key.replace(char, "\\"+char)
+
+    res = re.findall(key+".*", bulk)
+    if res:
+        print (re.match("("+key+")(.*)", res[0]).groups(1)[1])
+        return(0)
+    else:
+        return(1)
+
+'''test Main'''
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='parse pactl list information',
+                                     add_help=True, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument('--showsinks', action='store_true', default=False, help='show all sinks')
+    parser.add_argument('--showsources', action='store_true', default=False, help='show all sources')
+    parser.add_argument('--getsinkname', type=int, help='get the appointed sink Name')
+    parser.add_argument('--getsourcename', type=int, help='get the appointed source Name')
+    parser.add_argument('--getsinkcardname', type=int, help='get the card_name of appointed sink')
+    parser.add_argument('--getsourcecardname', type=int, help='get the card_name of appointed source')
+    parser.add_argument('--getsinkdeviceclass', type=int, help='get the appointed sink device.class')
+    parser.add_argument('--getsourcedeviceclass', type=int, help='get the appointed source device.class')
+    parser.add_argument('--getsinkactport', type=int, help='get the appointed sink active port name')
+    parser.add_argument('--getsourceactport', type=int, help='get the appointed source active port name')
+    parser.add_argument('--getsinkportinfo', help='get the appointed sink port information')
+    parser.add_argument('--getsourceportinfo', help='get the appointed source port information')
+    parser.add_argument('--version', action='version', version='%(prog)s 1.0')
+
+    ret_args = vars(parser.parse_args())
+
+    if ret_args['showsinks'] is True:
+        sinks = os.popen("pactl list short sinks").read()
+        for line in sinks.splitlines():
+            print (re.match("(\d+)(.*)", line).groups(1)[0])
+        exit(0)
+
+    if ret_args['showsources'] is True:
+        sources = os.popen("pactl list short sources").read()
+        for line in sources.splitlines():
+            print (re.match("(\d+)(.*)", line).groups(1)[0])
+        exit(0)
+
+    if ret_args.get('getsinkname') is not None:
+        sink = ret_args['getsinkname']
+        sinkinfo = get_sink(str(sink))
+        exit (get_value(sinkinfo, "Name: "))
+
+    if ret_args.get('getsourcename') is not None:
+        source = ret_args['getsourcename']
+        sourceinfo = get_source(str(source))
+        exit (get_value(sourceinfo, "Name: "))
+
+    if ret_args.get('getsinkcardname') is not None:
+        sink = ret_args['getsinkcardname']
+        sinkinfo = get_sink(str(sink))
+        exit (get_value(sinkinfo, "alsa.card_name = "))
+
+    if ret_args.get('getsourcecardname') is not None:
+        source = ret_args['getsourcecardname']
+        sourceinfo = get_source(str(source))
+        exit (get_value(sourceinfo, "alsa.card_name = "))
+
+    if ret_args.get('getsinkdeviceclass') is not None:
+        sink = ret_args['getsinkdeviceclass']
+        sinkinfo = get_sink(str(sink))
+        exit (get_value(sinkinfo, "device.class = "))
+
+    if ret_args.get('getsourcedeviceclass') is not None:
+        source = ret_args['getsourcedeviceclass']
+        sourceinfo = get_source(str(source))
+        exit (get_value(sourceinfo, "device.class = "))
+
+    if ret_args.get('getsinkactport') is not None:
+        sink = ret_args['getsinkactport']
+        sinkinfo = get_sink(str(sink))
+        exit (get_value(sinkinfo, "Active Port: "))
+
+    if ret_args.get('getsourceactport') is not None:
+        source = ret_args['getsourceactport']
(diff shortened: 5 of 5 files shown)


test-case/check-userspace-pa-status.sh
```diff
@@ -0,0 +1,35 @@
...
+
+func_opt_parse_option "$@"
+
+dlogi "check pulseaudio starts correctly or not"
+pid=$(pgrep -x pulseaudio)
+if [ "$?" != "0" ]; then
+    # let's try to start pulseaudio
+    pulseaudio --start
+    sleep 1
+    pid=$(pgrep -x pulseaudio)
```
marc-hb, line 28: No need to search with `pgrep`, `pactl stat` finds `pulseaudio` directly, unambiguously and performs more sanity checks.

test-case/check-userspace-packages.sh
```diff
@@ -0,0 +1,120 @@
...
+OPT_OPT_lst['p']='pulseaudio'     OPT_DESC_lst['p']='pulseaudio pakcages URL location'
+OPT_PARM_lst['p']=1         OPT_VALUE_lst['p']=''
+
+OPT_OPT_lst['u']='ucm'     OPT_DESC_lst['u']='UCM git repo URL location'
+OPT_PARM_lst['u']=1         OPT_VALUE_lst['u']=''
+
+func_opt_parse_option "$@"
+
+dat=`date +%Y%m%d`
+ART_URL=http://sof-ci-storage.sh.intel.com
```
marc-hb, line 32: This needs to be configurable to be usable outside Intel.

test-case/check-userspace-packages.sh
```diff
@@ -0,0 +1,120 @@
...
+## Preconditions:
+##    N/A
+## Description:
+##    check the userspace packages building and install the packages
+## Case step:
+##    1. download latest alsa-lib, pulseaudio debian packages and UCM
+##    2. install the userspace packages and UCM
+##    3. restart pulseaudio
+## Expect result:
+##    All packages and UCM are installed successfully and pulseaudio restarts successfuly
```
marc-hb, line 14: This is an installation script, it does not belong to the `test-case` directory. Can you submit this in a different Pull Request?

marc-hb (changes requested):
To save a lot of review time, please run `shellcheck -x` locally before submitting and start all scripts with `set -e` (see #312)

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,38 @@
...
+##    There is at least one available card for test
+##
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+func_opt_parse_option "$@"
+
+OLDIFS=$IFS
+cardlist=$(pactl list cards short)
```
xiulipan, line 23: Can you add some env check for pactl first? I think the pulseaudio need to be active for this cmd to be run

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,38 @@
...
+((available_card=0))
+IFS=$'\n'
+for card in $cardlist; do
+    dlogi "found card: $(echo $card|awk '{print $2}')"
```
xiulipan, line 27: Also it would be better to have some output example here. If the output format changed, we can easily debug and update our code here.

test-case/check-userspace-pa-status.sh
```diff
@@ -0,0 +1,35 @@
...
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+func_opt_parse_option "$@"
+
+dlogi "check pulseaudio starts correctly or not"
+pid=$(pgrep -x pulseaudio)
+if [ "$?" != "0" ]; then
+    # let's try to start pulseaudio
+    pulseaudio --start
```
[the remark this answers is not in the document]
xiulipan:
```
$ pactl stat
Connection failure: Connection refused
pa_context_connect() failed: Connection refused
```
I see some issue with pactl. If that means pa is not active?

test-case/check-userspace-packages.sh
xiulipan: I do not think this function should not be a test case. This should be some ENV check for our CI instead of a shared test case. Please use dpkg or apt for check only. Install should be in other part.

xiulipan (changes requested):
Please remove the install code part and update other checks.

test-case/check-userspace-pa-status.sh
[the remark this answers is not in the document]
marc-hb:
```
$ pactl stat
Connection failure: Connection refused
pa_context_connect() failed: Connection refused
```
>  If that means pa is not active?

Yes. Good error message;  good test utility.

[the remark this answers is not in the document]
libinyang: @marc-hb Do you mean to using `systemctl start pulseaudio` to start pulseaudio? I failed to start it on my test machine.

@xiulipan Yes, this means pulseaudio may not be started correctly

test-case/check-userspace-cardinfo.sh
libinyang: @xiulipan This test case runs after check-userspace-pa-status. In check-userspace-pa-status, it will check pulseaudio is started correctly or not.

test-case/check-userspace-cardinfo.sh
libinyang: @xiulipan OK, I will add a comment for the output example

libinyang:
> I do not think this function should not be a test case. This should be some ENV check for our CI instead of a shared test case. Please use dpkg or apt for check only. Install should be in other part.
> This is an installation script, it does not belong to the test-case directory. Can you submit this in a different Pull Request?

@xiulipan @marc-hb Thanks for raise this discussion. When I developed the check-userspace-packages test case at first, I did put it in sof-sh repo. From logic view, this should be put in sof-sh repo. However, I met some issues when I was trying to install the packages in sof-sh.
1. install pulseaudio, libasounds serial packages need a lot of dependency. This means we need more check whether they are installed successfully or not. From this point view, it's like a test case to check the dependency issue than just installation.
2. My final solution of this test case is a little complex logic, such as: Install libasound2 in Monday and test, install pulseaudio in Tuesday and test, install UCM in Wednesday and test. With this we can easily tell the failure is because of libasound2 update, pulseaudio update or UCM updated. It seems this logic is too complex for sof-sh.
3. I also want to test 'pulseaudio --kill' 'pulseaudio --start (or as Marc suggested use `systemctl`)' after install the new pulseaudio package. Of course, this should be not a big problem, we can split it a very simple test case which dedicates to test 'start' and 'stop'.

xiulipan:
@libinyang Please split set up test device and test scripts.

1. Setup scripts should be CI stuff, please add those into our CI scripts. 
For the test case dependency, we have `env-check.sh` to check and give advice. Also if the case is optional, we should add check before the test run.
2. Please do not install anything silently. We can add anything into our CI deploy scripts.
3. This should be OK. If the pulseaudio can be controlled now. Otherwise we will have conflict with other test cases.

Please try to set up a sync up meeting for this at any time you prefer

test-case/check-userspace-pa-status.sh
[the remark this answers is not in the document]
marc-hb: > Do you mean to using systemctl start pulseaudio to start pulseaudio?

No, I wrote `systemctl --user`

marc-hb:
1. Packages dependencies issues should be tested when new packages are built. Based on discussions in https://github.com/thesofproject/linux/issues/2211 It doesn't look like you want to move the debian code to build user space packages to sof-test, correct?

> 2. My final solution of this test case is a little complex logic, such as: Install libasound2 in Monday and test, install pulseaudio in Tuesday and test, install UCM in Wednesday and test. 

2. Again based on https://github.com/thesofproject/linux/issues/2211, we don't want to upgrade CI to newer version every week, do we? If correct then each CI upgrade will be manual, ideally performed in a PR. These PRs can be staggered as needed.

3. We should test pulseaudio continuously, not just when upgrading it and independently from it. pulseaudio could break because of a kernel or firmware change.

libinyang:
It seems 2 vs 1 to prefer to put it into the sof-sh. OK. I will move it into sof-sh.

> This should be OK. If the pulseaudio can be controlled now. Otherwise we will have conflict with other test cases.

No, pulseaudio will conflict with other driver test cases. So I setup 2 CI DUT dedicated for user space test.

> Packages dependencies issues should be tested when new packages are built. Based on discussions in thesofproject/linux#2211 It doesn't look like you want to move the debian code to build user space packages to sof-test, correct?

Build machine is different from the DUT. So we won't test the dependencies when building the packages on DUT. I use the debian code to build the alsa lib.

> Again based on thesofproject/linux#2211, we don't want to upgrade CI to newer user space package versions every week, do we? If correct then each CI upgrade will be manual, ideally performed in a PR. These PRs can be staggered as needed.

Based on Mengdong's requirement, this is used to test the upstreamed code. We will use this to make sure the latest alsa lib, pulseaudio and UCM won't break our Intel platform.  
In thesofproject/linux#2211, we offer the stable packages, which are for products release. Only with many tests (including automation test and manual test), these packages can be the stable packages.

> We should test pulseaudio continuously, not just when upgrading it and independently from it. pulseaudio could break because of a kernel or firmware change.  
Yes, this is why we add  the user space test case. Actually, the kernel change may cause UCM error, which will cause user space totally not work. I met such issue before. The user space test cases are not finished. I will submit another 2 test cases today of paplay and parecord.

marc-hb:
> OK. I will move it into sof-sh. [...] I use the debian code to build the alsa lib based on your suggestion.

Is this closed source location where the code to build the packages is right now? In https://github.com/thesofproject/linux/issues/2211 I shared a (pretty small), open-source prototype that anyone can use to rebuild alsa-lib packages. 

> No, pulseaudio will conflict with other driver test cases

I spent some more time trying the `systemctl` commands suggested in other pulseaudio isssues and it looks like it shouldn't be too hard to control pulseaudio like this. I believe no one surprisingly tried to use `systemctl` yet so maybe it's just the matter of using the right tools.

> Build machine is different from the DUT. So we won't test the dependencies when building the packages on DUT.

Both systems should be Ubuntu 20.04 configured similarly so I don't understand which dependencies could be different, can you share an example? SOF developers shouldn't have to use more than one system. They can but they shouldn't have too. https://github.com/thesofproject/linux/pull/2402#issuecomment-683123464

> Based on Mengdong's requirement,...  We will use this to make sure the latest alsa lib, pulseaudio and UCM won't break our Intel platform. In thesofproject/linux#2211, we provide the stable packages, which are for products release

Please write down and submit for review these test cycle and release requirements before submitting the corresponding code. I understand we don't want to wait for official ALSA releases but don't think we want to test every single ALSA commit either, do we?

libinyang:
> Is this closed source location where the code to build the packages is right now? In [thesofproject/linux#2211](https://github.com/thesofproject/linux/issues/2211) I shared a (pretty small), open-source prototype that anyone can use to rebuild alsa-lib packages.

It is closed source location, in sof-sh repo. Not sure attached the code address here is OK?

> 
> > No, pulseaudio will conflict with other driver test cases
> 
> I spent some more time trying the `systemctl` commands suggested in other pulseaudio isssues and it looks like it shouldn't be too hard to control pulseaudio like this. I believe no one surprisingly tried to use `systemctl` yet so maybe it's just the matter of using the right tools.

I didn't mean the conflict of using `systemclt`. I was answering Xiuli, that pulseaudio will conflict with aplay test cases.

> 
> > Build machine is different from the DUT. So we won't test the dependencies when building the packages on DUT.
> 
> Both systems should be Ubuntu 20.04 configured similarly so I don't understand which dependencies could be different, can you share an example? SOF developers shouldn't have to use more than one system. They can but they shouldn't have too. [thesofproject/linux#2402 (comment)](https://github.com/thesofproject/linux/pull/2402#issuecomment-683123464)

I don't have example here. But no example doesn't mean there will never be dependency issues, right? With the pulseaudio updated, it may require the related application's upgrade. If you try to install the latest pulseaudio, I believe you will meet such issue. However, we don't need to take too much effort on it now, this should not be our priority task. And if we really meet such issue, I think we can manually update the system.

> 
> > Based on Mengdong's requirement,...  We will use this to make sure the latest alsa lib, pulseaudio and UCM won't break our Intel platform. In [thesofproject/linux#2211](https://github.com/thesofproject/linux/issues/2211), we provide the stable packages, which are for products release
> 
> Please write down and submit for review these test cycle and release requirements before submitting the corresponding code. I understand we don't want to wait for official ALSA releases but don't think we want to test every single ALSA commit either, do we?

My plan is to test the user space package once every week. Do you think this test frequency is OK? Xiuli talked to me before that we can change to test every single commit.

libinyang:
update:
1. refine the patches based on Xiuli and Marc suggestion
2. remove check packages installation as Xiuli and Marc are strongly against it.
3. add the test case paplay and parecord

marc-hb:
> I was answering Xiuli, that pulseaudio will conflict with aplay test cases.

Hopefully not when disabled with `systemctl --user`. Then we can have a unique DUT configuration for everything - which is what developers need anyway. See previous discussions in https://github.com/thesofproject/sof-test/issues?q=+label%3Apulseaudio+

> I don't have example here. But no example doesn't mean there will never be dependency issues, right? With the pulseaudio updated, it may require the related application's upgrade

So a separate build system is solving a problem we don't know we have and that we haven't tried to solve yet...

> My plan is to test the user space package once every week. Do you think this test frequency is OK? 

That seems reasonable to me but it should be discussed with more people. To keep that particular discussion focused it shouldn't be diluted in this PR or in 2211. Instead it should be submitted as focused, documentation PR. You could also start a new topic at https://github.com/orgs/thesofproject/teams/sof-developers

libinyang:
a separate patch to deploy the user space packages patch has been submitted in sof-sh repo as Marc and Xiuli's feedback.

libinyang:
> 
> Hopefully not when disabled with `systemctl --user`. Then we can have a unique DUT configuration for everything - which is what developers need anyway. See previous discussions in https://github.com/thesofproject/sof-test/issues?q=+label%3Apulseaudio+

I checked the sof-sh. Currently we will remove pulseaudio. So it will not impact on our other DUTs in CI.

test-case/check-userspace-cardinfo.sh
xiulipan: @libinyang I think `check-userspace-pa-status` should be a lib instead of a test case.
Please also consider about single test case run situation.

test-case/check-userspace-parecord.sh
```diff
@@ -0,0 +1,104 @@
...
+if [[ ! -e $file ]]; then
+    dlogw "$file does not exist, use /dev/null as dummy capture output"
+    file=/dev/null
+fi
+
+# TODO: check the parameter is valid or not
+
+# go through all the sources
+# get all the sources name
+sourcekeys=$(pactlinfo.py | jq '.Source'|jq 'keys[]')
```
xiulipan, line 57: do we install jq on all DUTs?

tools/pactlinfo.py
```diff
@@ -0,0 +1,73 @@
...
+# Licensed under GPLv3.
+
+import re
+import os
+import subprocess
+import json
+
+# parse_pactl_list() is modified from the file _audio_data.py
+# in the package apport-symptoms
+def parse_pactl_list(pactlvalues):
```
xiulipan, line 12: A simple question here. If we have a help function, why still need `|` and jq for the string process? Why not do something like `sof-dump-status.py` to use args to directly output info you need.

tools/pactlinfo.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python3
+
+# Licensed under GPLv3.
```
plbossart, line 3: where does this come from? We should not blindly add stuff here, there's also no copyright information?

plbossart (commented):
let's be careful about licenses please?

test-case/check-userspace-parecord.sh
libinyang: No, we don't need install jq on all DUTs. It is only needed on user space DUTs. But I will write a patch to install jq on all DUTs.

tools/pactlinfo.py
libinyang: The script is copied and modified from the file _audio_data.py in the package apport-symptoms. The license is from that file. Is the GPLv2 OK for us?

test-case/check-userspace-cardinfo.sh
libinyang: > A simple question here. If we have a help function, why still need `|` and jq for the string process? Why not do something like `sof-dump-status.py` to use args to directly output info you need.

@xiulipan This is a good question. It's really a hard decision. Finally I decide to make the pactlinfo.py be a simple tool and let the logical analysis in the test case. This is because:
1. The similar function in  apport is “parse_cards()" and it seems it will call parse_cards() when it needs the information and don't have help function either. What's more:
2. We may need many informations from alsactl. If we use helper function, we may need a lot of such functions when we have more and more test cases later.

Both methods have pros and cons. I'm open for the final solution.

libinyang:
patch updated:
1. Move pactl stat check into lib.sh as a helper function
2. Not use jq. Add parameters in pactlinfo.py and make pactlinfo.py return specified information

xiulipan (approved):
Could we merge the playback and capture test later? I think we only need some parameter to split playback and catpure.

libinyang:
> Could we merge the playback and capture test later? I think we only need some parameter to split playback and catpure.

Sure. We can talk about it in detail.

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+if ! func_lib_check_pa; then
+    die "Please check whether pulseaudio runs correctly or not"
+fi
+
+func_opt_parse_option "$@"
+
+OLDIFS=$IFS
+dlogc "pactl list cards short"
+cardlist=$(pactl list cards short)
+((available_card=0))
```
marc-hb, line 30: This bashism not compatible with `set -e`. Change to the POSIX standard
```shell
: $((available_card=0))
```

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_04

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+IFS=$'\n'
+for card in $cardlist; do
+    # pactl list cards short format should be like:
+    # 0	alsa_card.pci-0000_00_1f.3-platform-sof_sdw	module-alsa-card.c
+    dlogi "found card: $(echo "$card" | awk '{print $2}')"
+    usbcard=$(echo "$card" |grep "[Uu][Ss][Bb]")
+    if [ -z "$usbcard" ]; then
+        ((available_card++))
```
marc-hb, line 38: See above

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+    die "Please check whether pulseaudio runs correctly or not"
+fi
+
+func_opt_parse_option "$@"
+
+OLDIFS=$IFS
+dlogc "pactl list cards short"
+cardlist=$(pactl list cards short)
+((available_card=0))
+IFS=$'\n'
```
marc-hb, line 31: Is this really needed?

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+    usbcard=$(echo "$card" |grep "[Uu][Ss][Bb]")
+    if [ -z "$usbcard" ]; then
+        ((available_card++))
+    fi
+done
+IFS=$OLDIFS
+
+if [ "$available_card" == "0" ]; then
+    # TODO: do more check to give hint why there is no available card
+    die "no available card for test" # There should be at least one card available
```
marc-hb, line 45: The comment is just repeating the error message.

test-case/check-userspace-paplay.sh
```diff
@@ -0,0 +1,114 @@
...
+## Expect result:
+##    paplay on the sinks successfully
+##
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# check pulseaudio runs properly or not
+if ! func_lib_check_pa; then
+    die "Please check whether pulseaudio runs correctly or not"
```
marc-hb, line 23: Avoid negations when not needed, see above. Shorter, clearer and looking more like an assert and plain English and like most other shell code:
```shell
func_lib_check_pa || die ...
```

test-case/check-userspace-paplay.sh
```diff
@@ -0,0 +1,114 @@
...
+func_opt_parse_option "$@"
+
+round_cnt=${OPT_VALUE_lst['r']}
+duration=${OPT_VALUE_lst['d']}
+file=${OPT_VALUE_lst['f']}
+format=${OPT_VALUE_lst['F']}
+rate=${OPT_VALUE_lst['R']}
+channel=${OPT_VALUE_lst['C']}
+
+if [[ ! -e $file ]]; then
```
marc-hb, line 53: Avoid negations

test-case/check-userspace-paplay.sh
```diff
@@ -0,0 +1,114 @@
...
+fi
+
+# TODO: check the parameter is valid or not
+
+# go through all the sinks
+# get all the sinks name
+sinkkeys=$(pactlinfo.py --showsinks)
+for round in $(seq 1 $round_cnt); do
+    for i in $sinkkeys; do
+        if ! sinkcard=$(pactlinfo.py --getsinkcardname "$i"); then
```
marc-hb, line 65: negation

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+## Preconditions:
+##    N/A
+## Description:
+##    Get the card name from pactl info
+##    check if there is an available card to do the following tests
+## Case step:
+##    1. run pactl to get the cards info
+##    2. check the card name
+## Expect result:
+##    There is at least one available card for test
```
marc-hb, line 14: I don't see what this test achieves. Other tests also check that pulseaudio is functional and they should also fail if zero card is found so I think this test can be removed.

test-case/check-userspace-paplay.sh
```diff
@@ -0,0 +1,114 @@
...
+            dlogw "failed to get sink $i card_name"
+            continue
+        fi
+
+        # Let's skip testing on USB card
+        usbcard=$(echo "$sinkcard" |grep "[Uu][Ss][Bb]")
+        if [ -n "$usbcard" ]; then
+            continue
```
marc-hb, line 73: usbcard is not needed, it's confusing.
```shell
if echo "$sinkcard" |grep '[Uu][Ss][Bb]'; then
```

tools/pactlinfo.py
marc-hb: >  Is the GPLv3 OK for us?

No, not in this case.

Even if it were OK you should never ever remove a copyright.

GPL aside, code re-use does not mean "copy/paste", you should try to import it first.

When really impossible re-use as is and copy cannot be avoided you should first commit the initial version and then the modifications in a different commit for version tracking purposes.

Pretty much everything is wrong.

marc-hb (changes requested):

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,46 @@
...
+##
+
```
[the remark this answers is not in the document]
libinyang: grep will fail if `set -e`.

test-case/check-userspace-cardinfo.sh
libinyang: there are several cards and we only care sof cards. Other user space test cases won't check it and return correctly.

case-lib/lib.sh
```diff
@@ -252,6 +252,16 @@ func_lib_get_tplg_path()
...
 }
 
+func_lib_check_pa()
+{
+    if ! pactl stat &>/dev/null; then
+        dloge "pulseaudio doesn't start correctly"
+        return 1
+    fi
+
+    return 0
```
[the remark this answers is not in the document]
libinyang: In the test cases, why do we care the interactive shell?

test-case/check-userspace-cardinfo.sh
libinyang: OK, I will remove it

test-case/check-userspace-paplay.sh
libinyang: will fix it in next version

tools/pactlinfo.py
libinyang: If so, it seems we can't reuse the codes. The original file is much complicated. Imposing it directly will cause failure, I did a little change in the original code. "When really impossible re-use as is and copy cannot be avoided you should first commit the initial version and then the modifications in a different commit for version tracking purposes." is a good suggestion.

Anyway, as we are concerning about the GPL. I will try to redesign the code and won't user GPLv3.

test-case/check-userspace-paplay.sh
xiulipan:
```suggestion
[[ -e $file ]] || (dlogw "$file does not exist, use /dev/zero as dummy playback source" && file=/dev/zero )
```

xiulipan (changes requested):
Please apply some suggestion from marc and run spellcheck for advice.
PS: we may take tare about scripts from other repo
If it is just some string handler and conversion, maybe write a version by our self is better.

case-lib/lib.sh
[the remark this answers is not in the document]
libinyang: In my new patch, I will use :
pactl stat &>dev/null || {
dloge ...
return 1
}

I still use "&>dev/null" because error message `pactl stat` doesn't show explicit information that `pactl stat` fails. It doesn't help for developer to identity which fails quickly. Instead, I will use `dloge` to export the useful information for sof-test developer.

libinyang:
patch updated:
1. fix the license issue
2. switch `if ! ...` to other code style
3. refine the code based on other review comments.

`set -e` is still not used because if it is used, the case will fail to run.

test-case/check-userspace-parecord.sh
```diff
@@ -0,0 +1,118 @@
...
+func_opt_parse_option "$@"
+
+round_cnt=${OPT_VALUE_lst['r']}
+duration=${OPT_VALUE_lst['d']}
+file=${OPT_VALUE_lst['f']}
+format=${OPT_VALUE_lst['F']}
+rate=${OPT_VALUE_lst['R']}
+channel=${OPT_VALUE_lst['C']}
+
+[[ -e $file ]] || { dlogw "$file does not exist, use /dev/zero as dummy playback source" && file=/dev/zero; }
```
xiulipan, line 51: capture should use `dev/null` as dummy file to write into.

test-case/check-userspace-parecord.sh
```diff
@@ -0,0 +1,118 @@
...
+sourcekeys=$(pactlinfo.py --showsources)
+for round in $(seq 1 $round_cnt); do
+    for i in $sourcekeys; do
+        sourcecard=$(pactlinfo.py --getsourcecardname "$i") || {
+            dlogw "failed to get source $i card_name"
+            continue
+        }
+
+        # Let's skip testing on USB card
+        if echo "$sourcecard" |grep "[Uu][Ss][Bb]" &>/dev/null; then
```
xiulipan, line 66: `grep -i usb` is enough
PS: do we need a ignore list or filter list? Three will also be legacy HDA card and other graphic HDMI card.

xiulipan (changes requested):

libinyang:
patch updated to add `set -e`

test-case/check-userspace-cardinfo.sh
```diff
@@ -0,0 +1,41 @@
...
+OLDIFS=$IFS
+dlogc "pactl list cards short"
+cardlist=$(pactl list cards short)
+: $((available_card=0))
+IFS=$'\n'
+for card in $cardlist; do
+    # pactl list cards short format should be like:
+    # 0	alsa_card.pci-0000_00_1f.3-platform-sof_sdw	module-alsa-card.c
+    dlogi "found card: $(echo "$card" | awk '{print $2}')"
+    echo "$card" |grep "[Uu][Ss][Bb]" &>/dev/null || ((available_card++))
```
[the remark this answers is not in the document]
libinyang: yes. USB is not our scope

test-case/check-userspace-parecord.sh
libinyang: I added a "TODO" for the filter list. Currently we don't meet such platform. How about we implement it when we meet such platforms?

test-case/check-userspace-parecord.sh
libinyang: good catch. Thanks. Fixed in the latest update

libinyang: > https://unix.stackexchange.com/questions/275794/iterating-over-multiple-line-string-stored-in-variable
echo "$variable" | while IFS= read -r line ; do echo $line; done

I don't why: if I use this such while xxx, the 'available_card' seems to be changed locally and it will remain to 0 out of the loop. So I still keep the original code.

test-case/check-userspace-cardinfo.sh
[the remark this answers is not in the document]
marc-hb: See detailed `set -e` guidelines in #312

test-case/check-userspace-cardinfo.sh
marc-hb: > Other user space test cases won't check it and return correctly.

This is wrong, all tests that need an sof card to test should either FAIL or SKIP (exit 2) when there is no SOF card. A test should never return PASS after testing nothing.

marc-hb (commented):
> In the test cases, why do we care the interactive shell?

It's often useful to copy/paste to the shell for interactive testing. Anyway that is not the main point: it's just good  practice to avoid negations because they require a little bit more reading/review effort. Avoid negations as a good practice also tends to avoid double negations (e.g: if `! nocodec_cmode` ) which require even more reading effort for no value. I don't think I would not approve a PR just because of negations (see what I just did?)

> I don't why: if I use this such while xxx, the 'available_card' seems to be changed locally and it will remain to 0 out of the loop. So I still keep the original code.

Please try to debug this, "I don't why it does not work" is not acceptable. If you can isolate and reproduce in a small function then share that function and I will debug it.

marc-hb (changes requested):
> I still use "&>dev/null" because error message pactl stat doesn't show explicit information that pactl stat fails. It doesn't help for developer to identity which fails quickly. Instead, I will use dloge to export the useful information for sof-test developer.

- pactl error messages can be googled. So even when they're really really bad (which I suspect is rare), they are still useful because the reader can very quickly see what command failed: a pactl command.

- If you think some pactl error message is really bad then feel free to ADD your own message but keep the original pact message too.

- Last but not least: how can you even know in advance all the possible error messages that this pactl command will produce? Did you read the entire pactl source code? People look at test logs only when something unexpected happened, do you have a complete list of all the unexpected things that can happen?

Never, ever discard stderr, period.

Don't discard stdout either before it has become an actual annoyance. Even in that case it's best to redirect it to a file.

This is test code here, we're not trying to build slick iPhone interfaces. No one looks at test logs unless the tests fail. When tests fail people need and want as much  original and googlable information as possible, they absolutely do not want alternative error messages hiding the original messages. No one care about "beautiful" test logs.

marc-hb:
Can you please start a new PR? This one is overloaded with many comments that don't apply any more.

Please also run `pylint` on the new Python script. It doesn't have to be completely warning-free but it should have a low enough number.

We don't have a defined policy for formatting python code but I recommend `black`

libinyang:
> try to debug this, "I don't why it does not work" is not

Answer from Google shows: When you pipe into a while loop in Bash, it creates a subshell.
https://stackoverflow.com/questions/4667509/shell-variables-set-inside-while-loop-not-visible-outside-of-it

libinyang:
> * pactl error messages can be googled. So even when they're really really bad (which I suspect is rare), they are still useful because the reader can very quickly see what command failed: a pactl command.
> * If you think some pactl error message is really bad then feel free to ADD your own message but keep the original pact message too.
> * Last but not least: how can you even know in advance all the possible error messages that this pactl command will produce? Did you read the entire pactl source code? People look at test logs only when something unexpected happened, do you have a complete list of all the unexpected things that can happen?

Do we need to understand entire source code when we use the tool? Sorry, I'm surprised to know that. What I think is that we and the user don't need to debug the pactl, pulseaudio tools. So I don't export all the error message to users. When I hide the error message, I don't say the error message is useless. The error message is useful to root cause why `pactl stat` fails. But we don't want to debug why `pactl stat` fails, do we? My opinion is to decide whether to show the error message case by case. It's not a good idea to mandate to always show the error messages or never show the error messages.

libinyang:
> Can you please start a new PR? This one is overloaded with many comments that don't apply any more.

Sure, I will create a new PR later.

> 
> Please also run `pylint` on the new Python script. It doesn't have to be completely warning-free but it should have a low enough number.

Good suggestion. I will try this tool. 

> 
> We don't have a defined policy for formatting python code but I recommend `black`

Sorry, I don't understand `black` means here. Could you please give a little more hints?

libinyang:
> It's often useful to copy/paste to the shell for interactive testing. Anyway that is not the main point: it's just good practice to avoid negations because they require a little bit more reading/review effort. Avoid negations as a good practice also tends to avoid double negations (e.g: if `! nocodec_cmode` ) which require even more reading effort for no value. I don't think I would not approve a PR just because of negations (see what I just did?)

Just find `if ! mycommand; then ...` is a recommend format in https://github.com/koalaman/shellcheck/wiki/SC2181

libinyang:
open a new thread https://github.com/thesofproject/sof-test/pull/479 and close this one

marc-hb:
> Answer from Google shows: When you pipe into a while loop in Bash, it creates a subshell.
https://stackoverflow.com/questions/4667509/shell-variables-set-inside-while-loop-not-visible-outside-of-it

There are a number of ways to solve this: https://mywiki.wooledge.org/BashFAQ/024

See process substitution example in `tools/CI/shellcheck-gitrange.bash` (commit faadb20a8548)


> Sorry, I don't understand black means here. Could you please give a little more hints?


apt-get install black
black something.py # re-formats in place


> Just find if ! mycommand; then ... is a recommend format in https://github.com/koalaman/shellcheck/wiki/SC2181

No, that page does not "recommend" this format because that page is about `$?` which is a different issue. This page uses `if ! command` because it's focuses between the difference between the "bad" and "good" code for the different 
issue and it does not want to distract the page from a much more minor and unrelated codestyle issue.

`if ! assert_something; then fail` works fine but `assert_something || fail` is more readable for a number of reasons, most that I already explained a couple times here and there. The main reason is that it reads like plain English ("do or die") and looks more like a C assert. It's also one of the most common shell idioms, try `grep -R -F ' || ' /usr/bin/` and see other examples in

https://github.com/koalaman/shellcheck/wiki/SC2164
https://github.com/koalaman/shellcheck/wiki/SC2065
https://github.com/koalaman/shellcheck/wiki/SC2103

https://mywiki.wooledge.org/BashPitfalls#cd_.2Ffoo.3B_bar


Code reviews include both required changed and not required changes, "avoid" negations is not required change, it's general advice to make your submissions easier and faster to review.

> set -e is still not used because if it is used, the case will fail to run.

I added detailed guidelines and some tips in #312. If that's not enough for some reason please ask me for help. I think this is very important so I will prioritize it.

marc-hb:
> Do we need to understand entire source code when we use the tool? 

Only if you discard stderr. Don't discard stderr if you don't want to have understand the whole `pactl` source code.

> So I don't export all the error message to users. When I hide the error message, I don't say the error message is useless.

You've seen ONE error message in one situation and you're wrongly assuming pactl has no other error message. The entire purpose of test code is to find as many errors as possible in as many places and situations as possible so considering only "the" one error message you've seen so far makes no sense.


> But we don't want to debug why pactl stat fails, do we? 

This PR was named "add user space test cases" yet you would do not want to observe and try to understand some unexpected pactl failure?!


>  It's not a good idea to mandate to always show the error messages or never show the error messages.

Especially for _test code_  yes: it is a "good idea" to never discard stderr except in some extremely rare (and well justified) cases. In fact it's absolutely critical for test code.

The user interface of test code is the exact opposite of the user interface of an iPhone: it has to be 100% transparent so users can totally and immediately see through absolutely everything it does. Users look at test logs ONLY when something unexpected happened and then they want to minimize guessing time as much as possible.

Such complete transparency is even more important when running in CI because it's often difficult to reproduce issues in different environment with no ability to interact directly.

marc-hb:
> The user interface of test code is the exact opposite of the user interface of an iPhone: it has to be 100% transparent so users can totally and immediately see through absolutely everything it does. Users look at test logs ONLY when something unexpected happened and then they want to minimize guessing time as much as possible.

> Such complete transparency is even more important when running in CI because it's often difficult to reproduce issues in different environment with no ability to interact directly.

Only one day later here's one example why sof-test should be as close as possible to interactive commands : https://github.com/thesofproject/sof-test/issues/495#issuecomment-722538385

PR #362, opened 2020-09-03
ignore_str filter updates from Pierre

marc-hb:
Re-submitting the first part of #354 . This only updates the `ignore_str` filters and does not have any actual code change.

PR #365, opened 2020-09-08
tools: ignore false error message for ICL

aiChaoSONG:
On ICL, the info "sof-audio-pci 0000:00:1f.3:
status = 0x00000000 panic = 0x00000000" causes
test failure because we detect the word "panic".

This patch ignores this pattern.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -50,6 +50,13 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
 ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. Restarting 0x0\.'
 ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
 
+# Test cases on some platforms fail because the false error message:
+#    sof-audio-pci 0000:00:1f.3: status = 0x00000000 panic = 0x00000000
+# Note that different platform may have different PCI ID, and the panic code
+# may not be 0x00000000.
+# Buglink: https://github.com/thesofproject/sof/issues/3395
+ignore_str="$ignore_str"'|sof-audio-pci 0000:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]: status = 0x[0-9a-f]{8} panic = 0x[0-9a-f]{8}'
+
 [[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
 # dmesg KB size buffer size
 #dmesg_config_define=$(awk -F '=' '/CONFIG_LOG_BUF_SHIFT/ {print $2;}' /boot/config-$(uname -r))


marc-hb (changes requested):
- I don't think we want to ignore every panic message like this but only the ones with 
`status = 0x00000000 panic = 0x00000000`.

- Are we sure all ICL devices are located on PCI `0000:00:1f.3`?

- @kv2019i do we have a bug about this?

- Can you please help review and merge #362 first? I bet it will conflict with this. #362 has been submitted for longer and this is much smaller so it will easier to rebase.

tools/sof-kernel-log-check.sh
```diff
@@ -40,6 +40,10 @@ ignore_str="$ignore_str"'|PM: Device i2c-DELL0955:00 failed to resume async: err
 # https://github.com/thesofproject/sof-test/issues/348
 ignore_str="$ignore_str"'|da7219 i2c-DLGS7219:00: SRM failed to lock'
 
+# Test cases on ICL fail because the false error message:
+#    sof-audio-pci 0000:00:1f.3: status = 0x00000000 panic = 0x00000000
```
xiulipan, line 44: We do see other fail on other devices with random status.
Please use regex for the pci id and dumped register value

xiulipan (changes requested):

aiChaoSONG:
@marc-hb 
> *I don't think we want to ignore every panic message like this but only the ones with `status = 0x00000000 panic = 0x00000000`.

Pattern like `status = 0x???????? panic = 0x????????` is just info, not error, and as Xiuli stated, we do see other code, so we need to ignore this pattern.

> Are we sure all ICL devices are located on PCI `0000:00:1f.3`

Yes, all ICL devices have the same PCI ID: 0000:00:1f.3, but other platforms may have a different PCI ID, and false error message also seen on other platform, so here I will use an RE also.

marc-hb:
> Pattern like status = 0x???????? panic = 0x???????? is just info, not error, 

Then why not just add `--level` like this in the right place:
```
dmesg --level=err,warn
```

Grepping for "panic" in **info** messages seems silly.

(In fact searching for keywords _at all_ instead of using levels seems a bit silly but let's not go there yet)

tools/sof-kernel-log-check.sh
aiChaoSONG: fixed

aiChaoSONG:
> why not just add --level like this in the right place

I regard this PR (or even other PR related to ignore some specified message) as a workaround. Just as you said, the best way is to use journalctl.  I will try `dmesg --level`, and see if it has bad impact. The only concern I have is that we may lose error context.

tools/sof-kernel-log-check.sh
```diff
@@ -50,6 +50,12 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
 ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. Restarting 0x0\.'
 ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
 
+# Test cases on some platforms fail because the false error message:
+#    sof-audio-pci 0000:00:1f.3: status = 0x00000000 panic = 0x00000000
```
kv2019i, line 54: @aiChaoSONG Please add a reference to https://github.com/thesofproject/sof/issues/3395 in the comment for this filter.

kv2019i (changes requested):
Please add a reference to the bug in the filter comment, but with that addressed, this seems good to go for me.

marc-hb:
> The only concern I have is that we may lose error context.

Agreed: we should use `--level` to _decide_ whether there is an error or not but we also want to _collect_ all levels anyway for the record. If this requires a big code change then this is off-topic, forget it for now.

`journalctl` will have the same requirement.

BTW #361 could be helpful here.

tools/sof-kernel-log-check.sh
aiChaoSONG: @kv2019i Audio controller on different platforms, they have dirrerent PCI ID.
BYT : `00:1b.0 Audio device: Intel Corporation Atom Processor Z36xxx/Z37xxx Series High Definition Audio Controller (rev 11)
`
APL: `00:0e.0 Multimedia audio controller: Intel Corporation Celeron N3350/Pentium N4200/Atom E3900 Series Audio Cluster (rev 0b)`
BDW: `00:03.0 Audio device: Intel Corporation Broadwell-U Audio Controller (rev 09)`
CML/TGL...: `00:1f.3 Multimedia audio controller: Intel Corporation Device a0c8 (rev 20)`

aiChaoSONG: @kv2019i Bug reference added.

xiulipan (approved):

xiulipan:
@kv2019i @marc-hb This workaround LGTM, so I will merge.
Let's force on the journalctl now. I think with journalctl we can have full support for the level and ignore the dmesg size limitation.

xiulipan (approved):

ranj063:
@aiChaoSONG can you please add a fix to this change to only ignore this message when it is prefixed by "error:" and not otherwise?

aiChaoSONG:
> @aiChaoSONG can you please add a fix to this change to only ignore this message when it is prefixed by "error:" and not otherwise?

We have to ignore this pattern even it is not prefixed by "error", because the word "panic" also causes test case failure.

marc-hb:
> Pattern like status = 0x???????? panic = 0x???????? is just info, not error, and as Xiuli stated, we do see other code, so we need to ignore this pattern.

Since this message appeared (see https://github.com/thesofproject/sof/issues/3395 for some history) I've never seen any other values than `status = 0x00000000 panic = 0x00000000`. Do you have examples of other values? I mean examples that should also be ignored for some (other?) well known reason? Which other reason besides 3395?

marc-hb:
> I will try dmesg --level, and see if it has bad impact.

@aiChaoSONG did you try `--level` before @xiulipan  merged this?

> Let's force on the journalctl now.

Do you mean "accelerate"?

> I think with journalctl we can have full support for the level 

I'm not sure what better `--level` option `journalctl` offers that we would use. The value of `journalctl` is (as you wrote) unlimited log size and also interleaving with user space errors.

aiChaoSONG:
@marc-hb I am afraid it doesn't work. 

`[[ "$begin_line" -eq 0 ]] && cmd="dmesg || cmd="sed -n '$begin_line,\$p' /var/log/kern.log"`

dmesg is used only when there is no new logs in /var/log/kern.log, even I add `dmesg --level=warn` here, it may not function.

that's why we want journalctl to replace all the log thing.

kv2019i:
@ranj063 wrote: 
> @aiChaoSONG can you please add a fix to this change to only ignore this message when it is prefixed by "error:" and not otherwise?

You meant the reverse, right @ranj063 ? We want to raise CI errors only if this trace is printed with "error:" prefix. I..e we should filter out traces that do _not_ have "error:" (the 0...2 attempts).

marc-hb:
> You meant the reverse, right @ranj063 ?

@ranj063 confirmed she did the reverse in another, private conversation. Anyway we don't have levels in `kern.log` so this will have to wait for `journalctl` as clarified by @aiChaoSONG

marc-hb:
After some more online and offline discussions it's clear this PR ignores way too many possible errors / potential "green failures". Also, hiding which platforms experience the issue and what status they print is not going to help address it in the long run. So I'm making this exclude much narrower in #372 so we can spot and record other occurrences, see source code comment there (@xiuli please wait until that all questions are answered and all discussions are finished before merging a PR)

BTW this is not a "false" error, this is a _known_ error.

PR #366, opened 2020-09-08
sof-kernel-log-check: fix last_line / begin_line off-by-one error

marc-hb:
If the last line of kern.log was an error then the _next_ test was failing.

Found thanks to the new fake_kern_error() function submitted separately.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #367, opened 2020-09-08
hijack.sh: fix sudo test for cached credentials or NOPASSWD

marc-hb:
On my system sudo fails to use cached credentials when invoked from
"expect" for some unknown reason. Rather than debugging this, let's
replace the 9 lines long `expect` test with a much simpler one-line "sudo
--non-interactive" test which works perfectly.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -123,16 +123,9 @@ func_hijack_setup_sudo_level()
     [[ "$SUDO_LEVEL" ]] && return 0
     # root permission, don't need to check
     [[ $UID -eq 0 ]] && SUDO_LEVEL=0 && return 0
-    # now check whether we need sudo passwd using expect
-    if expect >/dev/null <<END
-spawn $SUDO_CMD ls
-expect {
-    "password" {
-        exit 1
-    }
-exit 0
-}
-END
+
+    # Test for either cached credentials or NOPASSWD
+    if $SUDO_CMD --non-interactive true
     then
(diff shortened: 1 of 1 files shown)


xiulipan:
@marc-hb From your fix for the lib also see https://github.com/thesofproject/sof-test/pull/366 , I have some proposal for unit test for lib https://github.com/thesofproject/sof-test/issues/370
Could you try to have some test scripts for our lib? You can start with those 2 fail case you found first. This will make sure our future change do not bring regression.

marc-hb:
I'm not sure I understand the scope of #370 . In the mean time can you please review this? It merely replaces a complex, 9 lines-long expect script that doesn't work with a dead simple one-line `sudo` invocation that does.

case-lib/hijack.sh
```diff
@@ -123,16 +123,9 @@ func_hijack_setup_sudo_level()
...
-    "password" {
-        exit 1
-    }
-exit 0
-}
-END
+
+    # Test for either cached credentials or NOPASSWD
+    if $SUDO_CMD --non-interactive true
     then
```
xiulipan, line 129: Do we need else here to `exit 1`?

xiulipan (changes requested):
Some simple question here, do we need to exit 1 here?

case-lib/hijack.sh
marc-hb: The `exit` I removed was an `expect` command, not a `shell` command.

xiulipan (approved):

PR #369, opened 2020-09-08
Make func_lib_check_sudo() a warning cause sudo() message is better

marc-hb:
See long comment in the function header.

Also remove very bad SUDO_PASSWD suggestion from the error message and
replace it with less insecure NOPASSWD and cache credentials
suggestions.


marc-hb:
The fix for the single shellcheck warning in https://travis-ci.org/github/thesofproject/sof-test/builds/725414147 was already submitted in #361

marc-hb:
The only failure device-test failure is ICL issue #365

marc-hb:
Pure rebase, no code change.

xiulipan (approved):

PR #371, opened 2020-09-09
[test]verify-sof-firmware-load: add debug logs for fail case

xiulipan:
Dump last 50 lines from journalctl to check why the test fail.

Add debug logs to check why the case failed

### DIFF
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -33,5 +33,8 @@ if $cmd | grep -q " sof-audio.*version"; then
     $cmd | grep "Firmware debug build" -A3 | head -n 12
     exit 0
 else
+    journalctl --dmesg --lines 50
+    journalctl --dmesg | grep -C 1 -i version
+    printf '$cmd was %s\n' "$cmd"
     die "Cannot find the sof audio version"
 fi


test-case/verify-sof-firmware-load.sh
```diff
@@ -33,5 +33,6 @@ if $cmd | grep -q " sof-audio.*version"; then
     $cmd | grep "Firmware debug build" -A3 | head -n 12
     exit 0
 else
+    journalctl --dmesg |tail -50
```
marc-hb, line 36: `journalctl` has a (faster) built-in `--lines` option.

Print the value of `$cmd` too, bugs in `sof-kernel-dump.sh` are very likely considering how ugly the code is so we want to know which one ran.
```shell
journalctl --dmesg --lines 50
journalctl --dmesg | grep -C 1 -i version
printf '$cmd was %s\n' "$cmd"
```

marc-hb (changes requested):

test-case/verify-sof-firmware-load.sh
xiulipan: thanks for the suggestion, will update.

marc-hb (approved):

PR #372, opened 2020-09-10
@marc-hb sof-kernel-log-check: narrower ignore_str for ICL reboot issue

marc-hb:
2 commits. The main one:

Fixes ac415de ("tools: ignore a false error message with 'panic'")
which was ignoring way too many errors. See previous PR and long source
code comment.

Generally speaking, ignoring errors ("green failures") is extremely
dangerous and should be as narrow as possible.

More specifically here, not even knowing which platforms experience the
issue and what code they print is is really not going to help fix it.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -2,6 +2,9 @@
 
 begin_line=${1:-1}
 declare err_str ignore_str project_key
+
+platform=$(sof-dump-status.py -p)
+
 err_str="error|failed|timed out|panic|oops"
 
 # TODO explain
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -53,12 +53,29 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
 ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. Restarting 0x0\.'
 ignore_str="$ignore_str"'|wlo1: authentication with f4:f5:e8:6b:45:bb timed out'
 
-# Test cases on some platforms fail because the false error message:
+# Test cases on some platforms fail because the boot retry message:
+#
 #    sof-audio-pci 0000:00:1f.3: status = 0x00000000 panic = 0x00000000
-# Note that different platform may have different PCI ID, and the panic code
-# may not be 0x00000000.
+#    ...
+#    Attempting iteration 1 of Core En/ROM load...
+#
+# Despite the real boot failure the retry message is not at the error
+# level until after the last try. However we still use kern.log for now
+# and it has no log levels, so this may unfortunately hide this same
+# message at the 'error' level until we switch to journalctl
+# --priority. Hopefully other issues will cause the test to fail in that
+# case.
+#
+# For now the codes seem to be 0x00000000 and affected platforms have
+# PCI ID 1f.3. Before adding other values make sure you update the list
+# of affected systems in bug 3395 below.
+#
 # Buglink: https://github.com/thesofproject/sof/issues/3395
-ignore_str="$ignore_str"'|sof-audio-pci 0000:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]: status = 0x[0-9a-f]{8} panic = 0x[0-9a-f]{8}'
+case "$platform" in
+    icl|cml)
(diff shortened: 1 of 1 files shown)


marc-hb:
This PR doesn't not add any shellcheck warning in https://travis-ci.org/github/thesofproject/sof-test/builds/725765539, they were all already there.

I didn't to stress test ICL to test this change, #361 was enough. Please help review #361 too.

xiulipan:
@marc-hb I will close this PR. The issue can be seen on any device, but the reproduce rate on ICL is very high. So we should ignore the info for all platforms.

marc-hb:
> @marc-hb I will close this PR. 

You merged #365 before all discussions were finished and now you're closing this one before any discussion has started.

> but the reproduce rate on ICL is very high. So we should ignore the info for all platforms.

I can't see the how the second sentence relates to the first one.

("very high" doesn't mean much without some numbers)

xiulipan:
#365 is only a workaround. I did not think more discussion is needed for that. I think we have plan to use log level to filter the warning and error message to avoid keyword detection.

As the panic log is not platform specific one, I think we should ignore this info message for all platforms.

marc-hb:
> As the panic log is not platform specific one, I think we should ignore this info message for all platforms.

I've never seen it outside ICL across many PRs and the description of https://github.com/thesofproject/sof/issues/3395  is very specifically about ICL.

If/when it happens elsewhere then we want to know exactly when and how and not ignore failures "just in case they might happen". This code makes it trivial to add other platforms later, it's designed for that.

aiChaoSONG:
@marc-hb It also happened to CML_HEL_RT5682, check the failure on CML_HEL_RT5682 in https://sof-ci.sh.intel.com/#/result/planresultdetail/319

marc-hb:
> @marc-hb It also happened to CML_HEL_RT5682, check the failure on CML_HEL_RT5682 in https://sof-ci.sh.intel.com/#/result/planresultdetail/319

Thanks @aiChaoSONG , I just added `cml`. It took 4 additional characters.

I also updated https://github.com/thesofproject/sof/issues/3395

xiulipan:
@marc-hb As the panic info may show on any platforms, why we need to narrower the ignore list? I do not want to add a new platform name here once we see it on another platform.

marc-hb:
> I do not want to add a new platform name here once we see it on another platform.

People interested in finding and fixing issues are interested in when, where, how and how often they happened. They don't mind adding 4 characters in one file for that sort of information, that's a small price to pay.

marc-hb:
> People interested in finding and fixing issues are interested in when, where, how and how often they happened. They don't mind adding 4 characters in one file for that sort of information, that's a small price to pay.

Anyway this is not even the main point of this PR. The main purpose here is not to miss OTHER issues unrelated to https://github.com/thesofproject/sof/issues/3395

ranj063 (dismissed):
looks good. the commit title only talks of ICL when the patch is for ICL and CML. only minor though

tools/sof-kernel-log-check.sh
```diff
@@ -53,12 +53,29 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
...
+#
+# For now the codes seem to be 0x00000000 and affected platforms have
+# PCI ID 1f.3. Before adding other values make sure you update the list
+# of affected systems in bug 3395 below.
+#
 # Buglink: https://github.com/thesofproject/sof/issues/3395
-ignore_str="$ignore_str"'|sof-audio-pci 0000:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]: status = 0x[0-9a-f]{8} panic = 0x[0-9a-f]{8}'
+case "$platform" in
+    icl|cml)
+        ignore_str="$ignore_str"'|sof-audio-pci 0000:00:1f.3: status = 0x[0]{8} panic = 0x[0]{8}'
```
plbossart, line 76: don't you need a \ before the .?

marc-hb: It won't make any difference in this very particular case because the kernel will always print the exact same character in this position. However it is  good reminder for other cases when matching an actual dot matters so I'm adding it.

plbossart (approved):

ranj063 (approved):

aiChaoSONG:
Merge this for `sof-audio-pci 0000:00:1f.3: status = 0x00000000 panic = 0x00000000` is only seen on ICL and CML.  If we see this on  more platforms, we can add later.

Issue #297, opened 2020-07-28
Check for a valid trace for ALL tests


plbossart:
We need to double-check that the trace is functional and that sof-logger reports at least the firmware details and a dai trigger.

If there is no trace then there's no point in checking results and debugging further.

cc:
- #1075
- https://github.com/thesofproject/sof/pull/7955
- #1075
- https://github.com/thesofproject/sof-test/pull/1084

+ many others, follow the links.

marc-hb:
First part submitted in #373, please help review.

marc-hb:
From @xiulipan :
> it that behavior right? Do we need to fail the case if we could not get the logger? As we have standalone test case for logger.

marc-hb:
> Do we need to fail the case if we could not get the logger? 

I think what you're asking is: "is the logger a low priority feature"? I read the description of this request as a pretty clear "no" but I'd like to hear more from others.

> As we have standalone test case for logger.

It's risky to rely on a single test, if it's broken then the feature can be broken and no one will notice.  Even if the test is not broken, the logger could be broken only in some scenarios not tested by the logger test. I think the links above are some real-world examples of either.

A test is never meant to test only one thing, the more extras each test can test (in no extra time) and the better. Unexpected failures mean extra bugs are found which is great.

These questions sounds strange to me at a time when the logger seems to be working fine, I mean  all tests in #373 pass right now. I mean it sounds very strange to worry about failures we don't even have right now (this reminds me of recent #372) It almost sounds like like a desire to find FEWER bugs / and have more "green failures"?

xiulipan:
@marc-hb What I want to make clarify about is the **criterion** for our test case pass/fail.
Should we make aplay test fail if a debug feature is not working?

> A test is never meant to test only one thing, the more extras each test can test (in no extra time) and the better. 

I do not think we are adding test cases in this way. Each test case in this repo is focusing in a single feature. This will help us to narrow down the issue.

marc-hb:
> I do not think we are adding test cases in this way. Each test case in this repo is focusing in a single feature. This will help us to narrow down the issue.

This is :
- neither possible because sof-test are not unit tests. They all run from user space so each test involves many lines of code
- nor desirable unless you want to find fewer bugs - as it happened in the links above and elsewhere.

marc-hb:
>  I mean all tests in #373 pass right now.

[...]

PR #373, opened 2020-09-10
Stop ignoring various logger failures

marc-hb:
2 commits, see commit messages.

Green failures episode "the logger"

Incomplete fix for #297 , more coming.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -30,12 +30,23 @@ function func_exit_handler()
             func_lib_start_log_collect 1
             sleep 1s
         }
-        local loggerBin; loggerBin=$(basename "$SOFLOGGER")
-        sudo pkill -9 "$loggerBin" 2>/dev/null || {
+
+        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
+        # INT doesn't print any "Killed" message in non-interactive mode
+        sudo pkill -INT "$loggerBin" || {
             dloge "sof-logger was already dead"
             exit_status=1
         }
         sleep 1s
+        if pgrep "$loggerBin"; then
+            dloge "$loggerBin resisted pkill -INT, using -KILL"
+            sudo pkill -KILL "$loggerBin"
+            exit_status=2
+        fi
+        # logfile is set in a different file: lib.sh
+        # shellcheck disable=SC2154
+        wcLog=$(wc -l "$logfile")
+        dlogi "nlines=$wcLog"
     fi
     # when case ends, store kernel log
     # /var/log/kern.log format:
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -92,7 +92,7 @@ func_lib_start_log_collect()
         return 1
     }
 
-    local logfile="" logopt="-t"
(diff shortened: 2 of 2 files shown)


case-lib/hijack.sh
```diff
@@ -30,12 +30,23 @@ function func_exit_handler()
...
             sleep 1s
         }
-        local loggerBin; loggerBin=$(basename "$SOFLOGGER")
-        sudo pkill -9 "$loggerBin" 2>/dev/null || {
+
+        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
+        # logfile is set in a different file: lib.sh
+        # shellcheck disable=SC2154
+        wcLog=$(wc -l "$logfile")
+        dlogi "nlines=$wcLog"
```
xiulipan, line 38: maybe a more readable log here? This looks like a cmd.

xiulipan (changes requested):
Do we need any check if logger is not running before exiting?
This looks wired if test pass but logger failed and whole test case will exit with failure.

case-lib/hijack.sh
```diff
@@ -30,12 +30,23 @@ function func_exit_handler()
             func_lib_start_log_collect 1
             sleep 1s
         }
-        local loggerBin; loggerBin=$(basename "$SOFLOGGER")
-        sudo pkill -9 "$loggerBin" 2>/dev/null || {
+
+        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
+        # logfile is set in a different file: lib.sh
+        # shellcheck disable=SC2154
+        wcLog=$(wc -l "$logfile")
```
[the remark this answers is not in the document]
marc-hb: Good point, I'll move it right after killing the logger.

case-lib/hijack.sh
marc-hb: In the next PR I will implement a sanity check on the number of lines. For now this is not great looking but it's readable and I'd rather not write and test code that will change in the next PR.

marc-hb (commented):
Thanks @xiulipan !

> Do we need any check if logger is not running before exiting? This looks wired if test pass but logger failed and whole test case will exit with failure.

I don't understand this comment sorry. We already had for a few weeks  the "sof-logger was already dead" check that makes the test fail if the logger crashed. It works; I tested it before and after this PR.

xiulipan:
@marc-hb Eve- though we had such failures in past, I want to start a discussion here: it that behavior right? Do we need to fail the case if we could not get the logger? As we have standalone test case for logger.

marc-hb:
> it that behavior right? Do we need to fail the case if we could not get the logger? As we have standalone test case for logger.

The name of issue #297 is "Check for a valid trace for ALL tests" so I moved your question there.

plbossart:
> @marc-hb Eve- though we had such failures in past, I want to start a discussion here: it that behavior right? Do we need to fail the case if we could not get the logger? As we have standalone test case for logger.

yes we want to fail if the logger is gone during a test. There could be something with a specific test that gets in the way of the DMA trace, and we want to trap this error.

xiulipan (approved):

PR #376, opened 2020-09-15
test-case: add set -e to 5 basic test cases

RuiqingX:
add "set -e" for other shell

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
 ## Case Name: check alsabat
 ## Preconditions:
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
 ## Case Name: check-audio-equalizer.sh
 ## Preconditions:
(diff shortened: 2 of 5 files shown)


marc-hb:
Looks great in principle but a few caveats:
- Why are there 3 commits, 2 with a single change and a 3rd one with many changes? For a PR with so few lines changed I think a single commit would be OK as an exception to https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes
- Please mention the corresponding github issue number in the description.
- I think I remember we agreed in that issue to switch to `set -e` relatively slowly (only 4-5 scripts at once) to avoid too many regressions at once. `set -e` is vital BUT after adding it you can think everything works fine and then some time later realize that a few `|| true` are required to avoid breaking some unusual cases.
- Try to avoid empty commit messages. I agree there is not much to say here but you can always copy some sentences about the rationale from the corresponding github issue. Someone who doesn't know anything about shell scripts could be locally bisecting their unusual case and git bisect point at this commit.

aiChaoSONG:
@RuiqingX The commit title is too long. you don't need to mention every file name you have made changes to. Let's do it step by step as Marc suggested.:
- Please only change 4~5 scripts in this PR.
- Please install shellcheck locally with `sudo apt install shellcheck`, and fix the shellcheck issue introduced by adding set -e. (Eg, you add `set -e` to test-speaker.sh, just run `shellcheck test-speaker.sh` ), you will see shellcheck warnings and errors, there will be suggestions on how to fix them.

aiChaoSONG (commented):
the commit title: `test-case: add set -e to 5 basic test cases`

You don't need to capitalize the first letter for add, and you also need a prefix in the title to tell the reviewers of the code path you've made changes to.

marc-hb (approved):

PR #378, opened 2020-09-15
tools: fix a pipeline formats assignment error

aiChaoSONG:
check the comments in the code for a reason.

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -62,11 +62,16 @@ class clsTPLGReader:
                 # supported formats of playback pipeline in formats[0]
                 # supported formats of capture pipeline in formats[1]
                 formats = TplgFormatter.get_pcm_fmt(pcm)
+                # if capture is present, pcm['capture'] = 1, otherwise, pcm['capture'] = 0,
+                # same thing for pcm['playback']
                 pipeline_dict['fmts'] = " ".join(formats[pcm['capture']])
                 # use the first supported format for test
                 pipeline_dict['fmt'] = pipeline_dict['fmts'].split(' ')[0]
                 pipeline_dict['rate_min'], pipeline_dict['rate_max'] = self._key2str(cap, 'rate')
                 pipeline_dict['ch_min'], pipeline_dict['ch_max'] = self._key2str(cap, 'channels')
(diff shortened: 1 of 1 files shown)


fredoh9 (dismissed):
LGTM, thank you for adding comments as well

marc-hb:
@aiChaoSONG can you give an example of test failure or failed scenario? If not in the commit message then at least here in the PR comments.

aiChaoSONG:
@marc-hb The issue will not cause test failure, issue description is added in commit message.

marc-hb:
Interesting that we have no test that can detect this... Should we have one? Would it be difficult to have one or modify an existing one?

aiChaoSONG:
@marc-hb I don't think we should have a case for this, this is a one-time issue and now we fix it ! It is an issue of topology parser in sof-test, and will not impact any test. To enable a case for this, we need to compare the parsed formats by sof-tplgreader.py with the formats from topology conf or m4 source. Too much effort, doesn't worth a try.

marc-hb:
Thanks, so now I wonder how you found this but it's not important, never mind.

aiChaoSONG:
@marc-hb Keyon and I are working on TGL1.0 release this week. the new feature "codec adapter" can only work under S32_LE currently. So we need to run aplay with S32_LE, but the topology parser tells that there is no S32_LE support for the codec adapter pipeline. After checking the topology conf, the S32_LE is actually there, this drives me to debug the parser.

Assume this issue not fixed, and the playback pipeline supports all formats that the capture pipeline supports, we may never find it !

marc-hb:
> So we need to run aplay with S32_LE, but the topology parser tells that there is no S32_LE support for the codec adapter pipeline.

This is the (manual) "test case" I was asking about. It speeds up the understanding of the situation and the code review. It lets readers understand the code change even when they don't know and don't understand this codebase (and it introduces them to this codebase). It helps assess the severity of the bug fixed. Etc.

It's worth the commit message IMHO but having it in the PR is already much better than nothing.

xiulipan (approved):

xiulipan:
@marc-hb @aiChaoSONG This would be a good idea to let us do some check for our lib.
I think we can try something like manually compare the result of some tplg to make sure the tplg parser work as we expected.

PR #381, opened 2020-09-18
hijack.sh: don't fail yet when sof-logger resists kill -INT

marc-hb:
Mitigates #3433 for now.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -39,9 +39,10 @@ function func_exit_handler()
         }
         sleep 1s
         if pgrep "$loggerBin"; then
-            dloge "$loggerBin resisted pkill -INT, using -KILL"
+            # FIXME: https://github.com/thesofproject/sof/issues/3433
+            dlogw "$loggerBin resisted pkill -INT, using -KILL"
             sudo pkill -KILL "$loggerBin"
-            exit_status=2
+            # exit_status=1
         fi
         # logfile is set in a different file: lib.sh
         # shellcheck disable=SC2154


aiChaoSONG (approved):
LGTM, Thanks Marc, Let's wait for the CI test result.

xiulipan:
@marc-hb I merged the workaround now. Do we need to check why the logger resist `-INT`

marc-hb:
This has been broken for almost 24h, we could have waited a few more minutes for https://sof-ci.01.org/softestpr/PR381/build192/devicetest to complete. I hope we won't need another quick workaround on top of this other workaround that would be messy.

> Do we need to check why the logger resist -INT

I think so and I filed https://github.com/thesofproject/sof/issues/3433 for that.

marc-hb:
> we could have waited a few more minutes 

@xiulipan peeked at the logs before everything was finished and the results uploaded to https://sof-ci.01.org/softestpr/PR381/build192/devicetest. They are uploaded now and the vast majority of tests are passing.

xiulipan:
@marc-hb Checked the running logs, so I think this fix works.
Thanks for the issue reported, the issue is wired. I did not see this issue in previous test. Try to manually check the issue.

marc-hb:
I cannot reproduce locally but I got an idea:
```
2020-09-17 20:56:25 UTC [REMOTE_ERROR] sof-logger resisted pkill -INT, using -KILL
```
I don't do anything `REMOTE_` maybe that's why it's happening in CI and not locally.

From https://pubs.opengroup.org/onlinepubs/7908799/xsh/signal.h.html
> SIGINT | i | Terminal interrupt signal.

I've never done the REMOTE_ thing, no idea how to do it. @xiulipan or @aiChaoSONG could you compare REMOTE_ vs local for me?

PR #382, opened 2020-09-18
hijack.sh: switch first attempt to stop the logger from -INT to -TERM

marc-hb:
Commit 61271a3a03aa ("logger: stop discarding all error messages and fix
the cleanup code") made two mistakes:

1. It used the -INT signal which doesn't always work. This broke most
   tests.
2. It then misused exit_failure=2 where 2 means Not Applicable. This
  turned most broken tests to green.

Rushed commit a2669ff6591a ("hijack.sh: don't fail yet when sof-logger
resists kill -INT") then downgraded all this to warning to get the tests
passing again. However it didn't deal with the real issue: stopping the
logger on the first try. Switching to TERM does it and will hopefully
close https://github.com/thesofproject/sof/issues/3433

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -32,17 +32,18 @@ function func_exit_handler()
         }
 
         local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
-        # INT doesn't print any "Killed" message in non-interactive mode
-        sudo pkill -INT "$loggerBin" || {
+        # We need this to avoid the confusion of a "Terminated" message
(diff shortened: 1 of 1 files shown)


marc-hb:
CI results look good but I'm keeping this as a draft for a few hours to keep a "safe distance" with previous workaround #381 that was just merged.. If they are too close to each other it will be very difficult to determine which build used which sof-test version (there is no sof-test version information in test results yet)

xiulipan (approved):
LGTM.

PR #383, opened 2020-09-18
hijack.sh: fail if the logger trace is empty

### LINKED ISSUE #297: Check for a valid trace for ALL tests

marc-hb:
Does not completely fixes #297 but goes a long way.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -48,6 +48,12 @@ function func_exit_handler()
         # shellcheck disable=SC2154
         wcLog=$(wc -l "$logfile")
         dlogi "nlines=$wcLog"
+        local nlines; nlines=$(wc -l < "$logfile")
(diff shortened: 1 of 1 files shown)


marc-hb:
This demonstrates logger issues, more details in https://github.com/thesofproject/sof-test/issues/297#issuecomment-694661503

xiulipan:
@marc-hb Ping for update or close.

marc-hb:
Logger work fell down my priority list, will hopefully re-open in the future.

Issue #384, opened 2020-09-18
sof-logger issue w/ check-alsabat-headset-playback.sh for BYT_MB_NOCODEC


plbossart:
https://sof-ci.01.org/linuxpr/PR2452/build4513/devicetest/

check-alsabat-headset-playback.sh for BYT_MB_NOCODEC

````
BAT analysis: signal has 65536 frames at 44100 Hz, 1 channels, 2 bytes per sample.

Channel 1 - Checking for target frequency 997.00 Hz
Amplitude: 17958.7; Percentage: [54]
Detected peak at 988.51 Hz of 20.67 dB
 Total 20.7 dB from 988.51 to 988.51 Hz
Detected peak at 997.26 Hz of 36.36 dB
 Total 41.7 dB from 990.53 to 1002.64 Hz
 PASS: Peak detected at target frequency
Detected peak at 1003.99 Hz of 20.69 dB
 Total 41.8 dB from 1003.99 to 1003.99 Hz
Detected peak at 1006.00 Hz of 20.30 dB
 Total 41.8 dB from 1006.00 to 1006.00 Hz
Detected at least 4 signal(s) in total

Return value is -1003
2020-09-18 16:50:01 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2020-09-18 16:50:01 UTC [REMOTE_ERROR]  main()  @  /home/ubuntu/sof-test/test-case/check-alsabat.sh
2020-09-18 16:50:02 UTC [REMOTE_INFO] Starting /usr/bin/sof-logger  -l /etc/sof/sof-byt.ldc -o /home/ubuntu/sof-test/logs/check-alsabat/2020-09-18-16:49:54-26390/etrace.txt
error: in logger_read(), fread(..., /sys/kernel/debug/sof/etrace) failed: Inappropriate ioctl for device(25)
alsa-utils version 1.2.2
````
@fredoh9 @aiChaoSONG @marc-hb does this ring a bell?

marc-hb:
The error message and failed results are new because I've been busy "hardening" logger-related things per your request in #297 (and #298 )

The actual bug could be much, much older - we just don't know.

marc-hb:
Spending  more time looking at the logs, this test did NOT fail because the logger failed. It failed because alsabat returned -1003 (which may have caused the logger to fail)

plbossart:
Why should this:

````
[...]

PR #386, opened 2020-09-19
hijack.sh: don't start a second logger in the exit handler

marc-hb:
... because it makes no sense.

When a test using the logger fails it produces a confusing error
message like this:
```
2020-09-18 10:17:25 UTC [REMOTE_INFO] Starting /usr/bin/sof-logger
   -l /etc/sof/sof-byt.ldc -o
    /home/ubuntu/sof-test/logs/check-alsabat/<date>/etrace.txt
error: in logger_read(), fread(..., /sys/kernel/debug/sof/etrace)
    failed:  Inappropriate ioctl for device(25)
```
Reported by Pierre in #384

This mistake seems to have been there since the dawn of time however no
one noticed because most things logger-related (and others) have been
silenced so far - which is changing now with #297.

The problem this fixes can be reproduced trivially with this one-line
patch:
```diff
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -125,6 +125,7 @@ func_lib_start_log_collect()
     dlogi "Starting $loggerCmd"
     # Cleaned up by func_exit_handler() in hijack.sh
     sudo "$loggerCmd" &
+    exit 1
 }
```

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -25,11 +25,6 @@ function func_exit_handler()
 
     # when sof logger collect is open
     if [ "X$SOF_LOG_COLLECT" == "X1" ]; then
-        # when error occurs, exit and catch etrace log
-        [[ $exit_status -eq 1 ]] && {
-            func_lib_start_log_collect 1
-            sleep 1s
-        }
 
         local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
         # INT doesn't print any "Killed" message in non-interactive mode


case-lib/hijack.sh
```diff
@@ -25,11 +25,6 @@ function func_exit_handler()
 
     # when sof logger collect is open
     if [ "X$SOF_LOG_COLLECT" == "X1" ]; then
-        # when error occurs, exit and catch etrace log
-        [[ $exit_status -eq 1 ]] && {
-            func_lib_start_log_collect 1
-            sleep 1s
-        }
```
kv2019i, line 32: @marc-hb Not sure about this. This is not the normal logger run, but capture the separate etrace log in case of output (etrace.txt). Is this already covered somewhere else?

kv2019i (changes requested):

case-lib/hijack.sh
marc-hb: Thanks @kv2019i for the catch, really appreciated. I've been looking for information about the difference between `debug/trace` and `debug/etrace` but didn't find much, any good reference? @ktrzcinx ? I didn't expect both could be used simultaneously, in fact I now realize this PR is (was) an attempt to avoid a failure to use both simultaneously.

marc-hb: Found something here: https://github.com/thesofproject/sof/issues/3265#issuecomment-669802163
> thanks a lot for your information about traces. This is the kind of information we need to write it in sof-docs.

Issue #389, opened 2020-09-21
[CML] ignore non-audio failure caused by mei_me module


aiChaoSONG:
In the test of topic/sof-dev + stable-v1.6, we observed a false failure caused by mei_me module: `[ 6965.377054] mei_me 0000:00:16.4: hw_reset failed ret = -62`, this message should be filter out.

```
kernel: [ 6965.377054] mei_me 0000:00:16.4: hw_reset failed ret = -62
kernel: [ 6965.377099] mei_me 0000:00:16.4: unexpected reset: dev_state = ENABLED fw status = A4004103 00000000 00000000 00000000 00000000 00000000
```

The daily test report id is 457

aiChaoSONG:
In daily report 473, similar issue is observed on ICL_RVP_HDA and CML_RVP_SDW, `[ 6733.766202] mei_me 0000:00:16.0: hw_reset failed ret = -62`, but with a different ID.

kv2019i:
Still happening with CML_RDP_SDW. Made a patch to sof-test:
https://github.com/thesofproject/sof-test/pull/463

plbossart:
https://sof-ci.01.org/linuxpr/PR3774/build421/devicetest/?model=WHL_UPEXT_HDA_ZEPHYR&testcase=check-suspend-resume-5

````
===========================>>
[ 1160.148503] kernel: mei_me 0000:00:16.0: hbm: add client: state mismatch, [1, 3]
[ 1160.148507] kernel: mei_me 0000:00:16.0: mei_irq_read_handler ret = -71, state = 1.
[ 1160.149128] kernel: mei_me 0000:00:16.0: mei_irq_read_handler ret = -71, state = 1.
<<===========================
````

marc-hb:
Please re-open if observed again.

PR #390, opened 2020-09-21
sof-kernel-log-check: ignore an ime_me error on CML_RVP_SDW

aiChaoSONG:
Fixes: #389


xiulipan (approved):

marc-hb (commented):
Please move very platform-specific errors like this one to the
```shell
case "$platform" in
```
section so we can track them accurately.

marc-hb (commented):
Also, this error is already there, it only needs a wildcard:
```
ignore_str="$ignore_str"'|mei_me 0000:00:16\..: hw_start failed ret = -62'
```

aiChaoSONG:
@marc-hb No, not the same thing as the exist one, one is  hw_start, the other is hw_reset.

marc-hb:
Refined in #463

PR #392, opened 2020-09-22
wavetool.py: add THD+N analysis for audio quality check

aiChaoSONG:
This PR add a generic THD+N analysis for audio quality check, and a specific alsabat audio check. for alsabat audio check, the ramp data at the beginning will be skipped.

### DIFF
--- a/tools/wavetool.py
+++ b/tools/wavetool.py
@@ -292,7 +292,7 @@ def parse_cmdline():
     parser.add_argument('-R', '--recorded_wave', type=str, help='path of recorded wave')
     parser.add_argument('-Z', '--zero_threshold', type=float, default=-50.3, help='zero threshold in dBFS')
     parser.add_argument('-H', '--hb_time', type=float, default=2.1, help='history buffer size')
-    parser.add_argument('-T', '--threshold', type=float, default=-72.0, help='expected threshold')
+    parser.add_argument('-T', '--threshold', type=float, default=-65.0, help='expected threshold')
     return parser.parse_args()
 
 def main():
--- a/tools/wavetool.py
+++ b/tools/wavetool.py
@@ -212,6 +212,13 @@ def analyze_wav_wov(wave, fs):
 
 def analyze_wav_thdn(wave, fs_wav):
     wave = normalize(wave)
+    dbfs = calc_rms(wave)
+    print('RMS of specified wave is %s dB' % dbfs)
+    dbfs_result = dbfs > cmd.zero_threshold
+    if not np.all(dbfs_result):
+        print("Volume too low or only contains zero")
+        print("Valid volume level should be bigger than %f dBFS" % cmd.zero_threshold)
+        sys.exit(1003)
     thdn = calc_thdn(wave, fs_wav, cmd.freq[0])
     result = thdn < cmd.threshold
     print('THD+N of specified wave is %s dB' % thdn)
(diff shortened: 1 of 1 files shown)


tools/wavetool.py
```diff
@@ -242,11 +288,11 @@ def parse_cmdline():
     help='sample bits of generated wave')
     parser.add_argument('-o', '--output', type=str, help='path to store generated files', default='.')
     # wave comparison arguments
-    parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp', 'wov'], help='analyze recorded wave to give case verdict')
+    parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp', 'wov', 'thdn', 'alsabat'], help='analyze recorded wave to give case verdict')
     parser.add_argument('-R', '--recorded_wave', type=str, help='path of recorded wave')
     parser.add_argument('-Z', '--zero_threshold', type=float, default=-50.3, help='zero threshold in dBFS')
     parser.add_argument('-H', '--hb_time', type=float, default=2.1, help='history buffer size')
-    parser.add_argument('-T', '--threshold', type=float, default=-72.0, help='expected threshold')
+    parser.add_argument('-T', '--threshold', type=float, default=-65.0, help='expected threshold')
```
xiulipan, line 295: Any comment about this default value change?

tools/wavetool.py
```diff
@@ -205,6 +212,29 @@ def analyze_wav_wov(wave, fs):
         sys.exit(1002)
     print('wave analysis result: PASSED')
 
+def analyze_wav_alsabat(wave, fs_wav):
+    # Drop the ramp up data in unstable state
+    time_skip = 0.25
```
xiulipan, line 217: What is this value used here? Global value?

tools/wavetool.py
```diff
@@ -242,11 +288,11 @@ def parse_cmdline():
     help='sample bits of generated wave')
     parser.add_argument('-o', '--output', type=str, help='path to store generated files', default='.')
     # wave comparison arguments
-    parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp', 'wov'], help='analyze recorded wave to give case verdict')
+    parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp', 'wov', 'thdn', 'alsabat'], help='analyze recorded wave to give case verdict')
```
xiulipan, line 291: what is alsabat mode here? Does it means we have some predefined parameters here

xiulipan (changes requested):
I think we can make this feature more generic. Have some parameter for the skip things and add some predefined configure for some existing test.

tools/wavetool.py
aiChaoSONG: The first version I made it -72dB, but this threshold is high for our use case. eg, wov, so I have to lower the default value.

tools/wavetool.py
aiChaoSONG: There is a ramp in recorded wave in alsabat test case, and the duration is about 0.25 second

tools/wavetool.py
aiChaoSONG: No, The alsabat mode here means that we will skip the ramp, and start analysis from stable state.

xiulipan: OK, so this is only some configuration for test right? It is not some algorithm like `thdn`, right? So can we make the configurations more straightforward in some static format data.

aiChaoSONG:
close first, for no time to pick up this PR, will reopen later

PR #394, opened 2020-09-22
test-case: add more debug logs for verify-sof-firmware-load

xiulipan:
We can still see issue with verify-sof-firmware-load. Add more
debug logs to help root cause the issue.

### DIFF
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -34,8 +34,10 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
     $cmd | grep "Firmware debug build" -A3 | head -n 12
     exit 0
 else
+    $cmd | tail -n 500
+    printf ' ------\n $cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
+            "$cmd" "$DMESG_LOG_START_LINE"
     journalctl --dmesg --lines 50
-    journalctl --dmesg | grep -C 1 -i version
-    printf '$cmd was %s\n' "$cmd"
+    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
     die "Cannot find the sof audio version"
 fi


test-case/verify-sof-firmware-load.sh
```diff
@@ -35,7 +35,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
     exit 0
 else
     journalctl --dmesg --lines 50
-    journalctl --dmesg | grep -C 1 -i version
-    printf '$cmd was %s\n' "$cmd"
+    journalctl --dmesg | grep -q " sof-audio.*Firmware.*version";
+    printf '$cmd was %s, DMESG_LOG_START_LINE was %s\n' "$cmd" "$DMESG_LOG_START_LINE"
+    $cmd
```
marc-hb, line 40: The output of `$cmd` could be huge in case of a bug so how about:
- running it again first, immediately after the `else` so the other, shorter output don't get lost
- add a safety like: `$cmd | tail -n 10000`
- add some separators `---`.

```shell
else
     $cmd
    printf ' ------\n $cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
            "$cmd" "$DMESG_LOG_START_LINE"
    journalctl --dmesg --lines 50
    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
```

test-case/verify-sof-firmware-load.sh
```diff
@@ -35,7 +35,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
     exit 0
 else
     journalctl --dmesg --lines 50
-    journalctl --dmesg | grep -C 1 -i version
-    printf '$cmd was %s\n' "$cmd"
+    journalctl --dmesg | grep -q " sof-audio.*Firmware.*version";
```
marc-hb, line 38: I don't understand the `-q` and the lack of `|| true`, did you test this locally? It looks this was just copied from above.

marc-hb (changes requested):

xiulipan:
@marc-hb Updated. Thanks for the advice.

marc-hb (approved):

PR #397, opened 2020-09-22
kmod: add support for SoundWire SDCA 0.6 codecs

plbossart:
Modules need to be removed for kmod remove/insert tests to pass


marc-hb (approved):
I don't know anything about SoundWire myself but if we can't trust @plbossart for something that simple then we're in big trouble so I'm rubber-stamping this.

aiChaoSONG (approved):
LGTM

PR #398, opened 2020-09-23
sof-kernel-log-check: ignore an ime error for CML and ICL

aiChaoSONG:
Commit f764b59 ignore an ime error, and the similiar
ime error is observed again, but with a different PCI ID,
we need to ignore both.

Check daily report #473 / #457 suspend-resume test case for the error.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -20,10 +20,6 @@ ignore_str="$ignore_str"'|xhci_hcd 0000:00:14\.0: WARN Set TR Deq Ptr cmd failed
 ignore_str="$ignore_str"'|mei_me 0000:00:16\.0: wait hw ready failed'
 ignore_str="$ignore_str"'|mei_me 0000:00:16\.0: hw_start failed ret = -62'
 
-# On CML_RVP_SDW, suspend-resume test case failed due to "mei_me 0000:00:16.4: hw_reset failed ret = -62".
-# https://github.com/thesofproject/sof-test/issues/389
-ignore_str="$ignore_str"'|mei_me 0000:00:16\.4: hw_reset failed ret = -62'
-
 # CML Mantis has DELL touchpad i2c error on suspend/resume
 ignore_str="$ignore_str"'|i2c_designware i2c_designware\.0: controller timed out'
 ignore_str="$ignore_str"'|i2c_hid i2c-DELL0955:00: failed to change power setting'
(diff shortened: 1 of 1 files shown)


marc-hb (changes requested):
I don't understand how a kernel version can changed PCI IDs, can you explain?

This makes line 21 redundant, need to merge the two.

While doing this please address comments in the previous review (mainly use a platform dependent `case`)

tools/sof-kernel-log-check.sh
```diff
@@ -21,8 +21,9 @@ ignore_str="$ignore_str"'|mei_me 0000:00:16\.0: wait hw ready failed'
 ignore_str="$ignore_str"'|mei_me 0000:00:16\.0: hw_start failed ret = -62'
 
 # On CML_RVP_SDW, suspend-resume test case failed due to "mei_me 0000:00:16.4: hw_reset failed ret = -62".
+# The PCI ID in latest sof-dev differs from the PCI ID in linux-5.8.y
 # https://github.com/thesofproject/sof-test/issues/389
-ignore_str="$ignore_str"'|mei_me 0000:00:16\.4: hw_reset failed ret = -62'
+ignore_str="$ignore_str"'|mei_me 0000:00:16\..: hw_reset failed ret = -62'
```
xiulipan, line 26: maybe it is better to ignore the exact pci id here and match it with a pattern?

xiulipan (changes requested):

aiChaoSONG:
@marc-hb My bad, it is actually the same sof-dev kernel, but with different FW (stable 1.6 vs master).  Check the daily report 457/473, the IDs are different, but the error message are the same:
- [ 6733.766202] mei_me 0000:00:16.0: hw_reset failed ret = -62
- [ 6965.377054] mei_me 0000:00:16.4: hw_reset failed ret = -62

aiChaoSONG:
@xiulipan @marc-hb Let's first address why the IDs are different, any idea?

tools/sof-kernel-log-check.sh
aiChaoSONG: I think a `.` here is good enough.

marc-hb (approved):

xiulipan:
@aiChaoSONG The id is mapped to real hardware. The PCI bus is 0000:00:16 and the following number is the sub-bus. If the hardware is connected to different slot, the id may also change. So I think we may also need to confirm the mei_me is always on 0000:00:16

Please rebase and merge this PR

aiChaoSONG:
@xiulipan OK, I will check the log on every platform to see if the ME module got the same ID.

marc-hb:
> @xiulipan OK, I will check the log on every platform to see if the ME module got the same ID.

We can add this now if it solves a current problem and easily change it to a wildcard later if needed.

xiulipan:
> We can add this now if it solves a current problem and easily change it to a wildcard later if needed.

@marc-hb Agreed.  @aiChaoSONG Please rebase and merge this PR

marc-hb (changes requested):
I think you dropped the removal of the duplicate on line 25.

marc-hb (approved):

Issue #288, opened 2020-07-13
Python errors with check-mod-load-unload.sh on CHT device


plbossart:
````
root@Zotac:~/sof-test/test-case# ./check-kmod-load-unload.sh -l 10
2020-07-13 19:30:05 UTC [REMOTE_INFO] Pulseaudio disabled
2020-07-13 19:30:05 UTC [REMOTE_INFO] ===== Starting iteration 1 of 10 =====
2020-07-13 19:30:05 UTC [REMOTE_INFO] wait dsp power status to become suspended
Traceback (most recent call last):
  File "/root/sof-test/tools/sof-dump-status.py", line 405, in <module>
    print(sysinfo.sys_power['run_status'][ret_args['dsp_status']]['status'])
IndexError: list index out of range
Traceback (most recent call last):
  File "/root/sof-test/tools/sof-dump-status.py", line 405, in <module>
    print(sysinfo.sys_power['run_status'][ret_args['dsp_status']]['status'])
IndexError: list index out of range
````

aiChaoSONG:
@plbossart We don't have CHT device with us, Do you know someone have this device and share with us? then we can have a look.

plbossart:
> this

I can't share the device, so it's going to be interesting to debug. My Python skills are rusty, but can't we add some range checks on parameters here?

plbossart:
@bardliao @aiChaoSONG I also get this on a TGL device w/ SoundWire SDCA

````
~/sof-test/test-case# ./check-kmod-load-unload.sh 
2020-09-22 14:26:19 UTC [INFO] ===== Starting iteration 1 of 2 =====
2020-09-22 14:26:19 UTC [INFO] wait dsp power status to become suspended
Traceback (most recent call last):
  File "/root/sof-test/tools/sof-dump-status.py", line 405, in <module>
    print(sysinfo.sys_power['run_status'][ret_args['dsp_status']]['status'])
IndexError: list index out of range
Traceback (most recent call last):
  File "/root/sof-test/tools/sof-dump-status.py", line 405, in <module>
    print(sysinfo.sys_power['run_status'][ret_args['dsp_status']]['status'])
IndexError: list index out of range
Traceback (most recent call last):
  File "/root/sof-test/tools/sof-dump-status.py", line 405, in <module>
    print(sysinfo.sys_power['run_status'][ret_args['dsp_status']]['status'])
IndexError: list index out of range
````

I think it comes from the Power Status that doesn't exist:
````
Power Info:
[...]

PR #399, opened 2020-09-23
[don't merge]sof-dump-status: fix a PCI info acquisition error on CHT

aiChaoSONG:
Fixes: #288


marc-hb:
There seems to be some overlap with later #414 (cc: @plbossart )

plbossart:
this PR is clearly wrong, there are no PCI devices for the CHT device for which I reported the issue.
PR #414 is the 'right' way to fix this, we need to handle ACPI  and PCI devices separately - with the caveat that the entire use of lists in this python script needs to be reviewed completely.

I suggest this draft be closed since it's not going to help.

aiChaoSONG:
Let's rely on #414 to fix #288. close

PR #400, opened 2020-09-23
test-case: add set -e to 7 function test

RuiqingX:
Fixes: #312

### DIFF
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
 ## Case Name: check-pause-resume
 ## Preconditions:
--- a/test-case/check-reboot.sh
+++ b/test-case/check-reboot.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
 ## Case Name: test reboot
 ## Preconditions:
--- a/test-case/check-suspend-resume-with-audio.sh
+++ b/test-case/check-suspend-resume-with-audio.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 ##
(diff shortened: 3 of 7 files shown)


RuiqingX:
I have done some tests on sh-jsl-rvp-hda-01 and sh-tgl-shuri-hda-02.

marc-hb:
For a change like this I think we really need to wait for a green CI + make sure the next daily test run after merge is all green too.

aiChaoSONG:
I have checked the test result, The timeout on CML_RVP_SDW with alsabat test case is actually a failure, root cause is that the test framework doesn't know the exit code 21, so it fallback to timeout, but I don't think it is related to the PR. @RuiqingX Maybe you go to `jf-cml-rvp-sdw-3` or other CML_RVP_SDW and run a manual test with the alsabat case?

RuiqingX:
I've checked with Xiuli just now, and I have validated this issue on sh-cml-rvp-sdw-04, test result is pass.

keqiaozhang:
> I have checked the test result, The timeout on CML_RVP_SDW with alsabat test case is actually a failure, root cause is that the test framework doesn't know the exit code 21, so it fallback to timeout, but I don't think it is related to the PR. @RuiqingX Maybe you go to `jf-cml-rvp-sdw-3` or other CML_RVP_SDW and run a manual test with the alsabat case?

alsabat failed on CML-RVP-SDW is caused by the codec prefix_name change in machine description, so the amixer control names 
 have also changed accordingly. I have submitted a[ PR105](https://gitlab.devtools.intel.com/sof-ci/sof-framework/-/merge_requests/105) to fix it.

aiChaoSONG (approved):
LGTM， Thanks @keqiaozhang for correcting me on the alsabat fi, good to know that.

Issue #385, opened 2020-09-18
[BUG] sof-logger resisted pkill -INT, using -KILL


marc-hb:
https://github.com/thesofproject/sof-test/pull/373 was merged because https://sof-ci.01.org/softestpr/PR373/build182/devicetest was all green. However almost all tests were failing because the sof-logger resisted kill -INT. To make it worse I misused `exit_status=2` which has the special meaning Not Applicable which means no error and everything is green.

mengdonglin:
Keep this issue open before we root cause if this is a bug of sof-logger or test case. 
- If this is an sof-logger issue, will close this issue and report a bug in SOF repo.
- If this is a test case issue, we will use PR #401 and close this issue.

marc-hb:
I believe this has been fixed by #382 `hijack.sh: switch first attempt to stop the logger from -INT to -TERM`, closing.

PR #401, opened 2020-09-23
refine alsabat test script

keqiaozhang:


test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+            format_c=S16_LE
+            channel_c=1
+        else
+            format_c=$format
+            channel_c=$channel
+        fi
+
+        dlogc "alsabat -C $pcm_c -F $frequency -f $format_c -c $channel_c -r $rate"
+        alsabat -C "$pcm_c" -F "$frequency" -f "$format_c" -c "$channel_c" -r "$rate" || {
+            find /tmp -maxdepth 1 -type f -name "bat.wav.*" -size +0 -exec cp {} "$LOG_ROOT/" \;
```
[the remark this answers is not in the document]
marc-hb: You probably want :
```shell
alsabat ... || {
  find ... || true
  exit 1
}
```
... because of `set -e`. When `find` fails, it fails with an undocumented return value (!) which could cause serious issues like #385

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+            exit 1
+        }
+        timeout 5 tail --pid=$alsabat_P -f /dev/null || die "Failed to stop alsabat playback"
```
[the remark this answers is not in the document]
marc-hb: The "Failed to stop alsabat" message is misleading, it sounds like `kill` failed. You mean: "alsabat -P failed to stop"

marc-hb (changes requested):
Thanks for a shellcheck-clean PR! Saves a lot of review time.

Any idea about the many check-alsabat-headset-playback.sh failures in https://sof-ci.01.org/softestpr/PR401/build216/devicetest/ ? This doesn't look great in an alsabat PR :-(

I think you can write a much less vague PR title and summary than "refine", just summarize the commit messages.

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+
+        dlogc "alsabat -P $pcm_p --standalone -n $frames -F $frequency -c $channel -r $rate -f $format"
+        alsabat -P "$pcm_p" --standalone -n "$frames" -F "$frequency" -c "$channel" -r "$rate" -f "$format" & alsabat_P=$!
+        # playback may have low latency, add one second delay to aviod recording zero at beginning.
+        sleep 1
+
+        if [ "$codec_mode" != "nocodec" ]; then
+            # USB sound card only supports 1 channel S16_LE format.
+            format_c=S16_LE
+            channel_c=1
```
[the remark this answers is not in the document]
keqiaozhang: nocodec tplgs also support different formats like s16_le/s24_le and s32_le,  not passthrough mode. so we can check each format in the loop.

test-case/check-alsabat.sh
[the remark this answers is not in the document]
keqiaozhang: Good suggestion, I have added a new function in lib.sh to upload wav files.

test-case/check-alsabat.sh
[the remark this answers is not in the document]
keqiaozhang: updated.

case-lib/lib.sh
```diff
@@ -252,6 +252,12 @@ func_lib_get_tplg_path()
     return 0
 }
 
+func_check_nocodec_mode()
+{
+    uname -r |grep "nocodec" >> /dev/null &&
+    echo "nocodec"
+}
+
```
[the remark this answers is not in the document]
keqiaozhang: updated.

test-case/check-alsabat.sh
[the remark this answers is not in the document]
xiulipan: Do we real need alsabat for all fmts?

xiulipan (changes requested):

test-case/check-alsabat.sh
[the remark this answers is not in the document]
fredoh9: Checking all formats available is good idea. but S24_LE is always ignored, isn't it?

case-lib/lib.sh
```diff
@@ -252,6 +252,20 @@ func_lib_get_tplg_path()
...
+    uname -r |grep -q "nocodec"
+}
+
+func_upload_wav_file()
+{
+    local dir=$1
+    local file=$2
+
+    find "$dir" -maxdepth 1 -type f -name "$file" -size +0 -exec cp {} "$LOG_ROOT/" \;
+    return 0
```
marc-hb, line 266: The `return 0` will become a no-op when this script is converted to `set -e`. 

I recommend you remove it and you invoke the function like this below:

```shell
func_upload_wav_file "/tmp" "bat.wav.*" || true
```
This gives the caller of the function the choice.

(Cc: @RuiqingX )

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+        fi
+
+        dlogc "alsabat -C $pcm_c -F $frequency -f $format_c -c $channel_c -r $rate"
+        alsabat -C "$pcm_c" -F "$frequency" -f "$format_c" -c "$channel_c" -r "$rate" || {
+            func_upload_wav_file "/tmp" "bat.wav.*"
+            exit 1
+        }
+        wait $alsabatPID || die "alsabat -P failed to stop"
```
marc-hb, line 103: It's a bit more complicated, sorry I should have spotted this the first time. To be correct this requires two different error messages. Something like this:
```shell
wait $alsabatPID && ret=$?
if [ $ret = 124 ]; then
   die "alsabat -P failed to stop - we timed it out"
elif [ $ret != 0 ] ; then
   die "alsabat -P failed, returned $ret"
fi
```
Untested code, please double-check in `man timeout` and test by this code by temporarily setting a very short timeout or wrong parameters.

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+done
+
+exit 0
```
marc-hb, line 106: missing newline. There's probably a setting in your editor that you can change to avoid this.

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,51 @@ then
...
+channel=$(func_pipeline_parse_value 0 ch_max)
+rate=$(func_pipeline_parse_value 0 rate)
+fmts=$(func_pipeline_parse_value 0 fmts)
 
 # check the PCMs before alsabat test
 dlogi "check the PCMs before alsabat test"
-[[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && die "Failed to play on PCM: $pcm_p"
-[[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && die "Failed to capture on PCM: $pcm_c"
+[[ $(aplay -Dplug"$pcm_p" -d 1 /dev/zero -q) ]] && die "Failed to play on PCM: $pcm_p"
+[[ $(arecord -Dplug"$pcm_c" -d 1 /dev/null -q) ]] && die "Failed to capture on PCM: $pcm_c"
```
marc-hb, line 73: It's not something you're changing but it looks like this code was wrong. Can you take the opportunity to try to fix it and change it to:
```
aplay -Dplug"$pcm_p" -d 1 /dev/zero -q || die "Failed to play on PCM: $pcm_p"
```

marc-hb (changes requested):

test-case/check-alsabat.sh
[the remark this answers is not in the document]
keqiaozhang: Yes，S24_4LE is not supported by alsabat

test-case/check-alsabat.sh
keqiaozhang: You are right, die will never be executed.

test-case/check-alsabat.sh
keqiaozhang: Right, if the command times out, and --preserve-status is not set, then exit with status 124.

case-lib/lib.sh
keqiaozhang: fixed

test-case/check-alsabat.sh
marc-hb: > Untested code, please double-check in man timeout and test by this code by temporarily setting a very short timeout or wrong parameters.

With a clearer, Monday morning brain I see the untested code I gave you was wrong: `ret` never gets initialized when `wait` fails because of short circuit evaluation. I'm afraid this means you didn't inject any error locally, didn't test it and never saw these messages printed, this is worrying.

keqiaozhang: @marc-hb , I did performed the negative tests to verify this part of code, tried with wrong channel count "3" in alsabat command and returned 1.  if alsabat -P does not stop within the specified time, it will return 124. so which scenario you are talking above?

marc-hb: I mean this:
```shell
ret=0
(exit 5) && ret=$?
echo $ret

=> 0
```

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,56 @@ then
...
+channel=$(func_pipeline_parse_value 0 ch_max)
+rate=$(func_pipeline_parse_value 0 rate)
+fmts=$(func_pipeline_parse_value 0 fmts)
 
 # check the PCMs before alsabat test
 dlogi "check the PCMs before alsabat test"
-[[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && die "Failed to play on PCM: $pcm_p"
-[[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && die "Failed to capture on PCM: $pcm_c"
+aplay -Dplug"$pcm_p" -d 1 /dev/zero -q || die "Failed to play on PCM: $pcm_p"
+arecord -Dplug"$pcm_c" -d 1 /dev/null -q || die "Failed to capture on PCM: $pcm_c"
```
plbossart, line 73: this looks like an unrelated cleanup?

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,56 @@ then
...
+        # playback may have low latency, add one second delay to aviod recording zero at beginning.
+        sleep 1
+
+        if func_nocodec_mode; then
+            format_c=$format
+            channel_c=$channel
+        else
+            # USB sound card only supports 1 channel S16_LE format.
+            format_c=S16_LE
+            channel_c=1
```
plbossart, line 95: You guys need to buy better USB cards, this is not acceptable for channel swap tests.

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,56 @@ then
...
+        fi
+
+        dlogc "alsabat -C $pcm_c -F $frequency -f $format_c -c $channel_c -r $rate"
+        alsabat -C "$pcm_c" -F "$frequency" -f "$format_c" -c "$channel_c" -r "$rate" || {
+            func_upload_wav_file "/tmp" "bat.wav.*" || true
+            exit 1
+        }
+        wait $alsabatPID && ret=$?
+        if [ $ret == 124 ]; then
```
plbossart, line 104: what the hell does 124 mean?

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,56 @@ then
...
-fi
+for format in $fmts
+    do
+        if [ "$format" == "S24_LE" ]; then
+            dlogi "S24_LE is not supported, skip to test this format"
+            continue
+        fi
+
+        dlogc "alsabat -P $pcm_p --standalone -n $frames -F $frequency -c $channel -r $rate -f $format"
+        timeout -k 2 6 alsabat -P "$pcm_p" --standalone -n "$frames" -F "$frequency" -c "$channel" \
```
plbossart, line 84: what does 2 6 mean? seriously add comments on what this does.

test-case/check-alsabat.sh
```diff
@@ -50,39 +56,56 @@ then
...
+		-r "$rate" -f "$format" & alsabatPID=$!
+        # playback may have low latency, add one second delay to aviod recording zero at beginning.
```
plbossart, line 86: avoid

plbossart (changes requested):

test-case/check-alsabat.sh
marc-hb: The meaning of 124 is (again) in the `man timeout` page, in a previous review command and much better it is on the very next line: "alsabat -P failed to stop, we timed it out."

test-case/check-alsabat.sh
marc-hb: The meanings of 2 and 6 are at the top of the `man timeout` page. I don't think we need to copy/paste into this script the start of the man page of every unusual command but maybe we can have the  `--long` options like this:
```
timeout --kill-after 2 6 alsabat ...
```

test-case/check-alsabat.sh
marc-hb: I requested this bug fix, see earlier comments. Agreed it should ideally be in a different commit.

marc-hb (changes requested):
Still concerned about the lack of negative testing mentioned at
https://github.com/thesofproject/sof-test/pull/401#discussion_r502162170

test-case/check-alsabat.sh
keqiaozhang: I think we are overaction here. just put the alsabat -P in background and use wait to check the exit code, if exit code is non-zero， then exit with "alsabat -P failure".

test-case/check-alsabat.sh
keqiaozhang: i have separated it as a new commit.

test-case/check-alsabat.sh
xiulipan: Same idea here, we are looking for new cards now.

case-lib/lib.sh
```diff
@@ -252,6 +252,19 @@ func_lib_get_tplg_path()
     return 0
 }
 
+func_nocodec_mode()
```
xiulipan, line 255: Can we dump the result at first and only use the variable later? Three is no need to check this mode multiple time

case-lib/lib.sh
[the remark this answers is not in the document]
xiulipan: I do not see any update for this. but I have same idea with this.

xiulipan (changes requested):
Some update is needed. Please check

case-lib/lib.sh
marc-hb: > Can we dump the result at first and only use the variable later? 

This would be premature optimization. This command is only forking two additional and small processes per alsabat test, that's negligible.

case-lib/lib.sh
```diff
@@ -252,6 +252,19 @@ func_lib_get_tplg_path()
...
+{
+    uname -r |grep -q "nocodec"
+}
+
+func_upload_wav_file()
+{
+    local dir=$1
+    local file=$2
```
marc-hb, line 263: While double quotes are technically not required for assignments it's safer and good role modelling to just always quote.

test-case/check-alsabat.sh
```diff
@@ -48,39 +54,54 @@ then
 	exit 2
 fi
 
+pcmid_p=${pcm_p:-1}
```
marc-hb, line 57: Same as above, please double quote for consistency.

test-case/check-alsabat.sh
```diff
@@ -48,39 +54,54 @@ then
...
+            continue
+        fi
+
+        dlogc "alsabat -P $pcm_p --standalone -n $frames -F $frequency -c $channel -r $rate -f $format"
+        alsabat -P "$pcm_p" --standalone -n "$frames" -F "$frequency" -c "$channel" -r "$rate" \
+		-f "$format" & alsabatPID=$!
+        # playback may have low latency, add one second delay to aviod recording zero at beginning.
+        sleep 1
+
+        if func_nocodec_mode; then
```
marc-hb, line 87: `if nocodec_mode; then` would read like plain English.

test-case/check-alsabat.sh
```diff
@@ -48,39 +54,54 @@ then
...
+        fi
+
+        dlogc "alsabat -C $pcm_c -F $frequency -f $format_c -c $channel_c -r $rate"
+        alsabat -C "$pcm_c" -F "$frequency" -f "$format_c" -c "$channel_c" -r "$rate" || {
+            func_upload_wav_file "/tmp" "bat.wav.*" || true
+            exit 1
+        }
+        # check the alsabat -P exit code
+        if ! wait $alsabatPID; then
+            die "alsabat -P failure"
```
marc-hb, line 103: No need for the `!` negation
```shell
wait $alsabatPID || die "alsabat $alsabatPID failed"
```

test-case/check-alsabat.sh
marc-hb: AFAIK @keqiaozhang didn't recommend to remove the `timeout` command, he only recommended to handle timeout errors exactly like non-timeout errors.

Question for people with `alsabat` experience: can `alsabat` get stuck  when something goes wrong?

xiulipan (changes requested):
@keqiaozhang The PR is almost good, please fix some typo and address the advice to use a variable instead of call `func_nocodec_mode` multiple times

test-case/check-alsabat.sh
marc-hb: Internal issue 601 just filed about new USB cards, thanks @singalsu.

keqiaozhang: Thank you all. Let's hold this PR. Now we are trying to find a new USB cards w/ stereo capture support, then we can remove the plug completely.

xiulipan:
@keqiaozhang Ping for update or close.

keqiaozhang:
> @keqiaozhang Ping for update or close.

I just received the new USB card, will try with the new card and refine this patch soon.

keqiaozhang:
Submitted a new one https://github.com/thesofproject/sof-test/pull/554.

PR #410, opened 2020-10-06
Test-case: Add check volume levels

singalsu:
This patch adds a shell script that measures volume levels
to compute gains when executed with a nocodec topology where
SSP ports are connected to play -> capture loopback.

The measured gains are compared against test procedure and
error is issued if the gains deviate more than tolerance.

### DIFF
--- a/test-case/check-volume-levels.sh
+++ b/test-case/check-volume-levels.sh
@@ -1,0 +1,337 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+
+set -e
+
+##
+## Case Name: check-volume-levels
+## Preconditions:
+##    topology is nocodec with loopack in PCM0P -> PCM0C
+##    aplay should work
+##    arecord should work
+##    PCM0C capture PGA supports -50 - +30 dB range
+##    PCM0C capture PGA supports mute switch
+## Description:
+##    Set volume and mute switch to various values, measure
+##    volume gain from the actual levels and compare to set gains.
+## Case step:
+##    1. Start aplay
+##    2a. Capture 1st wav file
+##    2b. Capture 2nd wav file
+##    2c. Capture 3rd wav file
+##    3. Measure volume gains
+## Expect result:
+##    command line check with $? without error
+##
+
+TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+
+# shellcheck source=case-lib/lib.sh
+source "$TESTDIR/case-lib/lib.sh"
+
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+func_opt_parse_option "$@"
+
+tplg=${OPT_VALUE_lst['t']}
+
+TPLGREADER="$TESTDIR"/tools/sof-tplgreader.py
+PCM_ID=0
+CAP_FORMAT="S16_LE"
+CAP_RATE="48000"
+VOL_P30DB=80
+VOL_P10DB=60
+VOL_0DB=50
+VOL_M10DB=40
+VOL_M20DB=30
+VOL_M30DB=20
+VOL_MAX=$VOL_P30DB
+VOL_MIN=1
+VOL_MUTE=0
+VOL_NOM=$VOL_0DB
+APLAY_WAV=$(mktemp --suffix=.wav)
+ARECORD_WAV1=$(mktemp --suffix=.wav)
+ARECORD_WAV2=$(mktemp --suffix=.wav)
+ARECORD_WAV3=$(mktemp --suffix=.wav)
+
+#
+# Main test procedure
+#
+
+main () {
+    # Preparations
+    do_preparations
+    generate_sine
+
+    # Start sine playback
+    test -f "$APLAY_WAV" || die "Error: File $APLAY_WAV does not exist"
+    dlogi "Playing file $APLAY_WAV"
+    timeout -k 5 120 aplay -D "$PLAY_HW" "$APLAY_WAV" & aplayPID=$!
+    nap
+
+    # Do capture tests
+    run_test_1
+    run_test_2
+    run_test_3
+
+    # Stop sine playback and do cleanup
+    nap
+    dlogi "The test procedure is now complete. Killing the aplay process $aplayPID."
+    kill $aplayPID
+
+    # Measure, delete unnecessary wav files if passed
+    if measure_levels; then
+	dlogi "Deleting temporary files"
+	rm -f "$APLAY_WAV" "$ARECORD_WAV1" "$ARECORD_WAV2" "$ARECORD_WAV3"
+    fi
+}
+
+#
+# Test #1 volume and mute switch controls
+#
+
+#  0...1s channels 1-4 max gain
+#  1...2s channels 1-4 different gains
+#  2...3s channels 1-4 nominal gain
+#  3...4s channels 1-4 muted
+#  4...5s channels 1-4 nominal gain
+#  5...7s channels 1-4 muted
+#  7...8s channels 1-4 max gain
+#  8...9s channels 1-4 muted gain
+#  9..10s channels 1-4 min gain
+# 10..11s channels 1-4 different gains
+run_test_1 () {
+    cset_max; cset_unmute; nap
+
+    arecord_helper "$ARECORD_WAV1" "$CAP_CHANNELS" 11 & arecordPID=$!
+    nap
+    cset_volume_diff1 "$CAP_CHANNELS"; nap
+    cset_nom; nap
+    cset_mute; nap
+    cset_unmute; nap
+    cset_mute; nap
+    cset_max; nap
+    cset_unmute; nap
+    cset_mutevol; nap
+    cset_min; nap
+    cset_volume_diff2 "$CAP_CHANNELS"
+
+    # Wait for arecord process to complete
+    dlogi "Waiting $arecordPID"
+    wait $arecordPID
+    dlogi "Ready."
+}
+
+#
+# Test #2, check gains preservation from previous to next
+#
+
+# 0..1s channels 1-4 previous gains
+# 1..2s channels 1-4 different mute switches
+# 2..3s channels 1-4 all muted
+run_test_2 () {
+    arecord_helper "$ARECORD_WAV2" "$CAP_CHANNELS" 4 & arecordPID=$!
+    nap
+    cset_nom; cset_mute_diff1 "$CAP_CHANNELS"; nap
+    cset_mute; nap
+
+    dlogi "Waiting $arecordPID"
+    wait $arecordPID
+    dlogi "Ready."
+}
+
+#
+# Test #3, test mute switch preservation from previous to next
+#
+
+# 0..1s channels 1-4 muted
+# 1..2s channels 1-4  nominal gain
+# 2..3s different mute switches
+# 3..4s channels 1-4 nominal gain
+run_test_3 () {
+    arecord_helper "$ARECORD_WAV3" "$CAP_CHANNELS" 4 & arecordPID=$!
+    nap
+    cset_nom; nap
+    cset_mute_diff2 "$CAP_CHANNELS"; nap
+    cset_unmute
+
+    dlogi "Waiting $arecordPID"
+    wait $arecordPID
+    dlogi "Ready."
+}
+
+#
+# Helper functions
+#
+
+cset_nom () {
+    dlogi "Set nominal volume"
+    amixer cset name="$CAP_VOLUME" $VOL_NOM
+}
+
+cset_max () {
+    dlogi "Set maximum volume"
+    amixer cset name="$CAP_VOLUME" $VOL_MAX
+}
+
+cset_min () {
+    dlogi "Set minimum volume"
+    amixer cset name="$CAP_VOLUME" $VOL_MIN
+}
+
+cset_mutevol () {
+    dlogi "Set mute via volume"
+    amixer cset name="$CAP_VOLUME" $VOL_MUTE
+}
+cset_mute () {
+    dlogi "Set mute via switch"
+    amixer cset name="$CAP_SWITCH" off
+}
+
+cset_unmute () {
+    dlogi "Set unmute"
+    amixer cset name="$CAP_SWITCH" on
+}
+
+nap () {
+    dlogi "Sleeping"
+    sleep 1
+}
+
+cset_mute_diff1 () {
+    dlogi "Set switch pattern 1"
+    case "$1" in
+	4)
+	    amixer cset name="$CAP_SWITCH" off,on,on,off ;;
+	2)
+	    amixer cset name="$CAP_SWITCH" off,on ;;
+	1)
+	    amixer cset name="$CAP_SWITCH" off ;;
+    esac
+}
+
+cset_mute_diff2 () {
+    dlogi "Set switch pattern 2"
+    case "$1" in
+	4)
+	    amixer cset name="$CAP_SWITCH" on,off,off,on ;;
+	2)
+	    amixer cset name="$CAP_SWITCH" on,off ;;
+	1)
+	    amixer cset name="$CAP_SWITCH" on ;;
+    esac
+}
+
+cset_volume_diff1 () {
+    dlogi "Set volume pattern 1"
+    case "$1" in
+	4)
+	    amixer cset name="$CAP_VOLUME" $VOL_P10DB,$VOL_0DB,$VOL_M10DB,$VOL_M30DB ;;
+	2)
+	    amixer cset name="$CAP_VOLUME" $VOL_P10DB,$VOL_0DB ;;
+	1)
+	    amixer cset name="$CAP_VOLUME" $VOL_P10DB ;;
+    esac
+}
+
+cset_volume_diff2 () {
+    dlogi "Set volume pattern 2"
+    case "$1" in
+	4)
+	    amixer cset name="$CAP_VOLUME" $VOL_M10DB,$VOL_P10DB,$VOL_0DB,$VOL_M20DB ;;
+	2)
+	    amixer cset name="$CAP_VOLUME" $VOL_M10DB,$VOL_P10DB ;;
+	1)
+	    amixer cset name="$CAP_VOLUME" $VOL_M10DB ;;
+    esac
+}
+
+arecord_helper () {
+    dlogi "Capturing file $1"
+    timeout -k 5 30 arecord -D "$CAP_HW" -f $CAP_FORMAT -r $CAP_RATE -c "$2" -d "$3" "$1"
+}
+
+do_preparations () {
+
+    if [[ "$tplg" != *"nocodec"* ]]; then
+	# Return special value 2 with exit
+	dlogi "This test is executed only with nocodec topologies. Returning Not Applicable."
+	exit 2
+    fi
+
+    test -n "$(command -v octave)" || {
+	dlogi "Octave not found (need octave and octave-signal packages). Returning Not Applicable."
+	exit 2
+    }
+
+    # Get max. channels count to use from capture device
+    # remove xargs after the trailing blank from tplgreader is fixed
+    CAP_CHANNELS=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d ch_max -v)
+    CAP_HW=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d dev -v)
+    PLAY_HW=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:playback" -d dev -v)
+    CAP_PGA=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d pga -v)
+    PLAY_PGA=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:playback" -d pga -v)
+    export CAP_CHANNELS
+    export CAP_HW
+    export PLAY_HW
+
+    dlogi "Test uses $CAP_CHANNELS channels"
+    dlogi "Playback device is $PLAY_HW"
+    dlogi "Playback PGA is $PLAY_PGA"
+    dlogi "Capture device is $CAP_HW"
+    dlogi "Capture PGA is $CAP_PGA"
+
+    # Find amixer controls for capture, error if more than one PGA
+    numpga=${#CAP_PGA[@]}
+    test "$numpga" = 1 || die "Error: more than one capture PGA found."
+
+    tmp=$(amixer controls | grep -e "$CAP_PGA.*Volume" || true )
+    test -n "$tmp" || die "No control with name Volume found in $CAP_PGA"
+    search="name="
+    CAP_VOLUME=${tmp#*$search}
+    export CAP_VOLUME
+    dlogi "Capture volume control name is $CAP_VOLUME"
+
+    tmp=$(amixer controls | grep -e "$CAP_PGA.*Switch" || true )
+    test -n "$tmp" || die "No control with name Switch found in $CAP_PGA"
+    CAP_SWITCH=${tmp#*$search}
+    export CAP_SWITCH
+    dlogi "Capture switch control name is $CAP_SWITCH"
+
+    # Check needed controls
+    amixer cget name="$CAP_VOLUME" || die "Error: failed capture volume get command"
+    amixer cget name="$CAP_SWITCH" || die "Error: failed capture switch get command"
+
+    for pga in $PLAY_PGA; do
+	tmp=$(amixer controls | grep "$pga" | grep Volume)
+	search="name="
+	play_volume=${tmp#*$search}
+	dlogi "Set $play_volume to 100%"
+	amixer cset name="$play_volume" 100% || die "Error: failed play volume set command"
+    done
+}
+
+generate_sine () {
+    dlogi "Creating sine wave file"
+    cd "$TESTDIR"/tools
+    octave --silent --no-gui --eval "check_volume_levels('generate', '$APLAY_WAV')" ||
+	die "Error: failed sine wave generate."
+}
+
+measure_levels () {
+    dlogi "Measuring volume gains"
+    cd "$TESTDIR"/tools
+    octave --silent --no-gui --eval "check_volume_levels('measure', '$ARECORD_WAV1', '$ARECORD_WAV2', '$ARECORD_WAV3')" || {
+	dloge "Error: Failed one or more tests in volume levels check."
+	die "Please inspect files: $ARECORD_WAV1, $ARECORD_WAV2, and $ARECORD_WAV3."
+    }
+}
+
+#
+# Start test
+#
+
+main "$@"
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -64,6 +64,9 @@ tplg-binary
 sof-logger
 '
 
+# Requires Octave
+# testlist="$testlist volume_levels"
+
 main()
 {
 	# Default values overriden by the environment if any
@@ -190,6 +193,10 @@ test_pause-resume-capture()
 test_volume()
 {
 	"$mydir"/volume-basic-test.sh -l "$large_loop"
+}
+test_volume_levels()
+{
+	"$mydir"/check-volume-levels.sh
 }
 test_signal-stop-start-playback()
 {
--- a/tools/check_volume_levels.m
+++ b/tools/check_volume_levels.m
@@ -1,0 +1,253 @@
+function check_volume_levels(cmd, fn1, fn2, fn3)
+
+% check_volume_levels(cmd, fn1, fn2, fn3)
+%
+% Inputs
+%   cmd - Use 'generate' or 'analyze'
+%   fn1 - File name for sine wave to generate or first record file name to analyze
+%   fn2 - File name to analyze 2nd
+%   fn3 - File name to analyze 3rd
+%
+% E.g.
+%   check_volume_levels('generate','sine.wav');
+%   check_volume_levels('measure','rec1.wav','rec2.wav','rec3.wav');
+
+% SPDX-License-Identifier: BSD-3-Clause
+% Copyright(c) 2016 Intel Corporation. All rights reserved.
+% Author: Seppo Ingalsuo <seppo.ingalsuo@linux.intel.com>
+
+	addpath('../../sof/tools/test/audio/std_utils');
+	addpath('../../sof/tools/test/audio/test_utils');
+
+	if exist('OCTAVE_VERSION', 'builtin')
+		pkg load signal
+	end
+
+	switch lower(cmd)
+		case 'generate'
+			pass = generate(fn1);
+			if pass
+				fprintf(1, 'done\n');
+			else
+				error('FAIL');
+			end
+		case 'measure'
+			pass = measure(fn1, fn2, fn3);
+			if pass
+				fprintf(1, 'PASS\n');
+			else
+				error('FAIL');
+			end
+		otherwise
+			error('Invalid cmd')
+	endswitch
+
+end
+
+% Generate a 701 Hz and 1297 Hz -40 dBFS stereo sine wave
+% to test volume gain and muting
+
+function pass = generate(fn)
+	fprintf('Create sine wave file %s\n', fn)
+	fs = 48e3;
+	f1 = 701;
+	f2 = 1297;
+	a = 10^(-40/20);
+	t = 60;
+	x1 = multitone(fs, f1, a, t);
+	x2 = multitone(fs, f2, a, t);
+	x = [x1'; x2']';
+	sx = size(x);
+	d = (rand(sx(1), sx(2)) - 0.5)/2^15;
+	audiowrite(fn, x + d, fs);
+	pass = 1;
+end
+
+
+function pass = measure(fn1, fn2, fn3)
+
+	% General test defaults
+	lm.tgrid = 5e-3;              % Return level per every 5ms
+	lm.tlength = 10e-3;           % Use 10 ms long measure window
+	lm.sine_freqs = [701 1297];   % The stimulus wav frequencies
+	lm.sine_dbfs = [-40 -40];     % The stimulus wav dBFS levels
+
+	% Default gains for test 1
+	v1 = [+10   0 -10 -30];
+	v2 = [-10 +10   0 -20];
+	vmax = +30;
+	vnom = 0;
+	vmut = -100;
+	vmin = -49;
+	vol_ch1    = [ vmax v1(1) vnom vmut vnom vmut vmut vmax vmut vmin v2(1) ];
+	vol_ch2    = [ vmax v1(2) vnom vmut vnom vmut vmut vmax vmut vmin v2(2) ];
+	t1.vctimes = [ 0    1     2    3    4    5    6    7    8    9    10    ];
+	t1.volumes = [vol_ch1 ; vol_ch2 ]'; % Merge channels to matrix
+	t1.meas = [0.5 0.9]; % Measure levels 0.5s after transition until 0.9s
+	t1.vtol = 0.5; % Pass test with max +/- 0.5 dB mismatch
+
+	% Check test 1
+	pass1 = level_vs_time_checker(fn1, t1, lm, '1/3');
+
+	% Default gains for test 2
+	m1 = [vmut vnom vnom vmut];
+	m2 = [vnom vmut vmut vnom];
+	vol_ch1    = [ v2(1) m1(1) vmut ];
+	vol_ch2    = [ v2(2) m1(2) vmut ];
+	t2.vctimes = [ 0     1     2    ];
+	t2.volumes = [ vol_ch1 ; vol_ch2 ]'; % Merge channels to matrix
+	t2.meas = t1.meas; % Same as previous
+	t2.vtol = t1.vtol; % Same as previous
+
+	% Check test 2
+	pass2 = level_vs_time_checker(fn2, t2, lm, '2/3');
+
+	% Default gains for test 3
+	vol_ch1    = [ vmut vmut m2(1) vnom ];
+	vol_ch2    = [ vmut vmut m2(2) vnom ];
+	t3.vctimes = [ 0    1    2     3    ];
+	t3.volumes = [ vol_ch1 ; vol_ch2 ]';
+	t3.meas = t1.meas; % Same as previous
+	t3.vtol = t1.vtol; % Same as previous
+
+	% Check test 3
+	pass3 = level_vs_time_checker(fn3, t3, lm, '3/3');
+
+	if pass1 == 1 && pass2 == 1 && pass3 == 1
+		pass = 1;
+	else
+		pass = 0;
+	end
+
+end
+
+function pass = level_vs_time_checker(fn, tc, lm, id)
+	fprintf(1, 'File %s:\n', fn);
+
+	lev = level_vs_time(fn, lm);
+	%plot_levels(lev, tc, lm);
+	pass = check_levels(lev, tc, lm, 1);
+	if pass
+		fprintf(1, 'pass (%s)\n', id);
+	else
+		fprintf(1, 'fail (%s)\n', id);
+
+		% Swapped channels?
+		sine_freqs_orig = lm.sine_freqs;
+		lm.sine_freqs = sine_freqs_orig(end:-1:1);
+		lev = level_vs_time(fn, lm);
+		pass_test = check_levels(lev, tc, lm, 0);
+		if pass_test
+			fprintf(1,'Note: The test would pass with swapped channels.\n');
+			return
+		end
+
+		% Swapped controls?
+		lm.sine_freqs = sine_freqs_orig;
+		volumes_orig = tc.volumes;
+		tc.volumes = volumes_orig(:, end:-1:1);
+		lev = level_vs_time(fn, lm);
+		pass_test = check_levels(lev, tc, lm, 0);
+		if pass_test
+			fprintf(1,'Note: The test would pass with swapped controls.\n')
+			return
+		end
+
+		% Swapped controls and swapped channels
+		lm.sine_freqs = sine_freqs_orig(end:-1:1);
+		lev = level_vs_time(fn, lm);
+		pass_test = check_levels(lev, tc, lm, 0);
+		if pass_test
+			fprintf(1,'Note: The test would pass with swapped controls and swapped channels.\n')
+		end
+	end
+end
+
+function plot_levels(meas, tc, lm)
+	figure
+	plot(meas.t, meas.levels - lm.sine_dbfs);
+	grid on;
+
+	sv = size(tc.volumes);
+	hold on;
+	for j = 1:sv(2)
+		for i = 1:sv(1)
+			plot([tc.vctimes(i)+tc.meas(1) tc.vctimes(i)+tc.meas(2)], ...
+			     [tc.volumes(i,j)+tc.vtol tc.volumes(i,j)+tc.vtol], 'r--');
+			if tc.volumes(i,j) > -100
+				plot([tc.vctimes(i)+tc.meas(1) tc.vctimes(i)+tc.meas(2)], ...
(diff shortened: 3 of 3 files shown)


singalsu:
Update: I changed echo commands to dlogi and dloge.

singalsu:
Fixed environment variable declarations / exports.

singalsu:
For fun I added it to run-all-tests.sh. It will fail on most devices due to no Octave installed. Also the nocodec topologies do not create the required mute switch control. I'll do a separate SOF PR for it.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+# Stop sine playback and cleanup
+wait
+kill $aplay
+
+# Measure
+measure_levels
+ret=$?
+if [ $ret -eq 0 ]; then
+    rm -f "$APLAY_WAV"
+    rm -f "$ARECORD_WAV1"
```
marc-hb, line 391: You need this for compability with `set -e` (which is missing right now)
```shell
if measure_levels; then
 rm -f "$APLAY_WAV" "$ARECORD_WAV1"
fi
```
Not sure how this is not triggering https://github.com/koalaman/shellcheck/wiki/SC2181 ?!

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+
+measure_levels () {
+
+    dlogi Measuring volume gains
+    octave --silent --no-gui --eval "check_volume_levels('measure', '$ARECORD_WAV1', '$ARECORD_WAV2', '$ARECORD_WAV3')"
+    ret=$?
+    if [ $ret -ne 0 ]; then
+	dloge "Error: Failed one or more tests in volume levels check."
+	dloge "Please inspect files: $ARECORD_WAV1, $ARECORD_WAV1, and $ARECORD_WAV1."
+	exit $ret
```
marc-hb, line 360: You can't `exit $ret` because the test framework has some reserved values (I just got bit by this in commit 0af6a25bc85ff)
The right `exit` comes for free with `die`:
```shell
dloge "Error: Failed one or more tests in volume levels check."
die   "Please inspect files: $ARECORD_WAV1, $ARECORD_WAV1, and $ARECORD_WAV1."
```
The `die()` function is defined in `case-lib/lib.sh`.

test-case/run-all-tests.sh
```diff
@@ -18,6 +18,7 @@ speaker
 pause-resume-playback
 pause-resume-capture
 volume
+volume_levels
```
marc-hb, line 21: Comment it out at the end considering it requires octable/matlab

```shell
# Requires octave
# testlist="$testlist volume_levels"
```

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+    fi
+}
+
+#
+# Start test
+#
+
```
marc-hb, line 367: Please add a `main "$@"` function:
- all code is indented at the same level - minimizes future diffs
- allows sourcing the entire script by just commenting out one line
- functions can be defined anywhere / in any order

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+	fi
+    done
+}
+
+generate_sine () {
+
+    dlogi Creating sine wave file
+    octave --silent --no-gui --eval "check_volume_levels('generate', '$APLAY_WAV')"
+    ret=$?
+    if [ $ret -ne 0 ]; then
```
marc-hb, line 346: For compatibility with `set -e` (which is missing now)
```shell
octave --silent --no-gui --eval "check_volume_levels('generate', '$APLAY_WAV')" ||
  die "Error: failed sine wave generate"

```
Not sure how this is not triggering https://github.com/koalaman/shellcheck/wiki/SC2181 ?!

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+}
+
+wait_for_pid () {
+    dlogi Waiting for process "$1" to complete
+    tail --pid="$pid" -f /dev/null
+}
+
+arecord_in_background () {
+    dlogi Capturing file "$1"
+    arecord -D "$CAP_HW" -f $CAP_FORMAT -r $CAP_RATE -c "$2" -d "$3" "$1"  2> /dev/null &
```
marc-hb, line 245: Never discard stderr in a test script. Tests are meant to find errors, not hide them.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+    pid=$!
+    return $pid
```
marc-hb, line 247: Hardcoding the background decision in the function and returning the PID is unusual. I think it's much clearer to have the `&` and the `wait` at the same place:

```shell
my_arecord_function &  arecPID=$?
...
...
wait $arecPID

File: test-case/check-volume-levels.sh [hunk 8]
```diff
@@ -0,0 +1,394 @@
...
+}
+
+do_preparations () {
+
+    # Check that TPLG is set
+    if [ -z "$TPLG" ]; then
+	dloge "Error: Need to set environment variable TPLG to the used SOF topology"
+	exit 1
```
marc-hb, line 267: `OPT_OPT_lst['t']='tplg'` provides this check (and more?) for free. See example in `check-playback.sh` and others.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+    fi
+
+    if [[ "$TPLG" != *"nocodec"* ]]; then
+	dlogi "Note: this test is executed only with nocodec topologies."
+	exit 0
```
marc-hb, line 272: There is a special exit code that Jenkins interprets as "SKIP", I think it's 2 but please double check.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+    arecord -D "$CAP_HW" -f $CAP_FORMAT -r $CAP_RATE -c "$2" -d "$3" "$1"  2> /dev/null &
+    pid=$!
+    return $pid
+}
+
+aplay_in_background () {
+    if ! [ -f "$1" ]; then
+	dloge "Error: File $1 does not exist"
+	exit 1
+    fi
```
marc-hb, line 254: I believe `aplay` provides a decent error message for free, so once  you add `set -e` at the top you can get rid of this.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+	1)
+	    amixer cset name="$CAP_VOLUME" $VOL_M10DB
+	    ;;
+    esac
+}
+
+wait_for_pid () {
+    dlogi Waiting for process "$1" to complete
+    tail --pid="$pid" -f /dev/null
+}
```
marc-hb, line 241: Using `tail` like this is really a hack. It also fails to catch failures which is really bad for test code.

There's a clean way to do this:
```shell
timeout -k 5 10 some_background_ process & somePID=$!
...

wait_for_pid()
{
ret=0
wait $somePID || ret=$?
if [ $ret -eq 124 ] || [ $ret -eq 137 ]; then
 printf 'process %d didnt stop, we timed it out\n' "$somePID"
elif [ $ret -ne 0 ]
 printf 'process %d failed with status %d\n' "$somePID" "$ret"
fi
return $ret
}
```
See `man timeout`. Feel free to add a function like this to `case-lib/lib.sh`, that logic should be in use all across the tests and I don't think it is right now.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+
+# 0..1s channels 1-4 muted
+# 1..2s channels 1-4  nominal gain
+# 2..3s different mute switches
+# 3..4s channels 1-4 nominal gain
+run_test_3 () {
+    arecord_in_background "$ARECORD_WAV3" "$CAP_CHANNELS" 4
+    PID=$?
+    wait
```
marc-hb, line 126: `wait` waits for all processes so the purpose of `PID` is not obvious here.

marc-hb (changes requested):

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,394 @@
...
+##    1. Start aplay
+##    2a. Capture 1st wav file
+##    2b. Capture 2nd wav file
+##    2c. Capture 3rd wav file
+##    3. Measure volume gains
+## Expect result:
+##    command line check with $? without error
+##
+
+source "$(dirname "${BASH_SOURCE[0]}")/../case-lib/lib.sh"
```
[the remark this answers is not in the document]
singalsu: Thanks, it seems to be `# shellcheck source=case-lib/lib.sh`.

test-case/check-volume-levels.sh
singalsu: OK, there's a bunch of them.

test-case/check-volume-levels.sh
singalsu: Nice, thanks!

test-case/run-all-tests.sh
singalsu: Yep, this is a reference implementation for python version of the same. I think I'll add also ramp length (as milliseconds) check, then the Octave part should be ready.

test-case/check-volume-levels.sh
singalsu: I think that I mentioned working aplay and arecord in pre-requisites but yes you are right that it doesn't hurt to have more basics testing. I have a problem with background aplay/arecord under set -e.

test-case/check-volume-levels.sh
singalsu: I think that I could make such work by exporting aplayPID and arecPID.

test-case/check-volume-levels.sh
singalsu: Good idea. Though 2 seems to be unstable. I'll search more.

test-case/check-volume-levels.sh
singalsu: I'm not sure about aplay in background. Set -e complicates a lot things based on my first use experience.

test-case/check-volume-levels.sh
singalsu: Thanks, I'll try. Hope it solves the "aplay &" vs. "set -e" conflict.

test-case/check-volume-levels.sh
singalsu: I learned "wait" was a bad name for my own function. It's now called "pause ()". It uses "sleep 1" but if I get stable results with shorter level measurements I may shorten it for better test speed.

test-case/check-volume-levels.sh
singalsu: No success. Nohup might be an alternative too. I'm now upgrading Ubuntu in the test device in hope that my issue would be a bug in bash.

singalsu: Or put bash function background instead of aplay?

test-case/check-volume-levels.sh
singalsu: Done.

test-case/check-volume-levels.sh
singalsu: Done.

test-case/run-all-tests.sh
singalsu: Done.

test-case/check-volume-levels.sh
singalsu: Done.

test-case/check-volume-levels.sh
singalsu: The backgrounding works now and the stderr is preserved. Thanks for the help!

test-case/check-volume-levels.sh
singalsu: I didn't manage to collect the PID correctly with set -e. There's now `wait %2` in use. It works if there's the aplay in background as %1. I haven't tested killing it during the test but I'm pretty confident the levels checker will return fail if there's no signal for volume. But I'll keep this open until confirmed the script doesn't give false pass.

test-case/check-volume-levels.sh
singalsu: Good, I got it working.

test-case/check-volume-levels.sh
singalsu: I saw in another comment git commit text that mentioned value 2 as Non Applicable. I'll use that for now.

test-case/check-volume-levels.sh
singalsu: I'd prefer to keep this, it's a habit of mine.

test-case/check-volume-levels.sh
singalsu: I removed this function totally with use of wait %N. If %N is not reliable I need to return this this kind of function for PID. I'll test more the robustness with killed processes to see what happens.

test-case/check-volume-levels.sh
singalsu: Rename done.

test-case/check-volume-levels.sh
singalsu: Yep - done.

singalsu:
@marc-hb Thanks for all the help! I've tried to address all your comments now. Though there are two cases where I'm not yet sure the approach is the best one.

I'll still add to Octave part channels filtering for center frequencies. It will catch then also channels swaps. Also I'll test adding ramp length verification for the 0-100% transitions.

Can this be merged to sof-test as non-enabled test (while need for Octave)?

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    case "$1" in
+	4)
+	    amixer cset name="$CAP_SWITCH" off,on,on,off
+	    ;;
+	2)
+	    amixer cset name="$CAP_SWITCH" off,on
+	    ;;
+	1)
+	    amixer cset name="$CAP_SWITCH" off
+	    ;;
```
marc-hb, line 240: Again that seems like a lot of whitespace and scrolling here, how about something a bit more compact?
```
4) amixer ... ;;
```

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    amixer cget name="$CAP_VOLUME" || {
+	die "Error: failed capture volume get command"
+    }
+
+    amixer cget name="$CAP_SWITCH" || {
+	die "Error: failed capture switch get command"
+    }
+
+    for pga in $PLAY_PGA
+    do
```
marc-hb, line 356: Nit: the default syntax is:
```shell
for pga in $PLAY_PGA; do
  ..
```
This syntax is actually built in the shell itself. Most people don't know this but the shell has a built-in formatter, try these interactive commands:
```shell
$ f1 () { for i in a b ; do : ; done ; }
$ type f1
f1 is a function
f1 () 
{ 
    for i in a b;
    do
        :;
    done
}

File: test-case/check-volume-levels.sh [hunk 16]
```diff
@@ -0,0 +1,386 @@
...
+	tmp=$(amixer controls | grep "$pga" | grep Volume)
+	search="name="
+	play_volume=${tmp#*$search}
+	dlogi Set "$play_volume" to 100%
+	amixer cset name="$play_volume" 100% || {
+	    die "Error: failed play volume set command"
+	}
```
marc-hb, line 363: Nit: just like in C you don't need the brackets when there is only one statement. In this case you don't need to escape the new line either because `||` gives you the newline escape for free.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    done
+}
+
+generate_sine () {
+    dlogi Creating sine wave file
+    octave --silent --no-gui --eval "check_volume_levels('generate', '$APLAY_WAV')" || {
+	die "Error: failed sine wave generate."
+    }
```
marc-hb, line 371: Nit: no need for the brackets.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+	exit 2
+    fi
+
+    # Get max. channels count to use from capture device
+    # remove xargs after the trailing blank from tplgreader is fixed
+    CAP_CHANNELS=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d ch_max -v | xargs)
+    CAP_HW=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d dev -v | xargs)
+    PLAY_HW=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:playback" -d dev -v | xargs)
+    CAP_PGA=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d pga -v | xargs)
+    PLAY_PGA=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:playback" -d pga -v | xargs)
```
marc-hb, line 316: You can split these on multiple lines if you want. When a command is not complete the shell is generally smart enough not to interpret new lines (when it isn't it smart enough the error message will be clear and fast)

marc-hb (changes requested):
Thanks for your patience, I think we're getting very close with the shell script. Need someone to review the matlab code and logic.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+source "$(dirname "${BASH_SOURCE[0]}")/../case-lib/lib.sh"
+
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
+OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
+
+func_opt_parse_option "$@"
+
+tplg=${OPT_VALUE_lst['t']}
+
+TPLGREADER=../tools/sof-tplgreader.py
```
marc-hb, line 39: I missed this, found it when trying to run the script: test scripts must be runnable from anywhere. Fixed with this change:
```diff
@@ -26,8 +26,10 @@ set -e
 ##    command line check with $? without error
 ##
 
+
+TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 # shellcheck source=case-lib/lib.sh
-source "$(dirname "${BASH_SOURCE[0]}")/../case-lib/lib.sh"
+source "$TESTDIR/case-lib/lib.sh"
 
 OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
 OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
@@ -36,7 +38,9 @@ func_opt_parse_option "$@"
 
 tplg=${OPT_VALUE_lst['t']}
 
-TPLGREADER=../tools/sof-tplgreader.py
+TPLGREADER="$TESTDIR"/tools/sof-tplgreader.py
+
+
 PCM_ID=0
 CAP_FORMAT="S16_LE"
 CAP_RATE="48000"

```

test-case/check-volume-levels.sh
marc-hb: The problem with this `xargs` workaround: pipelines ignore errors and break `set -e`. I actually had a complete $TPLGREADER failure (see above) but the script kept going  anyway with empty values like nothing happened.

Please submit this as a different initial commit, seems to work:

```diff
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -250,7 +250,7 @@ if __name__ == "__main__":
                 output += str(value) + " "
             else:
                 output += key + "=" + str(value) + ";"
-        return output
+        return output.strip()
 
     def func_export_pipeline(pipeline_lst):
         length = len(pipeline_lst)

```

In the unlikely case this simple `sof-tplgreader.py` fix doesn't work there is also:

```shell
CAP_CHANNELS=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:capture" -d ch_max -v )
CAP_CHANNELS=${CAP_CHANNELS//[[:space:]]/}
```

marc-hb (changes requested):

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    if [ "$numpga" != 1 ]; then
+	die "Error: more than one capture PGA found."
+    fi
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Volume)
+    search="name="
+    CAP_VOLUME=${tmp#*$search}
+    export CAP_VOLUME
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Switch)
```
singalsu, line 338: This grep and above one for Volume need to be added || true to be able to next check for empty $tmp and in understandable way report what's the issue with topology.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    pause
+
+    cset_unmute
+    pause
+
+    cset_mutevol
+    pause
+
+    cset_min
+    pause
```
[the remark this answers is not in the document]
singalsu: OK

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+
+    cset_volume_diff2 "$CAP_CHANNELS"
+
+    # Wait for arecord process to complete. %1 is aplay and %2 is the just launched arecord.
+    dlogi "Waiting"
+    wait %2
```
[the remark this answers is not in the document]
singalsu: I tried clean and nice looking wait %% but it didn't work (hope I didn't have any other mess). It would be possible to grep the correct %N from jobs output but I didn't do that yet. Pid would be nicer.

test-case/check-volume-levels.sh
singalsu: OK

test-case/check-volume-levels.sh
singalsu: Yep, the latter looks good.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+
+    dlogi Playing file "$1"
+    timeout -k 5 120 aplay -D "$PLAY_HW" "$1" 
+}
+
+do_preparations () {
+
+    if [[ "$tplg" != *"nocodec"* ]]; then
+	dlogi "This test is executed only with nocodec topologies. Returning Not Applicable."
+	exit 2
```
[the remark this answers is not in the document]
singalsu: There's no exit code 2 possibility with die.

test-case/check-volume-levels.sh
singalsu: There's also problem with device like "hw:0,0 " not being accepted. I'll enter the above fix PR if you didn't already.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+    PLAY_PGA=$($TPLGREADER "$tplg" -f "id:$PCM_ID & type:playback" -d pga -v | xargs)
+    export CAP_CHANNELS
+    export CAP_HW
+    export PLAY_HW
+
+    dlogi Test uses "$CAP_CHANNELS" channels
+    dlogi Playback device is "$PLAY_HW"
+    dlogi Playback PGA is "$PLAY_PGA"
+    dlogi Capture device is "$CAP_HW"
+    dlogi Capture PGA is "$CAP_PGA"
```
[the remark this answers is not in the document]
singalsu: Yep

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,386 @@
...
+
+    # Find amixer controls for capture, error if more than one PGA
+    numpga=${#CAP_PGA[@]}
+    if [ "$numpga" != 1 ]; then
+	die "Error: more than one capture PGA found."
+    fi
```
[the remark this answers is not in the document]
singalsu: OK

test-case/check-volume-levels.sh
singalsu: Done.

test-case/check-volume-levels.sh
singalsu: Ah, I thought it's only for compacting. Shellcheck misses  this.

test-case/check-volume-levels.sh
singalsu: Yep.. shell syntax is confusing and didn't realize it can be done. I guess this is csh-ism.

test-case/check-volume-levels.sh
singalsu: Yep

singalsu:
This test code needs sof-test PR #417 to not fail with channel count and device strings and SOF PR https://github.com/thesofproject/sof/pull/3509 to have needed full-blow volume component controls instantiated.

marc-hb (commented):
I believe no one reviewed the Octave code yet :-( Who could do this?

I would be very worrying if we have only one person who can review Octave code, maybe we shouldn't even be using Octave in validation in that case! :-((

> This test code needs sof-test PR #417 to not fail with channel count and device strings and SOF PR thesofproject/sof#3509 to have needed full-blown volume component controls instantiated.

I still have a few minor comments but the shell script looks basically ready to me. However considering this test cannot run in CI (yet), we need IMHO evidence of at least one person other than the submitter who could run the test. It doesn't even have to pass, just to run and produce some meaningful audio results anywhere but on the submitter's system. I tried it last week but I couldn't run it on my platform in a reasonable time, not even with some local modifications/hacks. @singalsu should I wait until thesofproject/sof#3509 is merged before trying again?

marc-hb:
https://sof-ci.01.org/softestpr/PR410/build251/devicetest/ experienced a device failure on BSW_CYN_MAX98090
```
kernel: [    4.509348] EXT4-fs (mmcblk0p2): warning: maximal mount count reached, running e2fsck is recommended
kernel: [    4.515925] EXT4-fs (mmcblk0p2): re-mounted. Opts: errors=remount-ro
```

keyonjie:
@singalsu I am trying to run your test case on TGL nocodec, it fails as it has no 'Switch' kcontrol there, which tplg I should use to run it?

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,335 @@
...
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Volume || true )
+    [[ -z $tmp ]] && die "No control with name Volume found in $CAP_PGA"
+    search="name="
+    CAP_VOLUME=${tmp#*$search}
+    export CAP_VOLUME
+    dlogi "Capture volume control name is $CAP_VOLUME"
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Switch || true )
+    [[ -z $tmp ]] && die "No control with name Switch found in $CAP_PGA"
```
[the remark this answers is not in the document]
singalsu: Done

test-case/check-volume-levels.sh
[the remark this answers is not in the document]
singalsu: Good that you were persistent. It seems to work and I must have messed up with my previous try.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,335 @@
...
+    dlogi "Playback PGA is $PLAY_PGA"
+    dlogi "Capture device is $CAP_HW"
+    dlogi "Capture PGA is $CAP_PGA"
+
+    # Find amixer controls for capture, error if more than one PGA
+    numpga=${#CAP_PGA[@]}
+    test "$numpga" = 1 || die "Error: more than one capture PGA found."
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Volume || true )
+    [[ -z $tmp ]] && die "No control with name Volume found in $CAP_PGA"
```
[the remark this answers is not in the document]
singalsu: OK. I checked that test -n works for both success and fail.

singalsu:
@keyonjie There's need for https://github.com/thesofproject/sof/pull/3509 to add the missing switch controls.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,336 @@
...
+    # Start sine playback
+    aplay_helper "$APLAY_WAV" & aplayPID=$!
+    pause
+
+    # Do capture tests
+    run_test_1
+    run_test_2
+    run_test_3
+
+    # Stop sine playback and cleanup. %1 is the remaining aplay process.
```
marc-hb, line 78: > %1 is the remaining aplay process.

This comment is obsolete.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,336 @@
...
+    cset_mute; pause
+    cset_unmute; pause
+    cset_mute; pause
+    cset_max; pause
+    cset_unmute; pause
+    cset_mutevol; pause
+    cset_min; pause
+    cset_volume_diff2 "$CAP_CHANNELS"
+
+    # Wait for arecord process to complete. %1 is aplay and %2 is the just launched arecord.
```
marc-hb, line 120: Obsolete comment

marc-hb (commented):
The shell script looks good to me. Before merging I think this still needs:
- at least one other person who could run this (even if the tests don't pass)
- someone to review the Octave code.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,335 @@
...
+    test "$numpga" = 1 || die "Error: more than one capture PGA found."
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Volume || true )
+    [[ -z $tmp ]] && die "No control with name Volume found in $CAP_PGA"
+    search="name="
+    CAP_VOLUME=${tmp#*$search}
+    export CAP_VOLUME
+    dlogi "Capture volume control name is $CAP_VOLUME"
+
+    tmp=$(amixer controls | grep "$CAP_PGA" | grep Switch || true )
```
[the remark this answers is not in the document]
marc-hb: I'm fine with either option but I do expect every developer to understand a regular expression as basic as "$CAP_PGA.*Switch"!

juimonen (approved):
LGTM or the octave code

xiulipan (changes requested):
Could we add a check for octave in test case first to avoid accidently run fail for this case.

marc-hb:
> Could we add a check for octave in test case first to avoid accidently run fail for this case.

I think this could be useful later once we want to run this in automation but we would like to merge it earlier than that. For now I believe we still don't have any report yet of any other person besides the author who could run it manually.

singalsu:
@xiuli Yes, makes sense I'll add the check for octave.

@keyonjie Can you please add your review / report of what was need to get it running on your device.

xiulipan (dismissed):
LGTM, let's try to do some integration with CI framework.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,341 @@
...
+    amixer cset name="$CAP_SWITCH" off
+}
+
+cset_unmute () {
+    dlogi "Set unmute"
+    amixer cset name="$CAP_SWITCH" on
+}
+
+pause () {
+    dlogi "Pausing"
```
keyonjie, line 198: The wording "pause" is somewhat confusion, we are not aim to pause any playback/capture stream, maybe 'sleep_1s'?

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,341 @@
...
+    pause
+    dlogi "The test procedure is now complete. Killing the aplay process $aplayPID."
+    kill $aplayPID
+
+    # Measure, delete unnecessary wav files if passed
+    if measure_levels; then
+	dlogi "Deleting temporary files"
+	rm -f "$APLAY_WAV" "$ARECORD_WAV1" "$ARECORD_WAV2" "$ARECORD_WAV3"
+    fi
+}
```
keyonjie, line 88: Since we have known issue https://github.com/thesofproject/sof/issues/3520 for nocodec SSP LBM, and this test case is designed to test the volume level tuning only, maybe better to stop the playback between capture iterations to avoid hitting the not related issue, some changes like this, I verified it can pass 10 iterations on my TGL nocodec.
```
diff --git a/test-case/check-volume-levels.sh b/test-case/check-volume-levels.sh
index b4ff6ed..40fe4c1 100755
--- a/test-case/check-volume-levels.sh
+++ b/test-case/check-volume-levels.sh
@@ -67,18 +67,28 @@ main () {
     generate_sine
 
     # Start sine playback
-    aplay_helper "$APLAY_WAV" & aplayPID=$!
+    aplay_helper "$APLAY_WAV" &
     pause
 
-    # Do capture tests
+    # Do capture test 1
     run_test_1
+
+    # Restart playback for capture test 2
+    killall aplay
+    pause
+    aplay_helper "$APLAY_WAV" &
     run_test_2
+
+    # Restart playback for capture test 3
+    killall aplay
+    pause
+    aplay_helper "$APLAY_WAV" &
     run_test_3
 
     # Stop sine playback and do cleanup
     pause
-    dlogi "The test procedure is now complete. Killing the aplay process $aplayPID."
-    kill $aplayPID
+    dlogi "The test procedure is now complete. Killing the aplay process."
+    killall aplay
 
     # Measure, delete unnecessary wav files if passed
     if measure_levels; then
```

tools/check_volume_levels.m
```diff
@@ -0,0 +1,251 @@
...
+		ret.levels(i, :) = level_dbfs(x(i1:i2, :));
+	end
+	ret.levels_lin = 10.^(ret.levels/20);
+end
+
+function y = bandpass_filter(x, f, fs)
+	sx = size(x);
+	y = zeros(sx(1), sx(2));
+	c1 = 0.8;
+	c2 = 1/c1;
```
keyonjie, line 246: not sure if the -37dB value is related with this [0.8, 1.25] window, looks there could be some level captured when filtering 701Hz with a 1297Hz signal, 701 * [0.8, 1.25] = [560.8Hz, 876,25Hz]. So if we change this window to e.g. [0.9, 1.11] then the -37dB shift should become smaller like -50dB?

keyonjie (commented):
Thanks for the case, it looks professional with good coverage on the volume level.
For the pipeline coverage, it only change volume/switch for the capture one, is it worth to add these to playback pipeline also?

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,341 @@
...
+
+    # Do capture tests
+    run_test_1
+    run_test_2
+    run_test_3
+
+    # Stop sine playback and do cleanup
+    pause
+    dlogi "The test procedure is now complete. Killing the aplay process $aplayPID."
+    kill $aplayPID
```
keyonjie, line 81: This doesn't work for me sometimes, that's why I suggest to replace it with 'killall aplay'

tools/check_volume_levels.m
```diff
@@ -0,0 +1,251 @@
...
+				fprintf(1, 'FAIL\n');
+				exit(1);
+			end
+		case 'measure'
+			pass = measure(fn1, fn2, fn3);
+			if pass
+				fprintf(1, 'PASS\n');
+			else
+				fprintf(1, 'FAIL\n');
+				exit(1);
```
keyonjie, line 41: shall we remove those 2 'exit(1)' as they are making my octave hang when running it with gui mode, logging errors should be enough?

test-case/check-volume-levels.sh
singalsu: The test script really pauses execution so I prefer this. Just sleep is a reserved word and don't want to hard code with the name the length to 1s. It could be adjusted later to more than 1s for better accuracy or less than 1s for lower test time.

test-case/check-volume-levels.sh
singalsu: I used the continuous sine wave to lower the test execution time. It's especially useful if the test if electrical or acoustical. No need to wait two times more time before the playback is stable. Also there's more code "+" lines than "-" lines in your patch. I'm hesitating to change this just to avoid to trigger an issue. But no firm opinion. What do you think @marc-hb ?

tools/check_volume_levels.m
singalsu: I used quite relaxed band-pass because the filter design for IIR filters is not always robust. The narrower the band or the lower the frequencies are the more it's numerically sensitive. Octave does not support well the "z, p, k" and "sos" notation that is better than "b, a" so, to keep the function simple I kept the "b, a" filter coefficients with relaxed filter spec. The 37 dB mismatch when channels got swapped is quite large, well above error margin.

test-case/check-volume-levels.sh
singalsu: I didn't want use it if there would another test running with different sound device. What kind of problem did you see?

tools/check_volume_levels.m
singalsu: This script is not for GUI mode but for batch processing. For interactive tests with GUI I've commented out those lines. The GUI doesn't hang but it quits with exit().

singalsu: I'm not aware of simple ways in script to check if we are in Octave GUI. If such exist we could make a custom exit function.

marc-hb: >  This script is not for GUI mode but for batch processing. 

It would be great to find a solution that supports both but waiting for that should IMHO not block this "automation" PR.

test-case/check-volume-levels.sh
marc-hb: If this fails only "sometimes" then it sounds like this test found a real audio bug . @keyonjie can you please elaborate?

test-case/check-volume-levels.sh
marc-hb: I understand it is this test that helped find issue thesofproject/sof#3520 so it sounds crazy to me to modify it so it *hides* issue 3520 before we even have another test to reproduce 3520.

While tests are generally focused on one thing they always "accidentally" test many other things and that's very good. The more bugs a test can find (in a resonable time) and the better test it is.

If we want to test volume levels in automation before 3520 is fixed then something like this should do it:
```shell
platform=$(sof-dump-status.py -p)
case $platform in
   byt|others)
   dloge "skipping this test until http://#3520 is fixed
   exit 2 # 2 means SKIP for Jenkins ;;
esac
```

test-case/check-volume-levels.sh
marc-hb: `nap()` ? :-)

marc-hb (approved):
> This doesn't work for me sometimes, 

Considering this test is not enabled in automation yet, working "sometimes" for at least one person besides the submitter is IMHO enough "validation". Merging it will make fixing it easier.

test-case/check-volume-levels.sh
keyonjie: it reports error sometimes like this:
```
2020-10-19 05:39:45 UTC [INFO] The test procedure is now complete. Killing the aplay process 9360.
./check-volume-levels.sh: line 81: kill: (9360) - No such process
```

tools/check_volume_levels.m
keyonjie: @singalsu @marc-hb just removing this 'exit(1)' line will work for both GUI and batch processing per my testing.

test-case/check-volume-levels.sh
keyonjie: yes, either 'nap' / 'nop' or 'idle' sound better to me, or even exec_pause(), as we have test cases naming xxx_pause_release_xxx, I thought we are pausing the playback/capture pipeline at my first glance to the PR.

tools/check_volume_levels.m
keyonjie: Thanks for the info, though I am not quite understand it. :-)

test-case/check-volume-levels.sh
keyonjie: Yes, let's keep this, thanks.

tools/check_volume_levels.m
singalsu: The shell script thinks a failed test passed if I omit the exit(1).

test-case/check-volume-levels.sh
singalsu: nap sounds good to me!

tools/check_volume_levels.m
singalsu: Take some time to play in Octave with butter() (lowpass, highpass,bandpass) and freqz() and you'll see the issue. For deep dive plot the  poles (complex, roots of denominator polynomial) of the transfer function and an unit circle. The narrow and very low frequency filters can have numerical coefficients precision problems that make them non-stable.

test-case/check-volume-levels.sh
singalsu: If the recording process is not severely delayed it sounds like a bug. The playback sine wave duration is 60s and for me the test completes in less than 30s so there should be nearly 30s margin. Could the sample rates be severely off in the platform where you test (e.g. 10s recording takes much more time)?

tools/check_volume_levels.m
keyonjie: Okay, got you now @singalsu

test-case/check-volume-levels.sh
marc-hb: > ./check-volume-levels.sh: line 81: kill: (9360) - No such process

If the process is already dead then `killall` would: 1. hide what looks like a bug 2. possibly kill other, unrelated processes.

marc-hb:
jf-icl-rvp-hda-4 boot failed in  https://sof-ci.01.org/softestpr/PR410/build274/devicetest/

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,341 @@
...
+}
+
+arecord_helper () {
+    dlogi "Capturing file $1"
+    timeout -k 5 30 arecord -D "$CAP_HW" -f $CAP_FORMAT -r $CAP_RATE -c "$2" -d "$3" "$1"
+}
+
+aplay_helper () {
+    test -f "$1" || die "Error: File $1 does not exist"
+    dlogi "Playing file $1"
```
marc-hb, line 257: Just realized that `test && die()` is useless here because you always background `aplay_helper()`.

`aplay_helper()` seems to be used only once so you should probably inline it, get rid of the function and background the `timeout` instead. Who knows it may even help @keyonjie help debug his "no such process" issue?

tools/check_volume_levels.m
marc-hb: > The shell script thinks a failed test passed if I omit the exit(1).

Thanks @singalsu for "testing the test code" and making sure it can report failures. I found this to be incredibly rare in this project.

After only 2 minutes googling (sorry) I found "error()", did you try it? https://octave.org/doc/v4.0.1/Raising-Errors.html

tools/check_volume_levels.m
keyonjie: > Take some time to play in Octave with butter() (lowpass, highpass,bandpass) and freqz() and you'll see the issue. For deep dive plot the poles (complex, roots of denominator polynomial) of the transfer function and an unit circle. The narrow and very low frequency filters can have numerical coefficients precision problems that make them non-stable.

Thanks.

test-case/check-volume-levels.sh
keyonjie: @singalsu @marc-hb OK let's keep it as what it is today in the PR, as it doesn't introduce any real issue except seeing this warning, we can fix it later.

keyonjie (approved):
Looks good, we can add incremental fixes if minor issues existed.

singalsu:
I did the proposed and agreed updates. In addition I added a verbose flag to level check function to prevent additional reports of incorrect level printed while the script tries if swapped channels and swapped controls would pass. I always test the changes with manual add of error to controls or measurements side to see I get the pass and fail correctly.

marc-hb (approved):
> Looks good, we can add incremental fixes if minor issues existed.

+1 considering this doesn't run in CI yet.

test-case/check-volume-levels.sh
```diff
@@ -0,0 +1,337 @@
...
+
+do_preparations () {
+
+    if [[ "$tplg" != *"nocodec"* ]]; then
+	# Return special value 2 with exit
+	dlogi "This test is executed only with nocodec topologies. Returning Not Applicable."
+	exit 2
+    fi
+
+    test -n "$(command -v octave)" || {
```
marc-hb, line 265: Simpler and logging the location
```shell
type octave || {
   ...
}
```

marc-hb (approved):

keyonjie (approved):

test-case/check-volume-levels.sh
singalsu: I don't prefer this skip, and this is so far manual run only. If the fails become a nuisance we can add this later.

test-case/check-volume-levels.sh
singalsu: I didn't know that, it's very compact but not very intuitive. If you are OK to re-review I can change this.

Issue #412, opened 2020-10-08
check-sof-logger fails on CHT device w/ rt5640


plbossart:
````
root@Zotac:~/sof-test/test-case# ./check-sof-logger.sh 
2020-10-08 16:13:06 UTC [REMOTE_INFO] Found file: /usr/local/bin/sof-logger d7b6b0771ad4c00f8f81f63e1fb79226
2020-10-08 16:13:07 UTC [REMOTE_INFO] Checking ldc File: /etc/sof/sof-.ldc ...
2020-10-08 16:13:07 UTC [REMOTE_ERROR] File (/etc/sof/sof-.ldc) Not Found!
2020-10-08 16:13:07 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2020-10-08 16:13:07 UTC [REMOTE_ERROR]  die()  @  ./../case-lib/lib.sh
2020-10-08 16:13:07 UTC [REMOTE_ERROR]  main()  @  ./check-sof-logger.sh:44
2020-10-08 16:13:07 UTC [REMOTE_INFO] Test Result: FAIL!
````

Clearly it has no idea what the platform is

````
root@Zotac:~/sof-test/tools# ./sof-dump-status.py 
DMI Info:
	bios_vendor:		AmericanMegatrendsInc.
	bios_version:		5.11
	bios_date:		09/28/2016
	bios_:		5.11
	system_vendor:		ZOTAC
	product_name:		XXXXXX
	product_version:	XX
	board_vendor:		ZOTAC
	board_name:		CherryTrailFFD
	board_version:		XX
	chassis_vendor:		Defaultstring
	chassis_type:		3
	chassis_version:	Defaultstring

Couldn't detect for PCI device for audio
Card ID:		0
	Type:		rt5640
	Short name:	sof-bytcht rt5640
	Long name:	ZOTAC-XXXXXX-XX-CherryTrailFFD
	Codec:		NOCODEC
	PCM:
		ID:	0
		PCM:	PCM
		Type:	Playback
		ID:	0
		PCM:	PCM
		Type:	Capture
		ID:	1
		PCM:	PCM Deep Buffer
		Type:	Playback

Power Info:
	Current option:	s2idle
		option:	[s2idle]
	Wakeup Count:	1
Power Status:

````

plbossart:
the same script works fine on Chromebook / Cyan, so there's a missing autodetection or something.

````
[...]

PR #414, opened 2020-10-08
sof-dump-status: fix ACPI support

plbossart:
The ACPI support is currently problematic and only works if there is a
PCI device, which is true only on MinnowBoard and Chromebook Cyan used
for SOF CI, but false in general for all other BYT/CHT devices.

There is no reason why ACPI needs to be tied to PCI, devices can be
found in /sys/bus/acpi/devices/ directly.

### DIFF
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -7,6 +7,7 @@ class clsSYSCardInfo():
     def __init__(self):
         self.dmi={}
         self.pci_lst=[]
+        self.acpi_lst=[]
         self.proc_card={}
         # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c
         self._pci_ids={"0x119a":"tng","0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
@@ -73,7 +74,6 @@ class clsSYSCardInfo():
         # grep exit 1 means nothing matched
         if exit_code != 0:
             return
-        apci_key_lst = self._acpi_ids.keys()
         for line in output.splitlines():
             pci_info = {}
             pci_info['pci_id'] = line.split(' ')[0]
@@ -87,9 +87,32 @@ class clsSYSCardInfo():
                     break
             pci_info['hw_id']="0x" + tmp_line[2] + tmp_line[1] + " 0x" + tmp_line[4] + tmp_line[3]
             pci_info['hw_name'] = self._pci_ids["0x" + tmp_line[4] + tmp_line[3]]
-            if pci_info['hw_name'] in apci_key_lst:
-                pci_info['device_id'] = self._acpi_ids[pci_info['hw_name']]
             self.pci_lst.append(pci_info)
+
+    def loadACPI(self):
+        self.acpi_lst.clear()
+
+        for mach in self._acpi_ids.keys():
+            acpi_info = {}
+
+            # On devices such as MinnowBoard the ACPI subsystem
+            # creates devices 80860F28:00 and 80860F28:01. We first
+            # need to loop to find out which of the two is enabled.
+            device_index = -1
+            for i in range(0, 2):
+                # make sure the ACPI status is 15 (indicates device presence)
+                exit_code, output=subprocess.getstatusoutput(
+                    "cat /sys/bus/acpi/devices/%s:0%d/status" %
+                    (self._acpi_ids[mach], i))
+                if exit_code == 0 and output == "15":
+                    device_index = i
+                    break
+
+            if device_index != -1:
+                acpi_info['hw_name'] = mach
+                acpi_info['acpi_id'] = self._acpi_ids[mach]
+                acpi_info['acpi_id_suffix'] = device_index
+                self.acpi_lst.append(acpi_info)
 
     def loadProcSound(self):
         self.proc_card.clear()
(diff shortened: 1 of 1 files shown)


plbossart:
Fixes #412

tools/sof-dump-status.py
```diff
@@ -366,8 +402,15 @@ def dump_dapm(dapm, filter = "all"):
     sysinfo = clsSYSCardInfo()
     if ret_args['platform'] is True:
         sysinfo.loadPCI()
+        mach_name = None
         for pci_info in sysinfo.pci_lst:
-            print(pci_info['hw_name'])
+            mach_name = pci_info['hw_name']
```
marc-hb, line 407: It looks like `mach-name` will simply be equal to the last element `sysinfo.pci_lst[-1]['hw_name']`. So what does this `for` loop achieve?

Is there an indentation issue maybe?

tools/sof-dump-status.py
```diff
@@ -366,8 +402,15 @@ def dump_dapm(dapm, filter = "all"):
...
+        if (mach_name is None):
+            sysinfo.loadACPI()
+            for acpi_info in sysinfo.acpi_lst:
+                mach_name = acpi_info['hw_name']
```
marc-hb, line 411: Same comment: why the for loop instead of accessing the last element directly?

tools/sof-dump-status.py
```diff
@@ -200,6 +222,12 @@ def loadDAPM(self, filter = "all"):
                 continue
             self.dapm['ctrl_lst'].append({'id': pci_info['pci_id'], 'status':output})
 
+        for acpi_info in self.acpi_lst:
+            exit_code, output=subprocess.getstatusoutput("cat /sys/bus/acpi/devices/%s:00/power/control" % (acpi_info['acpi_id']))
```
marc-hb, line 226: Same as above

tools/sof-dump-status.py
```diff
@@ -174,22 +184,34 @@ def loadPower(self):
...
                 continue
             self.sys_power['run_status'].append({'map_id': pci_info['pci_id'], 'status': output})
 
+        if len(self.acpi_lst) == 0:
+            self.loadACPI()
+
+        for acpi_info in self.acpi_lst:
+            exit_code, output=subprocess.getstatusoutput("cat /sys/bus/acpi/devices/%s:00/power/runtime_status" % (acpi_info['acpi_id']))
+            if exit_code != 0:
+                continue
```
marc-hb, line 203: Same as above

tools/sof-dump-status.py
```diff
@@ -87,10 +87,20 @@ def loadPCI(self):
...
-            if pci_info['hw_name'] in apci_key_lst:
-                pci_info['device_id'] = self._acpi_ids[pci_info['hw_name']]
             self.pci_lst.append(pci_info)
 
+    def loadACPI(self):
+        self.acpi_lst.clear()
+
+        for mach in self._acpi_ids.keys():
+            acpi_info = {}
+            exit_code, output=subprocess.getstatusoutput("cat /sys/bus/acpi/devices/%s:00/status" % (self._acpi_ids[mach]))
```
marc-hb, line 97: `getstatusoutput()` has been deprecated for a long time. This shebang requires Python3 anyway.

I see it's already used left  and right but here no `subprocess` seems needed at all to just read a file?! So no need to copy/paste crap, instead:
```python
sys_name = "/sys/bus/acpi/devices/%s:00/status" % self._acpi_ids[mach]
try:
   with open(sys_name) as f:
       output = f.readline().strip()
except FileNotFoundError:
    print(sys_name + ' not found')
    continue
```

If this file should always exist then you can drop the `try: except` entirely and it becomes just 2.5 lines.

This will provide a good template to fix the rest later.

plbossart: I am not going to solve everything in this script. The suggested changes can be done separately.

tools/sof-dump-status.py
plbossart: don't ask  me, it's the same pattern everywhere.

tools/sof-dump-status.py
plbossart: I have no idea, this is what I started from. This script needs a lot of work, but for now the problem is to solve conceptual problems coming from PCI and ACPI confusions. How the code is written is a separate topic entirely.

tools/sof-dump-status.py
```diff
@@ -87,10 +87,20 @@ def loadPCI(self):
...
+            if exit_code != 0:
+                continue
+            acpi_info['hw_name'] = mach
+            acpi_info['acpi_id'] = self._acpi_ids[mach]
+            self.acpi_lst.append(acpi_info)
```
ranj063, line 102: Shouldnt we just break when one of the ID's is found?

tools/sof-dump-status.py
marc-hb: Copy/paste of code that doesn't make sense goes nowhere fast.

tools/sof-dump-status.py
marc-hb: > I am not going to solve everything in this script. The suggested changes can be done separately.

Good I'm not asking to change any existing code.

tools/sof-dump-status.py
plbossart: Don't ask me why a list is used. This is the same code we've been using *for months*, I just solved the ACPI/PCI confusion.

Whoever wrote that script made a lot of assumptions that made no sense, I just found that the DMI parsing is incorrect as well.

marc-hb: > Whoever wrote that script made a lot of assumptions that made no sense, I just found that the DMI parsing is incorrect as well.

Please at least add `FIXME: why is this a list?` comments so your time reverse engineering this is not completely wasted.

tools/sof-dump-status.py
ranj063: isnt os.path.exists(path) enough?

marc-hb: `cat` reads the file.

tools/sof-dump-status.py
plbossart: I'll add a separate bug, there are lists for PCI everywhere.

tools/sof-dump-status.py
plbossart: yes, and the missing part is actually to test that the status == 15, which indicates that the device wasn't disabled. Sometimes you see devices but the OS is not supposed to use them for some reason. I didn't add this part but it's needed for consistency

plbossart:
updated to fix kmod load/unload issues. The scripts assume the DSP device has index 0, which means the ACPI device, when present, needs to be listed first and PCI second (that would typically be the HDaudio/HDMI device).

plbossart:
SOFCI TEST

plbossart:
I don't get what happens with MinnowBoard Nocodec, I guess I'll have to debug myself what happens on this platform.

ranj063 (approved):

plbossart:
> I don't get what happens with MinnowBoard Nocodec, I guess I'll have to debug myself what happens on this platform.

Valid error reported by SOF CI, on MinnowBoard the ACPI device are listed as 80860F28:00 and 80860f28:01, only the second is enabled. Updated the code to loop and find the first enabled device.

plbossart:
@marc-hb @ranj063 @aiChaoSONG I am not going to spend more time on this script. it does what I needed and does not introduce any regression for CI devices.

I provided solutions to better handle ACPI devices, but didn't change the structure of this scripts. That is another topic for someone with more available time.

aiChaoSONG (approved):
Thanks Pierre for the ACPI part.

PR #417, opened 2020-10-09
Tools: Fix for printing extra blank after value

### LINKED ISSUE #408: sof-tplgreader.py -v returns blank after value
E.g. 

```
$ echo X$(./sof-tplgreader.py $TPLG -f "id:0 & type:capture & pga" -d ch_max -v)X
X2 X
```

This breaks in shell script simple check for channel count.

singalsu:
The just display value option "-v or --value" for sof-tplgreader.py
outputs after the actual value a space character. It can cause
unexpected issues when utilizing the values in test scripts. The added
strip fixes the issue.

Fixes #408

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -250,7 +250,7 @@ if __name__ == "__main__":
                 output += str(value) + " "
             else:
                 output += key + "=" + str(value) + ";"
-        return output
+        return output.strip()
 
     def func_export_pipeline(pipeline_lst):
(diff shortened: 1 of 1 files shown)


marc-hb (approved):
After looking around in sof-test, I think the reason no one needed this before is because the `noKey` option is never used. It's always `key=value;` + `eval` instead. If correct this also means this PR cannot break anything.

aiChaoSONG (approved):
@marc-hb Totally agreed

PR #418, opened 2020-10-09
run-all-tests: make the test durations somewhat variable

marc-hb:
Still a hack but better than nothing at all.

Barely any change to the previous, default values.

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -1,6 +1,25 @@
 #!/bin/bash
 set -e
 
+# This entire script is a hack. We need to re-use some existing test
+# runner from some existing test framework eventually
+
+# These can also be overriden by the environment
+very_short_tests()
+{
+	small_loop=1
+	medium_loop=1
+	large_loop=1
+
+	medium_count=5
+	large_count=10
+
+	long_duration=3
+
+	large_round=3
+}
+
+
 testlist="
 firmware-presence
 firmware-load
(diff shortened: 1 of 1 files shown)


marc-hb:
As requested by @plbossart 

For a quick run (~ 20 min?) comment the `very_short_tests` invocation back in. Default values are almost the same as before.

For a medium run copy/paste the `very_short_tests` function and tweak the values.

This also gives some more flexibility using optional environment variables.

It's all still a hack but should help us "survive" until we evaluate and use a proper test runner/framework.

test-case/run-all-tests.sh
```diff
@@ -1,6 +1,24 @@
...
+	:  "${large_loop:=1}"
+
+	:  "${medium_count:=5}"
+	:  "${large_count:=10}"
+
+	: "${long_duration:=3}"
+
+	: "${large_round:=3}"
+}
+# very_short_tests
```
plbossart, line 20: could we make this configurable with e.g. a -s option?

test-case/run-all-tests.sh
```diff
@@ -94,99 +128,99 @@ test_sof-logger()
...
 }
 test_kmod-load-unload()
 {
-	"$mydir"/check-kmod-load-unload.sh -l 50
+	"$mydir"/check-kmod-load-unload.sh -l "$medium_loop"
 }
 test_kmod-load-unload-after-playback()
 {
-	"$mydir"/check-kmod-load-unload-after-playback.sh -l 15
+	"$mydir"/check-kmod-load-unload-after-playback.sh -l "$small_loop"
```
plbossart, line 211: also we could disable these kmod tests with an option since they don't work on all devices (e..g Cyan)

plbossart (dismissed):
This is good enough for now, I am using the 'very-short-tests' and it's really useful for developers trying to check things with a quick smoke test. See comments below for improvements.

fredoh9 (approved):
Pretty nice we have a easy way to enable very_short_test by removing comment. And also define env variable to override number of loop!

plbossart:
Doesn't fully work, this adds errors in my tests. I suggest the following fix @marc-hb, one variable is not consistently set compared to the others, and there's a typo for the variable name.

````diff
diff --git a/test-case/run-all-tests.sh b/test-case/run-all-tests.sh
index 5afd673..0af04c2 100755
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -60,7 +60,7 @@ main()
        : "${medium_count:=50}"
        : "${large_count:=100}"
 
-       : "${large_duration:=100}"
+       : "${long_duration:=100}"
 
        : "${large_round:=50}"
 
@@ -132,11 +132,11 @@ test_ipc-flood()
 }
 test_playback-d100l1r1()
 {
-       "$mydir"/check-playback.sh -d ${long_duration} -l 1 -r 1
+       "$mydir"/check-playback.sh -d "$long_duration" -l 1 -r 1
 }
 test_capture-d100l1r1()
 {
-       "$mydir"/check-capture.sh -d ${long_duration} -l 1 -r 1
+       "$mydir"/check-capture.sh -d "$long_duration" -l 1 -r 1
 }
 test_playback-d1l100r1()
 {
````

plbossart:
diff file 
[duration.diff.txt](https://github.com/thesofproject/sof-test/files/5372358/duration.diff.txt)

test-case/run-all-tests.sh
marc-hb: Done: `-l 1`

test-case/run-all-tests.sh
marc-hb: How are they skipped in CI? If they SKIP with  `exit 2` then maybe we could catch this here just like Jenkins does.... in a later PR.

We also need to be conscious not to turn this hack into yet another NIH test runner.

marc-hb (commented):
Thanks a lot for fixing the bugs, applied and new option added. The first version was a bit "rushed" per your request, this one had more testing.

plbossart (approved):

marc-hb:
Only failure at https://sof-ci.01.org/softestpr/PR418/build267/devicetest is BSW Cyan does not boot, likely another eMMC issue. This script does not even run in CI, merging.

PR #422, opened 2020-10-09
opt.sh: remove dependency on "bc"

marc-hb:
We don't need "bc" to reverse a boolean, $((! )) does it just fine.

@plbossart reported that bc is not installed on Debian by default.


marc-hb:
The only failure is BYT_MB_NOCODEC that failed to boot. Unrelated.

xiulipan:
@marc-hb Looks good please make it ready for review.

marc-hb:
> @marc-hb Looks good please make it ready for review.

Thanks for the reminder I thought I did that.

xiulipan (approved):

PR #423, opened 2020-10-09
opt.sh: don't discard getopt stderr

marc-hb:
Why lose a good error message like this one!?
```
  getopt: invalid option -- 'u'
```
Never, ever discard stderr.


xiulipan (approved):
@marc-hb Nice catch! Will merge.

PR #424, opened 2020-10-09
opt.sh: add and rephrase comments. Zero code change.

marc-hb:
Add and rephrase comments. Zero code change.


xiulipan (approved):
Thanks for review the whole file and add comments.

Issue #396, opened 2020-09-22
Ignore i2c_hid i2c-DELL0955:00 error on Dell CML laptop


mengdonglin:
Please over unrelative error of i2c_hid i2c-DELL0955:00.

Fake failure observed in testing with SOF v.16 FW and linux-stable-5.8.10 kernel:
```
2020-09-21 08:08:48 UTC [ERROR] Caught dmesg error
===========================>>
Sep 21 08:06:27 sh-cml-mantis-hda-01 kernel: [    6.496237] i2c_hid i2c-DELL0955:00: failed to set a report to device.
<<===========================
2020-09-21 08:08:48 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
```

marc-hb:
Can you please provide a longer trace? Including some unrelated context before and after: this proves nothing related was ommitted.

If you're not sure please link to some CI test report and I'll try to do it.

PR #425, opened 2020-10-10
sof-kernel-log-check: ignore i2c error for DELL CML HDA laptop + i915 Type-C error

RuiqingX:
Fixes: #396

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -61,6 +61,11 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
 ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. Restarting 0x0\.'
 ignore_str="$ignore_str"'|: authentication with ..:..:..:..:..:.. timed out'
 
+# I915, issues reported by sof-test
+# BugLink: https://github.com/thesofproject/sof-test/issues/374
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -57,6 +61,15 @@ ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zon
 ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. Restarting 0x0\.'
 ignore_str="$ignore_str"'|: authentication with ..:..:..:..:..:.. timed out'
 
+# I915, issues reported by sof-test
+# BugLink: https://github.com/thesofproject/sof-test/issues/374
+ignore_str="$ignore_str"'|i915 0000:00:02\.0: \[drm\] ERROR TC cold unblock failed'
+ignore_str="$ignore_str"'|i915 0000:00:02\.0: \[drm\] ERROR TC cold block failed'
```
marc-hb, line 67: I have a couple questions about the Type-C errors, please check bug 374

marc-hb (changes requested):
Please move these to a platform specific section below:
```
case "$platform" in
```

marc-hb (approved):
In the future it would be better to guard more of these with `case $platform`, especially the ones with strings like "DELL0955" in the error message but Rome wasn't built in a day.

RuiqingX:
fixes and closed.

RuiqingX:
It was closed by mistake

PR #428, opened 2020-10-12
sof-dump-status: add TGL-H audio PCI ID

aiChaoSONG:
This patch keeps PCI ids in sof-dump-status.py and https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c aligned

### DIFF
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -11,7 +11,7 @@ class clsSYSCardInfo():
         # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c
         self._pci_ids={"0x119a":"tng","0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
             "0x9dc8":"cnl", "0xa348":"cfl", "0x9d71":"kbl", "0x9d70":"skl", "0x34c8":"icl", "0x38c8":"jsl", "0x4dc8":"jsl",
-            "0x02c8":"cml", "0x06c8":"cml", "0xa0c8":"tgl", "0x4b55":"ehl", "0x4b58":"ehl", 
+            "0x02c8":"cml", "0x06c8":"cml", "0xa0c8":"tgl", "0x43c8":"tgl-h", "0x4b55":"ehl", "0x4b58":"ehl",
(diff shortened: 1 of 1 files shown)


marc-hb:
BSW_CYN_MAX98090 failed to boot in https://sof-ci.01.org/softestpr/PR428/build250/devicetest

I believe jf-bsw-cyn-max98090-6 has been having eMMC issues, see for instance https://sof-ci.01.org/softestpr/PR410/build251/devicetest/

Unfortunately https://sof-ci.01.org/softestpr/PR428/build250/devicetest/ does not give the device hostname.

@aiChaoSONG can we disable this device for the moment?

aiChaoSONG:
@marc-hb I think jf-bsw-cyn-98090-3 also has eMMC issue, as you can check from CI issue 539. We have 4 bsw-cyn devices with ID 3,4,5,6.  if we disable 3 and 6, I am not sure two devices are enough for PR and daily test, let's involve more people.
@mengdonglin @xiulipan

marc-hb:
This is getting more confusing every day. https://github.com/thesofproject/sof/pull/3451/files has `tgplp` and `tgph`, https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c has `sof-tgl-h.ri`. Now this python script has `tglh`. 

@lgirdwood , @plbossart , @jajanusz , can we please fix P1 documentation issue https://github.com/thesofproject/sof/issues/3491 yesterday?

aiChaoSONG:
@marc-hb this python script should keep accordance with the kernel, and the name should be `tgl-h`.

marc-hb:
> @marc-hb this python script should keep accordance with the kernel

Do you mean: "... with the .ri filename in /lib/firmware/intel/"?

aiChaoSONG:
> > @marc-hb this python script should keep accordance with the kernel
> 
> Do you mean: "... with the .ri filename in /lib/firmware/intel/"?

You can understand like this.  If we run `sof-dump-status.py -p` to get platform name, like you do in sof-kernel-log-check, we will first acquire PCI ID, and get platform name by PCI ID, and this is the dict we use to do the mapping. and the platform name appears in the FW filename, sof-PLATFORM.ri

xiulipan (approved):
LGTM.

PR #429, opened 2020-10-13
sof-kernel-log-check: extend ignore_str for CNL platforms

aiChaoSONG:
The audio PCI ID on CML Mantis is 8086:9dc8, which is
defined as CNL in linux kernel.

The ignore_str used to apply to CML should also be
applied to CNL platforms.


marc-hb (approved):
None of the shellcheck warnings is new.

PR #430, opened 2020-10-13
sof-test: read /proc instead of parsing TPLG file for legacy HDA platforms

aiChaoSONG:
This PR enhance sof-test for legacy HDA platform. The main idea is to export pipeline parameters from proc file system, just like exporting pipeline parameters from topology.


case-lib/lib.sh
```diff
@@ -264,3 +264,12 @@ sudo sync -f || true
 if [ ! "$DMESG_LOG_START_LINE" ]; then
     DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
 fi
+
+is_sof_platform()
```
plbossart, line 268: this should be is_hda_legacy_platform()

and have additional checks for the PCI ID.

case-lib/pipeline.sh
```diff
@@ -2,6 +2,20 @@
...
+    if ! is_sof_platform; then
+        dlogi "None SOF platform detected, exporting pipeline parameters from proc file system"
+        local -a proc_pipeline
+        if ! readarray -t proc_pipeline < <(eval "sof-dump-status.py -e"); then
+            die "Failed to export pipeline parameters from proc file system"
+        fi
+        for line in "${proc_pipeline[@]}"
+        do
+            eval "$line"
+        done
```
plbossart, line 16: that looks absolutely horrible to me.

plbossart (changes requested):

marc-hb (changes requested):
> that looks absolutely horrible to me.

Agree that generating shell script code from Python just to pass some values is quite ugly. What @plbossart may have not noticed: @aiChaoSONG didn't create this but copied most of it from `tools/sof-tplgreader.py` and `case-lib/pipelline.sh`.

@aiChaoSONG one copy of this code is already painful enough, so Duplicating and Diverging a second time it is really not acceptable from a longer term maintenance perspective. Please submit first a pure refactoring where you only move the parts you need to some common files with no logical change. Once it's merged you can submit a second and much smaller PR for legacy HDA that re-uses existing code.

Note you will need to prepare the second PR before submitting the first but don't share the second until the first is merged to make the review easier.

case-lib/lib.sh
xiulipan: I think we should make this as is_sof_used or is_hda_legacy_used as we can switch to use both on same platform.

xiulipan (changes requested):

tools/sof-dump-status.py
```diff
@@ -309,12 +309,34 @@ def export_proc_sound(proc_card):
...
+
     def dump_cardinfo_pcm(card_info):
-        pcm_lst = card_info.get('pcm')
-        def _getStr(tmp_dict, key):
-            return '%s=%s'%(key, tmp_dict.get(key))
-        for pcm in pcm_lst:
-            print('%s;%s;%s;' % (_getStr(pcm, 'id'), _getStr(pcm, 'pcm'), _getStr(pcm, 'type')))
+        for pipeline in card_info.get('pcm'):
+            print(dump_pipeline(pipeline))
+        return 0
```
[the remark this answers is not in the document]
aiChaoSONG: the `return 0` is from precious code, it's useless, we don't even need this sentence. deleted

case-lib/lib.sh
aiChaoSONG: now I use /proc/aound/cards to tell if we are using SOF, if yes, dump pipelines from tplg, if not, dump pipelines from proc, it's much more dependable than just check the kernel module.

tools/sof-dump-status.py
```diff
@@ -359,6 +381,7 @@ def dump_dapm(dapm, filter = "all"):
     parser.add_argument('-P', '--fwpath', action='store_true', help='get firmware path according to DMI info')
     parser.add_argument('-S', '--dsp_status', type=int, help='get current dsp power status, should specify sof card number')
     parser.add_argument('-d', '--dapm', choices=['all', 'on', 'off', 'standby'], help='get current dapm status, this option need root permission to access debugfs')
+    parser.add_argument('-e', '--export', action='store_true', help='export pipeline parameters from proc file system')
```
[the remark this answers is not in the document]
aiChaoSONG: added

aiChaoSONG:
> a bit less ugly would be to source a temporary file. I couldn't think of any "cleaner" solution

@marc-hb I think this idea is good to eliminate the python generated bash code. But currently, I am working on enable Zephyr based SOF builds in CI, I think for the code refactoring part, I will file an issue, and return later, or we can let @RuiqingX  take part in this.
In fact, the pipeline filtering feature in sof-tplgreader.py is not enabled for pipelines dumped from proc, this can also be done when extract shared code into a new library file.

marc-hb:
>  I think this idea is good to eliminate the python generated bash code. But currently, I am working on enable Zephyr based SOF builds in CI, I think for the code refactoring part, I will file an issue, 

OK let's stop discussing changing the code.

On the other hand, copy/paste is still not acceptable and copy/paste of dubious code is even less acceptable. This creates a lot of difficult work for later because de-duplication requires a lot more work and a lot of testing.

Don't change the existing code but move it to another place (any place) where it can be imported and not duplicated.

aiChaoSONG:
> move it to another place (any place) where it can be imported and not duplicated

@marc-hb I move the common code to common.py, and there is no duplicated code, should be less painful.

tools/common.py
```diff
@@ -0,0 +1,27 @@
...
+    print('%s_COUNT=%d' % (keyword, length))
+    for idx in range(0, length):
+        # store pipeline
+        print('%s_LST[%d]="%s"' % (keyword, idx, format_pipeline(pipeline_lst[idx])))
+        # store pipeline to each list
+        print('unset %s_%d' % (keyword, idx))
+        print('declare -Ag %s_%d' % (keyword, idx))
+        for key, value in pipeline_lst[idx].items():
+            print('%s_%d["%s"]="%s"' % (keyword, idx, key, value))
+    return 0
```
marc-hb, line 27: missing newline, check your editor's settings.

marc-hb (changes requested):
Much cleaner, thanks. Nice git commit breakdown. Did you really test every commit? :-)

marc-hb:
SOFCI TEST

marc-hb:
About the massive https://sof-ci.01.org/softestpr/PR430/build305/devicetest/ failures see https://github.com/thesofproject/sof-test/issues/458

test-case/check-capture.sh
```diff
@@ -75,6 +75,10 @@ do
         type=$(func_pipeline_parse_value $idx type)
         snd=$(func_pipeline_parse_value $idx snd)
 
+        # Filtering feature is not enabled for pipelines dumped from proc
+        # This workaround is need for this case to run on legacy HDA platform.
```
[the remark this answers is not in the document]
aiChaoSONG: We cannot remove the filter here, because an empty filter means all pipelines, and filter "-f type:any" means all pipelines except WOV and Echo Reference, we cannot run arecord on these pipelines directly. I will change the filter to "-f type:any",  remove the puzzling comments, this also make this a FEATURE.

aiChaoSONG:
@marc-hb Yes, I have test the commits on one HDA platform and a SOF platform, both work well. All reviews are addressed

case-lib/pipeline.sh
```diff
@@ -2,6 +2,15 @@
 
 func_pipeline_export()
 {
+    # For legacy HDA platform, there is no topology, we have to exporting pipeline
+    # information from proc file system.
+    is_sof_used || {
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        source <(sof-dump-status.py -e) ||
+            die "Failed to export pipeline parameters from proc file system"
+        return 0
```
marc-hb, line 11: I would rather drop `return 0`. In the very unlikely case `source` fails then it's better not to return 0.

I just realize one drawback with this `source <(   )` solution: it can probably never fail. I mean even if `sof-dump-status.py` is not found or fails, `source <(   )` will still succeed. I hope most `sof-dump-status.py` error message / stack traces get at least printed to `stderr` so they will "escape" `< ( )` and be visible in the CI logs.

If you are not too worried about `sof-dump-status.py` failures then you can keep `source <( )`. If you are worried about some potential `sof-dump-status.py` failures then you should source an intermediate `mktemp` file instead of using `< ( )`

test-case/test-speaker.sh
```diff
@@ -41,6 +41,9 @@ do
     dev=$(func_pipeline_parse_value $idx dev)
     snd=$(func_pipeline_parse_value $idx snd)
 
+    type=$(func_pipeline_parse_value $idx type)
+    [ "$type" = "playback" ] || continue
```
marc-hb, line 45: I just realized `type` is pretty bad variable name because it's also a command name. Well, it seems to have worked so far...

marc-hb (commented):
I find the code much cleaner now, thanks @aiChaoSONG ! @xiulipan / @plbossart do you mind reviewing again?
This review could also use some topology experts, @ranj063 , @fredoh9 , others?

> @marc-hb Yes, I have test the commits on one HDA platform and a SOF platform, both work well.

I meant test each of the 6 commits individually. If you've done that then I'm impressed!

aiChaoSONG:
@marc-hb Yes, even the renaming commit.

test-case/test-speaker.sh
aiChaoSONG: we also have other scripts use "type" as variable name, we can do clean up later, I am not going to do any change in this PR.

case-lib/pipeline.sh
aiChaoSONG: I think we should care the sof-dump-status.py failures, it can help to detect system/sof-test issues. A intermediate temp file is used, and error handler is added.
The `return` is required, because we don't want the below code(tplg pipeline exporting) to be executed.

marc-hb: > The return is required, because we don't want the below code(tplg pipeline exporting) to be executed.

Ooops, sorry for the noise.

> A intermediate temp file is used

OK, then please use a `mktemp` (and delete it). I realize we don't support concurrency for now but we may want to in the future and this code is in a very generic and common function.

case-lib/pipeline.sh
```diff
@@ -2,6 +2,17 @@
 
 func_pipeline_export()
 {
+    # For legacy HDA platform, there is no topology, we have to exporting pipeline
+    # information from proc file system.
+    is_sof_used || {
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        tmp_pipeline_params=/tmp/pipeline_params.sh
```
xiulipan, line 9: Agreed with @marc-hb to use a mktemp. You can also add prefix if you'd like. Try to play with help.
But I think mktmp will not create same file, so you can leave the delete to the /tmp/ folder clean up
```suggestion
        tmp_pipeline_params=$(mktemp)
```

test-case/check-capture.sh
```diff
@@ -61,7 +61,7 @@ file_prefix=${OPT_VALUE_lst['f']}
 
 func_lib_setup_kernel_last_line
 func_lib_check_sudo
-func_pipeline_export $tplg "type:capture & ${OPT_VALUE_lst['S']}"
+func_pipeline_export $tplg "type:any & ${OPT_VALUE_lst['S']}"
```
xiulipan, line 64: Why not relay on the `func_pipeline_export` to do the type filter? I think make check to each end user is not a good idea. I think this would be a degenerate for our function.

xiulipan (changes requested):
The biggest problem to me is why to more the filter feature outside? Try to merge the feature in the function and keep minor change is always good idea.

test-case/check-capture.sh
aiChaoSONG: The pipeline filtering feature is provided by sof-tplgreader.py, and the filtering feature is not implemented for sof-dump-status.py.  If we need a common filtering feature shared between both, it requires a lot of code change, because the filter code in the sof-tplgreader.py is in a class, and requires `self`, which need some efforts to refactor.

This maybe done later

xiulipan: I checked the output from, I think add a simple filter for playback/capture will be very easy. (a simple grep will make it work)
```
cat /proc/asound/pcm
00-00: Port5 (*) :  : playback 1 : capture 1
00-01: DMIC (*) :  : capture 1
00-02: DMIC16kHz (*) :  : capture 1
00-04: Media Playback 4 (*) :  : playback 1
00-05: HDMI1 (*) :  : playback 1
00-06: HDMI2 (*) :  : playback 1
00-07: HDMI3 (*) :  : playback 1
```

marc-hb: Assuming the format of `/proc/asound/pcm` is stable and guaranteed then a simple `grep` in `func_pipeline_export()` seems like a cheap and good idea to me.

> PS: to me a patch that influence so many files is also a big risk. 

There is 1 new file which is just extracted from an existing one and only 4 other modified files. One of the modified file is just getting a one-line function.

> Changing API is a dangerous thing and need more test. So I would prefer to keep old ABI 

What specific API and ABI are changing here?

> and add simple implement for the legacy HDA.

I don't get this sorry, can you elaborate?

tools/sof-dump-status.py
```diff
@@ -413,6 +410,7 @@ def dump_dapm(dapm, filter = "all"):
...
```
plbossart: you change this two patches later, can we squash the changes in one location to avoid this tick-tock effect?

tools/sof-dump-status.py
```diff
@@ -435,7 +435,7 @@ def dump_dapm(dapm, filter = "all"):
             print(run_status['status'])
         exit(0)
 
-    if ret_args['export'] is True:
+    if ret_args['export'] is not None:
```
plbossart, line 438: same here, move this two patches earlier when this was added first.

tools/sof-dump-status.py
```diff
@@ -448,7 +448,15 @@ def dump_dapm(dapm, filter = "all"):
...
-        export_pipeline(pipeline_lst)
+        try:
+            pipeline_type = list(filter(lambda x: x.strip().startswith('type'), ret_args['export'].split(' ')))[0].split(':')[1]
+        except:
+            print('Invalid filter string')
+            exit(1)
+        if pipeline_type in ['playback', 'capture']:
+            export_pipeline(list(filter(lambda x: x['type'] == pipeline_type, pipeline_lst)))
+        else:
+            export_pipeline(pipeline_lst)
```
plbossart, line 459: a comment to explain what you are doing where wouldn't hurt.

Put yourself in the shoes of a reviewer who's not a Python expert and has no time to reverse engineer your work, thanks!

case-lib/pipeline.sh
```diff
@@ -2,6 +2,19 @@
 
 func_pipeline_export()
 {
+    # For legacy HDA platform, there is no topology, we have to exporting pipeline
+    # information from proc file system.
+    is_sof_used || {
+        filter_str="$@"["$#"]
```
plbossart, line 8: In English this almost looks like what people type when cursing....
Again explaining what this is supposed to do wouldn't hurt.

case-lib/pipeline.sh
```diff
@@ -2,6 +2,19 @@
...
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        tmp_pipeline_params=$(mktemp /tmp/pipeline-params.XXXXXXXX)
+        sof-dump-status.py -e "$filter_str" > "$tmp_pipeline_params" ||
+            die "Failed to export pipeline parameters from proc file system"
+        source "$tmp_pipeline_params" || return 1
```
plbossart, line 13: is there really no other way to do this? Why create a temp file that needs to be sourced?

plbossart (changes requested):
this can be improved quite a bit.

plbossart:
Another question i have is how would you deal with the TPLG env variable for legacy HDaudio? Is this even relevant?

case-lib/pipeline.sh
```diff
@@ -2,6 +2,19 @@
 
 func_pipeline_export()
 {
+    # For legacy HDA platform, there is no topology, we have to exporting pipeline
+    # information from proc file system.
+    is_sof_used || {
```
ranj063, line 7: will this test work where we have dmi info available and the card name doesnt contain "sof"? like the chromebooks?

case-lib/pipeline.sh
```diff
@@ -2,6 +2,19 @@
 
 func_pipeline_export()
 {
+    # For legacy HDA platform, there is no topology, we have to exporting pipeline
```
ranj063, line 5: exporting -> export

case-lib/pipeline.sh
```diff
@@ -2,6 +2,19 @@
...
+    # information from proc file system.
+    is_sof_used || {
+        filter_str="$@"["$#"]
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        tmp_pipeline_params=$(mktemp /tmp/pipeline-params.XXXXXXXX)
+        sof-dump-status.py -e "$filter_str" > "$tmp_pipeline_params" ||
+            die "Failed to export pipeline parameters from proc file system"
+        source "$tmp_pipeline_params" || return 1
+        rm "$tmp_pipeline_params" || true
```
marc-hb, line 14: `rm -f "$tmp_pipeline_params"` never fails if you ever need that but why would `rm "$tmp_pipeline_params"` fail anyway? If this simple `rm` fails then something must have gone terribly wrong and we probably want to let `set -e` stop there.

marc-hb (commented):
Thanks @plbossart  for the review, trying to partly answer some of your comments below (and add some more)

> Another question i have is how would you deal with the TPLG env variable for legacy HDaudio? Is this even relevant?

Just in case it helps: note the missing TPLG check is right _after_ the `is_sof_used ||` block, meaning there is no TPLG check unless `is_sof_used`.

case-lib/pipeline.sh
ranj063: For example : on my helios, card name is GoogleHeliosRev1. There's no sof prefix there.

ranj063: On the other hand, the procfs info for PCMs is valid for all cases. Why not just use that instead of topology?

case-lib/pipeline.sh
ranj063: > In English this almost looks like what people type when cursing....

I couldnt help laughing at this

marc-hb: > In English this almost looks like what people type when cursing....

You should be cursing when quoting, this is by design:

http://rg1-teaching.mpi-inf.mpg.de/unixffb-ss98/quoting-guide.html

case-lib/pipeline.sh
marc-hb: > On the other hand, the procfs info for PCMs is valid for all cases. Why not just use that instead of topology?

I guess part of the answer is in this code comment:

```
for pcm in card_info['pcm']:
                # There are limited pipeline parameters in the proc file system,
                # add some default parameters to make use of sof-test for legacy HDA test
                pcm['fmt'] = 'S16_LE'
                pcm['fmts'] = 'S16_LE S24_LE S32_LE'
                ...
```

tools/sof-dump-status.py
```diff
@@ -437,6 +435,30 @@ def dump_dapm(dapm, filter = "all"):
...
+                # There are limited pipeline parameters in the proc file system,
+                # add some default parameters to make use of sof-test for legacy HDA test
+                pcm['fmt'] = 'S16_LE'
+                pcm['fmts'] = 'S16_LE S24_LE S32_LE'
+                pcm['rate'] = '48000'
+                pcm['channel'] = '2'
+                pcm['dev'] = 'hw:{},{}'.format(card_id, pcm['id'])
+            pipeline_lst.extend(card_info['pcm'])
+        try:
+            pipeline_type = list(filter(lambda x: x.strip().startswith('type'), ret_args['export'].split(' ')))[0].split(':')[1]
```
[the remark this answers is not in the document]
ranj063: its actually direction more than type or filter

tools/sof-dump-status.py
```diff
@@ -448,7 +448,15 @@ def dump_dapm(dapm, filter = "all"):
                 pcm['channel'] = '2'
                 pcm['dev'] = 'hw:{},{}'.format(card_id, pcm['id'])
             pipeline_lst.extend(card_info['pcm'])
-        export_pipeline(pipeline_lst)
+        try:
+            pipeline_type = list(filter(lambda x: x.strip().startswith('type'), ret_args['export'].split(' ')))[0].split(':')[1]
+        except:
+            print('Invalid filter string')
```
ranj063, line 454: why do we need this filter exactly? What else can the type be?

case-lib/pipeline.sh
aiChaoSONG: @ranj063 I guess you are talking about Chromebook under ChromeOS,  because the content of /proc/asound/cards on my Helios looks like this:
```
ubuntu@helios:~$ cat /proc/asound/cards
 0 [sofcmlrt1011rt5]: sof-cml_rt1011_ - sof-cml_rt1011_rt5682
                      Google-Helios-rev2
```
Currently, sof-test cannot run on ChromeOS, because most infrastructure level python scripts cannot run on that OS.

ranj063: > @ranj063 I guess you are talking about Chromebook under ChromeOS, because the content of /proc/asound/cards on my Helios looks like this:
> 
> ```
> ubuntu@helios:~$ cat /proc/asound/cards
>  0 [sofcmlrt1011rt5]: sof-cml_rt1011_ - sof-cml_rt1011_rt5682
>                       Google-Helios-rev2
> ```
> 
> Currently, sof-test cannot run on ChromeOS, because most infrastructure level python scripts cannot run on that OS.

@aiChaoSONG no Im talking about linux with nocodec mode

ranj063: I dont know why but yesterday my card name was GoogleHeliosrev2 but its back to sof-nocodec now

tools/common.py
```diff
@@ -7,7 +7,7 @@ def func_dump_pipeline(pipeline, noKey=False):
             output += key + "=" + str(value) + ";"
     return output.strip()
 
-def func_export_pipeline(pipeline_lst):
+def export_pipeline(pipeline_lst):
```
ranj063, line 10: export_pipeline() is still a vague name. I was scratching my head as to where we're exporting the pipeline. What would make this clear is something like export_pipeline_env_vars() or something like that. And a comment above the function as to what it does, something along the lines of "export the enviromnent variables for pipeline info"

case-lib/pipeline.sh
marc-hb: > Currently, sof-test cannot run on ChromeOS, because most infrastructure level python scripts cannot run on that OS.

In theory the `dev_install` command can install python
https://chromium.googlesource.com/chromiumos/platform2/+/refs/heads/master/dev-install/

It does NOT mean everything would work but it would be a good start.

tools/common.py
aiChaoSONG: As a matter of fact, this function doesn't export any thing to env, it just generate some shell code which can be sourced later. I will add some comment here rather than changing the name, this will also make it clearer

tools/sof-dump-status.py
aiChaoSONG: some test case may only want to test on playback pipelines, or capture pipelines. so the type here may be one of the following: playback, capture, any.  As we want to make it compatible with the topology pipeline, so the filter string should be written like this: 'type:playback'

case-lib/pipeline.sh
plbossart: > @ranj063 I guess you are talking about Chromebook under ChromeOS, because the content of /proc/asound/cards on my Helios looks like this:
> 
> ```
> ubuntu@helios:~$ cat /proc/asound/cards
>  0 [sofcmlrt1011rt5]: sof-cml_rt1011_ - sof-cml_rt1011_rt5682
>                       Google-Helios-rev2
> ```
> 
> Currently, sof-test cannot run on ChromeOS, because most infrastructure level python scripts cannot run on that OS.

that's not a good-enough reason to screw-up the card name parsing. 

I think you guys are confusing things, the syntax is

cardIndex [mixer_name]: driver_name - card_name
          long_card_name

@ranj063 can you provide what you have on your device w/ Chrome?

tools/sof-dump-status.py
```diff
@@ -413,6 +410,7 @@ def dump_dapm(dapm, filter = "all"):
     parser.add_argument('-P', '--fwpath', action='store_true', help='get firmware path according to DMI info')
     parser.add_argument('-S', '--dsp_status', type=int, help='get current dsp power status, should specify sof card number')
     parser.add_argument('-d', '--dapm', choices=['all', 'on', 'off', 'standby'], help='get current dapm status, this option need root permission to access debugfs')
+    parser.add_argument('-e', '--export', type=str, help='export pipeline parameters of specified type from proc file system')
```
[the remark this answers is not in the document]
marc-hb: I meant add an example in the `--help` message. Looking at the code source shouldn't be needed to understand what sort of argument `--export` expects.

At this moment the only indication in the `--help` of what argument `--export` expects is: "`specified type from the proc file system`", I really cannot relate this to anything like `type:playback & pga:any & ~asrc:any | id:5`

tools/sof-dump-status.py
```diff
@@ -437,6 +437,35 @@ def dump_dapm(dapm, filter = "all"):
...
+                pcm['rate'] = '48000'
+                pcm['channel'] = '2'
+                pcm['dev'] = 'hw:{},{}'.format(card_id, pcm['id'])
+            pipeline_lst.extend(card_info['pcm'])
+        # The filter string may be very complex due to the compatibility with topology pipeline
+        # filter string, but we only implement a simple type filter here.
+        try:
+            # extract requested pipeline type from filter string, and ignore others.
+            # eg, the requested_type from filter string 'type:capture & pga:any | id:3' is 'capture'
+            requested_type = [f.split(':')[1] for f in ret_args['export'].split(' ') if f.strip().startswith('type')][0]
```
marc-hb, line 458: I still think one intermediate variable would make this much easier to read but the comment makes it very clear, thank you.

```python
filter_elements = ret_args['export'].split(' ')
first_requested_type = [f.split(':')[1] for f in filter_elements
    if f.strip().startswith('type')][0]
```

case-lib/pipeline.sh
marc-hb: I still don't understand the purpose of `filter_str="$@"["$#"]`. shellcheck doesn't understand either BTW. 

I just tested  `"$@"["$#"]` with `f() { printf '%s__' "$@"["$#"] ; }` to be sure and this is what it does, it turns:
`"a b"   "c d"   "e f"` into: `"a b"   "c d"   "e f[3]"`.

Is this really what you're trying to do here and why?

Why isn't the following enough? It would match the documentation you just added:

```shell
func_pipeline_export()
{
   local tplg="$1  filter="$2"
  
```

case-lib/pipeline.sh
```diff
@@ -1,7 +1,27 @@
 #!/bin/bash
 
+# This function will evaluate pipeline parameter related shell code generated by
```
marc-hb, line 3: Very useful comment considering this function should be called `something_IMPORT()` (import from Python script). Don't change the name in this PR.

case-lib/pipeline.sh
```diff
@@ -1,7 +1,27 @@
...
+# - sof-tplgreader.py (for SOF, pipeline parameters dumped from topology)
+# - sof-dump-status.py (for legacy HDA, pipeline paramters dumped from proc)
+# Args: $1: SOF topology path
+#       $2: Pipeline filter string
```
marc-hb, line 7: So @plbossart  was right, you need a topology path even when `is_sof_used_` is false? Which one?

I thought this function took different arguments depending on `is_sof_used`

marc-hb (changes requested):

ranj063:
@plbossart @aiChaoSONG @marc-hb this PR is in the right direction. I think we should decouple sof-test from the topology file even when testing with the SOF driver and use the procfs info for the card/pcm info and even for the topology graph using the asoc dapm info. This will also make sure that we dont have assumptions about the default card number. What do you think?

case-lib/pipeline.sh
ranj063: @plbossart this is what we see on chrome: 
localhost ~ # cat /proc/asound/cards
 0 [sofrt5682      ]: sof-rt5682 - sof-rt5682
                      Google-Delbin-1.0

ranj063: but for some reason when I was testing yesterday my card name was as below on my helios. I am unable to reproduce this. So I do not know where this was coming from. 
root@ranjani-Helios:~# cat /proc/asound/cards
 0 [GoogleHeliosrev]:  -
                      Google-Helios-rev2

tools/common.py
```diff
@@ -0,0 +1,30 @@
+def func_dump_pipeline(pipeline, noKey=False):
```
ranj063: an example of what is dumped here would be nice to see in the comment for the function. Imnot going to hold that against this PR though. Can be a separate one

tools/sof-dump-status.py
ranj063: thsi patch needs to be squashed with the previous one

case-lib/pipeline.sh
marc-hb: OK, I think I finally guessed what you're trying to do here. You're probably trying to define `filter` as the last element of the array, in other words you're trying to do this:
```shell
last_arg()
{
    array_from_args=("$@");
    echo "${array_from_args[${#array_from_args[@]}-1]}"
}
```
If I guessed correctly and taking the last function argument is really what you're trying then:
- why does the function header pretend you're using `$2`
- why silently ignore all the arguments between the first and the last?

marc-hb (changes requested):
> I think we should decouple sof-test from the topology file even when testing with the SOF driver and use the procfs

Interesting but please file a new github issue to discuss such a major sof-test design change otherwise this PR will become totally unreadable (it already  has a very large number of comments)


> that's not a good-enough reason to screw-up the card name parsing.

See also similar `Auxiliary devices in SOF break SOFCARD` #466 (thx @ranj063  for the link)

case-lib/pipeline.sh
ranj063: or "${@: -1}"? if we can assume the filter string will always be the last argument

marc-hb: > or "${@: -1}"?

No, that defaults to '1' when "$@" is empty (zero argument)

plbossart:
> @plbossart @aiChaoSONG @marc-hb this PR is in the right direction. I think we should decouple sof-test from the topology file even when testing with the SOF driver and use the procfs info for the card/pcm info and even for the topology graph using the asoc dapm info. This will also make sure that we dont have assumptions about the default card number. What do you think?

Agree that if we can use the /proc info then we have all the information needed for tests with PCM devices.
For probes it's actually the only solution since we expose a compressed interface that will *never* be shown in the topology but will be shown in the card properties (if it's not show it's a bug). The probes are like 'blue wires', they are not set in stone and not shown in the topology.

Having the topology would only help to show the graph as an information but would not have an functional input into the tests.

marc-hb:
> Agree that if we can use the /proc info then we have all the information needed for tests with PCM devices.

This PR implies the exact opposite.

Discussion transferred to new issue #471 `Remove sof-test dependency on topology file`

case-lib/pipeline.sh
aiChaoSONG: @marc-hb On legacy HDA, if you manually run test case, no TPLG is required. But CI will give a fake topology, as you can see from any report, this is how CI run a test case: `TPLG=sof-hda-generic-4ch-kwd.tplg ~/sof-test/test-case/verify-tplg-binary.sh`. Some compromise with sof-framework here. Here is how func_pipeline_export called in test case: ` func_pipeline_export $tplg "type:capture" `. so the filter string may be $1 (tplg is empty if manual run) or $2 (CI always give tplg), that's the reason for `"$@"["$#"]`

marc-hb: > func_pipeline_export $tplg "type:capture". so the filter string may be $1 (tplg is empty if manual run) or $2 (CI always give tplg

- This is _really_ missing from the function header!
- That's yet another demonstration that almost every parameter expansion must be quoted. Maybe it would be a too big change but with `func_pipeline_export "$tplg" "type:capture"` then `$1` would be `""` when `$tplg` is missing, the function would get a constant number of arguments and the filter would always be in `$2`. 

> that's the reason for "$@"["$#"]

... which doesn't seem to do what I understand you want. How was this tested!?

aiChaoSONG: I have run the test case on a legacy HDA platform, and an SOF platform with TPLG, both passed!

case-lib/pipeline.sh
xiulipan: @aiChaoSONG An inline comment would solve the issue.
I think what you want is 
`# to get the last argument only.`

case-lib/pipeline.sh
marc-hb: Maybe a broken `filter_str` does not matter?

aiChaoSONG: sof-tplgreader.py and sof-dump-status.py will help to handle the filter_str, so we don't need a handler here.

marc-hb: I meant: maybe a broken `filter_str` did not matter in your testing? So `filter_str` was never tested?

We digressing again and spent way too many comments and time on this. Trying to focus back on the code issues:

- `"$@"["$#"]` does not do anything anyone may want, it just doesn't work for any known definition of "work".
- `"$@"["$#"]` was not tested. Every new lines of code must be tested at least once
- The (failed) intent doesn't seem to match the function header which says nothing either about all the other, discarded arguments in the middle.
- Getting the last function argument is a bit complicated in shell and best avoided. 
- Getting the last argument is probably not a good idea in general, is it really needed here? Why not just pass an empty string first argument instead of none?

Before the next force-push what is your plan to address all these?

aiChaoSONG: @marc-hb I am going to drop the `"$@"["$#"]` thing, and add quote in every test case to `$tplg`, thus make sure the filter_str is always $2. So we don't need to think about how to get the last function argument in shell.

marc-hb: > add quote in every test case to $tplg

Sounds good. You probably want to merge that first and separately to avoid a "Big Bang Integration" (the opposite of _Continuous_ Integration)

marc-hb: I find that "fake topologies" sound bad too. On the other hand, an empty, **_dummy_ HDA topology file** would be OK if needed to keep sof-test APIs consistent or for some transition period. Such a dummy HDA topology should make any test FAIL FAST if the test tries using it by mistake.

case-lib/pipeline.sh
```diff
@@ -1,7 +1,26 @@
...
+        filter_str="$2"
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        tmp_pipeline_params=$(mktemp /tmp/pipeline-params.XXXXXXXX)
+        sof-dump-status.py -e "$filter_str" > "$tmp_pipeline_params" ||
+            die "Failed to export pipeline parameters from proc file system"
+        source "$tmp_pipeline_params" || return 1
+        rm "$tmp_pipeline_params"
+        return 0
+    }
+
```
marc-hb, line 23: I think the `if [ $# -lt 1 ]; then...` check should be rephrased and kept at the start of the function.

tools/sof-dump-status.py
```diff
@@ -413,6 +410,11 @@ def dump_dapm(dapm, filter = "all"):
     parser.add_argument('-P', '--fwpath', action='store_true', help='get firmware path according to DMI info')
     parser.add_argument('-S', '--dsp_status', type=int, help='get current dsp power status, should specify sof card number')
     parser.add_argument('-d', '--dapm', choices=['all', 'on', 'off', 'standby'], help='get current dapm status, this option need root permission to access debugfs')
+    # The filter string here is compatible with the filter string for pipeline from topilogy,
```
marc-hb, line 413: Nit: typo

marc-hb (commented):
This looks OK except it now depends on (part of) #475 to be merged first if I understood correctly

case-lib/pipeline.sh
aiChaoSONG: @plbossart @marc-hb Agreed. I think this requires some change in the CI framework, I will check.

aiChaoSONG:
SOFCI TEST

aiChaoSONG:
@xiulipan Ping!

case-lib/pipeline.sh
```diff
@@ -1,11 +1,32 @@
...
+    # For legacy HDA platform, there is no topology, we have to export pipeline
+    # parameters from proc file system.
+    is_sof_used || {
+        filter_str="$2"
+        dlogi "No SOF sound card found, exporting pipeline parameters from proc file system"
+        tmp_pipeline_params=$(mktemp /tmp/pipeline-params.XXXXXXXX)
+        sof-dump-status.py -e "$filter_str" > "$tmp_pipeline_params" ||
+            die "Failed to export pipeline parameters from proc file system"
+        source "$tmp_pipeline_params" || return 1
```
marc-hb, line 25: Add
```shell
# shellcheck disable=SC1090
```
`pipeline.sh` is one of our few files that are already shelcheck-clean, keep it that way.

marc-hb (changes requested):
One minor change to keep `pipeline.sh` shellcheck clean otherwise the code style LGTM

marc-hb:
The `aplay: pcm_write:2061: write error: Input/output error` in https://sof-ci.01.org/softestpr/PR430/build355/devicetest/ is in only one test (`check-playback-3rounds`) on only one system (`sh-bdw-wsb-rt286-01`) so I can't see how it could be related to this.

xiulipan (approved):
LGTM, CI error is known issue on legacy platforms

Issue #431, opened 2020-10-13
[CHT] suspend-resume fails due to unrelated USB error


plbossart:
Need to filter this out to pass tests.

````
===========================>>
[ 7702.164446] hub 2-2:1.0: hub_ext_port_status failed (err = -71)
[ 7702.502167] usb 2-2: Failed to suspend device, error -19
<<===========================
````

**Comment by marc-hb** [comment]
Thanks, can you be more specific than "CHT"?

Also: is this a recent kernel regression?

**Comment by mengdonglin** [comment]
@RuiqingX can you please fix this issue?

**Comment by marc-hb** [comment]
PR #432 already submitted

**Comment by keqiaozhang** [comment]
This issue happened again on CML_HEL_RT5682 in daily test ID1828.

**Comment by aiChaoSONG** [comment]
I think this is related to USB hub or card devices, because we never see this in PRC.

**Comment by marc-hb** [comment]
> This issue happened again on CML_HEL_RT5682 in case check-suspend-resume-50.sh, in CI daily test ID1828 on 2021-01-26:

What happened there is actually a cascade of 3 USB errors.

**Before** this USB error, there was another, previous USB failure two test earlier during the `check-kmod-load-unload.sh` audio test. I think it's already reported in another bug:
```
kernel: [ 7296.822125] usb 1-2.4: Device not responding to setup address.
kernel: [ 7297.030744] usb 1-2.4: device not accepting address 7, error -71
kernel: [ 7297.031055] usb 1-2-port4: attempt power cycle
kernel: [ 7297.658487] rt5682 i2c-10EC5682:00: Using default DAI clk names: rt5682-dai-wclk, rt5682-dai-bclk
kernel: [ 7297.658831] usb 1-2.4: new full-speed USB device number 8 using xhci_hcd
kernel: [ 7297.671781] usb 1-2.4: device descriptor read/8, error -32
kernel: [ 7297.787267] usb 1-2.4: device descriptor read/all, error -32
kernel: [ 7297.877766] usb 1-2.4: new full-speed USB device number 9 using xhci_hcd
kernel: [ 7297.877975] usb 1-2.4: Device not responding to setup address.
kernel: [ 7298.101181] usb 1-2.4: device descriptor read/8, error -32
kernel: [ 7298.220439] usb 1-2.4: device descriptor read/8, error -32
kernel: [ 7298.328195] usb 1-2-port4: unable to enumerate USB device
```

[...]

PR #432, opened 2020-10-13
sof-kernel-log-check: ignore USB hub issues

plbossart:
This is needed on CHT devices

BugLink: https://github.com/thesofproject/sof-test/issues/431

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -65,6 +65,10 @@ ignore_str="$ignore_str"'|: authentication with ..:..:..:..:..:.. timed out'
 # BugLink: https://github.com/thesofproject/sof-test/issues/374
 ignore_str="$ignore_str"'|i915 0000:00:02\.0: \[drm\] ERROR TC cold unblock failed'
 ignore_str="$ignore_str"'|i915 0000:00:02\.0: \[drm\] ERROR TC cold block failed'
+# CHT devices with USB hub, issues reported by sof-test
+# BugLink: https://github.com/thesofproject/sof-test/issues/431
+ignore_str="$ignore_str"'|hub 2-.: .'
+ignore_str="$ignore_str"'|usb 2-.: .'
 
 # Test cases on some platforms fail because the boot retry message:
 #


marc-hb (dismissed):

marc-hb:
Considering this seems like a generic USB error it would probably be better to guard this with `case $platform`.

plbossart:
> Considering this seems like a generic USB error it would probably be better to guard this with `case $platform`.

Why? there's nothing platform specific here? It's probably just the hub I used that's not handled.

marc-hb:
I thought that hub was embedded in the system, was it external? If yes can't you use another hub or ignore this error just locally? A bit concerned this file could grow into a database of defective peripherals...

Also not a fan of ignoring errors with random external hubs and hiding at the same time any errors with embedded, possibly Intel hubs.

plbossart:
@marc-hb I don't get why you are making things more complicated than they need to be. There's an issue on a device I use, the hub is actually in my keyboard, I don't have spare hubs and it is what is is. That device alone flagged major issues with sof-test, between a broken ACPI handling and broken card name parsing so you should thank me for going beyond the reasonable in hardening tests. End of discussion, nothing to see here.

marc-hb:
There's really nothing complicated in being concerned that this file becomes a dump of everyone's broken devices and trying to mitigate that a bit.

The quality level I'm trying to raise sof-test to is not 1/10th of the nitpicking found in a typical kernel code review. Of course that's because few people care about sof-test but that's a not a very good reason IMHO because role-modelling etc.

I'm "thanking you" by reviewing (and immediately approving!!) this PR and reviewing the majority of PRs in sof-test which is rare enough. I really appreciate that your involvement with sof-test is 10 times the SOF developer average's but first I think mine is still higher than yours and second the important topic here is not our respective levels of involvement in sof-test but the very low average level and how developers test their code. Anyway no one should be thanking anyone else for contributing to sof-test specifically because sof-test is not anyone's project, it's just one part of the larger SOF project which is much more your project than mine. So don't forget to thank _me_ :-)

aiChaoSONG:
The code change is good, but may add a backslash `\` to the dot `.` at the end?

aiChaoSONG (dismissed):

plbossart:
> The code change is good, but may add a backslash `\` to the dot `.` at the end?

no, this is a regexp, there's no actual point

plbossart:
rebased and added a wildcard to discard all usb2/hub2 issues.

aiChaoSONG (approved):

Issue #420, opened 2020-10-09
check_playback tests assume device0 is the SOF card - fails on BYT/CHT


plbossart:
On CHT device, the SOF card is not the first one:

````
root@Zotac:~/sof-test/test-case# aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: Audio [Intel HDMI/DP LPE Audio], device 0: HdmiLpeAudio [Intel HDMI/DP LPE Audi]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: Audio [Intel HDMI/DP LPE Audio], device 1: HdmiLpeAudio [Intel HDMI/DP LPE Audi]
  Subdevices: 0/1
  Subdevice #0: subdevice #0
card 0: Audio [Intel HDMI/DP LPE Audio], device 2: HdmiLpeAudio [Intel HDMI/DP LPE Audi]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: bytcrrt5640 [bytcr-rt5640], device 0: 3 []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: bytcrrt5640 [bytcr-rt5640], device 1: Deep-Buffer Audio (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
````

as a result, we test the wrong card:

````
 starting test_playback-d100l1r1 
+ test_playback-d100l1r1
+ /root/sof-test/test-case/check-playback.sh -d 3 -l 1 -r 1
2020-10-09 20:11:27 UTC Sub-Test: [REMOTE_INFO] Starting /usr/local/bin/sof-logger -t -l /etc/sof/sof-cht.ldc -o /root/sof-test/logs/check-playback/2020-10-09-15:11:27-24181/slogger.txt
error: Unable to open version file /sys/kernel/debug/sof/fw_version
2020-10-09 20:11:27 UTC Sub-Test: [REMOTE_INFO] no source file, use /dev/zero as dummy playback source
2020-10-09 20:11:27 UTC Sub-Test: [REMOTE_INFO] /root/sof-test/test-case/check-playback.sh will use topology /lib/firmware/intel/sof-tplg/sof-cht-rt5640.tplg to run the test case
2020-10-09 20:11:27 UTC Sub-Test: [REMOTE_INFO] Pipeline list to ignore is specified, will ignore 'pcm=HDA Digital,Media Playback' in test case
[...]

PR #434, opened 2020-10-14
lib.sh: fix SOF card ID acquisition logic

aiChaoSONG:
The SOF card is renamed in linux kernel,
our regular expression for old nameing style
only works under SoundWire and HDAudio. But
SOF card information in /proc/asound/cards
still contains 'sof-[a-z]' pattern.

This patch updates the regular expression for
acquiring SOF card ID.

Fixes: #420

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -35,7 +35,7 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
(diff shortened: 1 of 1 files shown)


xiulipan:
@aiChaoSONG Will this work on the byt/cht boards?

aiChaoSONG:
@xiulipan Yes, I have used the information from Pierre‘ BYT devices， you can see there is sof-[a-z] strings. so it will work on old SOF platforms.

xiulipan (approved):

PR #437, opened 2020-10-14
[WIP][RFC] Test-case: Set PGAs to unity gain in script check-alsabat.sh

singalsu:
This patch adds find of volume controls for used playback and capture
devices and sets them to unity gain (0 dB). The alsabat test fails
if e.g. capture PGA is set to their maximum value that can be up
to +30 dB. The test sine wave distorts and causes test fail.

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -42,6 +42,12 @@ OPT_HAS_ARG['n']=1             OPT_VAL['n']=240000
 OPT_NAME['s']='sof-logger'      OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
+OPT_NAME['t']='tplg'            OPT_DESC['f']='tplg file, default value is env TPLG: $''TPLG'
+OPT_HAS_ARG['t']=1             OPT_VAL['f']="$TPLG"
+
+TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+TPLGREADER="$TESTDIR"/tools/sof-tplgreader.py
+
 func_opt_parse_option "$@"
 
 pcm_p=${OPT_VAL['p']}
@@ -50,6 +56,7 @@ rate=${OPT_VAL['r']}
 channel_c=${OPT_VAL['C']}
 frequency=${OPT_VAL['f']}
 frames=${OPT_VAL['n']}
+tplg=${OPT_VAL['t']}
 
 if [ "$pcm_p" = "" ]||[ "$pcm_c" = "" ];
 then
@@ -72,10 +79,82 @@ function __upload_wav_file
     done
 }
 
+set_pga_to_unity_gain()
+{
+    tmp=$(amixer controls | grep "$1" | grep Volume)
+    search="name="
+    cname=${tmp#*$search}
+
+    # Get volume min and step to compute value for cset
+    # for 0 dB gain. The amixer line looks like
+    # "| dBscale-min=-50.00dB,step=1.00dB,mute=1"
+    scale=$(amixer cget name="$cname" | grep "dBscale" || true)
+    search="dBscale-min="
+    tmp=${scale#*$search}
+    min_db="${tmp%%dB*}"
+    search="step="
+    tmp=${scale#*$search}
+    step_db="${tmp%%dB*}"
+
+    # Get multiplied by 100 values by removing decimal dot
+    min_x100="${min_db//.}"
+    step_x100="${step_db//.}"
+    val=$(printf %d "$(((-min_x100) / step_x100))")
+
+    # Apply the computed value for requested gain
+    amixer cset name="$cname" "$val"
+}
+
+set_pgas_list_to_unity_gain()
+{
+    for pga in "$@"; do
+	dlogi "Set $pga"
+	set_pga_to_unity_gain "$pga"
+    done
+}
+
+get_snd_base()
+{
+    # Converts e.g. string hw:1,0 to /dev/snd/pcmC1D0
+    # the p or c for playback or capture is appended
+    # in the calling function.
+    tmp=${1#*"hw:"}
+    ncard="${tmp%%,*}"
+    ndevice="${tmp#*,}"
+    echo "/dev/snd/pcmC${ncard}D${ndevice}"
+}
+
+get_play_snd()
+{
+    tmp=$(get_snd_base "$1")
+    echo "${tmp}"p
+}
+
+get_capture_snd()
+{
+    tmp=$(get_snd_base "$1")
+    echo "${tmp}"c
+}
+
 # check the PCMs before alsabat test
 dlogi "check the PCMs before alsabat test"
 aplay   -Dplug$pcm_p -d 1 /dev/zero -q || die "Failed to play on PCM: $pcm_p"
 arecord -Dplug$pcm_c -d 1 /dev/null -q || die "Failed to capture on PCM: $pcm_c"
+
+# Set PGAs for PCMs to 0 dB value
+test -n "$(command -v "$TPLGREADER")" ||
+    die "Command $TPLGREADER is not available."
+
+test -n "$tplg" || die "Use -t or set environment variable TPLG to current topology"
+tplg_full_path=$(func_lib_get_tplg_path "$tplg")
+dlogi "Getting playback PGA information"
+play_snd=$(get_play_snd "$pcm_p")
+PLAY_PGAS=$($TPLGREADER "$tplg_full_path" -f "snd:$play_snd" -d pga -v)
+set_pgas_list_to_unity_gain $PLAY_PGAS
+dlogi "Getting capture PGA information"
+cap_snd=$(get_capture_snd "$pcm_c")
+CAP_PGAS=$($TPLGREADER "$tplg_full_path" -f "snd:$cap_snd" -d pga -v)
+set_pgas_list_to_unity_gain $CAP_PGAS
 
 # alsabat test
 # hardcode the channel number of playback to 2, as sof doesnot support mono wav.


test-case/check-alsabat.sh
```diff
@@ -64,11 +72,52 @@ function __upload_wav_file
     done
 }
 
+function set_pga_to_unity_gain
```
marc-hb, line 75: `function` is less portable across shells and adds no value.
For now we only support bash but it'd be great to be compatible with more.

The most portable is:
```shell
set_pga_to_unity_gain()
{
  ...
```

test-case/check-alsabat.sh
```diff
@@ -64,11 +72,52 @@ function __upload_wav_file
...
+    tmp=${scale#*$search}
+    min_db="${tmp%%dB*}"
+    search="step="
+    tmp=${scale#*$search}
+    step_db="${tmp%%dB*}"
+
+    # Get multiplied by 100 values by removing decimal dot
+    min_x100="${min_db//.}"
+    step_x100="${step_db//.}"
+    val=$(printf %d "$(((-$min_x100) / $step_x100))")
```
marc-hb, line 95: Nit: you don't need the `'$'` inside `$(( ))`

test-case/check-alsabat.sh
```diff
@@ -64,11 +72,52 @@ function __upload_wav_file
...
+    amixer cset name="$cname" "$val"
+}
+
 # check the PCMs before alsabat test
 dlogi "check the PCMs before alsabat test"
 [[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && die "Failed to play on PCM: $pcm_p"
 [[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && die "Failed to capture on PCM: $pcm_c"
 
+# Set PGAs for PCMs to 0 dB value
+test -n "$tplg" || die "Use -t or set environment variable TPLG to current topology"
```
marc-hb, line 107: Please double-check but I suspect `func_opt_parse_option` gives something like that for free (it should)

marc-hb (commented):
Interestingly https://sof-ci.01.org/softestpr/PR437/build271/devicetest failed to even start this test on all platforms except BSW_CYN_MAX98090 (which tends to have other, unrelated and severe eMMC issues, I digress).

 It suspect the test fails to grep some (PGA?) control and then does not start because of `set -e`, did you expect that? Either way I think you should add more explicit error handling because `grep` does not print anything when it fails to find anything, in other words:
```shell

scale=$(amixer cget name="$cname" | grep "dBscale" || true)
test -n "$scale" || die "dbScale $cname not found"
```

test-case/check-alsabat.sh
singalsu: Ah, yes I was following the convention in the script, agree. Maybe another PR could be done to update the rest.

test-case/check-alsabat.sh
singalsu: Thanks, I didn't know nor understand what shellcheck suggested!

singalsu:
@marc-hb @xiulipan Now I think I know what's the problem. The CI system runs the script with command
`￼￼https://github.com/thesofproject/sof-test/tree/master/test-case/check-alsabat.sh
`
It fails because sof-tplgreader.py can't find the file without proper path. Shouldn't the convention to be to have full path to topology file in TPLG environment variable?
`/home/ubuntu/sof-test/tools/sof-tplgreader.py sof-hda-generic-4ch.tplg
`

singalsu:
This would be simple to fix by testing with appended default path if topology file does not exist but I don't feel that is the proper thing to do. The call of this script does not seem to be in this repository.

aiChaoSONG:
@singalsu The bash function `func_lib_get_tplg_path` in lib.sh helps to get topology full path, whatever a full path or only topology name is specified, because the `/lib/firmware/intel/sof-tplg` is the default folder for TPLG on every computer that uses SOF. But the  sof-tplgreader.py is a general tool, developer may not put the topology in  /lib/firmware/intel/sof-tplg, so cannot suppose there is a default folder. I think this makes sense.

marc-hb (changes requested):
> tplg_full_path=$(func_lib_get_tplg_path $tplg)

+1

xiulipan:
@singalsu I think this would be good for nocodec test case. What about the codec cases? There maybe some volume control in codec. How could we handle those values for test?
PS: we also used this scripts for the USB codec, which may not be covered by TPLG.

@keqiaozhang Any comments here for codec test cases? I think in CI framework, we manually tuned each volume gain value to make sure we did not get peak off in record wavs.

Maybe we can set all value to 0db, so ideally we should get almost same wav we played.

singalsu:
@xiulipan The codec based (analog or digital) gains are better to set up per device to be suitable for the (usually hw:1,0) USB sound card and leave them that way. Since the script gets with this patch the PGAs names from topology it fortunately does not alter them at all. In some cases setting codec gain to 0 dB could possibly "overload" the USB sound card A/D converter.

There could be another script to automatically set the device D/A and USB A/D gains if adjustable via SW. The procedure would optimize THD+N performance of playback-capture loop.

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
...
+	dlogi "Set $pga"
+	set_pga_to_unity_gain "$pga"
+    done
+}
+
+function get_snd_base ()
+{
+    tmp=${1#*"hw:"}
+    ncard="${tmp%%,*}"
+    ndevice="${tmp#*,}"
```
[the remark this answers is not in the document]
singalsu: @marc-hb Sure!

@aiChaoSONG This e.g. hw:1,0 conversion to /dev/snd/PCMC0D0p is because I'm not able to filter with sof-tplgtool.py with search -f "dev:hw:1,0". Maybe it's a quoting issue by me? These functions get simpler if such search can be done.

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
     done
 }
 
+function set_pga_to_unity_gain ()
+{
+    tmp=$(amixer controls | grep "$pga" | grep Volume)
```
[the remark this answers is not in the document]
singalsu: Oops, how come has that appeared to work!

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
...
+
+    # Apply the computed value for requested gain
+    amixer cset name="$cname" "$val"
+}
+
+function set_pgas_list_to_unity_gain ()
+{
+    if [[ -z "$1" ]]; then
+	return
+    fi
```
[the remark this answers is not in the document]
singalsu: I think I got problems with "set -e" but I'll retry.

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
...
+
+    for pga in $1; do
+	dlogi "Set $pga"
+	set_pga_to_unity_gain "$pga"
+    done
+}
+
+function get_snd_base ()
```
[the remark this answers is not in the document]
singalsu: ok!

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
...
+
 # check the PCMs before alsabat test
 dlogi "check the PCMs before alsabat test"
 [[ $(aplay -Dplug$pcm_p -d 1 /dev/zero -q) ]] && die "Failed to play on PCM: $pcm_p"
 [[ $(arecord -Dplug$pcm_c -d 1 /dev/null -q) ]] && die "Failed to capture on PCM: $pcm_c"
 
+# Set PGAs for PCMs to 0 dB value
+test -n "$(command -v $TPLGREADER)" || {
+    die "Command $TPLGREADER is not available."
+}
```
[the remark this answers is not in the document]
singalsu: Yep, I forgot it again!

test-case/check-alsabat.sh
```diff
@@ -64,11 +73,84 @@ function __upload_wav_file
...
+    amixer cset name="$cname" "$val"
+}
+
+function set_pgas_list_to_unity_gain ()
+{
+    if [[ -z "$1" ]]; then
+	return
+    fi
+
+    for pga in $1; do
```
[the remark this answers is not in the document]
singalsu: This is confusing, but let's try.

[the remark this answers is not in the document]
singalsu: Shellcheck complains about line `set_pgas_list_to_unity_gain $CAP_PGAS` with SC2086: Double quote to prevent globbing and word splitting. It works OK though.

test-case/check-alsabat.sh
```diff
@@ -64,11 +72,83 @@ function __upload_wav_file
...
+
+set_pgas_list_to_unity_gain()
+{
+    for pga in "$@"; do
+	dlogi "Set $pga"
+	set_pga_to_unity_gain "$pga"
+    done
+}
+
+get_snd_base()
```
marc-hb, line 109: `get_snd_dev()`?

test-case/check-alsabat.sh
```diff
@@ -64,11 +72,83 @@ function __upload_wav_file
...
+    # Converts e.g. string hw:1,0 to /dev/snd/pcmC1D0
+    # the p or c for playback or capture is appended
+    # in the calling function.
+    tmp=${1#*"hw:"}
+    ncard="${tmp%%,*}"
+    ndevice="${tmp#*,}"
+    echo "/dev/snd/pcmC${ncard}D${ndevice}"
+}
+
+get_play_snd()
```
marc-hb, line 120: `get_play_snd_dev()`?

test-case/check-alsabat.sh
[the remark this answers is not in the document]
marc-hb: Yes, that's expected because you should ideally not do that either. You should ideally never use a single string to store multiple, whitespace-separated strings. My suggestion that you applied is only "the lesser of two evils", my suggestion was to do this bad thing only in a single place (here) keeping the `set_pgas_list_to_unity_gain()` function clean. Thanks for applying it.

So, ideally CAP_PGAS should be converted to an array. In practice we're going to assume that the output of `$TPLGREADER` is safe and you can tell shellcheck that with this:
```shell
CAP_PGAS=$($TPLGREADER "$tplg" -f "snd:$cap_snd" -d pga -v)
# Let's assume the output of $TPLGREADER is free of *, other special characters and spurious spaces.
# shellcheck disable=SC2086
set_pgas_list_to_unity_gain $CAP_PGAS
```

keqiaozhang:
> @singalsu I think this would be good for nocodec test case. What about the codec cases? There maybe some volume control in codec. How could we handle those values for test?
> PS: we also used this scripts for the USB codec, which may not be covered by TPLG.
> 
> @keqiaozhang Any comments here for codec test cases? I think in CI framework, we manually tuned each volume gain value to make sure we did not get peak off in record wavs.
> 
> Maybe we can set all value to 0db, so ideally we should get almost same wav we played.

I agree to set all PGA value to 0dB. But if we do this, I need to retune the related codec kcontrols(like headset volume, DAC volume etc.) for  the old platforms we added before to make sure the waveform does not overflow.

singalsu:
> > @keqiaozhang Any comments here for codec test cases? I think in CI framework, we manually tuned each volume gain value to make sure we did not get peak off in record wavs.
> > Maybe we can set all value to 0db, so ideally we should get almost same wav we played.
> 
> I agree to set all PGA value to 0dB. But if we do this, I need to retune the related codec kcontrols(like headset volume, DAC volume etc.) for the old platforms we added before to make sure the waveform does not overflow.

Does it mean that the firmware PGAs have been intentionally set to other than 0 dB gain to compensate for some (non-sufficient?) codec gain controls?

I made this PR because I noticed that the alsabat test failed when the capture pipeline was changed to other type (pipe-volume-switch-capture.m4) that has higher max. gain available.

singalsu:
SOFCI TEST

singalsu:
@keqiaozhang There's now two timeouts remaining, one for a CML platform and one for a GLK platform. Those seem to be due to excessive signal level (distortion, multiple sine waves detected). I wonder if I should try a lower digital gain for all platforms. Or can the analog DAC or headphone gains be tuned down in those platforms to enable unity digital gains to pass? The latter would be preferred if possible.

keqiaozhang:
> @keqiaozhang There's now two timeouts remaining, one for a CML platform and one for a GLK platform. Those seem to be due to excessive signal level (distortion, multiple sine waves detected). I wonder if I should try a lower digital gain for all platforms. Or can the analog DAC or headphone gains be tuned down in those platforms to enable unity digital gains to pass? The latter would be preferred if possible.

amixer setting for GLK:
amixer -c sofglkda7219max cset name='Headphone Gain Ramp Switch' 1
amixer -c sofglkda7219max cset name='Headphone Volume' 35 35
amixer -c sofglkda7219max cset name='Headphone Switch' 1
amixer -c sofglkda7219max cset name='Headphone ZC Gain Switch' 0
amixer -c sofglkda7219max cset name='Headphone Jack Switch' 1
amixer -c sofglkda7219max cset name='Playback Digital Volume' 111 111
amixer -c sofglkda7219max cset name='Playback Digital Switch' 1
amixer -c sofglkda7219max cset name='Playback Digital Gain Ramp Switch' 1
amixer -c sofglkda7219max cset name='Out DACR Mux' 3
amixer -c sofglkda7219max cset name='Out DAIR Mux' 0
amixer -c sofglkda7219max cset name='Mixer Out FilterL DACL Switch' 1
amixer -c sofglkda7219max cset name='Mixer Out FilterR DACR Switch' 1 10
amixer -c sofglkda7219max cset name='PGA1.0 1 Master Playback Volume' 32

For CML, which device has the timeout issue?

marc-hb:
There is a relatively small number of shellcheck warnings left and most of them seem easy to fix: just missing quotes. While you're testing this script, @singalsu would you mind applying the most obvious shellcheck's suggestions in a separate commit? Don't mind any complex warning or complex quoting issues, any progress is useful.

test-case/check-alsabat.sh
```diff
@@ -72,11 +79,83 @@ function __upload_wav_file
...
+    tmp=${scale#*$search}
+    min_db="${tmp%%dB*}"
+    search="step="
+    tmp=${scale#*$search}
+    step_db="${tmp%%dB*}"
+
+    # Get multiplied by 100 values by removing decimal dot
+    min_x100="${min_db//.}"
+    step_x100="${step_db//.}"
+    val=$(printf %d "$(((-min_x100) / step_x100))")
```
marc-hb, line 102: Nit, this would be slightly easier to read with some spaces:
```shell
val=$(printf %d "$(( (-min_x100) / step_x100))" )
```
Why do you need the printf at all?

test-case/check-alsabat.sh
```diff
@@ -72,11 +79,83 @@ function __upload_wav_file
...
+set_pga_to_unity_gain()
+{
+    tmp=$(amixer controls | grep "$1" | grep Volume)
+    search="name="
+    cname=${tmp#*$search}
+
+    # Get volume min and step to compute value for cset
+    # for 0 dB gain. The amixer line looks like
+    # "| dBscale-min=-50.00dB,step=1.00dB,mute=1"
+    scale=$(amixer cget name="$cname" | grep "dBscale" || true)
```
marc-hb, line 91: `|| true` means an empty `$scale` is OK. Is it really?

test-case/check-alsabat.sh
```diff
@@ -72,11 +79,83 @@ function __upload_wav_file
     done
 }
 
+set_pga_to_unity_gain()
+{
+    tmp=$(amixer controls | grep "$1" | grep Volume)
```
marc-hb, line 84: As you use very generic variable names, this first line will avoid any collision:
```
local tmp search cname scale ...
```

marc-hb (commented):
Just curious: would some UCM make all this easier to solve? I know @plbossart has been complaining about alsa settings since forever.

xiulipan:
> Just curious: would some UCM make all this easier to solve? I know @plbossart has been complaining about alsa settings since forever.

@singalsu @keqiaozhang @libinyang I remember you said the volume value can not be reset by UCM right? Please confirm here

marc-hb:
@singalsu did you forget about this? :-)

plbossart:
@singalsu is this still relevant or should we close this one?

marc-hb:
@singalsu please re-open when you get a chance.

test-case/check-alsabat.sh
```diff
@@ -72,11 +79,83 @@ function __upload_wav_file
...
 
+# Set PGAs for PCMs to 0 dB value
+test -n "$(command -v "$TPLGREADER")" ||
+    die "Command $TPLGREADER is not available."
+
+test -n "$tplg" || die "Use -t or set environment variable TPLG to current topology"
+tplg_full_path=$(func_lib_get_tplg_path "$tplg")
+dlogi "Getting playback PGA information"
+play_snd=$(get_play_snd "$pcm_p")
+PLAY_PGAS=$($TPLGREADER "$tplg_full_path" -f "snd:$play_snd" -d pga -v)
```
marc-hb, line 152: `-d pga` seems to be compatible with topology v1 only, same problem as in #1068. @ranj063 confirm?

ranj063: are we talking about the string "pga" in the kcontrol name here? Then yes it is compatible with the tplg2 as well. Take a look at the gain kcontrols in the nocodec tplg https://sof-ci.01.org/linuxpr/PR4937/build2360/devicetest/index.html?model=TGLU_RVP_NOCODEC-ipc4&testcase=verify-tplg-binary

marc-hb: When run on topology v1 files, `tools/sof-tplgreader.py` shows a field named `pga`. Example:
```
 ./tools/sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-apl-nocodec.tplg -d pga


pga=PGA9.0 PGA5.0;
pga=PGA6.0;
pga=PGA11.0 PGA5.0;
```
Note the blank lines stand for DMIC and other PCMs without a `pga=` field.

What would be the equivalent for topology v2? Is there a v2 compatibility issue with `sof-tplgreader.py`?

marc-hb: > Take a look at the gain kcontrols in the nocodec tplg https://sof-ci.01.org/linuxpr/PR4937/build2360/devicetest/index.html?model=TGLU_RVP_NOCODEC-ipc4&testcase=verify-tplg-binary

This device has at least 15 amixer controls with "Volu.." in the name (`name=` which may or may not be truncated). That's the list of kcontrols volume tests want, correct?

That log shows a list of pipelines, 3 of them which have `cap_name=Gain Capture`.

I don't see the connection.

ranj063: @marc-hb i see your problem. Sorry I misundertood the PGA in the picture. Looks like that's the tplgtool printing that the PGA prefix for the kcontrols. 

So, yes there's no "pga" string in tplg2 kcontrol names. Infact we consciously removed the widget name from the kcontrol name as well so you have "gain" either. So the question is how to figure out which kcontrols to test?

Can we do this during tplg parsing ie when as we parse the gain widgets, save the kcontrol names for the kcontrols associated with the widget and test only those?

marc-hb: > as we parse the gain widgets, save the kcontrol names for the kcontrols associated with the widget and test only those?

Sound like a plan. Let me take a look.

PR #438, opened 2020-10-15
test: adjust the vloume for KWD capture

keqiaozhang:
During the test, we found that KWD capture volume is a little
bit high. The volume will affect the THDN value and leads WOV
test fails occasionally. Lower the volume to make the WOV test
more stable.


xiulipan (approved):

aiChaoSONG:
I have synced with Seppo, the volume levels do impact the THD+N, so we need to tune the volume for test case

aiChaoSONG (approved):

aiChaoSONG:
The ICL failure has nothing to do with this PR.

PR #439, opened 2020-10-16
config.sh: enable media playback pipeline test

aiChaoSONG:
Media playback pipeline test is blocked in the
early stage of SOF development, maybe because it
caused a lot of trouble.

But as the SOF evolves, we don't have a reason to
continue blocking the test on media playback pipeline
any more.


xiulipan:
@aiChaoSONG Would you like to do a daily_lite test on more DUT?

xiulipan (approved):
An expand test on all CI DUT passed.

aiChaoSONG:
Thanks Xiuli for helping to run thorough test on this PR

plbossart:
If the "Media Playback" is going into a mixer, you are going to have issue https://github.com/thesofproject/sof/issues/3522

on which devices did you test this?

aiChaoSONG:
@plbossart Xiuli has run a lite test on every platform we have, and report ID is 566

PR #440, opened 2020-10-16
test-case: add kern.log debug logs in verify-sof-firmware-load

xiulipan:
With previous logs, we can norrow down the empty sof version info is related with
kern.log. Add debug logs to kern.log to help us root cause the issue.


aiChaoSONG (approved):
I guess when issue happens, the kern.log is empty, but let's check with this debug patch.

xiulipan:
@aiChaoSONG Yes, that was also my guess. Let's try to use the debug logs to confirm the issue while we are working on the job to replace kern.log with journalctl.

Issue #441, opened 2020-10-16
check-alsabat-headset-playback.sh TIMEOUT instead of FAIL


xiulipan:
Seen in latest test logs

```
Return value is -1003
2020-10-16 06:27:17 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=21, FUNCNAME stack:
2020-10-16 06:27:17 UTC [REMOTE_ERROR]  main()  @  /home/ubuntu/sof-test/test-case/check-alsabat.sh
2020-10-16 06:27:17 UTC [REMOTE_INFO] pkill -TERM sof-logger
Terminated
2020-10-16 06:27:18 UTC [REMOTE_INFO] nlines=878 /home/ubuntu/sof-test/logs/check-alsabat/2020-10-16-06:27:12-10002/slogger.txt
2020-10-16 06:27:18 UTC [REMOTE_INFO] Unknown exit code: 21
```

https://sof-ci.01.org/linuxpr/PR2501/build4685/devicetest/

While old alsabat test fail logs is 

```
Return value is -1003
2020-09-16 14:00:00 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2020-09-16 14:00:00 UTC [REMOTE_ERROR]  main()  @  /home/ubuntu/sof-test/test-case/check-alsabat.sh
2020-09-16 14:00:02 UTC [REMOTE_INFO] Test Result: FAIL!
```

xiulipan:
@marc-hb @aiChaoSONG I am thinking this issue is related with latest https://github.com/thesofproject/sof-test/pull/376.
The set -e changed our exception handler for this test case.

PR #442, opened 2020-10-16
test-case: check-alsabat: remove set -e

xiulipan:
There is error hanlder to upload the failure wav files in this test case.
set -e will skip them

And our lib could not handle the return value 23 from alsabat failure.

Fix https://github.com/thesofproject/sof-test/issues/441

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
-
 ##
 ## Case Name: check alsabat
 ## Preconditions:


xiulipan:
@RuiqingX @marc-hb @aiChaoSONG We may need special handler for the alsabat test.
Simply add `set -e` will break our return value check in `lib.sh` and skip our fail wav upload function.
@keqiaozhang Please help to check if `set -e` also influence our WoV test and echo reference test.

marc-hb:
I agree adding `set -e` on existing tests requires a fair amount of testing and could create issues like this one. We could need more time and more testing before using `set -e` to check-alsabat.sh

On the other hand, can you please provide more detailed reproduction steps because for now the connection between `set -e` and these TIMEOUTS is really not obvious to me.

fredoh9 (approved):
LGTM

xiulipan:
@marc-hb The TIMEOUT is actually from the CI framework as the sof-test output when alsabat test failed do not output a valid result. EG: `Test Result: FAIL!`
This will make framework think the scripts is not finished and report TIMEOUT.

marc-hb (approved):

marc-hb:
> The TIMEOUT is actually from the CI framework as the sof-test output when alsabat test failed do not output a valid result. EG: Test Result: FAIL!

Thanks for the explanation. Is there no way for the CI framework to tell the test script process has completed?

PR #443, opened 2020-10-16
sof-kernel-log-check.sh: move the DSP reset filter out of platform filter

xiulipan:
Now the DSP reset failur is seen on more platform. Move this filter out     of platform.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -84,29 +84,26 @@ ignore_str="$ignore_str"'|usb 2-.: .'
 # --priority. Hopefully other issues will cause the test to fail in that
 # case.
 #
-# For now the codes seem to be 0x00000000 and affected platforms have
-# PCI ID 1f.3. Before adding other values make sure you update the list
-# of affected systems in bug 3395 below.
-#
 # Buglink: https://github.com/thesofproject/sof/issues/3395
+
+ignore_str="$ignore_str"'|sof-audio-pci 0000:00:..\..: status = 0x[0]{8} panic = 0x[0]{8}'
+# There will be debug logs at each failed initializaiton of DSP before Linux 5.9
+#   sof-audio-pci 0000:00:1f.3: error: cl_dsp_init: timeout HDA_DSP_SRAM_REG_ROM_STATUS read
+#   sof-audio-pci 0000:00:1f.3: error: status = 0x00000000 panic = 0x00000000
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -79,34 +77,24 @@ ignore_str="$ignore_str"'|usb 2-.: .'
...
-        ignore_str="$ignore_str"'|error: iteration [01]'
-        ignore_str="$ignore_str"'|error: status'
-        ignore_str="$ignore_str"'|error: cl_dsp_init: timeout HDA_DSP_SRAM_REG_ROM_STATUS read'
-
-        # On CML_RVP_SDW, suspend-resume test case failed due to "mei_me 0000:00:16.4: hw_reset failed ret = -62".
-        # https://github.com/thesofproject/sof-test/issues/389
-        ignore_str="$ignore_str"'|mei_me 0000:00:16\..: hw_reset failed ret = -62'
-        ;;
-esac
+ignore_str="$ignore_str"'|sof-audio-pci 0000:00:1f\.3: status = 0x[0]{8} panic = 0x[0]{8}'
```
plbossart, line 81: this doesn't work on APL/GLK, the PCI ID is different...

marc-hb (changes requested):
> Most error logs come from shared issues,

I'm not sure in general but most of the error strings added recently were platform specific. They were (wrongly) added in the platform-independent part of the code.

> no need to map each error log with each platform.

We already discussed the value of this when the FW boot failure was added, see the corresponding PR.

marc-hb:
I just got some more context off-line. Please update https://github.com/thesofproject/sof/issues/3395 with the new information and move only the 3395-related messages out of the `case`, leave the other error inside the `case`.

> I'm not sure in general but most of the error strings added recently were platform specific. They were (wrongly) added in the platform-independent part of the code.

In fact just reading the comments in this file is enough to see that most errors are platform specific.

tools/sof-kernel-log-check.sh
xiulipan: Oops, I will try to update these with regex.

marc-hb (approved):
CI is broken, probably because of this kconfig change https://github.com/thesofproject/sof-test/issues/458

aiChaoSONG:
SOFCI TEST

aiChaoSONG:
A new round or CI test is passed, merge, The broken CI by kconfig change is fixed.

PR #444, opened 2020-10-19
wavetool: lower default threshold for WoV test case

aiChaoSONG:
For current WoV setup, the THD+N of a captured wave
without any distortion may be lower than -75dB, lower
the default threshold to make WoV test pass.


keqiaozhang (approved):

aiChaoSONG:
no regression introduced

Issue #245, opened 2020-06-04
test-speaker.sh: fix description


plbossart:
first line doesn't look so good:
````bash
# TODO: case & descrption step need to confirm
````

PR #447, opened 2020-10-19
test-speaker: delete TODO and case step is detail

RuiqingX:
Fixes:#245


aiChaoSONG:
please check $448, we can add some case description indeed.

PR #452, opened 2020-10-20
env-check.sh: show which file(s) lack exec permissions. Other cleanups

marc-hb:
It was not very useful to tell the user that some files lack execute
permission without saying which ones. Especially not useful when the
files missing execute permission are irrelevant files like foo.sh~
backup files. This is also simplifies the code.

Also rephrase the error messages and offer a simpler chmod fix

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -69,27 +69,29 @@ fi
 
 # check for the tools folder
 out_str="" check_res=0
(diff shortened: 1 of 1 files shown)


marc-hb:
The last other, unrelated shellcheck warnings in the group membership code are fixed in unrelated PR #454

xiulipan (approved):

PR #454, opened 2020-10-20
env-check.sh: rewrite complicated and duplicated group membership

marc-hb:
Simplify and remove duplication, use an array and for loop instead. 450
fewer characters.
Stop parsing files directly, switch to getent system interface instead.
Stop recommending editing file directly, recommend usermod system
interface instead.

Fix the last shellcheck warnings in this file:

In env-check.sh line 107:
```
[[ "$user" ]] && [[ ! $(awk -F ':' '/^adm:/ {print $NF;}' /etc/group|grep "$user") ]] && \
                      ^-- SC2143: Use grep -q instead of comparing output with [ -n .. ].
```
(3 times because of code duplication)

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -94,7 +94,8 @@ echo -ne "Checking for case folder:\t\t"
     echo -e "\e[31mWarning\e[0m\nSolution:$out_str"
 
 out_str="" check_res=0
-echo -ne "Checking the permission:\t\t"
+echo -ne "Checking group memberships:\t\t"
+
 if [[ "$SUDO_USER" ]]; then
     user="$SUDO_USER"
 elif [[ "$UID" -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


marc-hb:
The last shellcheck warnings in this file are fixed in (unrelated) PR #452

env-check.sh
```diff
@@ -94,32 +94,32 @@ echo -ne "Checking for case folder:\t\t"
...
-\tMissing permission to run command as sudo\n
-\t\tPlease use the following command to add current user to the sudo group:\n
-\t\e[31msed -i '/^sudo:/s:$:,$user:g' /etc/group\e[0m"
+declare -A grp_error
+grp_error=(
+[adm]='\tMissing permission to access /var/log/kern.log\n'
+[sudo]='\tMissing permission to run command as sudo\n'
+[audio]='\tMissing audio group membership to access /dev/snd/* devices\n'
+)
+for grp in "${!grp_error[@]}"; do
```
xiulipan, line 113: why not wrap this as a function.

```
check_group adm 'Missing permission to access /var/log/kern.log'
check_group sudo 'Missing permission to run command as sudo'
check_group audio 'Missing audio group membership to access /dev/snd/* devices' ```

xiulipan (changes requested):
Nice idea but I think it would be better with some function wrapper.

env-check.sh
marc-hb: Good idea, thanks! Done.

xiulipan (approved):
LGTM

PR #455, opened 2020-10-20
check-suspend-resume-with-audio: don't discard stderr

marc-hb:
Never discard stderr unless you don't want to find bugs. Who knows this
could help with https://github.com/thesofproject/linux/issues/2433?

This makes no output difference when the test is passing so not sure
what the point was. Maybe it was a failed attempt to silence the shell's
"Killed" message? Also adding a new dlogi() to make it clear this
"Killed" message is not a failure.


marc-hb:
No new shellcheck warning

xiulipan (approved):

Issue #458, opened 2020-10-21
kmod test failed on some platforms with SST config enabled


xiulipan:
After the SST config https://github.com/thesofproject/kconfig/pull/42 merged
Most kmod test failed https://sof-ci.01.org/linuxpr/PR2511/build4725/devicetest/

`rmmod: ERROR: Module snd_soc_acpi_intel_match is in use by: snd_soc_skl`

marc-hb:
Here are some (large) CI failures that seem related
https://sof-ci.01.org/softestpr/PR443/build307/devicetest
https://sof-ci.01.org/sofpr/PR3540/build7490/devicetest/
https://sof-ci.01.org/linuxpr/PR2518/build4731/devicetest/

marc-hb:
https://sof-ci.01.org/sofpr/PR3540/build7493/devicetest/ says

```
21/10/2020, **14:10** (UTC)
Linux: f8922c96 ASoC: rt711-sdca: Add RT711 SDCA vendor-specific driver
```
But the kernel commit f8922c96 does not include the next Linux commit 37740626ab which was merged at  2020-10-21 **02:21** UTC as shown by  `TZ=UTC git log --date=iso-local --format=fuller sof/topic/sof-dev` and the merged time shown in https://github.com/thesofproject/linux/pull/2487

So this SOF PR that ran at 14:10 UTC didn't include a 02:21 UTC kernel commit. I think this is proof that non-kernel Pull Requests use a nightly build of the kernel (for performance reasons?). @xiuli confirm?

If confirmed then this is enough to explain why most PRs failed today.

Warning: the SHA1s at https://sof-ci.01.org/ are _incomplete_ because the PR's SHA1 is actually combined with the latest-branch-of-the-moment before testing (which is good) but there's no indication of what exactly it was combined with (which is bad). I filed internal issue 498 to address this.

marc-hb:
Newer https://sof-ci.01.org/sofpr/PR3540/build7494/devicetest is mostly OK.

PR #459, opened 2020-10-21
tools: kmod: remove snd_soc_skl

xiulipan:
Now SST driver is enabled alongside with SOF drvier. Need to
remove the SST drivers as well.

Fix: #458


xiulipan:
@plbossart Please help review and merge this. This is also needed for the kconfig updated

plbossart (approved):
thanks @xiulipan, will merge to save on time

Issue #461, opened 2020-10-22
test scripts usage helper output in wrong format 


xiulipan:
When use `-h` option to print out usage, the format is in wrong shape.  Need to fix the output format.

```
$ ./test-case/check-alsabat.sh -h
Usage: ./test-case/check-alsabat.sh [OPTION]
    -s |  --sof-logger
        Open sof-logger trace the data will store at /home/pxl/work/cti/sof-test/logs/check-alsabat/2020-10-22-04:38:22-20564
        Default Value: On    -p parameter |  --pcm_p parameter
        pcm for playback. Example: hw:0,0
    -n parameter |  --frames parameter
        test frames
        Default Value: 240000    -f parameter |  --frequency parameter
        target frequency
        Default Value: 997    -c parameter |  --pcm_c parameter
        pcm for capture. Example: hw:1,0
    -h |  --help
        this message
```

xiulipan:
With fix we can get more clean format.

```
$ ./test-case/check-alsabat.sh -h
Usage: ./test-case/check-alsabat.sh [OPTION]
    -s |  --sof-logger
        Open sof-logger trace the data will store at /home/pxl/work/cti/sof-test/logs/check-alsabat/2020-10-22-07:26:46-19805
        Default Value: On
    -p parameter |  --pcm_p parameter
        pcm for playback. Example: hw:0,0
    -n parameter |  --frames parameter
        test frames
        Default Value: 240000
    -f parameter |  --frequency parameter
        target frequency
        Default Value: 997
    -c parameter |  --pcm_c parameter
        pcm for capture. Example: hw:1,0
    -h |  --help
        this message
```

PR #462, opened 2020-10-22
lib: opts: fix format in help

xiulipan:
Add missed newline for the default value in help usage
output.

fix: #461


aiChaoSONG (approved):
LGTM

marc-hb (approved):
For an output change like this it's good to share a sample output before versus after the change

Now this particular change is small enough so I just trust you blindly :-)

aiChaoSONG:
@marc-hb check it here: #461

marc-hb:
> @marc-hb check it here: #461

Thanks and sorry it was getting late!

(I would have not bothered with a bug and used the commit message directly BTW)

PR #263, opened 2020-06-28
lib: hijack: use journalctl command to dump kernel information

Bin-QA:
use journalctl instead of catch information from /var/log/kern.log file
1. DMESG_LOG_START_LINE => CASE_KERNEL_START_TIME
2. with this change, sof-get-kernel-line is not useful, remove it

Signed-off-by: Wu, BinX <binx.wu@intel.com>

PR #468, opened 2020-10-23
[RFC] use journalctl for kernel log

xiulipan:
Try to use journalctl to read kernel log from system.
Use UNIX style timestamp to split kernel logs from test case to test case.
Print same format like dmesg in journalctl kernel log.
Update all test case used to relay on demsg logs

### DIFF
--- a/test-case/check-ipc-flood.sh
+++ b/test-case/check-ipc-flood.sh
@@ -39,20 +39,20 @@ dlogi "Check sof debug fs environment"
 [[ "$(sudo file $ipc_flood_dfs|grep 'No such file')" ]] && dlogw "${BASH_SOURCE[0]} need $ipc_flood_dfs to run the test case" && exit 2
 dlogi "Checking ipc flood test!"
 
-# cleanup dmesg buffer before test
-sudo dmesg -c > /dev/null
-
 for i in $(seq 1 $loop_cnt)
 do
+    # cleanup dmesg buffer for each iteration
+    sudo dmesg -c > /dev/null
+    # set up timestamp for each iteration
+    func_lib_setup_kernel_last_timestamp
     dlogi "===== [$i/$loop_cnt] loop Begin ====="
     dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
     sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-    sof-kernel-log-check.sh 0 || die "Catched error in dmesg"
+    # check kernel log for each iteration to catch issues
+    sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP || die "Catch error in kernel log"
 
     dlogi "Dumping test logs!"
     dmesg | grep "IPC Flood count" -A 2
-done
 
-sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP
-exit 0
+done
--- a/test-case/check-ipc-flood.sh
+++ b/test-case/check-ipc-flood.sh
@@ -32,7 +32,7 @@ loop_cnt=${OPT_VALUE_lst['l']}
 
 [[ ! "$(sof-kernel-dump.sh|grep 'sof-audio'|grep 'Firmware debug build')" ]] && dlogw "${BASH_SOURCE[0]} need debug version firmware" && exit 2
 
-func_lib_setup_kernel_last_line
+func_lib_setup_kernel_last_timestamp
 func_lib_check_sudo
 
 dlogi "Check sof debug fs environment"
@@ -54,5 +54,5 @@ do
     dmesg | grep "IPC Flood count" -A 2
 done
 
-sof-kernel-log-check.sh $KERNEL_LAST_LINE
+sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP
 exit 0
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -75,9 +75,10 @@ esac
 [[ -z $file_name ]] && file_name=$dummy_file
 
 func_pipeline_export $tplg "type:$test_mode & ${OPT_VALUE_lst['S']}"
-func_lib_setup_kernel_last_timestamp
 for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
 do
+    # set up timestamp for each iteration
+    func_lib_setup_kernel_last_timestamp
     channel=$(func_pipeline_parse_value $idx channel)
     rate=$(func_pipeline_parse_value $idx rate)
     fmt=$(func_pipeline_parse_value $idx fmt)
@@ -122,9 +123,6 @@ END
         [[ $? -ne 0 ]] && dlogw "Kill process catch error"
         exit $ret
     fi
-    # sof-kernel-log-check script parameter number is 0/Non-Number will force check from dmesg
-    sof-kernel-log-check.sh 0 || die "Catch error in dmesg"
+    # check kernel log for each iteration to catch issues
+    sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP || die "Catch error in kernel log"
 done
-
-sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP
-exit $?
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -75,7 +75,7 @@ esac
 [[ -z $file_name ]] && file_name=$dummy_file
 
 func_pipeline_export $tplg "type:$test_mode & ${OPT_VALUE_lst['S']}"
-func_lib_setup_kernel_last_line
+func_lib_setup_kernel_last_timestamp
 for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
 do
     channel=$(func_pipeline_parse_value $idx channel)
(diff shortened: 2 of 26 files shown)


marc-hb:
Previous attempts (and previous mistakes) in #263 and #354

"Those who ignore history are doomed to repeat it"

PS: I did NOT look at this newer attempt yet

xiulipan:
@marc-hb Yep, I reviewed both #263 and #354 and some patch is cherry-pick from previous work.
Still trying to find a clear way to get the log collection feature work with journalctl.

test-case/verify-sof-firmware-load.sh
```diff
@@ -36,12 +34,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
...
-    printf ' ------\n  debuging with /var/log/kern.log  \n ---- \n'
-    ls -alht /var/log/kern.log
-    grep -na "Linux version" /var/log/kern.log || true
-    printf ' ------\n  cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
-            "$cmd" "$DMESG_LOG_START_LINE"
-    journalctl --dmesg --lines 50 --no-pager
-    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
+    journalctl journalctl -k -q --no-pager --utc --output=short-iso --no-hostname --lines 50
     die "Cannot find the sof audio version"
+    printf ' ------\n  debuging with systemd journalctl  \n ---- \n'
```
ranj063, line 39: typo debugging

test-case/verify-sof-firmware-load.sh
```diff
@@ -36,12 +34,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
...
     exit 0
 else
-    printf ' ------\n  debuging with /var/log/kern.log  \n ---- \n'
-    ls -alht /var/log/kern.log
-    grep -na "Linux version" /var/log/kern.log || true
-    printf ' ------\n  cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
-            "$cmd" "$DMESG_LOG_START_LINE"
-    journalctl --dmesg --lines 50 --no-pager
-    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
+    journalctl journalctl -k -q --no-pager --utc --output=short-iso --no-hostname --lines 50
```
ranj063, line 37: duplicate journalctl?

tools/sof-get-default-tplg.sh
```diff
@@ -1,14 +1,12 @@
 #!/bin/bash
+#
+# read kernel log to get topology file loaded by SOF driver
+# Example from an apl up2 pcm512x device:
+#
+# sof-audio-pci 0000:00:0e.0: loading topology:intel/sof-tplg/sof-apl-pcm512x.tplg
+#
+# sof-apl-pcm512x.tplg will be return
```
ranj063, line 8: returned

tools/sof-kernel-log-check.sh
```diff
@@ -237,32 +234,19 @@ case "$platform" in
...
-# now decide using which to catch the kernel log
-#[[ $kernel_buffer_size -lt $dmesg ]] && cmd="dmesg" || cmd="tail -n  +${begin_line}  /var/log/kern.log"
-
-# confirm begin_line is number, if it is not the number, direct using dmesg
-[[ "${begin_line//[0-9]/}" ]] && begin_line=0
-[[ "$begin_line" -eq 0 ]] && cmd="dmesg" || cmd="tail -n  +${begin_line}  /var/log/kern.log"
-
-#echo "run $0 with parameter '$*' for check kernel message error"
+# confirm begin_timestamp is in UNIX timestamp format, otherwise search full log
+re='^[0-9]{10}'
```
ranj063, line 238: re? can we please use a better name?

ranj063: regex maybe

test-case/verify-sof-firmware-load.sh
xiulipan: copy paste issue.

tools/sof-kernel-log-check.sh
xiulipan: remove the variable and use regex directly.

test-case/check-pause-resume.sh
```diff
@@ -122,9 +123,6 @@ END
         [[ $? -ne 0 ]] && dlogw "Kill process catch error"
         exit $ret
     fi
-    # sof-kernel-log-check script parameter number is 0/Non-Number will force check from dmesg
-    sof-kernel-log-check.sh 0 || die "Catch error in dmesg"
+    # check kernel log for each iteration to catch issues
+    sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP || die "Catch error in kernel log"
```
ranj063, line 127: Caught error instead of Catch error

test-case/check-ipc-flood.sh
```diff
@@ -39,20 +39,20 @@ dlogi "Check sof debug fs environment"
...
+    sudo dmesg -c > /dev/null
+    # set up timestamp for each iteration
+    func_lib_setup_kernel_last_timestamp
     dlogi "===== [$i/$loop_cnt] loop Begin ====="
     dlogc "sudo bash -c 'echo $lpc_loop_cnt > $ipc_flood_dfs'"
     sudo bash -c "'echo $lpc_loop_cnt > $ipc_flood_dfs'"
 
-    sof-kernel-log-check.sh 0 || die "Catched error in dmesg"
+    # check kernel log for each iteration to catch issues
+    sof-kernel-log-check.sh $KERNEL_LAST_TIMESTAMP || die "Catch error in kernel log"
```
ranj063, line 53: same comment here. lets fix it everywhere

marc-hb (changes requested):
This is fixing an extremely critical feature so this must be broken down in smaller pull requests and merged progressively = "continuous" integration. This is especially important considering the regular stream of other, unrelated  CI changes and issues will not stop and will still need to be investigated and addressed at the same time. I mean we cannot stop everyone else (including you!) from making other CI fixes and improvements while performing this critical change. We also need the ability to quickly revert the last PR if anything goes wrong - and from past experience a number of things will go wrong.

- As long as you use the `-k` flags you do not need to modify the `ignore_str` at the same time. The last thing we want is to change the implementation and the ignore_str at the same time. So let's forget about any ignore_str changes for now.

- The --priority flag must also be postponed.

- The first PR to submit is the PR that renames `func_lib_setup_kernel_last_line` to `func_lib_setup_kernel_last_timestamp` and KERNEL_LAST_LINE to KERNEL_LAST_TIMESTAMP without making any real code change. That will first get a lot of changed lines out of the way with no risk. The name will not match the old implementation for the transition period but we don't care.

- The second PR to submit is quoting all the instances of $KERNEL_LAST_TIMESTAMP and then wait a day or two to make sure everything still works.

- Then we can start flipping the switch with high impact but small number of lines PR.


Most importantly: what sort of (fake) errors did you test this with? Testing a change like this requires at least two throw-away PRs:
- one throw-away PR with strategically placed fake errors + old code
- a second throw-away PR with the exact same fake errors + the new code

The test results for both must be exactly the same.

xiulipan:
@marc-hb Thanks for the tips, I am also think about how to spilt the patches. Please help to check my plan:

1. replace LINE with TIMESTAMP, log check with no `--priority` but still with `err_str`
2. quoting clean up for the `$KERNEL_LAST_TIMESTAMP`
3. remove `err_str` and use `--priority`, add new `ignore_str`
4. clean up test case to remove duplicate kernel log check.

some discussion I want to have here:

- What `priority` level we want to track here? @plbossart @ranj063 In this RFC, I have tried `--priority=err` Is this enough? Will warning be overkill?
- verify-sof-firmware-load.sh seems still have issue even with `sudo journalctl --sync --flush`, I could not tell what happened to it
and we still got a fail https://sof-ci.01.org/softestpr/PR468/build344/devicetest/ on BSW. I will try to fallback to dmesg for this test.

Keep this PR for testing and I will start with task 1.

xiulipan:
close with latest part 1 PR #480

marc-hb:
> Please help to check my plan:

Much finer-grained, much better! 

For such a "core" change like this we really need "Totally Continuous" Integration: small, well tested changes, quick reviews, and then at least a couple days careful observation between merged PRs. See my request in #480.
 One day observation is enough after non-functional changes like renames, more days needed to observe more impactful changes.

> What priority level we want to track here? @plbossart @ranj063 In this RFC, I have tried --priority=err Is this enough? Will warning be overkill?

We can talk about changing levels and filtering last, after everything else is in place. Then we should involve many more people. For now the priority is to switch to journalctl while preserving everything else exactly the same - and then find and fix everything that broke with just that before even tweaking the filtering.

marc-hb:
> verify-sof-firmware-load.sh seems still have issue 

Still now after fixing all clock and timezone issues?

> even with sudo journalctl --sync --flush, I could not tell what happened to it

These options make absolutely no difference except if you use them a split second before of system crash or sudden power loss. We are not really dealing with crashes anyway so don't use them, they're just confusing.

> and we still got a fail https://sof-ci.01.org/softestpr/PR468/build344/devicetest/ on BSW. 

BSW has been experiencing a lot of storage issues (see internal issue 539 + links), could that be the cause?

> I will try to fallback to dmesg for this test.

Please no, that  sounds brittle and complex. Let's understand what the real issue is first and/or temporarily disable BSW in the mean time.

PR #469, opened 2020-10-24
travis: add pylint

marc-hb:
It's much easier to review commits one by one because github fails to detect the file rename when reviewing all commits at once, this makes the diff much bigger than it really is.


marc-hb:
A earlier revision included a dummy, one-line .py file change for pure testing purposes. Here's how the output looked like: https://travis-ci.org/github/thesofproject/sof-test/jobs/738901369

tools/CI/check-gitrange.bash
```diff
@@ -0,0 +1,77 @@
...
+    local checked_ftype="$1"; shift || usage
+    file --list | grep -qF "$checked_ftype" ||
+        die 'The file command does not know what %s is\n' "$checked_ftype"
+
+    local checker="$1"; shift || usage
+    type "$checker" || die 'Checker %s not found\n' "$checker"
+
+    # Triple dot "git log A...B" includes commits not relevant to triple
+    # dot "git diff A...B"
+    local logrange=${diffrange/.../..}
```
xiulipan, line 40: That was a complex path for this three dot. Can we move this into travis.yml and keep the what is same here? I think put three dot as argument my need some explanation. This did not follow user experience with `git log`

xiulipan (changes requested):
Thanks for adding the support for python code scan.

tools/CI/check-gitrange.bash
marc-hb: Travis provides only a diffrange syntax (triple dots)

This script needs both the logrange syntax (double dots) and the diffrange syntax (triple dots). 

If a conversion is added in `.travis.yml` then:
- it will have to be done on every travis line
- the reverse conversion will have to be done in this script anyway.

This means many more conversions, more tedious and error-prone than just only one conversion in only one place.

xiulipan (approved):

PR #470, opened 2020-10-24
sof-kernel-log-check.sh: add (long) ignore_str guidelines

marc-hb:
This is overdue.


marc-hb:
Thanks @kv2019i for mentioning the (much more complex) https://gitlab.freedesktop.org/gfx-ci/cibuglog error filtering system

xiulipan:
@marc-hb Nice document.
For the https://gitlab.freedesktop.org/gfx-ci/cibuglog maybe we can create a new issue to check if we can reuse it.

xiulipan (approved):

PR #474, opened 2020-10-26
lib.sh: revise card number acquisition code

aiChaoSONG:
Auxiliary devices in SOF feature is introduced in
thesofproject/linux#2501, which adds a new sof-probes
card in /proc/asound/cards. Old SOF card number
acquisition cannot work any more due to this change.
    
This patch excludes the sof-probes card with a grep.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -35,7 +35,7 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
-    SOFCARD=$(grep 'sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
+    SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{print $1;}')
 fi
 
 func_lib_setup_kernel_last_line()


aiChaoSONG:
@ranj063 Please add your comment here, will the name 'sof-probes' change later?

case-lib/lib.sh
```diff
@@ -35,7 +35,7 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
-    SOFCARD=$(grep 'sof-[a-z]' /proc/asound/cards|awk '{print $1;}')
+    SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{print $1;}')
```
ranj063, line 38: @aiChaoSONG im my opinion this should be handled slightly differently. We should run sof-test for all sof-* cards. This is coming soon too after the auxiliary bus patch is accepted. And anyway with sof-probes, there will really be nothing to do (just yet) becaue it doesnt add any pcm devices, only compress devices.

aiChaoSONG: We cannot acquire the PCM card mapping from TPLG, right? @xiulipan We need to carefully deal with this change.

xiulipan: @aiChaoSONG @ranj063 Yes, I would say this patch is only a simple workaroud for the PR2501.
We have some coming change that will spilt existing PCMs to different card. Let's have some sync up meeting first to decided what should be done for the update.

xiulipan:
@aiChaoSONG I am OK with this workaround, please also add some inline comments and update the commit to clarify this is a workaround.

aiChaoSONG:
@xiulipan Done, and tested on Helios, works fine.

marc-hb (commented):
```shell
# SOFCARD is the _first_ sof card found
SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards |
                             awk '/sof-[a-z]/ {print $1; exit; }')
```

xiulipan (approved):

PR #475, opened 2020-10-27
test-case: add quotes to function arguments for func_export_pipeline

aiChaoSONG:
If a shell function takes a variable as its argument,
and the variable is empty, then the argument will
degrade to nothing in the function.

This patch adds quotes to function arguments to avoid
argument degrading.

### DIFF
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -38,7 +38,7 @@ duration=${OPT_VALUE_lst['d']}
 loop_cnt=${OPT_VALUE_lst['l']}
 
 # import only EQ pipeline from topology
-func_pipeline_export $tplg "eq:any"
+func_pipeline_export "$tplg" "eq:any"
 sofcard=${SOFCARD:-0}
 
 # Test equalizer
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -61,7 +61,7 @@ file_prefix=${OPT_VALUE_lst['f']}
 
 func_lib_setup_kernel_last_line
 func_lib_check_sudo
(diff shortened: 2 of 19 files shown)


marc-hb (changes requested):
Totally agree with the change except it is way too big. For instance is possible some function calls rely sometimes on white space inside the argument, which is bad (it should use a bash array instead) but working and this would break it.

Please resubmit with only the changes you need right now and with a separate commit per function.

aiChaoSONG:
@marc-hb Updated, I only need the change for `func_pipeine_export`

xiulipan (dismissed):

marc-hb:
Looks good except you kept the older and now misleading commit subject.

aiChaoSONG:
@xiulipan Please help to approve this PR again, as I change the misleading commit subject as Marc suggested. Once this get merged, I will test https://github.com/thesofproject/sof-test/pull/430, if the result is good, it can be merged too.

marc-hb (changes requested):
>  No functional change, only add quotes.

It is a big functional change when the `$tplg` argument is empty or not defined.

marc-hb (dismissed):

marc-hb:
Travis builds delayed everywhere because of the required migration to travis.com https://travis-ci.community/t/builds-hang-in-queued-state/10250  shellcheck would not have been very helpful in this PR because of the number of files changed. I reviewed every line carefully.

The only failure in https://sof-ci.01.org/softestpr/PR475/build346/devicetest/ is a RTC wakeup failure on just one system, this can't be related.

Merging this to unblock #430

marc-hb (approved):

PR #480, opened 2020-11-02
change all line num with timestamp and use journalctl in kernel-log-check

xiulipan:
Remove all line num related with /var/log/kern.log
use journalctl to replace the kernel-log-check default kernel log reader.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -20,9 +20,9 @@ declare -A TPLG_IGNORE_LST
 TPLG_IGNORE_LST['pcm']='HDA Digital'
 
 # Will be set by the lib function, don't need to set
-# Catches the last line of /var/log/kern.log, which will be used by
-#   sof-kernel-log-check.
-# KERNEL_LAST_LINE
+# Catches the timestamp for a test case start used by
+# sof-kernel-log-check with journalctl
+# KERNEL_LAST_TIMESTAMP
 
 # If not set will be automatically set by logging_ctl function
 # Test case log root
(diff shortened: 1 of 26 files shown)


xiulipan:
@marc-hb Looks like the first part did not get any regression. All issue is known issue.

marc-hb (commented):
Can you please submit a first PR with _only_ the name changes (search and replace) and absolutely no functional change? We can review this first PR very quickly.

Yes, it will make the new names inconsistent with the old code but only temporarily. In return, it will make the next, functional changes much smaller, much easier to review, much easier to test and much easier to revert if something goes wrong.

xiulipan:
@marc-hb I do not see the need to spilt this PR again. As the name only change did not even need review. The only difference will be name change done by search and replace.

marc-hb:
> I do not see the need to spilt this PR again.

Read more slowly.

> As the name only change did not even need review. 

Correct, it's the next PR that will be much easier to handle

xiulipan:
@marc-hb I do not like the proposal here to have commit with only name change, it did not follow the patch style.
You can try with file filter to review the function very easily.

marc-hb:
>  I do not like the proposal here to have commit with only name change, it did not follow the patch style.

This part of the code is probably the most critical in the entire framework. We have a long track record of missing ("green") failures. We have a long track record of unreliable CI. So "patch style" is secondary for this particular work.

> You can try with file filter to review the function very easily.

Anyone should be able to review these changes easily anywhere at any time even in the future  without any special instructions like "please look only at this file".

This is not just about the code review, this is also about easily reverting  without conflict if needed for testing purposes etc.

https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

PR #481, opened 2020-11-02
test-case: check the wov in D0i3

keqiaozhang:
add a new test case to check wov when DSP is in D0i3 state.


test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+fmts=$(func_pipeline_parse_value 0 fmts)
+
+test_dir="/tmp"
+def_blob="$test_dir"/detertor_default_config.blob
+
+# get wov kcontrol ID
+wov_kctl_id=$(amixer controls |grep "Detector" |awk -F= '{print $2}' |cut -d"," -f1)
+
+# get wov PGA volume kcontrol name
+wov_kctl_pga=$(amixer controls |grep "KWD Capture Volume" |awk -F= '{print $4}')
```
plbossart, line 79: why can't we use 'amixer name="KWD Capture Volume"' commands. This looks like a hack really.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+
+# adjust wov PGA volume
+amixer cset name="$wov_kctl_pga" 40
```
plbossart, line 82: why? We should never change UCM values. 
Please stop using 'mixer' scripts.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+        dlogi "default blob:"
+        cat "$def_blob"
+    fi
+}
+
+# get default preamble time and history depth
+_get_default_pt_hd()
+{
+    def_pt=$(awk -F "," '{print $4}' < "$def_blob")
+    def_hd=$(awk -F "," '{print $7}' < "$def_blob")
```
plbossart, line 98: What is this supposed to do and how do you expect anyone to maintain this?

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+    fi
+}
+
+# update the blob
+_update_blob(){
+    new_blob=$test_dir"/new_blob"
+    [ "$preamble_time" -ge "$history_depth" ] || {
+        die "Warning: invalid arguments, preamble_time must be greater than or equal to history_depth"
+    }
+    awk -F, -v OFS=, '{if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; $7='"$history_depth"'}1' \
```
plbossart, line 114: What? This is just ridiculously complicated for no good reason.

plbossart (changes requested):
Can we dial-down the complexity a notch please? 
Scripts are supposed to be self-explanatory, this is way too convoluted and with hard-coded assumptions.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+preamble_time=${OPT_VALUE_lst['p']}
+duration=${OPT_VALUE_lst['d']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+func_pipeline_export "$tplg" "kpbm:any"
+func_lib_setup_kernel_last_line
+
+if test "$PIPELINE_COUNT" != "1"; then
+    die "detected $PIPELINE_COUNT wov pipeline(s) from topology, but 1 is needed"
```
marc-hb, line 63: Once again, the negation is not necessary:
```shell
test "$PIPELINE_COUNT" = "1" || 
     die ...
```
This reads like plain English and looks like `assert(PIPELINE_COUNT == 1)` in C which is exactly what it does.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+channel=$(func_pipeline_parse_value 0 ch_max)
+rate=$(func_pipeline_parse_value 0 rate)
+dev=$(func_pipeline_parse_value 0 dev)
+fmts=$(func_pipeline_parse_value 0 fmts)
+
+test_dir="/tmp"
+def_blob="$test_dir"/detertor_default_config.blob
+
+# get wov kcontrol ID
+wov_kctl_id=$(amixer controls |grep "Detector" |awk -F= '{print $2}' |cut -d"," -f1)
```
marc-hb, line 76: `awk` has `grep`, `cut` and other features built-in, please use them and use a single tool instead of jumping around too many of them. Example:
```shell
# get the FIRST kcontrol ID, ignore any other one after that
amixer controls |
  awk -F, '/Detector/ { split($1, numid_val, "="); print numid_val[2]; exit; }'
```
Splitting with the `=` equal sign _before_ the comma `,` is very confusing because it doesn't follow the actual structure of the output. Don't do that, split with the comma first.

Code readability relies mostly on good names for intermediate variables (of functions) You must spend a lot of of time choosing these names because good English words  make the code readable even for people who don't know the language.

test-case/check-keyword-detection-in-D0i3.sh
marc-hb: > why can't we use `amixer name="KWD Capture Volume"` commands.

I've been looking for something like this but this one doesn't work. Didn't find anything in the man page.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+        sof-ctl -Dhw:0 -n "$wov_kctl_id" -r -s "$def_blob"
+    fi
+}
+
+# update the blob
+_update_blob(){
+    new_blob=$test_dir"/new_blob"
+    [ "$preamble_time" -ge "$history_depth" ] || {
+        die "Warning: invalid arguments, preamble_time must be greater than or equal to history_depth"
+    }
```
marc-hb, line 113: Nit: just like in C, you don't need the `{ }`

test-case/check-keyword-detection-in-D0i3.sh
marc-hb: This deserves at least a comment.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+    awk -F, -v OFS=, '{if ( $4 == '"$def_pt"' && $7 == '"$def_hd"' ) $4='"$preamble_time"'; $7='"$history_depth"'}1' \
+        "$def_blob" > "$new_blob"
+    dlogi "kwd config blob is updated to:"
+    cat "$new_blob"
+    dlogi "write back the new kwd config blob"
+    sof-ctl -Dhw:0 -n "$wov_kctl_id" -r -s "$new_blob" >> /dev/null || die "Failed to write back the new kwd config bolb"
```
marc-hb, line 119: Nit: `> /dev/null`  is enough and the common practice.

You can split such long lines very easily, the shall auto escapes new lines after `||`
```shell
sof-ctl -Dhw:0 -n "$wov_kctl_id" -r -s "$new_blob" >> /dev/null ||
     die "Failed to write back the new kwd config bolb"
```
How about packing some more information in the error message?
```shell
die "Failed to write back $new_blob ..."
```

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+}
+
+for i in $(seq 1 "$loop_cnt")
+do
+    for fmt in $fmts
+    do
+        if [ "$fmt" == "S24_LE" ]; then
+            dlogi "S24_LE is not supported, skip to test this format"
```
marc-hb, line 127: > dlogi "S24_LE is not supported, skip to test this format"

English: "S24_LE is not supported, skipping this format"

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+        wavetool.py -g wov -A 0.25 1. -D 4. 6. -B "${fmt%_*}" -o "$test_file"
+
+        # run arecord on keyword detection pipeline first
+        dlogc "arecord -D$dev -M -N -r $rate -c $channel -f $fmt --buffer-size=$buffer_size -d $duration $recorded_file"
+        arecord -D"$dev" -M -N -r "$rate" -c "$channel" -f "$fmt" --buffer-size="$buffer_size" -d "$duration" \
+		"$recorded_file" -vvv 2>/dev/null & WOVPID=$! 
+        
+	sleep 6 # wait for DSP enter to D0i3
+	# check D0i3 status
+	sudo hexdump /sys/kernel/debug/sof/hda |grep "0001040" |awk -F ' ' '{print $7}' |grep "04" || die "DSP failed to enter D0i3"
```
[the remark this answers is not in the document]
marc-hb: Indeed, magic numbers most definitely deserves some comment(s) at the very least.

`awk` has built-in features, use them. See above.

I don't understand what the second grep achieves considering you already grepped for 0001040. Explain.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+
+        # play the test wav to trigger kwd
+        dlogc "aplay -Dplughw:1,0 $test_file"
+        aplay -Dplughw:1,0 "$test_file"  || die "Failed to play the test wav on USB sound card"
+        dlogi "Testing: iteration $i with $fmt format"
+
+        timeout 5 tail --pid="$WOVPID" -f /dev/null || die "Failed to stop arecord on kwd pipeline"
```
marc-hb, line 152: We already discussed this pattern a few times:
- Don't use `tail --pid` to wait for a $WOVPID, use `wait $WOVPID`. 
- Use the timeout on the arecord command itself, not on the wait.
- Also use the `--kill-after=` long option, it's safer and more English makes the code more readable.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+        # analyze the recorded wav file
+        wavetool.py -a "wov" -R "$recorded_file" || {
+            # upload the failed wav file
+            find /tmp -maxdepth 1 -type f -name "wov_*_record.wav" -size +0 -exec cp {} "$LOG_ROOT/" \;
+            exit 1
```
marc-hb, line 157: This looks like a copy/paste/diverge of some other test... is it? If it is then see if you can make it a re-used function.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+# get wov kcontrol ID
+wov_kctl_id=$(amixer controls |grep "Detector" |awk -F= '{print $2}' |cut -d"," -f1)
+
+# get wov PGA volume kcontrol name
+wov_kctl_pga=$(amixer controls |grep "KWD Capture Volume" |awk -F= '{print $4}')
+
+# adjust wov PGA volume
+amixer cset name="$wov_kctl_pga" 40
+
+# store the default config blob
```
marc-hb, line 84: It's not clear to me what "store" means here. How about "extract to filesystem" maybe? Sorry I'm not familiar with WoV.

marc-hb (changes requested):

test-case/check-keyword-detection-in-D0i3.sh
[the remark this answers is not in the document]
keqiaozhang: This register is defined in the spec. I will drop you an email to explain this.

test-case/check-keyword-detection-in-D0i3.sh
```diff
@@ -0,0 +1,164 @@
...
+tplg=${OPT_VALUE_lst['t']}
+loop_cnt=${OPT_VALUE_lst['l']}
+buffer_size=${OPT_VALUE_lst['b']}
+history_depth=${OPT_VALUE_lst['s']}
+preamble_time=${OPT_VALUE_lst['p']}
+duration=${OPT_VALUE_lst['d']}
+
+[[ ${OPT_VALUE_lst['s']} -eq 1 ]] && func_lib_start_log_collect
+
+func_pipeline_export "$tplg" "kpbm:any"
```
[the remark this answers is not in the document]
keqiaozhang: Keywords used to filter out any kwd pipelines.

test-case/check-keyword-detection-in-D0i3.sh
keqiaozhang: Yes, I prefer to use UCM instead of amixer, but UCM is not ready yet on TGL/JSL RVP HDA platforms.  For the volume change, we synced this with @singalsu  before, the volume levels can impact the THD+N. In order to make the test results more stable, we need to tune this PGA volume properly.

test-case/check-keyword-detection-in-D0i3.sh
keqiaozhang: Restore may be more accurate. we need to restore the wov config blob to default value after the test. so it won't impact the next run of test.

test-case/check-keyword-detection-in-D0i3.sh
plbossart: @keqiaozhang You don't want to modify volumes as part of a test script, unless you are testing the volume change itself and restoring the configurations afterwards.
If you modify volumes, you do it outside of this script. This is confusing platform setup and test configuration. This is wrong on multiple counts.

test-case/check-keyword-detection-in-D0i3.sh
plbossart: again it's not right for a test script to alter the test target configuration. You are confusing layers. Volume settings should not be part of this script.

xiulipan:
@keqiaozhang Ping for update or close.

plbossart:
@keqiaozhang it's rather clear to me this PR will never be updated or used, can we close?

Issue #482, opened 2020-11-02
Filter out irrelevant USB 3.1 errors


plbossart:
Found on one test for suspend/resume on TigerLake device

````
2020-11-02 03:47:25 UTC [ERROR] Caught dmesg error
===========================>>
[   28.276757] usb 3-10: device descriptor read/64, error -71
[   28.495719] usb 3-10: device descriptor read/64, error -71
[   28.825732] usb 3-10: device descriptor read/64, error -71
[   29.047728] usb 3-10: device descriptor read/64, error -71
[   30.189618] usb 3-10: device not accepting address 13, error -71
[   30.717629] usb 3-10: device not accepting address 14, error -71
<<===========================
````

we need to filter this out.

aiChaoSONG:
close for the fix $483 merged.

RuiqingX:
@aiChaoSONG This issue found on TGLU_ANDR_SDCA device again.

https://sof-ci.sh.intel.com/#/result/planresultdetail/789?model=TGLU_ANDR_SDCA&testcase=check-suspend-resume-50

plbossart:
the filter hasn't changed and worked before, so something else must have changed? @aiChaoSONG @marc-hb

marc-hb:
> the filter hasn't changed and worked before

Has it worked before really? Sorry for being sceptical but I understand these errors are intermittent, which means it's very easy to get the filter wrong and not notice anything. I recommended multiple times to locally test every `ignore_str` change with `fake_kern_error()` but each of these recommendations has been followed by suspicious silence.

For instance fix 3e9df9e16dd4d9 shows that the previous version of the filter was clearly never tested. The commit message mentions "markdown" (?!) which suggests the code change wasn't even written on a real device.

`fake_kern_error()` is also required to test future variations of the error message (e.g. different USB addresses) which are impossible to test with real errors only.

Code that is not tested is code that does not work.

aiChaoSONG:
@plbossart Problem of the previous regular expression is in the `.`, it can only match one character.

marc-hb:
Closing in bug scrub because no one saw this recently and @aiChaoSONG thinks the regex has been fixed. Please re-open if you see this again.

PR #483, opened 2020-11-02
sof-kernel-log-check: filter out USB 3.1 suspend/resume issues

plbossart:
This is unrelated to SOF, add to filters.

BugLink: https://github.com/thesofproject/sof-test/issues/482


marc-hb (commented):
If this is really specific to TGL (any idea why?) can you please guard with:
```shell
case "$platform" in
  tgl_lp_something) # https://github.com/thesofproject/sof/issues/3491
  ignore_str=...
  ;;
esac
```
See why at the top of this file.

plbossart:
> If this is really specific to TGL (any idea why?) can you please guard with:

I don't think it's TGL-specific. it's rather the peripheral that's problematic. same as some hubs before, we should not confuse host and device for USB issues. We don't care about device issues.

marc-hb (approved):

PR #486, opened 2020-11-04
wavetool: enhance sinusoid generation 

aiChaoSONG:
The tool now support to generate different sine wave to different channel in amplitude, frequency and phase.

- generate wave with different frequency in different channel: `wavetool.py -g sinusoid -F 500 1000 1500 2000 -C 4`
- generate wave with different frequency and amplitude in different channel: `wavetool.py -g sinusoid -A 0.4 0.6 0.8 1.0 -F 500 1000 1500 2000 -C 4`

Screenshot from 2020-11-04 17-01-01

### DIFF
--- a/test-case/check-smart-amplifier.sh
+++ b/test-case/check-smart-amplifier.sh
@@ -97,7 +97,7 @@ do
         tmp_dir="/tmp"
         file="$tmp_dir/smart_amp_test_${fmt%_*}.wav"
         recorded_file="$tmp_dir/smart_amp_recorded_${fmt%_*}.wav"
-        wavetool.py -gsine -A0.8 -B"${fmt%_*}" -o"$file"
+        wavetool.py -gsinusoid -A0.8 -B"${fmt%_*}" -o"$file"
         dlogc "aplay -D$pb_dev -r $pb_rate -c $pb_chan -f $fmt -d $duration -v -q $file &"
         aplay -D"$pb_dev" -r "$pb_rate" -c "$pb_chan" -f "$fmt" -d "$duration" -v -q "$file" &
         dlogc "arecord -D$cp_dev -r $cp_rate -c $cp_chan -f $fmt -d $duration -v -q $recorded_file"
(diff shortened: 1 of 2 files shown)


marc-hb:
There's only one shellcheck warning (lack of quotes, the usual) and it seems easy to fix, can you fix? In a new commit if needed.

marc-hb:
There are a few `pylint --disable=C` warnings, please take a look at them to be on the safe side:
```
  ----- pylint --disable=C tools/wavetool.py ----

************* Module wavetool
tools/wavetool.py:175:0: W0301: Unnecessary semicolon (unnecessary-semicolon)
tools/wavetool.py:81:12: E1136: Value 'data.shape' is unsubscriptable (unsubscriptable-object)
tools/wavetool.py:81:48: E1136: Value 'data.shape' is unsubscriptable (unsubscriptable-object)
tools/wavetool.py:253:4: W0601: Global variable 'cmd' undefined at the module level (global-variable-undefined)
tools/wavetool.py:13:0: W0611: Unused import os (unused-import)
```

https://travis-ci.org/github/thesofproject/sof-test/jobs/741339555 is not running `pylint`, not sure why not. I fetched this PR and ran the exact same command than Travis ran and I got the output above.

aiChaoSONG:
@marc-hb shellcheck and pylint warning fixed. 
On pylint not run: I think pylint already ran, but there is no pylint warnings, this may be because travis environment is using python2-pylint. After checking the travis configuration, seems we are using Ubuntu Bionic, Python2 is still the default version for Bionic. Upgrade to Focal?

marc-hb:
> https://travis-ci.org/github/thesofproject/sof-test/jobs/741339555 did not run pylint, not sure why not.

Fixed in #493 , please review. It's just removing two spaces :-)

> Upgrade to Focal?

Why not but not required for this.

xiulipan (approved):

marc-hb:
@xiulipan I don't think we should merge non-trivial changes with only one approval _and_ less than 24h in review. I don't have any simple rule to recommend but this seems really too quick.

marc-hb:
... especially too quick considering how unstable and unreliable our CI system is in general. It's very rare to have pull requests without any issue unrelated to the pull request! So we should go slow and focus on getting some stabilization first. Then we may go faster (but still allow longer than 24h for  reviews)

PR #488, opened 2020-11-04
tools: add nhlt-parse.py tool

kv2019i:
Add a simple tool to parse ACPI NHLT table. This only parses
relevant information for SOF PDM dmic detection.

Example usage:

 cat /sys/firmware/acpi/tables/NHLT >nhtl.bin
 ./nhtl-parse.py nhtl.bin

Implemented according to Intel® Smart Sound Technology NHLT Specification
version 0.8.1:
https://01.org/sites/default/files/595976_intel_sst_nhlt.pdf

### DIFF
--- a/tools/nhlt-parse.py
+++ b/tools/nhlt-parse.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+#
+# usage: nhlt-parse.py <nhtl-dump>
+#
+# examples:
+#   cat /sys/firmware/acpi/tables/NHLT >nhtl.bin
+#   ./nhtl-parse.py nhtl.bin
+#
+# references:
+#  - https://01.org/sites/default/files/595976_intel_sst_nhlt.pdf
+
+from collections import namedtuple
+import struct
+import sys
+
+if len(sys.argv) == 1:
+    infile = '/sys/firmware/acpi/tables/NHLT'
+else:
+    infile = sys.argv[1]
+
+f = open(infile, 'rb')
+
+record = f.read()
+
+Link = namedtuple('Link', 'DescriptorLength LinkType InstanceID VendorID DeviceID RevID SubsystemID DevType Direction VirtualBusId')
+
+link_types = { 0:'HDA', 1:'DSP', 2:'PDM', 3:'SSP', 4:'Reserved4', 5:'SoundWire' }
+
+# skip ACPI table header
+p = 36
+n_endpoints = struct.unpack('<B', record[p:p+1])
+
+p = p+1
+
+print('NHLT %u endpoints' % n_endpoints)
+
+dmic_count = 0
+
+for i in range(0,n_endpoints[0]):
+    # ENDPOINT_DESCRIPTOR
+    ffmt = '<IBBHHHIBBB'
+    fsize = struct.calcsize(ffmt)
+    endpoint_descriptor = record[p:p+fsize]
+    link = Link._make(struct.unpack(ffmt, endpoint_descriptor))
+    link_type = link._asdict()['LinkType']
+    print(link)
+    print('NHLT link type %s' % link_types[link_type])
+    if link_type == 2:
+        dmic_count += 1
+    p = p + fsize
+
+    # SPECIFIC_CONFIG
+    ffmt = '<I'
+    fsize = struct.calcsize(ffmt)
+    cap_size = struct.unpack(ffmt, record[p:p+fsize])
+    print('NHLT SPECIFIC_CONFIG size %u' % cap_size)
+    p = p + cap_size[0] + fsize
+
+    # FORMATS_CONFIG
+    ffmt = '<B'
+    fsize = struct.calcsize(ffmt)
+    n_formats = struct.unpack(ffmt, record[p:p+fsize])
+    print('NHLT FORMAT CONFIGS %u' % n_formats)
+    p = p + cap_size[0] + fsize
+
+print('NHTL: %d dmics described' % dmic_count)


marc-hb (changes requested):
Oooohhhh, yet another parser to maintain in sof-test :-( Isn't there some more "popular" git repo this could be maintained in? alsa-utils maybe? Eventually?

Could you add one sample output in the commit message? This would be much easier to review because it's not easy to infer it from the code :-) Or even better: can you add in some new subdirectory one sample binary input and the corresponding output? This will act as both an example and a manual test.

> cat /sys/firmware/acpi/tables/NHLT >nhtl.bin
> ./nhtl-parse.py nhtl.bin

Anything wrong with a direct `./nhtl-parse.py /sys/firmware/acpi/tables/NHLT` ?

Please use the standard `__main__` trick so it can be `import`ed as a library. Otherwise we will soon see another script pop-up that re-parses parts of the output of this script :-D
This is also great for interactive debug and development BTW.

There are a good few `pylint` codestyle warnings that will show up eventually in
https://travis-ci.org/github/thesofproject/sof-test/builds/741373473

(Travis' capacity was reduced since the migration https://travis-ci.community/t/builds-hang-in-queued-state/10250)

marc-hb:
> There are a good few pylint codestyle warnings that will show up eventually in
https://travis-ci.org/github/thesofproject/sof-test/builds/741373473

After 5 hours of wait, Travis corrected me the minute I posted this comment: there is no pylint warning at this link. But only because I configured it in a lenient (`--disable=C`) mode for now considering the not so good scripts we already have in the repo. Please give pylint a try locally while not disabling anything.

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2020 Intel Corporation. All rights reserved.
+#
+# usage: nhlt-parse.py <nhtl-dump>
+#
+# examples:
+#   cat /sys/firmware/acpi/tables/NHLT >nhtl.bin
+#   ./nhtl-parse.py nhtl.bin
```
xiulipan, line 9: can we directly read the /sys/firmware/acpi/tables/NHLT here?

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
...
+# references:
+#  - https://01.org/sites/default/files/595976_intel_sst_nhlt.pdf
+
+from collections import namedtuple
+import struct
+import sys
+
+if len(sys.argv) == 1:
+    infile = '/sys/firmware/acpi/tables/NHLT'
+else:
```
xiulipan, line 20: could you use some our existing arguments parsing style to add more extendable options here? like `-h` and `-o INFILE`

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
...
+    infile = sys.argv[1]
+
+f = open(infile, 'rb')
```
xiulipan, line 23: Do we need some try catch here to avoid wrong path used.

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
...
+
+record = f.read()
+
+Link = namedtuple('Link', 'DescriptorLength LinkType InstanceID VendorID DeviceID RevID SubsystemID DevType Direction VirtualBusId')
+
+link_types = { 0:'HDA', 1:'DSP', 2:'PDM', 3:'SSP', 4:'Reserved4', 5:'SoundWire' }
+
+# skip ACPI table header
+p = 36
```
xiulipan, line 32: Is this hardcode skip? can we have more generic check here? like search for some magic value

xiulipan (changes requested):

tools/nhlt-parse.py
marc-hb: `argparse` is great but a bit overkill for only one argument. Later?

tools/nhlt-parse.py
marc-hb: I don't think so, Python has very good error messages by default.

tools/nhlt-parse.py
cujomalainey: python can use absolute paths, this is fine

cujomalainey:
I do have to ask, why not parse this in the kernel and print it to plain text in the debugfs

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
...
+    p = p + cap_size[0] + fsize
+
+    # FORMATS_CONFIG
+    ffmt = '<B'
+    fsize = struct.calcsize(ffmt)
+    n_formats = struct.unpack(ffmt, record[p:p+fsize])
+    print('NHLT FORMAT CONFIGS %u' % n_formats)
+    p = p + cap_size[0] + fsize
+
+print('NHTL: %d dmics described' % dmic_count)
```
ranj063, line 68: typo NHLT

kv2019i: @ranj063 Argghg, in the first version I had a 50/50% split between NHTL and NHLT. :)

plbossart:
> I do have to ask, why not parse this in the kernel and print it to plain text in the debugfs

we do parse this already in the kernel, that's how we determine the number of dmics (for non-Chrome devices). 
But that's a standalone tool that helps find out this number based on what users report. In the bug report, we received the table as is. Think of it as a diagnosis tool, not a companion to the kernel at all.

tools/nhlt-parse.py
plbossart: I keep doing the same typo as well, I don't know why this sequence of letters is so typo-prone

kv2019i:
@marc-hb wrote:
> Oooohhhh, yet another parser to maintain in sof-test :-( Isn't there some more "popular" git repo this could be maintained in? alsa-utils maybe? Eventually?

That's a very valid question. I mean this is exactly why I ended up uploading this to sof-test. I was contemplating whether I just keep this to myself. But. but, why not share. But where to put. Hmm, let's look at sof-test/tools . Ok, there's lots of scripts here, so let's just add here.

But you do have a point -- merging will just invite more similar scripts to be added. If this is not used by SOF tests, it shouldn't be in the tools, that is simpler.

I'll find a new home for this.

plbossart:
we should ask @tiwai and @perexg for advice, maybe this script could go in one of the ALSA repo (alsa-utils?).

ranj063:
Thanks for this PR @kv2019i. it finally made me read the NHLT specification which I've been meaning to for a while now!

tools/nhlt-parse.py
```diff
@@ -0,0 +1,68 @@
...
+    # ENDPOINT_DESCRIPTOR
+    ffmt = '<IBBHHHIBBB'
+    fsize = struct.calcsize(ffmt)
+    endpoint_descriptor = record[p:p+fsize]
+    link = Link._make(struct.unpack(ffmt, endpoint_descriptor))
+    link_type = link._asdict()['LinkType']
+    print(link)
+    print('NHLT link type %s' % link_types[link_type])
+    if link_type == 2:
+        dmic_count += 1
```
plbossart, line 51: this doesn't seem to work with the attachment in https://bugzilla.kernel.org/attachment.cgi?id=293599

We should be getting an invalid config and we get

````
NHLT 1 endpoints
Link(DescriptorLength=3054, LinkType=2, InstanceID=0, VendorID=32902, DeviceID=44576, RevID=4332, SubsystemID=270864439, DevType=1, Direction=1, VirtualBusId=0)
NHLT link type PDM
NHLT SPECIFIC_CONFIG size 2
NHLT FORMAT CONFIGS 1
NHTL: 1 dmics described
```

xiulipan:
@kv2019i Ping for update or close.

kv2019i:
I'll close this for now. The firmware team is planning to release a full-blown decoder that work son Linux, so let's wait for that. I'll keep the python script in my Github tree.

PR #490, opened 2020-11-04
lib.sh: make func_lib_lsof_error_dump() compatible with set -e

marc-hb:
See #312 for more details.

This bug was neither painting failures green nor successes red but it
was at least interrupting the tests before manual error messages could
get printed.


aiChaoSONG (approved):
LGTM

PR #492, opened 2020-11-05
[TEST IGNORE] Debug pylint

marc-hb:


marc-hb:
Looking at https://travis-ci.org/github/thesofproject/sof-test/jobs/741523416 identifies `tools/wavetool.py` as text/plain. It runs Ubuntu 2018 "bionic".

Issue #374, opened 2020-09-14
i915 0000:00:02.0: [drm] ERROR TC cold [un]block failed in check-suspend-resume.sh 


RanderWang:
suspend-resume test was failed by these error

```
[ 7854.607440] i915 0000:00:02.0: [drm] *ERROR* TC cold unblock failed
[ 7854.616848] i915 0000:00:02.0: [drm] *ERROR* TC cold block failed
[ 7854.616910] i915 0000:00:02.0: [drm] *ERROR* TC cold unblock failed
[ 7854.616926] i915 0000:00:02.0: [drm] *ERROR* TC cold block failed
[ 7854.616996] i915 0000:00:02.0: [drm] *ERROR* TC cold unblock failed
[ 7854.617011] i915 0000:00:02.0: [drm] *ERROR* TC cold block failed
[ 7854.617097] i915 0000:00:02.0: [drm] *ERROR* TC cold unblock failed
[ 7854.617116] i915 0000:00:02.0: [drm] *ERROR* TC cold block failed
[ 7854.617175] i915 0000:00:02.0: [drm] *ERROR* TC cold unblock failed
[ 7854.631418] atkbd serio0: Failed to deactivate keyboard on isa0060/serio0
[ 7855.007595] atkbd serio0: Failed to enable keyboard on isa0060/serio0
[ 7855.020982] sof-audio-pci 0000:00:1f.3: ipc rx: 0x90020000: GLB_TRACE_MSG
[ 7855.020999] sof-audio-pci 0000:00:1f.3: ipc rx done: 0x90020000: GLB_TRACE_MSG

```

marc-hb:
For now this bug has zero configuration and version information in the description, neither about hardware, firmware or software. Please add some.

There are three different and likely unrelated parts in this log.

The `sof-audio-pci` messages seem harmless and hopefully unrelated.

The `atkbd serio0` can hopefully be ignored. Can you try disabling legacy mouse/legacy keyboard support in the BIOS and see if it goes away?

The Type-C errors are more worrying and could affect test results if they rely on some USB or Display+audio device. Can you confirm there is no such dependency? They should also be reported to the i915 team with of course the exact kernel version.

Hopefully none of these errors leave the kernel in a bad state.

marc-hb:
https://cgit.freedesktop.org/drm/drm/log/?h=05e31dd78e26 looks relevant:
```
[...]

PR #496, opened 2020-11-06
sof-kernel-log-check:ignore i915 false failure(update)

RuiqingX:
Fixes: #374

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -186,8 +186,8 @@ ignore_str="$ignore_str"'|acpi device:00: Failed to change power state to D3hot'
 
 # I915, issues reported by sof-test
 # BugLink: https://github.com/thesofproject/sof-test/issues/374
-ignore_str="$ignore_str"'|i915 0000:00:02\.0: \[drm\] ERROR TC cold unblock failed'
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
I don't fully get the point here, why add a pair of raw `*`, could you please explain a little bit?

marc-hb:
Use `fake_kern_error()` in `case-lib/lib.sh` to locally test `ignore_str` regular expressions  when the real error is intermittent.

tools/sof-kernel-log-check.sh
```diff
@@ -186,8 +186,8 @@ ignore_str="$ignore_str"'|acpi device:00: Failed to change power state to D3hot'
 
 # I915, issues reported by sof-test
 # BugLink: https://github.com/thesofproject/sof-test/issues/374
```
xiulipan, line 188: please also put origin error logs here. A link it not enough

xiulipan (changes requested):
Please add more commit message, a simple fix could not tell anything. Also the commit message should be something start with fix.

marc-hb:
The code change is good. The commit message says basically nothing. 

Besides an English description of the change it should also contain "Fixes: 589920e6ee83"

aiChaoSONG (approved):
LGTM

PR #497, opened 2020-11-06
multiple-pipeline: fix set -e compatibility and add it

marc-hb:
Notably remove top-level "expr" because "expr 0" is surprisingly a
failure and expr is antiquited anyway (SC2003)

Also add a dlogi() before killing so no one mistakes the deaths of aplay
and arecord for failures.

Progress towards #312.

### DIFF
--- a/test-case/multiple-pipeline-capture.sh
+++ b/test-case/multiple-pipeline-capture.sh
@@ -22,6 +22,9 @@
 ##    all pipelines are alive and without kernel error
 ##
 
+set -e
+
+# shellcheck source=case-lib/lib.sh
 source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
 
 OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
@@ -87,8 +90,8 @@ func_run_pipeline_with_type()
         dlogc "${APP_LST[$1]} -D $dev -c $channel -r $rate -f $fmt ${DEV_LST[$1]} -q"
         "${APP_LST[$1]}" -D $dev -c $channel -r $rate -f $fmt "${DEV_LST[$1]}" -q &
 
-        tmp_count=$(expr $tmp_count - 1 )
-        [[ $tmp_count -le 0 ]] && return
+        : $((tmp_count--))
+        if [ "$tmp_count" -le 0 ]; then return 0; fi
     done
 }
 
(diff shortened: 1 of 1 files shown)


marc-hb:
The shellcheck warnings in https://travis-ci.org/github/thesofproject/sof-test/jobs/741782133 predates this PR.

The only test failure in https://sof-ci.01.org/softestpr/PR497/build367/devicetest is known bug https://github.com/thesofproject/sof/issues/3462 and  this test code change cannot cause this sort of failure, especially not on just one platform.

test-case/multiple-pipeline-capture.sh
```diff
@@ -87,8 +89,8 @@ func_run_pipeline_with_type()
         dlogc "${APP_LST[$1]} -D $dev -c $channel -r $rate -f $fmt ${DEV_LST[$1]} -q"
         "${APP_LST[$1]}" -D $dev -c $channel -r $rate -f $fmt "${DEV_LST[$1]}" -q &
 
-        tmp_count=$(expr $tmp_count - 1 )
-        [[ $tmp_count -le 0 ]] && return
+        : $((tmp_count--))
+        if [ "$tmp_count" -le 0 ]; then return 0; fi
```
xiulipan, line 93: @marc-hb what is the preferred style `[[ CONDTION  ]] && DO_SOMETHING` or `if [ CONDTION   ]; then DO_SOMETHING;fi`
I am quite confused here.

test-case/multiple-pipeline-capture.sh
```diff
@@ -147,7 +149,7 @@ do
     [[ $? -eq 1 ]] && func_error_exit "Catch the abnormal process status of aplay"
 
 
-    # kill all arecord
+    dlogi 'pkill -9 aplay arecord'
```
xiulipan, line 152: `dlogc` here maybe? dlogi should be some message, but this looks like more like exactly cmd.

xiulipan (changes requested):

test-case/multiple-pipeline-capture.sh
marc-hb: Maybe you have in mind "avoid negations" comments I made in other reviews? There's no negation here in either case, these "avoid negations" comments don't apply.

`A || B` and `if A; then B; fi` never  have any compatibility problem with `set -e` (#312)

`A && B` works too most of the time but sometimes it can cause issues with `set -e` in some  corner cases. This is one of these corner cases because the statement is last in the function.

Even when it works, `A && B` is "less obviously" compatible with `set -e` because the exit value of the list can be non-zero even when everything is fine.

I'm genuinely sorry `set -e` can be complicated but in shell this is really the best we get.

xiulipan: @marc-hb I think that was only some style preference here. But I do see issue here, the `set -e` will exit the script when `tmp_count` for example is 0, right?

marc-hb: Yes `expr 0` was returning a failure. `expr` is bad (also for other reasons), shellcheck warns about it.

`:` always succeeds.

marc-hb:
https://sof-ci.01.org/softestpr/PR497/build372/devicetest is 100% successful and no shellcheck warning in https://travis-ci.org/github/thesofproject/sof-test/builds/742370356 is new

marc-hb:
https://sof-ci.01.org/softestpr/PR503/build373/devicetest/?model=BYT_MB_NOCODEC&testcase=check-alsabat-headset-playback looks like https://github.com/thesofproject/sof/issues/3462 again. This PR doesn't change this test

test-case/multiple-pipeline-capture.sh
marc-hb: > Even when it works, A && B is "less obviously" compatible with set -e because the exit value of the list can be non-zero even when everything is fine.

Explanation at http://redsymbol.net/articles/unofficial-bash-strict-mode/#short-circuiting if you're interested

aiChaoSONG (approved):
LGTM, failures on APL and BDW is not related.

marc-hb:
This change broke the test in some configurations, tentative fix in #521

Issue #489, opened 2020-11-04
Use aplay --fatal-errors to stop on underruns or errors


plbossart:
In test scripts, we often have cases where there is no issue reported in firmware or driver logs, but nasty error messages in userspace, e.g. pcm_write error, bad descriptor, etc.

aplay provides a --fatal-errors command line option, and when set will stop processing and return an error code.

I would think this is really required for CI/QA tests (except maybe in rare resume cases where a .prepare operation is unavoidable). We would also need to test what aplay returns.

Rather than forcing this option in the scripts, I would prefer an environment variable, same as KCFLAGS='-Werror' when compiling

plbossart:
@marc-hb @kv2019i @aiChaoSONG @xiulipan what do you think?

marc-hb:
On one hand, anyone serious and professional about validation should love every possibility to find more errors because that's the whole purpose of testing. I didn't know this option existed, I think it should have been used in tests always.

We most likely want to combine this with `set -e` https://github.com/thesofproject/sof-test/issues/312 for it to be really useful.

On the other hand, the main purpose of Continuous Integration  is to find _new_ regressions, NOT to constantly report the same known bugs. When some PR check is red most of the time then most people stop looking at it. The only error example listed in`man aplay` and `man arecord` is "xrun" and I'm afraid (and sad) that these would be both non-deterministic and frequent enough to make entire CI results too painful to use?

> Rather than forcing this option in the scripts, I would prefer an environment variable, same as KCFLAGS='-Werror' when compiling

Something optional/flexible would be a great way to start. Unfortunately `aplay` does not support empty arguments like `aplay "$maybe_empty"` so this would have to be `aplay $maybe_empty` which implies a lot of shellcheck warnings to manually disable :-(

Is there no `~/.aplayrc` of some kind?

plbossart:
@marc-hb if we get an xrun, we should stop. period. There's no point in continuing further, something really bad just happened and ignoring it is not an option. If we don't then we push the detection of issues to the alsa-bat case, where we will indirectly detect the xrun with a bad THD level.

We *could* ignore errors to see how resilient the error handling is, but that's a different test IMHO.

We can address the issues above with aplay by having a variable instead of a hard-coded 'aplay'.

marc-hb:
> We can address the issues above with aplay by having a variable instead of a hard-coded 'aplay'.

[...]

PR #498, opened 2020-11-06
New aplay_opts() and arecord_opts() to change flags across all tests

marc-hb:
"All problems in computer science can be solved by another level of
indirection"

First step towards #489

Default `SOF_ALSA_OPTS` to `--fatal-errors`, can be overriden with:
```
 SOF_ALSA_OPTS="" ./scripts/run-all-tests.sh
```

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -253,6 +253,47 @@ func_lib_get_tplg_path()
     return 0
 }
 
+# We must not quote SOF_ALSA_OPTS and disable SC2086 below for two reasons:
+#
+# 1. We want to support multiple parameters in a single variable, in
+#    other words we want to support this:
+#
+#      SOF_ALSA_OPTS="--foo --bar"
+#      aplay $SOF_ALSA_OPTS ...
+#
+# 2. aplay does not ignore empty arguments anyway, in other words this
+#    does not work anyway:
+#
+#      SOF_ALSA_OPTS=""
+#      aplay "$SOF_ALSA_OPTS" ...
+#
+# This is technically incorrect because it means our SOF_ALSA_OPTS
+# cannot support whitespace, for instance this would be split in two
+# options: --bar="aaaa and bbbb"
+#
+#      SOF_ALSA_OPTS='--foo --bar="aaaa bbbb"'
+#
+# To do this "correctly" SOF_ALSA_OPTS etc. should be arrays.
+# From https://mywiki.wooledge.org/BashGuide/Arrays "The only safe way
+# to represent multiple string elements in Bash is through the use of
+# arrays."
+#
+# However, 1. arrays would complicate the user interface 2. ALSA does not
+# seem to need arguments with whitespace or globbing characters.
+
+aplay_opts()
+{
+    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
+    # shellcheck disable=SC2086
+    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+}
+arecord_opts()
+{
+    dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+    # shellcheck disable=SC2086
+    arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+}
+
 die()
 {
     dloge "$@"
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -95,8 +95,7 @@ do
                     dlogi "using $file as capture output"
                 fi
 
-                dlogc "arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
-                arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q
+                arecord_opts -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q
                 if [[ $? -ne 0 ]]; then
                     func_lib_lsof_error_dump $snd
                     die "arecord on PCM $dev failed at $i/$loop_cnt."
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ To run a test, call the scripts directly
 
 Example:
 ```
-~/sof-test/test-case$ ./check-playback.sh -h
+~/sof-test/test-case$ SOF_ALSA_OPTS='-q --fatal-errors' ./check-playback.sh -h
 Usage: ./check-playback.sh [OPTION]
 
     -F |  --fmts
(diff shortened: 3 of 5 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR498/build369/devicetest was successful.

One new quoting issue caught by shellcheck in https://travis-ci.org/github/thesofproject/sof-test/builds/741801847, my bad Ijust force-pushed a fix.

test-case/check-capture.sh
```diff
@@ -95,8 +95,7 @@ do
                     dlogi "using $file as capture output"
                 fi
 
-                dlogc "arecord -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q"
```
aiChaoSONG, line 98: The dlogc line is used to print the command we are running to logs, no reason to remove this.

xiulipan: @aiChaoSONG the warpper `arecord_opts` did the dlogc already.

aiChaoSONG: @marc-hb sorry, my bad, ignore this comment.

case-lib/lib.sh
```diff
@@ -253,6 +253,50 @@ func_lib_get_tplg_path()
...
+# To do this "correctly" SOF_ALSA_OPTS should be an array.
+# From https://mywiki.wooledge.org/BashGuide/Arrays "The only safe way
+# to represent multiple string elements in Bash is through the use of
+# arrays."
+# However 1. it would complicate the user interface 2. We don't seem to
+# need arguments with whitespace or globbing characters.
+
+# Default value, can be overriden by the environment. Empty string does
+# override.
+: "${SOF_ALSA_OPTS=--fatal-errors}"
```
xiulipan, line 285: @marc-hb maybe consider to put this into the `case-lib/config.sh` file? That config file is designed for overridable ENV values.

case-lib/lib.sh
```diff
@@ -253,6 +253,50 @@ func_lib_get_tplg_path()
...
+
+aplay_opts()
```
xiulipan, line 287: I think `opts` means options? maybe a more good name for the wrapper function

case-lib/lib.sh
```diff
@@ -253,6 +253,50 @@ func_lib_get_tplg_path()
...
+{
+    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
+    # shellcheck disable=SC2086
+    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
```
xiulipan, line 291: where will define `SOF_APLAY_OPTS`? from environment? I think this can also be added to `case-lib/config.sh`

case-lib/lib.sh
```diff
@@ -253,6 +253,50 @@ func_lib_get_tplg_path()
...
+}
+arecord_opts()
+{
+    dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+    # shellcheck disable=SC2086
+    arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
```
xiulipan, line 297: same for `SOF_ARECORD_OPTS` like `SOF_APLAY_OPTS`

xiulipan (dismissed):
Nice patch for a generic wrapper for aplay/arecord call. But I think we may need more README and a correct file for the overridable ENV

case-lib/lib.sh
plbossart, line 291: we also want to test the return status since it contains an error code when it fails. If you just stop I think the rest of the script will probably think nothing bad happened.

marc-hb: The vast majority of the time `set -e` (errexit #312) will immediately abort the script, demonstration:

```shell
$ SOF_ALSA_OPTS=--bogus ./test-case/check-playback.sh

2020-11-06 19:05:09 UTC [COMMAND] sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-apl-pcm512x.tplg -f 'type:playback & id:any' -b ' pcm:HDA Digital' -s 0 -e
2020-11-06 19:05:09 UTC [INFO] ===== Testing: (Round: 1/1) (PCM: Port5 [hw:0,0]<playback>) (Loop: 1/3) =====
2020-11-06 19:05:09 UTC [COMMAND] aplay --bogus  -Dhw:0,0 -r 48000 -c 2 -f S16_LE -d 10 /dev/zero -v -q
aplay: unrecognized option '--bogus'
Try `aplay --help' for more information.
2020-11-06 19:05:09 UTC [ERROR] aplay on PCM hw:0,0 failed at 1/3.
2020-11-06 19:05:09 UTC [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2020-11-06 19:05:09 UTC [ERROR]  die()  @  ./test-case/../case-lib/lib.sh
2020-11-06 19:05:09 UTC [ERROR]  main()  @  ./test-case/check-playback.sh:99
```
Same with `./test-case/check-capture.sh`

The other test `SOF_ALSA_OPTS=--bogus ./test-case/multiple-pipeline-capture.sh` doesn't stop immediately because `multiple-pipeline-capture.sh` backgrounds the `aplay` and `arecord` commands. Testing the error code inside this function would add complexity and not make any difference because the entire function is backgrounded.

What makes a difference with backgrounded commands is to collect ASAP their exit status with `wait $aplayPID`. This is something sof-test rarely does for now but that I systematically requested in every code review involving backgrounded command(s). `wait $aplay_optsPID` will collect the exit status of the last command in `aplay_opts()` function which is of course `aplay`. `set -e` applies to `wait` exactly the same and if desired it can be overriden with a better message (#312) the same.

Even without considering backgrounding issues, the caller of `aplay_opts` caller can generally provide more useful and relevant context if a custom error message is desired.

case-lib/lib.sh
marc-hb: I like config files but I see a number of issues with the `case-lib/config.sh` design. I'll remove the default value here and resubmit so we can postpone the potentially complex discussion about config files and focus on these wrappers first. One change  at a time!

Some issues with `case-lib/config.sh`:
- Except for two actual values, all the rest is made of no-op `AAAA=${AAAA:-}` statements. That's very misleading from a coding perspective.
- What would be more useful is a local config file that is not in git so it doesn't pollute `git diff` and `git status` (and even [`git add *`](https://github.com/thesofproject/rimage/commit/0d2af68a9e2bc) - never do that)
- I'm not even sure a single file collecting all default values is a good idea. Most software seems to have good defaults even when run with empty config files.

case-lib/lib.sh
marc-hb: I've been looking for a better but still short name but I didn't find any, suggestions welcome. It has  to be  kept relatively short because it will be used a lot.

case-lib/lib.sh
marc-hb: Environment yes because that's very convenient for  interactive testing and possibly config file too but let's postpone the config file discussion.

marc-hb (commented):
>  But I think we may need more README 

Good point I'll add something to the README.md

case-lib/lib.sh
xiulipan: I create a new issue https://github.com/thesofproject/sof-test/issues/502 for the later discussion. Nice point here

case-lib/lib.sh
xiulipan: Same feeling here. If I had a better name, I should type it already.

plbossart (approved):
I have zero understanding of the bash/shellcheck considerations below, if that detects issues and does not add false-positives that sounds good to me.

marc-hb:
@xiulipan , others I think I addressed or answered all your comments, can you please check?

This change is a lot of English to explain the direction and very little code change for the moment :-)

xiulipan (approved):
Thanks for the update, let's update the usage once we find it is needed for other test cases.

PR #499, opened 2020-11-06
lib: rename last_line to checkpoint

xiulipan:
What we need is to set a checkpoint for kernel log to let
us start check kernel. Rename last_line to a more generic
name checkpoint.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -22,7 +22,7 @@ TPLG_IGNORE_LST['pcm']='HDA Digital'
 # Will be set by the lib function, don't need to set
 # Catches the last line of /var/log/kern.log, which will be used by
 #   sof-kernel-log-check.
-# KERNEL_LAST_LINE
+# KERNEL_CHECKPOINT
 
 # If not set will be automatically set by logging_ctl function
 # Test case log root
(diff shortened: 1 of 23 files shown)


xiulipan:
@marc-hb Done, please have a look or blind approval.

xiulipan:
@marc-hb Ping

marc-hb (approved):
Thanks, the change is good and the name "checkpoint" sounds great but the commit message is a bit incomplete: It's not obvious why `sof-kernel-log-check.sh $VAR` is the same as `sof-kernel-log-check.sh "$VAR"` and the commit message should mention why they have the same effect in this particular codebase.

`sof-kernel-log-check.sh $VAR` is the same as `sof-kernel-log-check.sh "$VAR"` only because `sof-kernel-log-check.sh` starts with `${1:-1}`.

If `sof-kernel-log-check.sh` were for instance starting with `${1-1}` then `sof-kernel-log-check.sh $VAR` is the same as `sof-kernel-log-check.sh "$VAR"` would be a real functional change.

In general one cannot assume that ` $VAR` is the same as  "$VAR"`. This difference is not a purely theoretical issue, see real example in https://github.com/thesofproject/sof-test/pull/430#pullrequestreview-518224500 . Shellcheck warnings are rarely just about "codestyle" , most shellcheck warnings are about functional changes.

PR #507, opened 2020-11-10
multiple-pipeline-capture: show aplay/record processes left on fail

marc-hb:
Only on a test failure.

Sample output:
```
15248 aplay -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/zero -q
15209 arecord -D hw:0,1 -c 2 -r 48000 -f S16_LE /dev/null -q
15218 arecord -D hw:0,2 -c 2 -r 16000 -f S16_LE /dev/null -q
15228 arecord -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/null -q
```
Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #508, opened 2020-11-11
multiple-pipeline-playback.sh: show aplay/arecord processes left on fail

fredoh9:
Display PID of aplay/arecord on failure.

I was so thrilled to see https://github.com/thesofproject/sof-test/pull/507. Missed playback test need same fix.


marc-hb:
> I was so thrilled to see #507. Missed playback test need same fix.

The amount of copy/paste/diverge between the two is just unbelievable.

marc-hb (approved):

PR #509, opened 2020-11-11
use journalctl as default kernel log reader

xiulipan:
Use timestamp as checkpoint for each test case.
use journalctl to read kernel logs in sof-kernel-log-check.sh
use journalctl in other kernel read scripts.

### DIFF
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -20,13 +20,11 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
 func_opt_parse_option "$@"
 
-if ! alias | grep -q 'Sub-Test'; then
-    # hijack DMESG_LOG_START_LINE which refer dump kernel log in exit function
-    DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
-    cmd="sof-kernel-dump.sh"
-else
-    cmd="journalctl --dmesg --no-pager"
-fi
+# hijack DMESG_LOG_START_LINE to dump kernel from file start is not Sub-Test
+# TODO: clean up Sub-Test feature
+alias | grep -q 'Sub-Test' || DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh | tail -n 1 | awk '{print $1;}' )
+
+cmd="journalctl_cmd"
 
 dlogi "Checking SOF Firmware load info in kernel log"
 if $cmd | grep -q " sof-audio.*Firmware.*version"; then
@@ -36,12 +34,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
     $cmd | grep "Firmware debug build" -A3 | head -n 12
     exit 0
 else
-    printf ' ------\n  debuging with /var/log/kern.log  \n ---- \n'
-    ls -alht /var/log/kern.log
-    grep -na "Linux version" /var/log/kern.log || true
-    printf ' ------\n  cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
-            "$cmd" "$DMESG_LOG_START_LINE"
-    journalctl --dmesg --lines 50 --no-pager
-    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
+    journalctl_cmd --lines 50 || true
+    printf ' ------\n  debugging with systemd journalctl  \n ---- \n'
+    systemctl status systemd-journald* || true
     die "Cannot find the sof audio version"
 fi
--- a/tools/sof-get-default-tplg.sh
+++ b/tools/sof-get-default-tplg.sh
@@ -1,14 +1,12 @@
 #!/bin/bash
+#
+# read kernel log to get topology file loaded by SOF driver
+# Example from an apl up2 pcm512x device:
+#
+# sof-audio-pci 0000:00:0e.0: loading topology:intel/sof-tplg/sof-apl-pcm512x.tplg
+#
+# sof-apl-pcm512x.tplg will be returned
+#
 
-# Current: load TPLG file order:
-# dmesg
-# journalctl
-# /var/log/syslog (sof-kernel-dump.sh)
-# Future: possibly be loaded from elsewhere
-# Notice: Only verified on Ubuntu 18.04
-tplg_file=$(dmesg |grep -i topology|awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
-[[ ! "$tplg_file" ]] && \
-    tplg_file=$(journalctl -k |grep -i topology |awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
-[[ ! "$tplg_file" ]] && \
-    tplg_file=$(sof-kernel-dump.sh |grep -i topology|awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
+tplg_file=$(journalctl -k |grep -i topology |awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
 [[ "$tplg_file" ]] && basename "$tplg_file" || echo ""
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -126,11 +126,14 @@
 # Append some garbage to an ignore pattern to turn it off. Much easier
 # than deleting it.
 
-
-begin_line=${1:-1}
+# fallback to use dmseg when set to 0.
+# TODO: update the fallback option when drop dmesg
+begin_timestamp=${1:-0}
 declare err_str ignore_str
 
-platform=$($(dirname "$0")/sof-dump-status.py -p)
+# pwd resolves relative paths
+my_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+platform=$("$my_dir"/sof-dump-status.py -p)
 
 err_str="error|failed|timed out|panic|oops"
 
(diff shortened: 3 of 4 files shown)


marc-hb (changes requested):
I didn't spend a lot of time on every line yet but this looks OK. However and as already requested in previous PRs: please submit independent bugs fixes and cleanups in other PRs. Either in PR(s) before this switch to journalctl, or in other PRs after this switch to journalctl but just not in the same PR.

This switch is a small number of lines but with a very high impact so it needs to be integrated alone so it can very easily be tested (I will test it locally with fake_kern_error), observed  for a few days in CI and very easily reverted if needed. I'm running out of ways to repeat myself.

tools/sof-kernel-log-check.sh
```diff
@@ -127,10 +127,10 @@
 # than deleting it.
 
 
-begin_line=${1:-1}
+begin_timestamp=${1:-0}
 declare err_str ignore_str
 
-platform=$(sof-dump-status.py -p)
+platform=$($(dirname "$0")/sof-dump-status.py -p)
```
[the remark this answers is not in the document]
marc-hb: Or in two slightly more readable and typical steps:
```shell
# pwd resolves relative paths
my_dir=$(cd dirname "${BASH_SOURCE[0]}" && pwd)
...

platform=$("$my_dir"/sof-dump-status.py -p)

File: tools/sof-kernel-log-check.sh [hunk 2]
```diff
@@ -246,32 +246,21 @@ case "$platform" in
...
 
-[[ ! "$err_str" ]] && echo "Missing error keyword list" && exit 0
-# dmesg KB size buffer size
-#dmesg_config_define=$(awk -F '=' '/CONFIG_LOG_BUF_SHIFT/ {print $2;}' /boot/config-$(uname -r))
-#dmesg_buffer_size=$( echo $(( (1<<$dmesg_config_define) / 1024 )) )
-# kernel file log buffer size
-#kernel_buffer_size=$(du -k /var/log/kern.log |awk '{print $1;}')
-# now decide using which to catch the kernel log
-#[[ $kernel_buffer_size -lt $dmesg ]] && cmd="dmesg" || cmd="tail -n  +${begin_line}  /var/log/kern.log"
+# confirm begin_timestamp is in UNIX timestamp format, otherwise use dmesg
```
marc-hb, line 249: Please add a comment explaining why we still need  to support a different begin_timestamp format.

tools/sof-kernel-log-check.sh
```diff
@@ -246,32 +246,21 @@ case "$platform" in
...
+journalctlflag="-k -q --no-pager --utc --output=short-iso --no-hostname"
+[[ $begin_timestamp =~ ^[0-9]{10} ]] && cmd="journalctl $journalctlflag --since=@$begin_timestamp" || cmd="dmesg"
```
marc-hb, line 251: Why stick to `dmesg`? Its output looks very different. Why not just drop the `--since=` parameter?

marc-hb (changes requested):

tools/sof-kernel-log-check.sh
xiulipan: @marc-hb We need to keep dmesg as some user still use dmesg way. See https://github.com/thesofproject/sof-test/pull/468
commits like https://github.com/thesofproject/sof-test/pull/468/commits/8db6f8c18a81d79b082284bc452d18132e2552e2

tools/sof-kernel-log-check.sh
xiulipan: I do not think we need **a different begin_timestamp format.** What I mean here is to keep the fallback we used to have to dmesg now to cover the small migration steps.

tools/sof-kernel-log-check.sh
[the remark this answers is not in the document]
xiulipan: Will update.

tools/sof-kernel-log-check.sh
```diff
@@ -127,10 +127,10 @@
 # than deleting it.
 
 
-begin_line=${1:-1}
+begin_timestamp=${1:-0}
```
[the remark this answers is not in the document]
xiulipan: Because I drop the usage for line number now. And with my comments below
`# confirm begin_timestamp is in UNIX timestamp format, otherwise use dmesg`

We will always fallback to use demsg if no checkpoint is provided or wrong format checkpoint is provided

xiulipan:
updated and rebased @marc-hb @aiChaoSONG Ping for review.

test-case/verify-sof-firmware-load.sh
```diff
@@ -20,13 +20,11 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
...
-    DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
-    cmd="sof-kernel-dump.sh"
-else
-    cmd="journalctl --dmesg --no-pager"
-fi
+# hijack DMESG_LOG_START_LINE to dump kernel from file start is not Sub-Test
+# TODO: clean up Sub-Test feature
+alias | grep -q 'Sub-Test' || DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh | tail -n 1 | awk '{print $1;}' )
+
+cmd="journalctl_cmd"
```
marc-hb, line 27: I'm not 100% sure this has to be in the same PR than the rest.

test-case/verify-sof-firmware-load.sh
```diff
@@ -36,12 +34,8 @@ if $cmd | grep -q " sof-audio.*Firmware.*version"; then
...
-    printf ' ------\n  debuging with /var/log/kern.log  \n ---- \n'
-    ls -alht /var/log/kern.log
-    grep -na "Linux version" /var/log/kern.log || true
-    printf ' ------\n  cmd was %s, DMESG_LOG_START_LINE was %s  \n ---- \n' \
-            "$cmd" "$DMESG_LOG_START_LINE"
-    journalctl --dmesg --lines 50 --no-pager
-    journalctl --dmesg | grep -C 1 " sof-audio.*Firmware.*version" || true
+    journalctl_cmd --lines 50 || true
+    printf ' ------\n  debugging with systemd journalctl  \n ---- \n'
+    systemctl status systemd-journald* || true
```
marc-hb, line 39: This looks OK but it does really not belong to this PR. This is just debug information for one specific test, it's not related to the checkpoint code at all. This code does not need to change right now. This can be changed much later after the dust of the `sof-kernel-log-check` switch has settled.

tools/sof-get-default-tplg.sh
```diff
@@ -1,14 +1,12 @@
...
-# journalctl
-# /var/log/syslog (sof-kernel-dump.sh)
-# Future: possibly be loaded from elsewhere
-# Notice: Only verified on Ubuntu 18.04
-tplg_file=$(dmesg |grep -i topology|awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
-[[ ! "$tplg_file" ]] && \
-    tplg_file=$(journalctl -k |grep -i topology |awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
-[[ ! "$tplg_file" ]] && \
-    tplg_file=$(sof-kernel-dump.sh |grep -i topology|awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
+tplg_file=$(journalctl -k |grep -i topology |awk -F ':' '/tplg/ {print $NF;}'|tail -n 1)
```
marc-hb, line 11: Impressive simplification but I don't think it's related to the checkpointing change, is it? Can it be submitted in a later PR?

tools/sof-kernel-log-check.sh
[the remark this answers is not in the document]
marc-hb: > We will always fallback to use demsg if no checkpoint is provided or wrong format checkpoint is provided

The only user of this script is ourselves, so to reduce complexity a lot let's just fix the user instead of providing a fallback. Fail here when some test does not provide a valid checkpoint and then calls this script, which seems very wrong!? Let's fix all these broken tests and make them consistent _before_ switching the checkpoint to journalctl

tools/sof-kernel-log-check.sh
```diff
@@ -250,32 +253,22 @@ case "$platform" in
...
-if [ "$err" ]; then
-    echo "$(date -u '+%Y-%m-%d %T %Z')" "[ERROR]" "Caught dmesg error"
+[[ -z "$err" ]] || {
+    echo "$(date -u '+%Y-%m-%d %T %Z')" "[ERROR]" "Caught kernel log error"
     echo "===========================>>"
     echo "$err"
     echo "<<==========================="
     builtin exit 1
-fi
+}
```
marc-hb, line 274: Nice cleanup but independent from the checkpointing so submit this in a different PR.

git tip: if you have say 4 commits then you can rotate your commits with `git rebase -i` and then submit `HEAD~3` 4 times. This creates 4 separate PRs that can be reviewed and validated separately.

marc-hb (changes requested):
OK, now I understand that there is some "dmesg fallback" for some "legacy users". Not sure what these are but please fix the legacy users first so we don't need any fallback and can keep the code simple.

test-case/verify-sof-firmware-load.sh
xiulipan: I named this PR as `use journalctl as default kernel log reader`, so I think put all change there is OK.
I split them into different commit. If you want single commit in one PR I am OK with that.

tools/sof-kernel-log-check.sh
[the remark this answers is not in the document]
xiulipan: I thought we have an agreement in https://github.com/thesofproject/sof-test/pull/468#issuecomment-718311303
> @marc-hb Thanks for the tips, I am also think about how to spilt the patches. Please help to check my plan:
> 
> 1. replace LINE with TIMESTAMP, log check with no `--priority` but still with `err_str`
> 2. quoting clean up for the `$KERNEL_LAST_TIMESTAMP`
> 3. remove `err_str` and use `--priority`, add new `ignore_str`
> 4. clean up test case to remove duplicate kernel log check.

Now we have done half of the step **1** by changing the LINE into CHECKPOINT in https://github.com/thesofproject/sof-test/pull/499. Now I am not sure if I can do clean up in step **4** before we change to use journalctl with use `--priority`, it would bring more unrelated error in our test now.


Or we can do something more simpler, merge a big PR https://github.com/thesofproject/sof-test/pull/468 I have already made and tested and fix bugs if we find any.

Any suggestions here @marc-hb  @aiChaoSONG @plbossart @mengdonglin 
I am not sure what we are reviewing here. I do not see any feedback about codes or implementation. I only see how to spilt patches. For such big implementation change, I am not sure if small steps can make thing easier because I need to do some un-needed migration clean up for each small change. EG: @marc-hb The **fallback** here is one of those things I do not need to have in big PR. But I have to keep it here for migration.

If we do not make progress before my vacation next week. I will use my super power to merge all the change in one PR.

PR #510, opened 2020-11-11
kmod-load-reload: don't hide FW load check logs

aiChaoSONG:
The output of sub-case verify-sof-firmware-load.sh
gives hints on why test case failed, should not hide
its output.


marc-hb (approved):
Good catch, thanks.

Pretty amazing discard considering this test is entirely about module loading and unloading!

fredoh9 (approved):
Looks good. thanks

PR #511, opened 2020-11-12
func_exit_handler: print 'Test Result: FAIL!' for unknown exit codes

marc-hb:
We must always print some 'Test Result' otherwise some callers will time
out, see alsabat example in
https://sof-ci.01.org/linuxpr/PR2501/build4685/devicetest/

Fixes: #441

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -93,6 +93,8 @@ function func_exit_handler()
         exit_status=$ret
     fi
 
+    # We must always print some 'Test Result' otherwise some callers
+    # will time out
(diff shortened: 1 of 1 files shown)


marc-hb:
- https://sof-ci.01.org/softestpr/PR511/build389/devicetest/?model=BYT_MB_NOCODEC&testcase=multiple-pipeline-playback looks like https://github.com/thesofproject/sof/issues/3522

- Our cyan devices are all dying :-(

xiulipan (approved):
@marc-hb Thanks for the fix

PR #512, opened 2020-11-12
alsabat cleanups

marc-hb:
See commit messages.

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -73,7 +73,8 @@ dlogi "check the PCMs before alsabat test"
 # different PCMs may support different audio formats(like samplerate, channel-counting, etc.).
 # use plughw to do the audio format conversions. So we don't need to specify them for each PCM.
 dlogc "alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency"
-alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency &
+alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency & playPID=$!
+
 # playback may have low latency, add one second delay to aviod recording zero at beginning.
 sleep 1
 dlogc "alsabat -Cplug$pcm_c -F $frequency"
@@ -83,3 +84,4 @@ alsabat -Cplug$pcm_c -F $frequency || {
 	exit 1
 }
 
+wait $playPID
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -55,10 +55,11 @@ function __upload_wav_file
     # upload the alsabat wav file
     for file in /tmp/bat.wav.*
     do
-	    size=`ls -l $file | awk '{print $5}'`
-	    if [[ $size -gt 0 ]]; then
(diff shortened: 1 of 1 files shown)


test-case/check-alsabat.sh
```diff
@@ -55,10 +56,11 @@ function __upload_wav_file
...
     for file in /tmp/bat.wav.*
     do
-	    size=`ls -l $file | awk '{print $5}'`
-	    if [[ $size -gt 0 ]]; then
-		    cp $file $LOG_ROOT/
-	    fi
+	# alsabat has a bug where it creates an empty record in playback
+	# mode
+	if test -s "$file"; then
+	    cp "$file" ~/ # $LOG_ROOT/
```
xiulipan, line 62: why not copy files to `$LOG_ROOT`？

test-case/check-alsabat.sh
```diff
@@ -71,16 +73,16 @@ dlogi "check the PCMs before alsabat test"
 # different PCMs may support different audio formats(like samplerate, channel-counting, etc.).
 # use plughw to do the audio format conversions. So we don't need to specify them for each PCM.
 dlogc "alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency"
-alsabat -Pplug$pcm_p --standalone -n $frames -F $frequency &
+timeout --kill-after 3 20 alsabat \
```
xiulipan, line 76: a timeout watchdog for the alsabat? I think we have clean up for alsabat aplay arecord thread in exit.

xiulipan (changes requested):

test-case/check-alsabat.sh
marc-hb: oops, that was for debug and slipped. Good catch thank you.

test-case/check-alsabat.sh
marc-hb: > a timeout watchdog for the alsabat? I think we have clean up for alsabat aplay arecord thread in exit.

No test should leave random processes behind, each test should clean up after itself because only the test knows exactly what was started. Cleaning up in the exit function is great to give the next test a chance but finding processes still hung at the exit stage should be treated as a test failure.

The `timeout` and  `wait` are unrelated to each other.

`wait` is the only way to collect errors from background commands. Every background command should have a corresponding wait.

The `timeout` is just me being  paranoid and be prepared in case alsabat gets stuck. If there's no precedence of alsabat ever getting stuck then it's probably overkill and I can remove it.

marc-hb: > The timeout and wait are unrelated to each other.

Actually they are related, sorry for the confusion: `alsabat -P &` getting stuck  would stall the test only when there is a `wait`. But `alsabat -C` is not in the background so it could stall the test too. To be consistent there should be a `timeout` on either both `-P` and `-C` or none.

I removed the `timeout`, there are already 3 different commits in this PR that's enough for one PR.

xiulipan: @marc-hb Got that, you also want to get the logs from alsabat.

xiulipan (approved):

Issue #472, opened 2020-10-26
multiple-pipeline-capture: Fail to find 4 pipelines still alive, only 3 left


mengdonglin:
This is a random issue on multiple platforms.

e.g. on EHL RVP and CML-U notebook Olym* in daily test on 2020-10-25, report ID 613. 

```
2020-10-25 23:07:04 UTC [REMOTE_INFO] /home/ubuntu/sof-test/test-case/multiple-pipeline-capture.sh will use topology /lib/firmware/intel/sof-tplg/sof-ehl-rt5660.tplg to run the test case
2020-10-25 23:07:04 UTC [REMOTE_INFO] Pipeline list to ignore is specified, will ignore 'pcm=HDA Digital' in test case
2020-10-25 23:07:04 UTC [REMOTE_INFO] Run command to get pipeline parameters
2020-10-25 23:07:04 UTC [REMOTE_COMMAND] sof-tplgreader.py /lib/firmware/intel/sof-tplg/sof-ehl-rt5660.tplg -f 'type:playback' -b ' pcm:HDA Digital' -s 0 -e
2020-10-25 23:07:04 UTC [REMOTE_INFO] Testing: Headset [hw:0,0]
2020-10-25 23:07:04 UTC [REMOTE_COMMAND] aplay -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/zero -q
2020-10-25 23:07:04 UTC [REMOTE_INFO] pipeline start sleep 0.5s for device wakeup
2020-10-25 23:07:09 UTC [REMOTE_INFO] checking pipeline status
2020-10-25 23:07:09 UTC [REMOTE_INFO] preparing sleep 5
2020-10-25 23:07:14 UTC [REMOTE_INFO] checking pipeline status again
/home/ubuntu/sof-test/test-case/multiple-pipeline-capture.sh: line 103: 66312 Killed                  "${APP_LST[$1]}" -D $dev -c $channel -r $rate -f $fmt "${DEV_LST[$1]}" -q
/home/ubuntu/sof-test/test-case/multiple-pipeline-capture.sh: line 103: 66321 Killed                  "${APP_LST[$1]}" -D $dev -c $channel -r $rate -f $fmt "${DEV_LST[$1]}" -q
[...]

PR #516, opened 2020-11-16
add debug info in multiple-pipeline-playback/capture to root cause #472

aiChaoSONG:


marc-hb (changes requested):
Please help review #497 (and others) first.

Also let's not add too much to the already ridiculous amount of copy/paste across these two files (they seem 90% identical!!!). You wrote a function so it should be re-usable without copy/paste/diverge...

marc-hb (changes requested):
Sorry I just realized you are reducing copy/paste/diverge a bit: from 4 instances to 2 instances. How about finishing the job and going to just one instance?

xiulipan:
@aiChaoSONG Ping for update, also rebase is needed.

aiChaoSONG:
@xiulipan @marc-hb Updated, I going to merge the multiple-pipeline-playback.sh and multiple-pipeline-capture.sh into one script later, so some common code is still there. the function func_error_exit can be used by other scripts later,so it is moved to lib.sh

test-case/multiple-pipeline-playback.sh
```diff
@@ -92,21 +90,25 @@ func_run_pipeline_with_type()
...
 }
 
 for i in $(seq 1 $loop_cnt)
 do
     dlogi "===== Testing: (Loop: $i/$loop_cnt) ====="
     # clean up dmesg
     sudo dmesg -C
+    # this variable is modified in func_run_pipeline_with_type,
+    # need to reassign at every iteration
+    tmp_count=$max_count
```
xiulipan, line 111: use local for the tmp_count if you do not want it to be used outside.

test-case/multiple-pipeline-playback.sh
```diff
@@ -92,21 +90,25 @@ func_run_pipeline_with_type()
...
-    pgrep -a arecord
-    pkill -9 aplay
-    pkill -9 arecord
-    exit 1
+    arecord_count=$(pidof arecord | wc -w)
+    dlogi "$arecord_count capture process is running"
+    aplay_count=$(pidof aplay | wc -w)
+    dlogi "$aplay_count playback process is running"
+    overall_count=$((arecord_count + aplay_count))
+    [[ $overall_count -eq $max_count ]] ||
```
xiulipan, line 100: use `local` for the `overall_count`

xiulipan (changes requested):
try to use local in loops or functions to avoid wrong usage.

marc-hb:
> Please help review #497 (and others) first.

I screwed up #497, apologies. Tentative (and urgent) fix in #521. 

It's not practical to work on the same code at the same time because it will cause git conflicts. @aiChaoSONG please pause your work on the multiple pipelines tests for a couple days until I'm done with:
- urgent fix #521 
- waiting a day or two to make sure I didn't screw up even more
- re-synchronizing multiple-pipeline-playback with multiple-pipeline-capture (I totally missed the massive copy/paste)

Once I'm all done I promise these tests will be all yours to deduplicate and clean-up! Appreciated.

marc-hb:
> re-synchronizing multiple-pipeline-playback with multiple-pipeline-capture (I totally missed the massive copy/paste)

Submitted in #525

marc-hb:
I think I got the root cause, see candidate fix in #538

aiChaoSONG:
@472 is already root caused and fixed, close.

PR #517, opened 2020-11-17
kmod-after-playback: remove variant LOG_ROOT and improve set -e compatibility

aiChaoSONG:


test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -55,33 +55,22 @@ pb_duration=${OPT_VALUE_lst['d']}
...
 if [ ${OPT_VALUE_lst['p']} -eq 1 ];then
     func_lib_disable_pulseaudio
 fi
 
-export LOG_ROOT=$log_root/round-0
-$(dirname ${BASH_SOURCE[0]})/check-playback.sh -l 1 -t $tplg -d $pb_duration
-ret=$?
-export LOG_ROOT=$log_root
-[[ $ret -ne 0 ]] && dloge "aplay check failed" && exit $ret
+$(dirname ${BASH_SOURCE[0]})/check-playback.sh -l 1 -t $tplg -d $pb_duration ||
```
marc-hb, line 63: While you're changing these lines, compute this once at the top of the file.
```shell
mydir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
```

marc-hb (changes requested):
Removing code is always great but do you know a bit more about why this feature was here and are you sure nothing uses it any more?

test-case/check-kmod-load-unload-after-playback.sh
aiChaoSONG: The feature is to generate logs of each iteration to different folder, but our CI is not aware of these `round-X` folder, so logs cannot show in report. And I am sure no one is really aware of the logs in `round-X` folder.

I will update `mydir`, thanks

xiulipan (changes requested):
@aiChaoSONG I do not think remove the round here is a good idea. Need to know, we do not need full log for pass round if we have stress test like 1000 round. Maybe some clean up here to keep the last 3 round logs if the test case fail and 1 round log for the pass test or no logs for the passed test cases.

aiChaoSONG:
> @aiChaoSONG I do not think remove the round here is a good idea. Need to know, we do not need full log for pass round if we have stress test like 1000 round. Maybe some clean up here to keep the last 3 round logs if the test case fail and 1 round log for the pass test or no logs for the passed test cases.

We are using sub-case in this test,  instead of putting logs of every round into different round-X folder, we will always have the latest round of log, whether if pass or fail, I think this is good.

xiulipan:
@aiChaoSONG We may need to keep last few rounds logs as some error is not triggered in the last logs but may in previous few logs. We need some comparation to see what is the different between fail and pass round.

aiChaoSONG:
@xiulipan  We already have big problem in this test case: if you look into the case, we have two sub-case in it, playback case and kmod case,  the log even in round-X will be overridden, this is inevitable. so only keep the latest logs is OK.

>  triggered in the last logs

For this case, I think the condition may not happen.

>We need some comparation to see what is the different between fail and pass round.

a pass log can always be extracted from manual test or previous test.

xiulipan:
@aiChaoSONG Could you do some local test with some log file tree to show the update you have done here?
I think the sub-test logs will not shown in the wrapper case folder and we will ignore the logs in our CI.

aiChaoSONG:
@xiulipan before this change:
```
├── dmesg.txt
├── round-0
│   ├── dmesg.txt
│   ├── slogger.txt
│   └── version.txt
├── round-1
│   ├── dmesg.txt
│   ├── slogger.txt
│   └── version.txt
├── round-2
│   ├── dmesg.txt
│   ├── slogger.txt
│   └── version.txt
└── version.txt
```
 after this change, only the last round of log kept
```
.
├── dmesg.txt
├── slogger.txt
└── version.txt
```
Sub-test log will show in the folder too. so even with current code, the logs are overrided.

aiChaoSONG:
SOFCI TEST

aiChaoSONG:
@xiulipan I re-trigger the test, everything is good, If you are OK, please approve, seems Pierre wants this to see the dmesg when case fails in SDW PV

xiulipan (approved):

PR #520, opened 2020-11-18
lib.sh: make process state check a function

aiChaoSONG:

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -318,3 +318,39 @@ journalctl_cmd()
    journalctl -k -q --no-pager --utc --output=short-monotonic \
      --no-hostname  "$@"
 }
+
+check_process_state()
+{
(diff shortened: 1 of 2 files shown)


xiulipan:
@aiChaoSONG Ping for update or close.

aiChaoSONG:
@xiulipan I decide to use check_process_state function to replace sof-process-state.sh,  I think we don't need a single file for this small feature. same thing for sof-process-kill.sh.

xiulipan:
@aiChaoSONG @marc-hb @keqiaozhang @libinyang To me the key point here is do we need standalone run for this function. Do we need this as a tool to debug with aplay/arecord process.

aiChaoSONG:
I don't think anyone have really used this script, and for debug purpose, it is no better than the `ps`

aiChaoSONG:
I will leave the process check as a single file, and improve the code in that file

PR #521, opened 2020-11-18
multiple-pipeline-capture.sh: fix recent set -e related regression

marc-hb:
Fixes 202507fe0c71 ("multiple-pipeline: fix set -e compatibility and add
it") that didn't pay attention to the fact that sof-process-state.sh
returns 2 when there is no process match and broke the test when there
are only aplay or only arecord commands but not both at the same time.

This regression was observed (among others) at:
 https://sof-ci.01.org/linuxpr/PR2565/build4887/devicetest/?model=APL_UP2_NOCODEC&testcase=multiple-pipeline-capture

Also fixes more minor set -e compatibility issues.

I missed the failure in
 https://sof-ci.01.org/softestpr/PR497/build372/devicetest/?model=APL_UP2_NOCODEC&testcase=multiple-pipeline-capture
because it was reported as "N/A" and not FAIL because Jenkins interprets
an exit status '2' as N/A. I should have paid more attention.

### DIFF
--- a/test-case/multiple-pipeline-capture.sh
+++ b/test-case/multiple-pipeline-capture.sh
@@ -97,10 +97,12 @@ func_run_pipeline_with_type()
 func_error_exit()
 {
     dloge "$*"
-    pgrep -a aplay
-    pgrep -a arecord
-    pkill -9 aplay
-    pkill -9 arecord
+    ( set +e
+      pgrep -a aplay
+      pgrep -a arecord
+      pkill -9 aplay
+      pkill -9 arecord
+    )
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


test-case/multiple-pipeline-capture.sh
```diff
@@ -97,10 +97,12 @@ func_run_pipeline_with_type()
...
-    pgrep -a aplay
-    pgrep -a arecord
-    pkill -9 aplay
-    pkill -9 arecord
+    ( set +e
+      pgrep -a aplay
+      pgrep -a arecord
+      pkill -9 aplay
+      pkill -9 arecord
+    )
```
aiChaoSONG, line 105: how about `! pgrep -a aplay ||  pkill -9 aplay`, I think this should be compatible with set -e. the (set +e ...) looks like some hack.

aiChaoSONG (changes requested):
other part look good

test-case/multiple-pipeline-capture.sh
marc-hb: I think this a good idea and better code however I don't think it's worth delaying this very urgent fix and I don't think (set +e ...) is a hack at all

Is it enough if I promise to submit a follow up commit with your change ASAP after this?

marc-hb:
I inspected https://sof-ci.01.org/softestpr/PR521/build405/devicetest/?model=APL_UP2_HDA&testcase=multiple-pipeline-capture which is N/A and it looks OK. This test is PASS on every other platform.

test-case/multiple-pipeline-capture.sh
aiChaoSONG: @marc-hb sure, this is already good enough

aiChaoSONG (approved):
LGTM

test-case/multiple-pipeline-capture.sh
marc-hb: > how about ! pgrep -a aplay || pkill -9 aplay

#523

PR #524, opened 2020-11-18
wavetool.py: add THD+N and RMS calculation for wave analysis

aiChaoSONG:
@keyonjie @keqiaozhang @xiulipan Please help to review, you can try this PR with any wave you have:
`wavetool.py -a thdn -R your-wave-file-path`

### DIFF
--- a/tools/wavetool.py
+++ b/tools/wavetool.py
@@ -21,6 +21,9 @@ import scipy.io.wavfile as wavefile
 # is wrong with firmware scheduler.
 SMART_AMP_DELAY_THRESHOLD = 8
 
+# used to avoid zero division
+eps = 0.000000000001
+
 # Module level global variable which will store command line parameters later
 cmd = None
 
@@ -112,6 +115,8 @@ def do_wave_analysis():
         analyze_wav_smart_amp(wave, fs_wav)
     if cmd.analyze == 'wov':
         analyze_wav_wov(wave, fs_wav)
+    if cmd.analyze == 'thdn':
+        analyze_wav_thdn(wave, fs_wav)
 
 # remove digital zeros in two sides
 def trim_wave(wave):
@@ -205,6 +210,16 @@ def analyze_wav_wov(wave, fs):
         sys.exit(1002)
     print('wave analysis result: PASSED')
 
+def analyze_wav_thdn(wave, fs_wav):
+    wave = normalize(wave)
+    thdn = calc_thdn(wave, fs_wav, cmd.freq[0])
+    result = thdn < cmd.threshold
+    print('THD+N of specified wave is %s dB' % thdn)
+    if not np.all(result):
+        print('THD+N too high, wave analysis result: FAILED')
+        sys.exit(1003)
+    print('wave analysis result: PASSED')
+
 def calc_thdn(wave, fs, freq):
     """
     This function calculates Total Harmonic Distortion plus Noise (THD+N) of a wave.
@@ -217,14 +232,15 @@ def calc_thdn(wave, fs, freq):
 
     Returns
     ----------
-    A list of THD+N values for each channel
+    A list of THD+N values for each channel, unit: dB
     """
     # skip 20ms to avoid the pulse due to filtering
     time_skip = 0.02
     samples_skip = int(time_skip * fs)
     filtered = stdnotch(wave, freq, fs)
-    filtered_cut = filtered[samples_skip:,:]
-    return 10 * np.log10(np.mean(np.power(filtered_cut, 2), axis=0))
+    filtered_cut = filtered[samples_skip:]
+    return 10 * np.log10(np.mean(np.power(filtered_cut, 2), axis=0) + eps)
+
 
 def parse_cmdline():
     parser = argparse.ArgumentParser(add_help=True, formatter_class=argparse.RawTextHelpFormatter,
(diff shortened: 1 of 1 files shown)


tools/wavetool.py
```diff
@@ -112,6 +115,8 @@ def do_wave_analysis():
         analyze_wav_smart_amp(wave, fs_wav)
     if cmd.analyze == 'wov':
         analyze_wav_wov(wave, fs_wav)
```
xiulipan, line 117: I think we should use `if elif` here

tools/wavetool.py
```diff
@@ -205,6 +210,17 @@ def analyze_wav_wov(wave, fs):
...
 
+def analyze_wav_thdn(wave, fs_wav):
+    wave = normalize(wave)
+    check_rms(wave)
+    thdn = calc_thdn(wave, fs_wav, cmd.freq[0])
+    result = thdn < cmd.threshold
+    print('THD+N of specified wave is %s dB' % thdn)
+    if not np.all(result):
+        print('THD+N too high, wave analysis result: FAILED')
+        sys.exit(1003)
```
xiulipan, line 221: why the exit code is 1003

xiulipan (changes requested):

tools/wavetool.py
aiChaoSONG: The exit code is selected, start from 1000, 1001 for smart_amp failure, 1002 for wov failure and 1003 for thdn failure. this should not matter, I think.

tools/wavetool.py
aiChaoSONG: `parser.add_argument('-a', '--analyze', type=str, choices=['smart_amp', 'wov', 'thdn'] ...`. argparse will guarantee us when code goes here, cmd.analyze is one of `['smart_amp', 'wov', 'thdn'] `, so with or without using `if elif` should not matter. if we give something else to cmd.analyze, argparse will exit the program and dump help info.

tools/wavetool.py
paulstelian97: Define some constants and return them instead of hardcoding 1003?

tools/wavetool.py
```diff
@@ -21,6 +21,9 @@
 # is wrong with firmware scheduler.
 SMART_AMP_DELAY_THRESHOLD = 8
 
+# used to avoid zero division
+eps = 0.000000000001
```
singalsu, line 25: Nit, can you use 1e-12 instead?

tools/wavetool.py
```diff
@@ -112,6 +115,8 @@ def do_wave_analysis():
         analyze_wav_smart_amp(wave, fs_wav)
     if cmd.analyze == 'wov':
         analyze_wav_wov(wave, fs_wav)
+    if cmd.analyze == 'thdn':
+        analyze_wav_thdn(wave, fs_wav)
```
singalsu, line 119: Could you add DC offset calculation? Take average value of the clip to analyse and report as dBFS (repeat for every channel). You can omit the start transient. The fail criteria default should be > -60 dBFS. With 16 bit audio it should be normally around -90 dBFS.

tools/wavetool.py
singalsu: Use binary bits coding for multiple fails?

tools/wavetool.py
```diff
@@ -217,14 +232,15 @@ def calc_thdn(wave, fs, freq):
...
+    A list of THD+N values for each channel, unit: dB
     """
     # skip 20ms to avoid the pulse due to filtering
     time_skip = 0.02
     samples_skip = int(time_skip * fs)
     filtered = stdnotch(wave, freq, fs)
-    filtered_cut = filtered[samples_skip:,:]
-    return 10 * np.log10(np.mean(np.power(filtered_cut, 2), axis=0))
+    filtered_cut = filtered[samples_skip:]
+    return 10 * np.log10(np.mean(np.power(filtered_cut, 2), axis=0) + eps)
```
singalsu, line 243: I think we should call this "THD+N ratio" since it was measured with sine wave input. Extracts from AES spec:

_"The test signal shall be a sine wave. The analyzer shall include the standard low-pass filter (see 5.2.5). The total
signal level at the EUT output is measured and recorded. The fundamental component of the test signal present
in the output shall be removed with the standard notch filter (see 5.2.8), and the unweighted level of the residual
shall be measured. The THD+N ratio shall be the difference between the residual level and the total signal
level, expressed in dB."

and

"The test signal level shall be a sine wave at -1 dB relative to the maximum input level (see 6.2.1). "

So, the used sine level should be subtracted from the result. E.g. -81 - (-1) dB gives -80 dB.

tools/wavetool.py
aiChaoSONG: @singalsu The exit code just denote failure, I just want to use different code to denote different failure. this script is called from bash, we only care it is PASS of FAIL. Btw, I have printed some very useful logs.

tools/wavetool.py
aiChaoSONG: > The test signal level shall be a sine wave at -1 dB relative to the maximum input level 

I think we may not use the THD+N ratio here,  because our test signal is not a standard -1 dB sine wave.  our loopback wave varies in level, so an absolute value (thdn) is better than a relative value (thdn ratio)

singalsu: The problem without ratio is that a muted / a lot attenuated test might a really good THD+N value. You need to set the limit to such that an OK performing system passes it. I would never pass a result worse than -60 dB (equivalent to 10 bits performance - our implementation is 16/24/32 bits).

aiChaoSONG: @singalsu For a good quality sine wave with -30dB level, and -80dB thdn, the THD+N ratio is on -50dB. it is hard to maintain a 0 dB volume control in our loopback route, how can we deal with this condition? maybe we should keep the calculation of thdn and thdn ration, and let user to choose what they want to use?

singalsu: With current test system (headphone out -> headset adapter mic in) I think it's not expected to get any better performance. Could the pass criteria be something like -48 dB until the CI hardware can be upgraded? Testing it is much better than nothing though small issues will be left unnoticed. The THD+N alone is not good because a broken system can get a good result.

@mengdonglin The -50 dB THD+N ratio is performance of a 80's C-cassette player. Please try to escalate the need for better audio validation hardware. We really should be able to use something like -70 to -80 dB THD+N ratio as automatic test pass/fail criteria.

aiChaoSONG:
This PR will need to wait https://github.com/thesofproject/sof-test/pull/539 to be merged first, and also some clean up with wavetool.py, let's concentrate on #539, after cleaning up wavetool.py, this will be picked up.

plbossart:
@aiChaoSONG @singalsu is this still relevant or should we close this?

aiChaoSONG:
@plbossart  we can close this first,  no time to come back now

PR #526, opened 2020-11-18
check-alsabat: use aplay/arecord exit status in precheck

marc-hb:
... as opposed to assuming that any -q output is an error message.


xiulipan (approved):
This would catch some hardware issues as early as possible, thanks @marc-hb

Issue #329, opened 2020-08-11
Don't use default period-size and buffer-size


plbossart:
We need to control better what the period and buffer sizes are in all sof-tests.

aplay and arecord can be configured with --period-time=5000 and --buffer-time=10000 for e.g 2 periods of 5ms.

It would be good enough to have an environment variable for each value, so that we can configure the tests at will, and even set stricter values for specific tests.

xiulipan:
@marc-hb @plbossart will https://github.com/thesofproject/sof-test/pull/527 can work on this request?

marc-hb:
It looks like SOF_ALSA_OPTS from (not finished) #489 would work for this too. @plbossart  can you confirm?

PR #527, opened 2020-11-19
check-pause-resume: add support for SOF_ALSA/APLAY/ARECORD_OPTS

marc-hb:
2 commits: 1. shellcheck clean-up, 2. SOF_ALSA_OPTS

### DIFF
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -60,10 +60,12 @@ rnd_range=$(( rnd_max -  rnd_min ))
 case $test_mode in
     "playback")
         cmd=aplay
+        cmd_opts="$SOF_APLAY_OPTS"
         dummy_file=/dev/zero
     ;;
     "capture")
         cmd=arecord
+        cmd_opts="$SOF_ARECORD_OPTS"
         dummy_file=/dev/null
     ;;
     *)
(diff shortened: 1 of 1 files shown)


marc-hb:
On my APL_UP2_NOCODEC I get this error message from time to time:
```
(1/10) Wait for 155 ms before pause                | 00%
=== PAUSE ===           aplay: do_pause:1545: pause push error: File descriptor in bad state
(1/10) Wait for 197 ms before pause                | 00%                  
(1/10) Wait for 116 ms before pause                                      
(1/10) Wait for 118 ms before resume
```

Yet `SOF_ALSA_OPTS='--fatal-errors'`  does NOT stop `aplay` after it prints this error and it does not make the test fail #greenfailures

cc: @plbossart

marc-hb:
Pure rebase (after # 529), no code change.

test-case/check-pause-resume.sh
```diff
@@ -75,24 +78,25 @@ esac
...
+    rate=$(func_pipeline_parse_value "$idx" rate)
+    fmt=$(func_pipeline_parse_value "$idx" fmt)
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    snd=$(func_pipeline_parse_value "$idx" snd)
 
     # expect is tcl language script
     #   expr rand(): produces random numbers between 0 and 1
     #   after ms: Ms must be an integer giving a time in milliseconds.
     #       The command sleeps for ms milliseconds and then returns.
+    dlogi "Entering expect script with:
```
xiulipan, line 95: @marc-hb I see we have duplicated logs, example:

https://sof-ci.01.org/softestpr/PR527/build423/devicetest/?model=ICL_RVP_HDA&testcase=check-pause-resume-playback-10
```
2020-11-23 22:32:27 UTC [REMOTE_INFO] Entering expect script with:
      aplay   -D hw:0,5 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
spawn aplay -D hw:0,5 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
```

xiulipan (approved):
Looks good, but we seems to have duplicated logs output with expect.

test-case/check-pause-resume.sh
marc-hb: The log was already duplicated, I just added a newline that made it more obvious. Before:

```shell
git checkout origin/master
./test-case/check-pause-resume.sh

2020-12-02 04:25:33 UTC [INFO] Entering expect script with: aplay -D hw:0,0 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
spawn aplay -D hw:0,0 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
```
I can easily escape the newline so the output is exactly like before.

I'm a bit reluctant to trust and rely entirely on `expect` for the logging... maybe that's also why someone else doubled this?

xiulipan: Good point here, let's compare the cmd here with both logs.

marc-hb:
For the record this should help with #329 and #489

PR #528, opened 2020-11-19
kmod: add max98373_i2c codec driver name

keqiaozhang:
max98373_i2c driver is spilt from max98373 codec driver.


aiChaoSONG (approved):
LGTM

PR #529, opened 2020-11-20
journalctl patchset version 5

xiulipan:
- dmesg usage clean up
- journalctl replacement
- new error log ignore list‘

TODO:
kernel log dump and some test case dmesg usage.

PR #530, opened 2020-11-24
fake_kern_error: add missing \n when writing to /dev/kmsg

marc-hb:
This wasn't really a problem as long as sof-kernel-log-check.sh was
reading /var/log/kernel.log, which just changed in PR #529


xiulipan (approved):

xiulipan:
I thought the issue is caused by we need a flush. It seems we missed a '\n'

marc-hb:
> I thought the issue is caused by we need a flush. It seems we missed a '\n'

I think it's "both". `dmesg` is line-based so I think nothing gets printed until there is a newline character. Otherwise concurrent writers could be interleaved, it would look horrible.

I didn't know you noticed this, please tell me next time.

PR #523, opened 2020-11-18
multiple-pipeline-capture: simplify func_error_exit()

marc-hb:
As suggested by Chao in #521

Also remove no-op exit $? on the last line.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #531, opened 2020-11-24
multiple-pipeline-playback: remove dmesg -C call

marc-hb:
This was apparently missed in #529

Together with #523 this makes it identical to multiple-pipeline-capture.sh
again.


xiulipan:
@marc-hb Thanks for the clean up. I will have an overall demsg clean up.

xiulipan (approved):

PR #532, opened 2020-11-24
De-duplicate multiple-pipeline-capture/playback

marc-hb:
4 commits, commits MUST be reviewed individually otherwise github shows a very large code change when it's actually quite small.


marc-hb:
https://sof-ci.01.org/softestpr/PR532/build427/devicetest/ passed for the most part but this change confused Jenkins which timed out. I think this because this change makes the test looks like a "Sub-Test", so the last line printed is: `Sub-Test: [REMOTE_INFO] Test Result: PASS!` which is not enough for Jenkins. @xiulipan, @aiChaoSONG  would this explanation make sense?

marc-hb:
> I think this because this change makes the test looks like a "Sub-Test", 

Fixed  in https://sof-ci.01.org/softestpr/PR532/build438/devicetest/ thanks to `exec`

test-case/multiple-pipeline.sh
```diff
@@ -35,6 +33,10 @@ OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
 OPT_OPT_lst['c']='count'    OPT_DESC_lst['c']='test pipeline count'
 OPT_PARM_lst['c']=1         OPT_VALUE_lst['c']=4
 
+OPT_OPT_lst['f']='first'
+OPT_DESC_lst['f']='Fill either playback (p) or capture (c) first'
+OPT_PARM_lst['f']=1         OPT_VALUE_lst['f']='p'
+
```
aiChaoSONG, line 39: @marc-hb The option `f ` means first, that's good for now.  In fact, I want one more option for `all pipelines`,  then the `first` may not be proper.

aiChaoSONG (approved):

marc-hb:
> The option f means first, that's good for now. In fact, I want one more option for all pipelines, then the first may not be proper.

Good point, how about `-o(rder)`?

I'm going to merge this because it's been submitted for 7 days ago, should make no functional change and it's blocking further, actual functional changes. I can submit a follow-up commit that just renames the option as the option does not have any external user yet.

PR #533, opened 2020-11-24
tools: sof-kernel-log-check: fix bracket in ACPI errors

xiulipan:
In ACPI related error logs, missed escaping for '(' and ')'

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -179,7 +179,7 @@ ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. R
 ignore_str="$ignore_str"'|: authentication with ..:..:..:..:..:.. timed out'
 
 # Dell TGL laptop with SoundWire, issues reported by sof-test
-ignore_str="$ignore_str"'|ACPI BIOS Error (bug):'
+ignore_str="$ignore_str"'|ACPI BIOS Error \(bug\):'
 ignore_str="$ignore_str"'|ACPI Error:'
 ignore_str="$ignore_str"'|acpi device:00: Failed to change power state to D3hot'
 


libinyang (approved):

tools/sof-kernel-log-check.sh
```diff
@@ -179,7 +179,7 @@ ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Microcode SW error detected\. R
 ignore_str="$ignore_str"'|: authentication with ..:..:..:..:..:.. timed out'
 
 # Dell TGL laptop with SoundWire, issues reported by sof-test
-ignore_str="$ignore_str"'|ACPI BIOS Error (bug):'
+ignore_str="$ignore_str"'|ACPI BIOS Error \(bug\):'
```
aiChaoSONG, line 182: Normally, there should be more info after the `:`, eg, `ACPI BIOS Error (bug): Could not resolve symbol [\_SB.PC00.PEG0._PS3.HPEX]`, I think `ACPI BIOS Error (bug):` match exactly "ACPI BIOS Error (bug):", maybe add a `*` ?

aiChaoSONG (commented):
please also remove the `ignore_str="$ignore_str"'|ACPI Error:'` I think never worked.

tools/sof-kernel-log-check.sh
xiulipan: @aiChaoSONG I think we did not need that, this is already a regex.

xiulipan:
@aiChaoSONG full error log looks like

```
kernel: [ 6669.927574] ACPI BIOS Error (bug): Could not resolve symbol [\_SB.PC00.PEG0._PS3.HPEX], AE_NOT_FOUND (20200925/psargs-330)
kernel: [ 6669.927588] ACPI Error: Aborting method \_SB.PC00.PEG0._PS3 due to previous error (AE_NOT_FOUND) (20200925/psparse-531)
kernel: [ 6669.927608] acpi device:00: Failed to change power state to D3hot
```

marc-hb:
```shell
echo '1 a(b)c' | grep     'a(b)c'   # match 
echo '2 a(b)c' | grep -E  'a(b)c'   # no match 
echo '3 a(b)c' | grep -E  'a\(b\)c' # match 
```

PR #535, opened 2020-11-27
kernel: add a string to ignore

lyakh:
Add one more DRM error message to the list of ignored errors.


aiChaoSONG (approved):

marc-hb:
According to @kv2019i  pipe A is typically for the eDP panel which does not have audio.

PR #537, opened 2020-11-30
multiple pipeline TEST HACK

marc-hb:

PR #538, opened 2020-12-01
multiple-pipeline: replace pidof->ps to catch uninterruptible sleep

marc-hb:
pidof ignores processes in uninterruptable sleep by default, probably
because sending them signals is futile. pidof has a -z option but it's
just simpler to use "ps".

Also de-duplicate code into new ps_checks() function

Fixes: #472

Any aplay or arecord process can be in uninterruptable sleep when doing
I/O but the easiest way to reproduce is to wait a few seconds and let the
DSP go to sleep.

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -66,8 +66,6 @@ DEV_LST['playback']='/dev/zero'
 APP_LST['capture']='arecord_opts'
 DEV_LST['capture']='/dev/null'
 
-tmp_count=$max_count
-
 # define for load pipeline
 func_run_pipeline_with_type()
 {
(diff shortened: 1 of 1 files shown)


marc-hb:
The only way to make sure races are fixed is to make them worse. This is what TEST #537 did.

marc-hb:
https://sof-ci.01.org/softestpr/PR538/build443/devicetest/ is all green

marc-hb:
Note @keqiaozhang moved uninterruptible sleep to the "normal" state category in commit 180b5d761f41

xiulipan (approved):
Nice catch, I think we already find the process status issue in https://github.com/thesofproject/sof-test/blob/686d273e460e801c6915b7ecf3d669a48f68a194/tools/sof-process-state.sh#L40-L41
But the difference of `pidof` and `ps` is really difficult to debug.

fredoh9:
This is really good catch and fix! Thanks Marc!

marc-hb:
Follow-up commit #543 should save about 20 minutes in nightly tests.

PR #540, opened 2020-12-02
Travis: update OS and add numpy and scipy packages

aiChaoSONG:


marc-hb:
Unfortunately we need to wait a few hours :-(

aiChaoSONG:
@marc-hb The travis job is still queuing, is there any problem with travis or the code change is not correct?

marc-hb:
I think Travis has just been incredibly slow. On November 19 they reduce the capacity from 500 down to 400 :-(
https://www.traviscistatus.com/#month

I will look at github actions soon

marc-hb (approved):
https://travis-ci.org/github/thesofproject/sof-test/builds/747115385 successful about 30 hours (!) after submission...

PR #541, opened 2020-12-02
tools: sof-kernel-log-check: fix wrong usage of condition check

xiulipan:
use if then else fi to replace wrong usage of condition check with
&& and ||

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -313,11 +313,19 @@ ignore_str="$ignore_str"'|rt700 sdw:.:.*:700:.: Bus clash detected'
 
 # confirm begin_timestamp is in UNIX timestamp format, otherwise search full log
 journalctlflag="-k -q --no-pager --utc --output=short-monotonic --no-hostname"
-[[ $begin_timestamp =~ ^[0-9]{10} ]] && cmd="journalctl $journalctlflag --since=@$begin_timestamp" || cmd="journalctl $journalctlflag"
+if [[ $begin_timestamp =~ ^[0-9]{10} ]]; then
+    cmd="journalctl $journalctlflag --since=@$begin_timestamp"
+else
+    cmd="journalctl $journalctlflag"
+fi
 
 declare -p cmd
 # check priority err for error message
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -313,11 +313,19 @@ ignore_str="$ignore_str"'|rt700 sdw:.:.*:700:.: Bus clash detected'
 
 # confirm begin_timestamp is in UNIX timestamp format, otherwise search full log
 journalctlflag="-k -q --no-pager --utc --output=short-monotonic --no-hostname"
-[[ $begin_timestamp =~ ^[0-9]{10} ]] && cmd="journalctl $journalctlflag --since=@$begin_timestamp" || cmd="journalctl $journalctlflag"
+if [[ $begin_timestamp =~ ^[0-9]{10} ]]; then
+    cmd="journalctl $journalctlflag --since=@$begin_timestamp"
+else
+    cmd="journalctl $journalctlflag"
+fi
```
marc-hb, line 320: This could be
```shell
cmd="journalctl $journalctlflag"
if [[ $begin_timestamp =~ ^[0-9]{10} ]]; then
   cmd="$cmd --since=@$begin_timestamp"
fi
```

tools/sof-kernel-log-check.sh
```diff
@@ -313,11 +313,19 @@ ignore_str="$ignore_str"'|rt700 sdw:.:.*:700:.: Bus clash detected'
...
 
 declare -p cmd
 # check priority err for error message
-[[ "$ignore_str" ]] && err=$($cmd --priority=err | grep -vE "$ignore_str") || $($cmd --priority=err)
+if [[ "$ignore_str" ]]; then
```
marc-hb, line 324: In general `[[ -n "$ignore_str" ]]` feels safer to me because it does not matter how `ignore_str` starts but off-topic nitpicking.

marc-hb (approved):
Approved if the shellcheck warnings are gone.

xiulipan:
@marc-hb local run `shellcheck the tools/sof-kernel-log-check.sh` shows no warning or error.

PR #543, opened 2020-12-02
multiple-pipeline: remove initial sleep and clarify second sleep

marc-hb:
Having the same -w parameter for both this initial sleep and the test
duration made no sense: fixed by removing the first sleep.

It's not clear what the purpose of this first sleep was, it has been in
multiple-pipeline-capture/playback.sh since the dawn of time (= commit
39cd964ddc776c) For sure this first sleep was helping hide
"uninterruptible state" bug #472 by giving the DSP more time to wake up!

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -47,7 +47,7 @@ OPT_NAME['f']='first'
 OPT_DESC['f']='Fill either playback (p) or capture (c) first or any (a) for all pipelines'
 OPT_HAS_ARG['f']=1         OPT_VAL['f']='p'
 
-OPT_NAME['w']='wait'     OPT_DESC['w']='perpare wait time by sleep'
+OPT_NAME['w']='wait'     OPT_DESC['w']='duration of one (sub)test iteration'
 OPT_HAS_ARG['w']=1         OPT_VAL['w']=5
 
 OPT_NAME['r']='random'   OPT_DESC['r']='random load pipeline'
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR543/build480/devicetest is all green.

This PR should save about 20 minutes in nightly testing.

It's ready for review but let's not merge it until we have confirmation that #472 is fixed.

marc-hb:
No new shellcheck warning in https://travis-ci.org/github/thesofproject/sof-test/jobs/747267512

test-case/multiple-pipeline.sh
```diff
@@ -149,13 +149,10 @@ do
             die "Wrong -f argument $f_arg, see -h"
     esac
 
-    dlogi "sleep ${OPT_VALUE_lst['w']}s for sound device wakeup"
-    sleep ${OPT_VALUE_lst['w']}
```
xiulipan, line 153: I think this sleep is used to avoid delay in PM resume.
Not sure if this is still a blocker for us.
@aiChaoSONG @keqiaozhang Any idea？

marc-hb: ping?

fredoh9 (approved):
Logically making sense. But I do remember it was introduced due to some errors in some cases. (don't know why we put same value with before/after ps_check).
By the device test result, it looks good.

keqiaozhang (approved):

marc-hb:
SOFCI TEST

marc-hb:
The only failure in https://sof-ci.01.org/softestpr/PR543/build676/devicetest/?model=BYT_MB_NOCODEC&testcase=simultaneous-playback-capture is some input/output error that sounds familiar but that is more importantly in a different test not modified at all by this PR. Merging

keqiaozhang:
@marc-hb , I added more logs and got below information. From the output, the processe ID 4986 marked in red was not cleaned up during the next round of testing, that's why it failed with:
```
2021-04-28 09:02:43 UTC [ERROR] Running process count is 5, but 4 is expected
```
Capture

So the sleep time between each round of testing is necessary to ensure that the all processes of the previous round of testing is completely cleared

marc-hb:
Reverted in #669 and new version submitted in #671

PR #544, opened 2020-12-02
lib.sh: log git version of the tests at the start of every test

marc-hb:
While there is generally in test results some version information about
firmware and kernel (except for force-pushes, merges with target
branch,... I digress) we had absolutely none about the version of
sof-test itself. This made it very difficult to monitor sof-test fixes
and other sof-test changes. This is also useful when someone shares the
log of a local test run.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -21,6 +21,28 @@ source "$SCRIPT_HOME/case-lib/pipeline.sh"
 # shellcheck source=case-lib/hijack.sh
 source "$SCRIPT_HOME/case-lib/hijack.sh"
 
+sof_test_version()
+{
+    SOF_TEST_VERSION=$(cd "$SCRIPT_HOME" &&
+                           git describe --always --long --tags --dirty) || {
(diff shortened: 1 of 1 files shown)


xiulipan:
@marc-hb This version logging looks good. Could you help to check 
https://github.com/thesofproject/sof-test/blob/8a21c09bd0d0e349ea769cfde317cb519de7a582/case-lib/opt.sh#L144-L156
And make some update and unify for the version dump to file feature?

marc-hb:
Thanks, I did not notice that. Will take a look.

marc-hb:
SOFCI TEST

plbossart:
@marc-hb is this still relevant? I see no update since December but a test run mid-May?

marc-hb:
There is a similar "version" tab in Jenkins results now, example:
https://sof-ci.01.org/sofpr/PR4389/build9449/devicetest/?model=GLK_BOB_DA7219&testcase=verify-tplg-binary

It's Jenkins-specific but probably enough, closing.

PR #545, opened 2020-12-02
sof-kernel-log-check: drop $journalctlflag duplicating journalctl_cmd()

marc-hb:
Include case-lib/lib.sh and use journalctl_cmd() wrapper instead

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -130,11 +130,15 @@ begin_timestamp=${1:-0}
 declare ignore_str
 
 # pwd resolves relative paths
-my_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
-platform=$("$my_dir"/sof-dump-status.py -p)
+test_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -312,11 +316,10 @@ ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
...
-journalctlflag="-k -q --no-pager --utc --output=short-monotonic --no-hostname"
 if [[ $begin_timestamp =~ ^[0-9]{10} ]]; then
-    cmd="journalctl $journalctlflag --since=@$begin_timestamp"
+    cmd="journalctl_cmd --since=@$begin_timestamp"
 else
-    cmd="journalctl $journalctlflag"
+    cmd="journalctl_cmd"
 fi
 
 declare -p cmd
```
xiulipan, line 325: maybe it is also time to remove the debug log here.

xiulipan (approved):
Thanks for the optimization.

tools/sof-kernel-log-check.sh
marc-hb: #529 is less than one week old so I would rather wait a bit longer.

PR #549, opened 2020-12-08
sof-kernel-log-check: fix to filter out USB3.1 suspend/resume error

fredoh9:
'.' is single charactor. Template string doesn't need error number
to have.

Still test will be failed due to this errors without this fix.
```
[   28.970627] kernel: usb 3-10: device descriptor read/64, error -71
[   29.193611] kernel: usb 3-10: device descriptor read/64, error -71
[   29.526684] kernel: usb 3-10: device descriptor read/64, error -71
[   29.746289] kernel: usb 3-10: device descriptor read/64, error -71
[   30.887534] kernel: usb 3-10: device not accepting address 13, error -71
[   31.415512] kernel: usb 3-10: device not accepting address 14, error -71
```

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -204,8 +204,8 @@ ignore_str="$ignore_str"'|usb 2-.: .'
 
 # TGL devices with USB 3.1 devices, issues reported by sof-test
 # BugLink: https://github.com/thesofproject/sof-test/issues/482
-ignore_str="$ignore_str"'|usb 3-.: device descriptor read/64, error .'
-ignore_str="$ignore_str"'|usb 3-.: device not accepting address ., error .'
+ignore_str="$ignore_str"'|usb 3-.+: device descriptor read/64, error .+'
+ignore_str="$ignore_str"'|usb 3-.+.: device not accepting address .+, error .+'
 
 # Test cases on some platforms fail because the boot retry message:
 #


tools/sof-kernel-log-check.sh
```diff
@@ -204,8 +204,8 @@ ignore_str="$ignore_str"'|usb 2-.: .'
 
 # TGL devices with USB 3.1 devices, issues reported by sof-test
 # BugLink: https://github.com/thesofproject/sof-test/issues/482
-ignore_str="$ignore_str"'|usb 3-.: device descriptor read/64, error .'
-ignore_str="$ignore_str"'|usb 3-.: device not accepting address ., error .'
+ignore_str="$ignore_str"'|usb 3-..: device descriptor read/64, error'
```
aiChaoSONG, line 207: I don't think this is going to work well,  the `usb 3-..` is not going to work if we have `usb 3-4`, and we have!!!
This is better:
```
ignore_str="$ignore_str"'|usb 3-.+: device descriptor read/64, error .+'
ignore_str="$ignore_str"'|usb 3-.+: device not accepting address .+, error .+'
```

there is one more usb error in report 1021
```
[ 6688.142149] kernel: usb usb3-port10: unable to enumerate USB device
```
add an re like this
```
usb usb.-port.+: unable to enumerate USB device
```

aiChaoSONG (commented):
@fredoh9 Let have a quick fix and merge

tools/sof-kernel-log-check.sh
aiChaoSONG: all above re are tested locally.

marc-hb (changes requested):
The trailing `.+` is misleading because AFAIK we don't perform full line matching, can you please use `fake_kern_error` and re-test without it?

It still works but generates the confusion that we perform full line matching and will be copied by future users.

aiChaoSONG:
@marc-hb I have tested old RE with a single dot, confirmed not work.

aiChaoSONG (approved):

marc-hb:
@aiChaoSONG 

> The TRAILING (last) `.+` is misleading 

If the TRAILING `.+` was required, many of the other filters wouldn't work.

@fredoh9 can you test and submit a fixup removing the trailing `.+`?

marc-hb:
I think the `.+.` in the middle was a but, fixed in #624

PR #550, opened 2020-12-08
sof-kernel-log-check: ignore a usb error on TGL

aiChaoSONG:
After using journalctl, third-party error message
which doesn't contain predefined keywords will be
collected, too. Ignore a USB error on TGL in this
patch.


fredoh9:
Sorry, I saw this log but I just ignored it because I was under impression this is not an error. I was living in old log filtering style.
Thank you for good catch.

fredoh9 (approved):

PR #554, opened 2020-12-14
remove plughw in alsabat test

keqiaozhang:
Add channel and rate options support in alsabat test and
remove use of plughw to run alsabat test at hw level w/o
any conversions done by alsa-lib.

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -24,22 +24,30 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 OPT_NAME['p']='pcm_p'     	OPT_DESC['p']='pcm for playback. Example: hw:0,0'
 OPT_PARM_lst['p']=1          	OPT_VALUE_lst['p']=''
 
+OPT_NAME['C']='channel_c'       OPT_DESC['C']='channel number for capture.'
+OPT_PARM_lst['C']=1             OPT_VALUE_lst['C']='1'
+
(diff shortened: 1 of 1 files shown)


marc-hb:
Sorry I don't understand why the number of channels is hardcoded for playback but not for capture.

I also find the code comment not visible enough, I think a `dlogw` or at the very least a `dlogi` warning should be printed.

keqiaozhang:
> Sorry I don't understand why the number of channels is hardcoded for playback but not for capture.
> 
> I also find the code comment not visible enough, I think a `dlogw` or at the very least a `dlogi` warning should be printed.

This script is for sof playback test, as our sof doesn't support 1 channel(mono wav) playback, so I set the channel number to 2.  

Since we use different USB sound cards in CI, old USB sound cards only support 1 channel capture and new USB sound cards support 2 channels, so I set the channel number of capture as an option( -C ), we can specify it in the alsabat-playback.csv file for each platform.   
We plan to buy more USB sound cards to replace the old one in CI. After that, we can remove this option and test 2 channels for both playback and capture.

aiChaoSONG (dismissed):
some usb card now support stereo channel, thanks for the fix.

xiulipan:
@keqiaozhang Need to take care nocodec platforms, they do not use USB loopback.

xiulipan (approved):

aiChaoSONG:
SOFCI TEST

aiChaoSONG:
The failure on APL_NOCODEC is not caused by this PR, but need to fine tune.

aiChaoSONG (approved):

PR #559, opened 2020-12-17
TEST GH actions

marc-hb:

Issue #558, opened 2020-12-17
[EHL]multiple pause resume failed on EHL RVP I2S.


WYL-12:
In daily test report 1229 on EHL RVP I2S, below error observed in multiple pause resume test:

 failed case:
multiple-pause-resume-25.sh

Error message：
```
2020-12-16 23:12:58 UTC [REMOTE_ERROR] Still have expect process not finished after wait for 250
ubuntu     82034   80047  0 23:12 ?        00:00:00 grep -E aplay|arecord
2020-12-16 23:12:58 UTC [REMOTE_ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2020-12-16 23:12:58 UTC [REMOTE_ERROR]  main()  @  /home/ubuntu/sof-test/test-case/multiple-pause-resume.sh
```
[1217_multiple-pause-resume_etrace.txt](https://github.com/thesofproject/sof-test/files/5707314/1217_multiple-pause-resume_etrace.txt)

Recipe:
Kernel: topic/sof-dev 71ae5193-2
SOF : master 8d277b5b159e-1
TPLG: sof-ehl-rt5660.tplg
Platform: EHL RVP I2S

aiChaoSONG:
close for not reproduced

PR #561, opened 2020-12-17
New .github/workflows/pull_request.yml converted from .travis.yml

marc-hb:
Seems to achieve 100% parity with .travis.yml

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,45 @@
+---
+# Tools that can save round-trips to github and a lot of time:
+#
+# yamllint -f parsable pull_request.yml
+# pip3 install ruamel.yaml.cmd
+# yaml json pull_request.yml
+# yaml merge-expand pull_request.yml exp.yml &&
+#    diff -b -u pull_request.yml exp.yml
+
+name: Pull Requests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+
+  shellcheck:
+    runs-on: ubuntu-20.04
+    steps:
+      - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
+      - name: shellcheck
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        # Triple-dotted 'origin/master...HEAD'
+        run: ./tools/CI/check-gitrange.bash origin/${BASE_REF}...HEAD
+               text/x-shellscript shellcheck -x
+
+  pylint:
+    runs-on: ubuntu-20.04
+    steps:
+      # :-( https://github.community/t/support-for-yaml-anchors/16128
+      - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
+      - uses: actions/setup-python@v2
+      - name: get python libs
+        # FIXME: apt-get succeeds but 'import numpy' still fails!?
+        run: sudo apt-get -y install python3-numpy python3-scipy pylint
+      - name: pylint
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        # let's re-enable 'C'onventions once we got the other numbers down
+        run: ./tools/CI/check-gitrange.bash origin/${BASE_REF}...HEAD
+               text/x-python pylint --disable=C


marc-hb:
Relatively well tested in #559 , see shellcheck and pylint actually running there.

paulstelian97 (commented):
I'm pretty sure that this is the right direction, but I want to know that everything works well before approving.

marc-hb:
> I'm pretty sure that this is the right direction, but I want to know that everything works well before approving.

Can you be more specific? The only known regression compared to Travis is the "import numpy" warning, see test #559.

On December 31 we're going to lose Travis across all repos, at that point in time  everything will _not_ work.

paulstelian97:
> > I'm pretty sure that this is the right direction, but I want to know that everything works well before approving.
> 
> Can you be more specific? The only known regression compared to Travis is the "import numpy" warning, see test #559.
> 
> On December 31 we're going to lose Travis across all repos, at that point in time everything will _not_ work.

I would have expected more than just Shellcheck and Pylint if this replaces Travis.

marc-hb:
> I would have expected more than just Shellcheck and Pylint if this replaces Travis.

I think there's some confusion: this is the `sof-test` repo, not the `sof` or `linux` repo. I converted the _entire_ (and small) `.travis.yml` file. I started with `sof-test` because it's small, so a good guinea pig.

PS: one of the reasons why `.travis.yml` is small in `sof-test` is because everything that can live outside `.travis.yml` does, in a `.sh` file. Much easier maintenance in general, [unlike](https://github.com/thesofproject/linux/pull/2301#issuecomment-661479045) the `.travis.yml` file in the `linux` repo`. I digress.

paulstelian97 (approved):

cujomalainey:
Just curious what the motivation is for the switch?

marc-hb:
> Just curious what the motivation is for the switch?

https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

I am _not_ endorsing the opinions in this blog, I just found this particular one to also have a very good description of the history and facts. There are many other articles on the same topic for anyone interested in a more complete picture.

This very first Github Actions PR is both a 97% functional demo and a heads-up/"poll" about how the post-`travis.org` era could/should look like; "send patches" as they say :-)

There is little code in this PR and CI code rarely draws a lot of nit-picking so I think we can use this PR to host higher level pros and cons discussions if needed, hopefully no need to create a separate github issue to separate the two.

cujomalainey:
Fine by me, that backlog drove me mad. I had builds that take sub 2min on another project get backlogged for 3 days

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,45 @@
...
+        # Triple-dotted 'origin/master...HEAD'
+        run: ./tools/CI/check-gitrange.bash origin/${BASE_REF}...HEAD
+               text/x-shellscript shellcheck -x
+
+  pylint:
+    runs-on: ubuntu-latest
+    steps:
+      # :-( https://github.community/t/support-for-yaml-anchors/16128
+      - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
+      - uses: actions/setup-python@v2
```
cujomalainey, line 36: is this using python2? is that why your import is failing

marc-hb: Thanks, you pushed me to take a closer look and I found that `ubuntu-latest` still means 18.04 on github. I changed it to `ubuntu-20.04` (what `.travis.yml` had) and that single known regression is now gone! See latest result in #558 

`latest` wasn't a good idea  anyway, very poor "version control" I'm glad this made me change it. I bet github hasn't moved it because they're too afraid to break too much stuff in one go :-)

Gory details: `pylint` was a python2 script in 18.04 and apparently not capable of analyzing a `#!/usr/bin/python3` script.

xiulipan (approved):

xiulipan:
@marc-hb I think we can create next PR to delete `.travis.yml`

dbaluta:
+1 for github actions. thanks @marc-hb

PR #563, opened 2020-12-22
multiple-pipeline: add more comments

aiChaoSONG:
This patch adds more comments to multiple-pipeline.sh
to describe how the script runs.

No functonal change.

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -1,25 +1,33 @@
 #!/bin/bash
 
 ##
-## Case Name: Run multiple pipelines for playback or capture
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@xiulipan this is only comment update, but if you read the comment, you should know how to run all pipelines in parallel.

xiulipan:
@aiChaoSONG We will need to increase the count to a very large number, right?

xiulipan (approved):

aiChaoSONG:
The failure on BDW has nothing to do with this pr, no new shellcheck warnings introduced, merge.

Issue #565, opened 2020-12-28
Filter out irrelevant asix error.


keqiaozhang:
CI detected this error when testing “check-suspend-resume-50” on TGLH platform.
```
kernel: [ 7321.220312] asix 3-12.1:1.0 enx000ec668ad2a: Failed to read reg index 0x0000: -113
kernel: [ 7321.220314] asix 3-12.1:1.0 enx000ec668ad2a: Failed to write reg index 0x0000: -113
kernel: [ 7321.220315] asix 3-12.1:1.0 enx000ec668ad2a: Failed to enable software MII access
```

1371?model=TGLH_RVP_HDA&testcase=check-suspend-resume-50

keqiaozhang:
Looks like this a network hardware issue, asix_bind return error when hw reset fails. I have switched the network to another network port on motherboard. if this doesn't help, I will try to use the USB network adapter.

keqiaozhang:
CI detected the similar asix USB Ethernet issue on CML Chrome device when testing suspend/resume.  
```
kernel: [ 7974.604354] ax88179_178a 2-2.3:1.0 enx000ec66a7f60: Failed to read reg index 0x0002: -19
kernel: [ 7974.604367] ax88179_178a 2-2.3:1.0 enx000ec66a7f60: Failed to write reg index 0x0002: -19
kernel: [ 7974.619850] ax88179_178a 2-2.3:1.0 enx000ec66a7f60 (unregistered): Failed to write reg index 0x0002: -19
kernel: [ 7974.619870] ax88179_178a 2-2.3:1.0 enx000ec66a7f60 (unregistered): Failed to write reg index 0x0001: -19
kernel: [ 7974.619881] ax88179_178a 2-2.3:1.0 enx000ec66a7f60 (unregistered): Failed to write reg index 0x0002: -19
kernel: [ 7974.622730] xhci_hcd 0000:00:14.0: WARN Set TR Deq Ptr cmd failed due to incorrect slot or ep state.
```
https://sof-ci.sh.intel.com/#/result/planresultdetail/1547?model=CML_HEL_RT5682&testcase=check-suspend-resume-with-playback-25

marc-hb:
> I have switched the network to another network port on motherboard.

@keqiaozhang can you confirm `asix 3-12.1:1.0 enx000ec668ad2a` is on the motherboard? Are the network interfaces on the motherboard all USB? `lsusb -tv` might help.

I see other USB errors, maybe a generic USB issue.

[...]

PR #566, opened 2020-12-29
sof-kernel-log-check: ignore net related error

### LINKED ISSUE #564: Filter out irrelevant e1000e network adapter error.
CI detected an irrelevant network adapter error, need to filter it out in our test case.
`[ 6928.068820] kernel: e1000e 0000:00:1f.6 enp0s31f6: Hardware Error`
1371?model=CML_RVP_HDA&testcase=check-suspend-resume-50

xiulipan:
add ignore for asix and e1000e seriers network adapter

Fix #564 #565

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -308,6 +308,12 @@ ignore_str="$ignore_str"'|proc_thermal 0000:00:..\..: No auxiliary DTSs enabled'
 # elan_i2c i2c-ELAN0000:00: invalid report id data (ff)
 ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 
+# net related error logs can be ignored
+# origin logs seen on TGL platforms
+ignore_str="$ignore_str"'|asix .*:'
+# origin logs seen on CML platforms
+ignore_str="$ignore_str"'|e1000e .*:'
+
 #
 # SDW related logs
 #


keqiaozhang (approved):

tools/sof-kernel-log-check.sh
```diff
@@ -308,6 +308,12 @@ ignore_str="$ignore_str"'|proc_thermal 0000:00:..\..: No auxiliary DTSs enabled'
 # elan_i2c i2c-ELAN0000:00: invalid report id data (ff)
 ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 
+# net related error logs can be ignored
+# origin logs seen on TGL platforms
+ignore_str="$ignore_str"'|asix .*:'
+# origin logs seen on CML platforms
+ignore_str="$ignore_str"'|e1000e .*:'
```
aiChaoSONG, line 315: `*`: match zero or more characters
`.`: match exactly one character
`.+`: match one or more characters
if you want to ignore all asix and e1000e error messages, below is enough, means any error messages contains asix or e1000e will be ignored.
```
ignore_str="$ignore_str"'|asix'
ignore_str="$ignore_str"'|e1000e'
```

aiChaoSONG (changes requested):

marc-hb (changes requested):
I don't think we can ignore ALL network errors unconditionally because we depend on the network for everything. Especially not now that we filter with error levels thanks to `journalctl`, which means all errors are actual, severe errors and not just warning or info messages with "error" in the name.

What specific errors were observed?

keqiaozhang:
> I don't think we can ignore ALL network errors unconditionally because we depend on the network for everything. Especially not now that we filter with error levels thanks to `journalctl`, which means all errors are actual, severe errors and not just warning or info messages with "error" in the name.
> 
> What specific errors were observed?

We have 2 kinds of net errors:https://github.com/thesofproject/sof-test/issues/565 and https://github.com/thesofproject/sof-test/issues/564.  
I assigned the first one to myself and trying to fix it now.

xiulipan:
close as we need to discuss if we would ignore the net related error.

PR #570, opened 2021-01-07
multiple-pipeline: run all pipelines with -f a option

aiChaoSONG:
Running all pipelines in parallel can be completed
by specifying a very bigger pipeline count to -c option,
but it is too indirect.

This patch adds support for running all pipelines in
parallel with '-f a' option

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -35,9 +35,9 @@
 set -e
 
 # shellcheck source=case-lib/lib.sh
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
-OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_NAME['t']='tplg'     OPT_DESC['t']='Topology path, default to environment variable: TPLG'
 OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
 
 OPT_NAME['c']='count'    OPT_DESC['c']='test pipeline count'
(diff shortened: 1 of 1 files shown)


test-case/multiple-pipeline.sh
```diff
@@ -34,9 +34,9 @@
 set -e
 
 # shellcheck source=case-lib/lib.sh
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
-OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg path, default to environment variable: TPLG'
```
xiulipan, line 39: I will add this common string change into my task list. Will ping you for overall clean up.

test-case/multiple-pipeline.sh
```diff
@@ -8,14 +8,15 @@
...
 ##         pipeline count is less than requested pipeline count defined by -c option,
 ##         capture pipelines will be started to ensure requested number of pipelines are
 ##         running in parallel.
-##      b. capture first mode: same with playback mode, but capture pipelines will be
+##      2. capture first mode: same with playback mode, but capture pipelines will be
 ##         started first.
-##      c. all pipelines mode: if requested pipeline count is greater than the number of
-##         pipelines defined in topology, all pipelines will be started.
+##      3. all pipelines mode: run all pipelines in parallel by
+##          3.a use option '-f a' OR
```
xiulipan, line 18:
```
a.
b. 
```
will looks better. You already have the intend.

xiulipan (changes requested):
Few style change.

test-case/multiple-pipeline.sh
aiChaoSONG: I will keep this change in this PR for the multiple-pipeline.sh, because leave only one shellcheck warning makes me feel uncomfortable. also, I will update `tplg` in the description to `topology`.

test-case/multiple-pipeline.sh
aiChaoSONG: I think we should align on this, how about
level 1: use number, 1.
level 2: use lower case character,  a.
level 3: use upper case character, A. ???

xiulipan (changes requested):
Please rebase

xiulipan (approved):

test-case/multiple-pipeline.sh
marc-hb: I have barely tested it but this seems to work better and without warning:
```shell
OPT_DESC['t']="tplg file, default value is env TPLG: $TPLG"
```
If the above is not possible for some unknown reason, here's a quoting trick to avoid the warning:
```shell
OPT_DESC['t']='tplg file, default value is env TPLG: $''TPLG'
```

PR #573, opened 2021-01-08
sof-test: rename OPT_OPT_lst and OPT_DESC_lst

aiChaoSONG:
This PR renames two variable names used to define script command line options, `OPT_OPT_lst` to `OPT_NAME`, `OPT_DESC_lst` to `OPT_DESC`

No functional change, only repo wide search and replace. check discussion https://github.com/thesofproject/sof-test/issues

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -3,26 +3,26 @@
 # Before using this, you must define these option arrays in your test
 # script. They must all be indexed by some unique, one-character
 # codename for each of your option.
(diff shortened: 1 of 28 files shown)


marc-hb:
thanks for this! As long as `git grep` shows that nothing was missed this should be good.

xiulipan:
@aiChaoSONG Let me do a local check with cmd from @marc-hb

xiulipan (approved):
LGTM, no missing name from my local test.

xiulipan:
I will merge this and let's track the runner status closely if we missed any corner case.

Issue #572, opened 2021-01-07
[RFC] Rename variables for defining script's options


aiChaoSONG:
We have for variables named `OPT_DESC_lst` `OPT_OPT_lst` `OPT_PARM_lst` `OPT_VALUE_lst` used for defining script's options. And I just think they are not very to the point, so I decide to rename them to:
`OPT_NAME`, `OPT_ARG_COUNT`, `OPT_ARG_VAL` and `OPT_DESC`

Currently, our option definition code looks like this:
```
OPT_OPT_lst['t']='tplg'     OPT_DESC_lst['t']='tplg file, default value is env TPLG: $''TPLG'
OPT_PARM_lst['t']=1         OPT_VALUE_lst['t']="$TPLG"
```
Two sentences are written in one line, that's not clean and tidy, either.
I want to first rename the variable names, and rearrange the code for script options. After this, we should be less painful.
I'd like to hear from you,  @xiulipan @marc-hb  @fredoh9 @keqiaozhang

marc-hb:
I agree the names are confusing, I have to check the comments in `opt.sh` every time (copied below for convenience).

`OPT_NAME` LGTM and keeping `OPT_DESC` too.

`OPT_ARG_COUNT` can only be 0 or 1, it's basically a boolean so how about `OPT_HAS_ARG`?

For `OPT_ARG_VAL` I prefer `OPT_VAL` because it's used  even when there is no ARG (as another boolean)


```shell
# option setup && parse function
func_opt_parse_option()
{
    # OPT_OPT_lst     (long) option name
    # OPT_DESC_lst    short sentence describing the option
    # OPT_PARM_lst    0 or 1: number of argument required
    # OPT_VALUE_lst   default value overwritten by command line
    #                 input if any. Set to 0 or 1 when PARM=0

    # for example
    # OPT_OPT_lst['r']='remote'
    # OPT_DESC_lst['r']='Run for the remote machine'
    # OPT_PARM_lst['r']=1
    # OPT_VALUE_lst['r']='' ## if PARM=1, must be set, if PARM=0, can ignore

    # h & help is default option, so don't need to add into option list
    OPT_OPT_lst['h']='help'
    OPT_DESC_lst['h']='this message'
    OPT_PARM_lst['h']=0
    OPT_VALUE_lst['h']=0

```

aiChaoSONG:
[...]

PR #574, opened 2021-01-08
sof-test: rename OPT_PARM_lst and OPT_VALUE_lst

aiChaoSONG:
This PR renames two variable names used to define script command line options, OPT_PARM_lst to OPT_HAS_ARG, OPT_VALUE_lst to OPT_ARG

No functional change, only repo wide search and replace. check discussion https://github.com/thesofproject/sof-test/issues

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -3,27 +3,27 @@
 # Before using this, you must define these option arrays in your test
 # script. They must all be indexed by some unique, one-character
 # codename for each of your option.
-declare -A OPT_DESC OPT_NAME OPT_PARM_lst OPT_VALUE_lst
+declare -A OPT_DESC OPT_NAME OPT_HAS_ARG OPT_VALUE_lst
 
 # option setup && parse function
 func_opt_parse_option()
 {
     # OPT_NAME     (long) option name
     # OPT_DESC    short sentence describing the option
-    # OPT_PARM_lst    0 or 1: number of argument required
+    # OPT_HAS_ARG    0 or 1: number of argument required
     # OPT_VALUE_lst   default value overwritten by command line
     #                 input if any. Set to 0 or 1 when PARM=0
 
     # for example
     # OPT_NAME['r']='remote'
     # OPT_DESC['r']='Run for the remote machine'
-    # OPT_PARM_lst['r']=1
+    # OPT_HAS_ARG['r']=1
     # OPT_VALUE_lst['r']='' ## if PARM=1, must be set, if PARM=0, can ignore
 
     # h & help is default option, so don't need to add into option list
     OPT_NAME['h']='help'
     OPT_DESC['h']='this message'
-    OPT_PARM_lst['h']=0
+    OPT_HAS_ARG['h']=0
     OPT_VALUE_lst['h']=0
 
     local _op_temp_script
@@ -58,7 +58,7 @@ func_opt_parse_option()
                 _op_long_lst["--${OPT_NAME[$i]}"]="$i"
             fi
             # append ':' if the option accepts an argument
-            [ ${OPT_PARM_lst[$i]} -eq 1 ] && _short_opt=$_short_opt':' && _long_opt=$_long_opt':'
+            [ ${OPT_HAS_ARG[$i]} -eq 1 ] && _short_opt=$_short_opt':' && _long_opt=$_long_opt':'
         done
 
         if  ! _op_temp_script=$(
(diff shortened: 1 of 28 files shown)


case-lib/opt.sh
```diff
@@ -3,28 +3,28 @@
...
 # option setup && parse function
 func_opt_parse_option()
 {
     # OPT_NAME     (long) option name
     # OPT_DESC    short sentence describing the option
-    # OPT_PARM_lst    0 or 1: number of argument required
-    # OPT_VALUE_lst   default value overwritten by command line
+    # OPT_HAS_ARG    0 or 1: number of argument required
+    # OPT_ARG   default value overwritten by command line
     #                 input if any. Set to 0 or 1 when PARM=0
```
xiulipan, line 15:
```suggestion
    #                 input if any. Set to 0 or 1 when OPT_HAS_ARG    =0
```

And try to refine the text here to make it more easy to understand

case-lib/opt.sh
```diff
@@ -3,28 +3,28 @@
...
 
     # for example
     # OPT_NAME['r']='remote'
     # OPT_DESC['r']='Run for the remote machine'
-    # OPT_PARM_lst['r']=1
-    # OPT_VALUE_lst['r']='' ## if PARM=1, must be set, if PARM=0, can ignore
+    # OPT_HAS_ARG['r']=1
+    # OPT_ARG['r']='' ## if PARM=1, must be set, if PARM=0, can ignore
```
xiulipan, line 21: PARM in the comments also need to be changed

case-lib/opt.sh
```diff
@@ -82,19 +82,19 @@ func_opt_parse_option()
                 # display short option
                 [ "$i" ] && printf '    -%s' "$i"
                 # have parameter
```
xiulipan, line 84:
```suggestion
                # have argument
```

xiulipan (changes requested):
Please also go through the inline comments to see if we need any update.

case-lib/opt.sh
aiChaoSONG: Updated

marc-hb:
As I wrote in #572, `OPT_ARG` is misleading because it also stores the boolean state for options that have no arg.

case-lib/opt.sh
```diff
@@ -3,28 +3,28 @@
...
+    # OPT_ARG   default value overwritten by command line
+    #                 input if any. Set to 0 or 1 when OPT_HAS_ARG=0
 
     # for example
     # OPT_NAME['r']='remote'
     # OPT_DESC['r']='Run for the remote machine'
-    # OPT_PARM_lst['r']=1
-    # OPT_VALUE_lst['r']='' ## if PARM=1, must be set, if PARM=0, can ignore
+    # OPT_HAS_ARG['r']=1
+    # OPT_ARG['r']='' ## if OPT_HAS_ARG=1, must be set, if OPT_HAS_ARG=0, can ignore
```
marc-hb, line 21: This comment is very misleading (I think it did misled you)

aiChaoSONG:
> As I wrote in #572, `OPT_ARG` is misleading because it also stores the boolean state for options that have no arg.

but OPT_HAS_ARG has already done this for us, I mean if OPT_HAS_ARG is 0, we can leave OPT_ARG unspecified. this requires some changes in opt.sh(can be done later).  for option require extra arg, we do
```
OPT_NAME['o']='option_name'
OPT_HAS_ARG['a']=1
OPT_ARG['a']=some_name
OPT_DESC['a']='some description'
```
and for option doesn't require extra arg, we do
```
OPT_NAME['o']='option_name'
OPT_HAS_ARG['a']=0
OPT_DESC['a']='some description'
```

case-lib/opt.sh
aiChaoSONG: Yes, I think I will delete the demo comment here sometime later. and as this PR require rebase, I will drop the comment fix commit. The clean up of opt.sh is just startup.

marc-hb:
> but OPT_HAS_ARG has already done this for us, I mean if OPT_HAS_ARG is 0, we can leave OPT_ARG unspecified. 

Then how do you know if the flag is used?

aiChaoSONG:
> Then how do you know whether the flag has been used?

This is what explained in opt.sh by Bin

marc-hb (changes requested):
> This is what explained in opt.sh by Bin

The code is from Bin but the explanation / comment is from me (check the git blame): options without arg use OPT_VALUE_lst as a boolean, that's where the information is stored and AFAIK nowhere else. This code is not the easiest to understand (this comment cost me a lot of time) but it's very well tested and it works so let's not bother changing it unless we really need to.

This PR and issue started as a large but very simple and mechanical rename easy to review but now we're mixing that up with a code change tangent.

aiChaoSONG:
> This PR and issue started as a large but very simple and mechanical rename easy to review but now we're mixing that up with a code change tangent.

Ok, Let's use OPT_VAL currently.

xiulipan (approved):
LGTM, let's only do name change first. Local git grep checked.

aiChaoSONG:
The failure on TGLH_RVP_HDA is not related to this PR. and reviews are all resolved.

PR #577, opened 2021-01-18
[RFC]tools: kmod: change PCI module order

xiulipan:
snd_sof_intel_hda_common is also a helper that depends on snd_sof_pci
move it

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -33,6 +33,9 @@ remove_module snd_sof_acpi_intel_bdw
 #-------------------------------------------
 # Helpers
 #-------------------------------------------
+# change from https://github.com/thesofproject/linux/pull/2683
+# snd_sof_intel_hda_common has dependency on snd_sof_pci
+remove_module snd_sof_intel_hda_common
 remove_module snd_sof_acpi
 remove_module snd_sof_pci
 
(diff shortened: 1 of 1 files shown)


plbossart:
@xiulipan it's not good if you break existing solutions and don't describe what the problem was in the first place with the kernel changes?

marc-hb:
Most (all?) failures in https://sof-ci.01.org/softestpr/PR577/build547/devicetest/ look like:
```
rmmod: ERROR: Module snd_sof_intel_hda_common is in use by: snd_sof_pci
```

fredoh9:
Let's wait and see where https://github.com/thesofproject/linux/pull/2683 will go.

xiulipan:
@plbossart I have reported the dependency change issue in https://github.com/thesofproject/linux/pull/2683#issuecomment-761974449
Here is an RFC to show why and how the dependency change will break our existing test.

If we make sure https://github.com/thesofproject/linux/pull/2683 has to change the dependency, I will refine this PR commit that we need to change the order by breaking old ones.

plbossart:
alternate solution at https://github.com/thesofproject/sof-test/pull/589 to avoid breaking CI

xiulipan:
close via https://github.com/thesofproject/sof-test/pull/589

PR #579, opened 2021-01-19
CI: remove travis support

xiulipan:
migration to GitHub actions now.


marc-hb:
Is Travis causing any issue right now?

Of course we don't want to maintain the same, duplicate code across travis and github but there is practically no code duplicated (it's all in the same single script as it should be) and there's been no maintenance recently.

xiulipan:
@marc-hb I see we have very slow response from Travis. so let's remove it for now before it stop the service.

marc-hb (approved):
OK, I thought response time got better in 2021 with everyone moving but maybe I was just lucky: https://www.traviscistatus.com/#week

We had good  "test coverage" of these github actions by now.

Note Travis hasn't stopped their service, they're moving to a different place. There's still a possibility that they will automatically migrate everyone to travis.com

PS: SOF migration to github actions is at: https://github.com/thesofproject/sof/pull/3761

xiulipan:
@marc-hb So we do not need to bother with that anymore as we have Jenkins and GitHub actions now.
Reviewing the sof part.

PR #580, opened 2021-01-19
lib: remove the fake kernel device prefix

xiulipan:
The fake_kern_error need to test with not only sof related issues.
before fix:
[683402.712924] sof-audio-pci 0000:66:66.6: asix 3-12.1:1.0 enx000ec668ad2a: Failed to read reg index 0x0000: -113 >/dev/kmsg
after fix:
[683541.396907] asix 3-12.1:1.0 enx000ec668ad2a: Failed to write reg index 0x0000: -113 >/dev/kmsg

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -54,12 +54,13 @@ func_lib_setup_kernel_checkpoint()
 # more critical for testing changes to the test framework and especially
 # error handling code.
 #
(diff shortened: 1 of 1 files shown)


marc-hb (changes requested):
Thanks, can you move the removed prefix to the sample usage line just before the function?

xiulipan:
@marc-hb updated.

marc-hb (approved):

Issue #578, opened 2021-01-18
Filter out irrelevant iwlwifi error


lyakh:
```
[ 6572.558792] kernel: iwlwifi 0000:00:14.3: No beacon heard and the time event is over already...
```

lyakh:
should be fixed by https://github.com/thesofproject/sof-test/pull/581

marc-hb:
I think we should ignore everything from `iwlwifi` (or even better: blacklist it). Hopefully it's not going to crash the system.

However #581 also ignores errors from wired network adapters.

xiulipan:
Close via #601 merged.

PR #581, opened 2021-01-19
Clean up and add new filter rules for net related error logs

### LINKED ISSUE #564: Filter out irrelevant e1000e network adapter error.

xiulipan:
Add asix error log filter
Add e1000e error log filter
Clean up and add iwlwifi error log filter

Fix #564 #564 #578

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -326,6 +326,9 @@ ignore_str="$ignore_str"'|iwlwifi .*: Direct firmware load for .* failed with er
 # BugLink: https://github.com/thesofproject/sof-test/issues/307
 # iwlwifi 0000:00:14.3: Microcode SW error detected. Restarting 0x0.'
 ignore_str="$ignore_str"'|iwlwifi .*: Microcode SW error detected\. Restarting 0x0\.'
+# BugLink: https://github.com/thesofproject/sof-test/issues/578
+# iwlwifi 0000:00:14.3: No beacon heard and the time event is over already...
+ignore_str="$ignore_str"'|iwlwifi .*: No beacon heard and the time event is over already'
 
 #
 # SDW related logs
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -311,6 +311,18 @@ ignore_str="$ignore_str"'|proc_thermal 0000:00:..\..: No auxiliary DTSs enabled'
 ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 
 #
+# net work related logs
+#
+# asix net adaptor logs can be ignored
+# origin logs seen on TGLH platforms
+# BugLink: https://github.com/thesofproject/sof-test/issues/565
+# asix 3-12.1:1.0 enx000ec668ad2a: Failed to read reg index 0x0000: -113
+# asix 3-12.1:1.0 enx000ec668ad2a: Failed to write reg index 0x0000: -113
+# asix 3-12.1:1.0 enx000ec668ad2a: Failed to enable software MII access
+ignore_str="$ignore_str"'|asix .*: Failed to .* reg index .*'
+ignore_str="$ignore_str"'|asix .*: Failed to enable software MII access'
+
+#
 # SDW related logs
 #
 
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -322,6 +322,12 @@ ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 ignore_str="$ignore_str"'|asix .*: Failed to .* reg index .*'
 ignore_str="$ignore_str"'|asix .*: Failed to enable software MII access'
 
+# e1000e net adaptor logs can be ignored
+# origin logs seen on CML platforms
+# BugLink: https://github.com/thesofproject/sof-test/issues/564
+# e1000e 0000:00:1f.6 enp0s31f6: Hardware Error
+ignore_str="$ignore_str"'|e1000e .*: Hardware Error'
+
 #
 # SDW related logs
 #
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -166,20 +166,10 @@ ignore_str="$ignore_str"'|cfg80211: failed to load regulatory\.db'
 ignore_str="$ignore_str"'|EXT4-fs \(nvme0n1p6\): re-mounted\. Opts: errors=remount-ro'
 ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U1 failed\.'
 ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U2 failed\.'
-ignore_str="$ignore_str"'|iwlwifi 0000:00:14\.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-56\.ucode failed with error -2'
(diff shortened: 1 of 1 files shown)


marc-hb:
Can't we blacklist all these network drivers instead?

Networking is critical for everything we do so we should not use any network interface that reports errors.

xiulipan:
@marc-hb I think those network adaptor may those we are using to do remote test. But some got into stability issue.
@keqiaozhang Any idea here?

keqiaozhang:
I think we can filter out the error information of network adaptors first, since the network is still working after these errors and no side effect.

lyakh (approved):

marc-hb:
> since the network is still working after these errors and no side effect.

Then something will change in the test patterns and someone (not me) will waste days with IT wondering why the internal DNS server is not reliable. Some release maybe be delayed. All this because of a $15 adapter, very professional.

marc-hb (changes requested):
If the `asix` errors are specific to some adapters of a given type, then they should be thrown to the trash and replaced by a different model.

Else if the asix errors are caused by some platform-specific (and hopefully temporary) USB issue, these errors should be in a `$platform` specific section to avoid the risk of ignoring network issues else where. Discuss in #565 which one it is.

The network is too critical and must just work, we cannot tolerate even a small risk of network failure. Troubleshooting network issues (whether they're actually network-related or not) is way too time consuming, see internal issues 539, 631, 653, 654 for some very time consuming examples.

Similarly, please confirm we don't even try to use the `e1000e` interface that complains about a "Hardware Error"! (#564) I hope we use a USB adapter instead. This one should also be in a $platform specific section to avoid the risk of ignoring network issues else where.

Wifi errors are different since we don't use wifi. It would be better to blacklist the wifi drivers to avoid the risk of wifi failures contaminating the rest of the kernel (error handling is never tested) but that risk seems IMHO more acceptable than actually using and having a critical dependency on something that reports errors.

xiulipan:
@marc-hb So what should we do now for the showing error logs from net adaptors?
I do not want those nonblocking errors to block CI filter now.

@keqiaozhang @aiChaoSONG @mengdonglin any comments here? What should we do about those not used hardware or always breaking hardware.

marc-hb:
> @marc-hb So what should we do now for the showing error logs from net adaptors?

It depends on whether these are caused by platform-specific USB issues or by faulty USB dongles. I recommended a solution for each case, see my previous comment.

xiulipan:
@marc-hb I agree we have hardware issues here, but we should be aware the filter is for public. We could not try to blacklist or remove problematic USB or net adaptor for all DUT not maintained by us.

@keqiaozhang @aiChaoSONG @fredoh9 Please check if we can have some replacement here soon. I think we should have a quick decision here to avoid seeing these known error logs block our test.

PS: We could not replace all USB device in a short time and we does not known which USB will be better. Worst case here is most USB switch or net adaptor used same IC and had similar issue.

marc-hb:
> but we should be aware the filter is for public. We could not try to blacklist or remove problematic USB or net adaptor for all DUT not maintained by us.

I don't understand this sorry. No one should ever waste their time with broken network hardware when running tests; public or private makes no difference. So in general this sof-test file doesn't need to ignore errors from broken network hardware and it shouldn't start becoming a database of all the broken hardware devices in the world.

Are we really sure the problem is broken hardware and not with the USB host? I asked 6 days ago in #565 but no answer. I'm wondering because no one said this network is new and these errors are new and apparently showing only on some platforms. Swapping network hardware across platforms should be enough to test and answer #565

> We could not replace all USB device in a short time

If all USB devices are the same and affected, why is this problem new?

xiulipan:
So the root cause is also a mystery for us, here are some guess:
- kernel code base changed, driver regression
- random hardware issue
- broken single hardware 
- problematic/deprecated hardware

As we could not tell what happen with the hardware and these logs are harmless for our tests. I think filter them out would be nice idea and that is the initial propose for our `tools/sof-kernel-log-check.sh` script: **filter all error logs not related to our test.**

marc-hb:
>  filter all error logs not related to our test.

Networking is absolutely critical for our tests. We already wasted a LOT of time in the past with (unrelated) network or network-looking issues.

xiulipan:
@marc-hb Yes, network is critical. But as we can get those logs through network, I think those error logs are not that important and blocking our test cases.

@plbossart @keqiaozhang @aiChaoSONG @kv2019i 
Actually I think we are in a wrong direction now, we should only care error logs from AUDIO related drivers. Three will be un-limited error logs from unrelated modules: keyboard, wifi, BlueTooth, ME and etc. And we are backporting upstream changes every few weeks, all upstream regressions will influence our log check function.

tools/sof-kernel-log-check.sh
```diff
@@ -328,6 +318,15 @@ ignore_str="$ignore_str"'|asix .*: Failed to enable software MII access'
 # e1000e 0000:00:1f.6 enp0s31f6: Hardware Error
 ignore_str="$ignore_str"'|e1000e .*: Hardware Error'
 
+# iwlwifi net adaptor logs can be ignored
+# origin logs seen on CML platforms
+# iwlwifi 0000:00:14.3: Direct firmware load for iwlwifi-QuZ-a0-hr-b0-56.ucode failed with error -2
+# iwlwifi 0000:00:14.3: Direct firmware load for iwl-debug-yoyo.bin failed with error -2
+ignore_str="$ignore_str"'|iwlwifi .*: Direct firmware load for .* failed with error -2'
```
aiChaoSONG, line 325: For this kind of scenario, I suggest to use `.+`, because `*` matches to everything, and may not proper to use here. 
just
```
ignore_str="$ignore_str"'|iwlwifi .+'
```

aiChaoSONG (changes requested):

tools/sof-kernel-log-check.sh
xiulipan: Yep, I think we can ignore all iwlwifi related error logs from our discussion. and `.+` would be better.
Actually `.*` and `.+` will both match everything, but `.*` will match even when there is nothing while `.+` need at least one char.

xiulipan:
@marc-hb From our previous discussion, will update this PR to mention that all added filter rule will not break the network functionality and only show with PM.
And all `iwlwifi` error logs will be filtered.

tools/sof-kernel-log-check.sh
aiChaoSONG: The problem here is our ignore_str will look like `iwifi: .*|some-other-ignore-str`,  I am not sure if the `*` will eat the `|`, which used to connect two ignore_strs.

marc-hb:
> But as we can get those logs through network, I think those error logs are not that important

The network is not either working perfectly or not working at all. There are plenty intermediate and unreliable states and these are the most time consuming.

> will update this PR to mention that all added filter rule will not break the network functionality

Asserting this  requires careful examination of the error message and corresponding source code.

> And all iwlwifi error logs will be filtered.

Please submit the wifi patch separately. #565 is still being investigated.

marc-hb:
To filter all iwlwifi errors but only iwlwifi errors and nothing else:
```shell
echo 'kernel: iwlwifi 0000:00:14.3: No beacon heard and the time event is over already...' |
    grep -E '^kernel: iwlwifi [[:digit:].:]+: '
```

tools/sof-kernel-log-check.sh
marc-hb: The `|` is not really "connecting" anything, it's a logical OR part of the regular expression.
The `*` affects only the previous expression.

See regex suggestion in the main thread.

PR #583, opened 2021-01-19
tplgtool: implement enum kcontrol parsing function

aiChaoSONG:
Previously, the _enum_ctrl_parse() is left unimplemented
because we don't have enum kcontrol in existing topologies.

The new component TDFB is introduced recently, and requires
enum kcontrol to be parsed.

This patch implements the _enum_ctrl_parse function to
support enum kcontrol parsing.

### DIFF
--- a/tools/tplgtool.py
+++ b/tools/tplgtool.py
@@ -130,9 +130,56 @@ class TplgParser():
         bytes_data = bytes_data[156 + priv_size:]
         return mixer, bytes_data
 
-    # no such kcontrol in the binary tplg till now, leave this unimplemented
+    def _tplg_chan_parse(self, bytes_data):
+        chan_fields = ["size", "reg", "shift", "id"]
+        values = struct.unpack("IIII",bytes_data[0:16])
+        return dict(zip(chan_fields, values))
+
+    # parse snd_soc_tplg_enum_control struct
     def _enum_ctrl_parse(self, bytes_data):
-        pass
+        enum_ctrl_fields = ["size", "num_channels", "channel", "items", "mask", "count", \
+            "texts", "values", "priv"]
+        values = []
+        # size and num_channels
+        values.append(struct.unpack("I",bytes_data[0:4])[0])
+        values.append(struct.unpack("I",bytes_data[4:8])[0])
+        # channel
+        channel = []
+        SOC_TPLG_MAX_CHAN = 8
+        for idx in range(SOC_TPLG_MAX_CHAN):
+            channel.append(self._tplg_chan_parse(bytes_data[8+idx*16: 24+idx*16]))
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
@singalsu The previously version of this tool doesn't implement the enum kcontrol parsing feature, and caused the issue you raised up. This patch add the implementation, and align with the existing coding style.

@thesofproject/test-maintainers This tool can date back to the early days after I join Intel several months, this tool is my starter project to SOF and ALSA. the code may looks ugly because I am not very familiar with ASoC structures.  I have plan to optimize or rewrite this tool, don't review so much.

xiulipan (approved):
For byte reader, I will suggest you to take a look at https://github.com/thesofproject/sof/blob/master/tools/sof_ri_info/sof_ri_info.py

That would avoid to use hard code offset by read the binary as a stream.

marc-hb:
> The new component TDFB is introduced recently, and requires enum kcontrol to be parsed.

@aiChaoSONG  does this mean there will be some CI coverage for this new code? If not it will "bitrot".

> https://github.com/thesofproject/sof/blob/master/tools/sof_ri_info/sof_ri_info.py
> That would avoid to use hard code offset by read the binary as a stream.

@xiulipan do you mean relative offsets as opposed to absolute offsets in `bytes_data[]`?

 `sof_ri_info.py` would be much shorter and easier to read if it were using the reader concept for relative offsets _and_`struct` too inside its functions. For instance `parse_css_manifest_4()` could half the size if it were using `read_bytes()` at the start and then a few `struct` internally.

Afraid another problem with  `sof_ri_info.py` is the lack of test coverage, is there any?

aiChaoSONG:
> does this mean there will be some CI coverage for this new code? If not it will "bitrot"

@marc-hb Currently, no. All of our existing typologies don't contains any component that requires enum kcontrol, so this function is left unimplemented before, because it will never be called. But recently, @singalsu is working on TDFB component and finds the tool cannot work, so I add implementation in this PR, and test it with his topology. As stated, it is will not be called before any topology with enum kcontrol is used.
sof-cml-rt1011-rt5682-tdfb_68mm-2ch

marc-hb:
Thanks, so rephrasing my question: no automated TDFB test that will regularly run this PR in the (distant?) future? Just wondering.

aiChaoSONG:
> Thanks, so rephrasing my question: no automated TDFB test that will regularly run this PR in the (distant?) future? Just wondering.

It depends on when will Seppo's TDFB component get merged to SOF. once merged, we will have topology with TDFB which have enum kcontrol inside, so the code in this PR will tested daily.

singalsu (approved):
Thanks! The decoded graphical topology view is correct.

PR #584, opened 2021-01-21
opt.sh: refine option definition and overriding

aiChaoSONG:
1. move common option definition to a function, common options can be define here(eg, help, and tplg). TODO, call this in script explicitly(rather than in opt.sh), or some options may be hidden.
2. split  function in_func_create_tmpbash to _fill_opt_vars and _validate_and_format_cmd_line for better code structure.
3. fix unclear comments and add more comments

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -23,50 +23,36 @@ add_common_options()
 # option setup && parse function
 func_opt_parse_option()
 {
-    local _op_temp_script
-    local -A _op_short_lst _op_long_lst
-
     _func_case_dump_description()
     {
         grep '^##' "$SCRIPT_NAME" | sed 's/^## //g'
     }
 
-    # Asks getopt to validate command line input and generate $_op_temp_script
-    #
-    # Also initialize the global _op_short_list and _op_long_list
-    # "reversed maps" (reversed compared to the user's OPT_ maps) that
-    # map valid command line input to the corresponding
-    # one-character option code name. Example:
-    #   _op_short_lst=
-    #        ( [-t]="t" [-d]="d" [-l]="l" [-h]="h" )
-    #   _op_long_lst=
-    #        ( [--help]="h" [--duration]="d" [--tplg]="t" [--loop]="l" )
-    _func_create_tmpbash()
+    # This function helps to fill below four variables
+    # - short_opt_str/long_opt_str: short/long option in getopt format, used
+    #   for command line validation
+    # - short_opt_lst/long_opt_lst: list of short/long option map which maps
+    #   command line option to its corresponding short/long name, used for
+    #   overriding default OPT_VAL, eg, short_opt_lst=([-h]="h" [-t]="t")
+    #   long_opt_lst=([--help]="h" [--tplg]="t")
+    _fill_opt_vars()
     {
-         # options in getopt format
-        local i _short_opt _long_opt
-        for i in ${!OPT_DESC[*]}
+        local opt
+        for opt in ${!OPT_DESC[*]}
         do
-            _short_opt=$_short_opt"$i"
-            _op_short_lst["-$i"]="$i"
-            [ "$_long_opt" ] && _long_opt="$_long_opt"','
-            if [ "${OPT_NAME[$i]}" ]; then
-                _long_opt=$_long_opt"${OPT_NAME[$i]}"
-                _op_long_lst["--${OPT_NAME[$i]}"]="$i"
+            short_opt_str="$short_opt_str""$opt"
+            short_opt_lst["-$opt"]="$opt"
+            [ "$long_opt_str" ] && long_opt_str="$long_opt_str"','
+            if [ "${OPT_NAME[$opt]}" ]; then
+                long_opt_str="$long_opt_str""${OPT_NAME[$opt]}"
+                long_opt_lst["--${OPT_NAME[$opt]}"]="$opt"
             fi
-            # append ':' if the option accepts an argument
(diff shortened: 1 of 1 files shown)


case-lib/opt.sh
```diff
@@ -13,48 +13,46 @@ declare -A OPT_NAME OPT_HAS_ARG OPT_VAL OPT_DESC
...
+            }
         done
+    }
 
-        if  ! _op_temp_script=$(
-                getopt -o "$_short_opt" --long "$_long_opt" -- "$@"); then
-            # here output for the End-User who don't care about this function/option couldn't to run
-            printf 'Error parsing options: %s\n' "$*"
-            # To debug uncomment these lines:
-            # printf 'DEBUG parsing options: short_opt: %s / long_opt: %s / args: %s\n' "$_short_opt" "$_long_opt" "$*" >&2
```
marc-hb, line 62: I left this `printf DEBUG` because it found it very useful, please keep it.

case-lib/opt.sh
```diff
@@ -98,21 +96,26 @@ func_opt_parse_option()
...
+    # the option map used for OPT_VAL overriding
+    local -A short_opt_lst long_opt_lst
+
+    _fill_opt_vars
+    _validate_and_format_cmd_line "$@"
 
-    # FIXME: what is this supposed to do!?
-    eval set -- "$_op_temp_script"
+    # set the contents in formatted_cmd_opts to position arguments $1, $2, ...
+    eval set -- "$formatted_cmd_opts"
```
marc-hb, line 115: Thanks for the new comment! Can you add a (small) `"$formatted_cmd_opts"` example here?

case-lib/opt.sh
```diff
@@ -13,48 +13,46 @@ declare -A OPT_NAME OPT_HAS_ARG OPT_VAL OPT_DESC
...
-    # Asks getopt to validate command line input and generate $_op_temp_script
-    #
-    # Also initialize the global _op_short_list and _op_long_list
-    # "reversed maps" (reversed compared to the user's OPT_ maps) that
-    # map valid command line input to the corresponding
-    # one-character option code name. Example:
-    #   _op_short_lst=
-    #        ( [-t]="t" [-d]="d" [-l]="l" [-h]="h" )
-    #   _op_long_lst=
-    #        ( [--help]="h" [--duration]="d" [--tplg]="t" [--loop]="l" )
```
marc-hb, line 39: I don't see your replacement for this comment as an improvement, you removed a fair bit for no clear reason and you broke the indentation.

marc-hb (changes requested):
Thanks for cleaning up this code, I always found it very hard to follow.

I see there are only 2 shellcheck warnings left, they seem easy to fix and you're already touching that code anyway.


(pylint failed because of some apt-get failure; there's no python code change in this PR)

case-lib/opt.sh
xiulipan: I think the intend is fixed in `_func_case_dump_description`

case-lib/opt.sh
```diff
@@ -71,71 +60,93 @@ func_opt_parse_option()
...
-            printf '\tthis message\n'
-            _func_case_dump_description
-            trap - EXIT
-            exit 2
-        }
+        printf '    -h |  --help\n'
+        printf '\tthis message\n'
+        _func_case_dump_description
+        trap - EXIT
+        exit 1
```
xiulipan, line 91: I am thinking about we return as error with help message?

xiulipan (changes requested):

case-lib/opt.sh
aiChaoSONG: @xiulipan agreed on this. this function is called if `-h` is specified, or command line parsing error. we should exit success for the first condition and failure for the second condition.

xiulipan (dismissed):

aiChaoSONG:
This failures on TGLH and BYT has nothing to do with this code change.

case-lib/opt.sh
```diff
@@ -146,5 +163,15 @@ func_opt_parse_option()
...
+}
+
+# Define common options among test scripts.
+add_common_options()
+{
+    # The help option
+    OPT_NAME['h']='help'
+    OPT_HAS_ARG['h']=0
+    OPT_VAL['h']=0
+    OPT_DESC['h']='show help information'
```
marc-hb, line 176: Can you move this function back to the top of the file? It's a great example so a great way to start the file.

case-lib/opt.sh
```diff
@@ -1,141 +1,158 @@
...
+                [ "X${OPT_HAS_ARG[$i]}" != "X1" ] || printf ' parameter'
+            fi
+            printf '\n\t%s\n' "${OPT_DESC[$i]}"
+            if [ "${OPT_VAL[$i]}" ]; then
+                if [ "${OPT_HAS_ARG[$i]}" -eq 1 ]; then
+                    printf '\tDefault Value: %s\n' "${OPT_VAL[$i]}"
+                elif [ "${OPT_VAL[$i]}" -eq 0 ]; then
+                    printf '\tDefault Value: Off\n'
+                else
+                    printf '\tDefault Value: On\n'
```
marc-hb, line 84: Next time please don't mix in the same PR such a big whitespace change with actual code changes because it's practically impossible to see whether any code has changed there.

marc-hb (approved):

xiulipan:
@aiChaoSONG Please try to update the simple fix as @marc-hb required and merge or merge directly and send FIX later.

marc-hb (approved):

xiulipan (approved):

PR #587, opened 2021-01-22
github-actions: update repository before installing

aiChaoSONG:
Signed-off-by: Chao Song <chao.song@intel.com>

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: actions/setup-python@v2
       - name: get python libs
         # FIXME: apt-get succeeds but 'import numpy' still fails!?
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
this PR should help to fix pylint issue, https://github.com/thesofproject/sof-test/pull/584/checks?check_run_id=1747087212

marc-hb:
Good catch. It's very unfortunate we have to do `apt update`, it's very slow and discouraged: https://github.com/actions/virtual-environments/issues/1757

This #588 alternative is 30% faster and it is more likely to break in the future which will give us a reminder to remove the hack and be fast again.

xiulipan:
close as we merged the workaround https://github.com/thesofproject/sof-test/pull/588

PR #589, opened 2021-01-22
kmod: work-around dependency changes

plbossart:
PR https://github.com/thesofproject/linux/pull/2683 introduces a
dependency change. There is still one case there a module is used
differently before and after this PR, and renaming isn't really an
option.

This patch suggests a temporary workaround where we try to remove a
module (with errors disabled), and if it doesn't work we keep going.
In practice the snd_sof_hda_common module is made mostly of helper
routines and tables, without any PCI stuff, so there is a limited risk
of issues.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -14,6 +14,14 @@ remove_module() {
     fi
 }
 
+try_remove_module() {
+    set +e
+
+    remove_module $1
+
+    set -e
+ }
+
 # SOF CI has a dependency on usb audio
 remove_module snd_usb_audio
 
@@ -29,6 +37,11 @@ remove_module snd_sof_pci_intel_icl
 remove_module snd_sof_pci_intel_tgl
 remove_module snd_sof_acpi_intel_byt
 remove_module snd_sof_acpi_intel_bdw
+
+#-------------------------------------------
+# temporary dependency
+#-------------------------------------------
+try_remove_module snd_sof_intel_hda_common
 
 #-------------------------------------------
 # Helpers


tools/kmod/sof_remove.sh
```diff
@@ -30,6 +38,11 @@ remove_module snd_sof_pci_intel_tgl
 remove_module snd_sof_acpi_intel_byt
 remove_module snd_sof_acpi_intel_bdw
 
+#-------------------------------------------
+# temporary dependency
+#-------------------------------------------
+try_remove_module snd_sof_intel_hda_common
```
marc-hb, line 44: If this is temporary then the new function is overkill, this is enough and IMHO faster to understand because no indirection:
```shell
( set +e; remove_module snd_sof_intel_hda_common )
```

marc-hb (commented):
How temporary is this? In other words, for how long do we keeping testing old kernels?

marc-hb (changes requested):
`set +e; cmds; set -e` means no one can temporarily disable `set -e` for the whole script any more. A subshell `(set +e; cmds )` solves this, or in this simple case the much simpler:
```diff
- remove_module snd_sof_intel_hda_common
+ # need this  until we stop testing kernel versions < 5.x
+ # because of https://github.com/thesofproject/linux/pull/2683
+ remove_module snd_sof_intel_hda_common || true
```

EDIT: The subshell `(set +e ; cmds)` technique works... except for the last of the `cmds`. So it does not work at all here. Use `|| true` as usual.

plbossart:
I for one have no idea what 'cmd || true' could possibly mean, and we've encouraged others developers NOT to use this pattern in the past.

I still don't see what was wrong with my proposal. if you enable error checks in all cases, the script will fail. That's precisely what we want to prevent. There is a known failure case which needs to be squelched.

marc-hb:
> I for one have no idea what 'cmd || true' could possibly mean,

It runs `cmd` and returns success (`true`) whether `cmd` fails or not. This is the shortest, simplest and most common way to solve this problem.

```shell
sudo grep -r -F '|| true' /etc | wc -l
105 # 105 lines in 30 files
```

> and we've encouraged others developers NOT to use this pattern in the past.

I don't remember this.

> I still don't see what was wrong with my proposal.

It's not local and changes the global state.

marc-hb:
Re-submitted in #594

PR #590, opened 2021-01-22
sof-kernel-log-check: ignore ACPI _ADR workaround

plbossart:
PR https://github.com/thesofproject/linux/pull/2701 introduces a
workaround which leads to errors being thrown when the DSDT contains a
device with the 0x701 part_id.

Ignore the error since it's a normal error flow.

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -310,6 +310,11 @@ ignore_str="$ignore_str"'|proc_thermal 0000:00:..\..: No auxiliary DTSs enabled'
 # elan_i2c i2c-ELAN0000:00: invalid report id data (ff)
 ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 
+# SoundWire workaround using RT701
+# https://github.com/thesofproject/linux/pull/2701
+ignore_str="$ignore_str"'|rt711 sdw:[0-3]:25d:701:0: Probe of rt711 failed: -19'
+ignore_str="$ignore_str"'|rt1308 sdw:[0-3]:25d:701:0: Probe of rt1308 failed: -19'
+
 #
 # SDW related logs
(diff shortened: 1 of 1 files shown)


plbossart:
@marc-hb GitActions issue here unrelated to this PR?

tools/sof-kernel-log-check.sh
```diff
@@ -310,6 +310,11 @@ ignore_str="$ignore_str"'|proc_thermal 0000:00:..\..: No auxiliary DTSs enabled'
 # elan_i2c i2c-ELAN0000:00: invalid report id data (ff)
 ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 
+# SoundWire workaround using RT701
+# https://github.com/thesofproject/linux/pull/2701
+ignore_str="$ignore_str"'|rt711 sdw:[0-3]:25d:701:0: Probe of rt711 failed: -19'
+ignore_str="$ignore_str"'|rt1308 sdw:[0-3]:25d:701:0: Probe of rt1308 failed: -19'
+
```
marc-hb, line 317: Can you guard this with:
```shell
if [ "$platform" = xxx ] ; then
  ignore_str="$ignore_str"'|rt711 ...'
  ignore_str="$ignore_str"'|rt1308 ...'
fi 
```
The PR URL is super useful but... very long! Can you also add just a couple words like:
```shell
# some BIOSes are  buggy so we need to probe like mad.
```
This helps a lot with future maintenance of this file.

marc-hb (changes requested):
> @marc-hb GitActions issue here unrelated to this PR?

Pylint failure is indeed unrelated; you didn't submit any Python. The Github Ubuntu VM is out of date, see #588 hack if you're interested (you're not)

plbossart:
no longer needed, we chose a different implementation in the Linux kernel.

PR #592, opened 2021-01-25
sof_remove.sh: show pulseaudio on exit if it is running

marc-hb:
Even when idle, pulseaudio uses modules "non-deterministically". So even
on success show when it is running cause it could make this script fail
the next time.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -14,6 +14,22 @@ remove_module() {
     fi
 }
 
+exit_handler()
+{
+    local exit_status="$1"
+    # Even without any active audio, pulseaudio can use modules
+    # "non-deterministically". So even if we are successful this time,
+    # warn about any running pulseaudio because it could make us fail
+    # the next time.
+    if pgrep -a pulseaudio; then
+        printf 'ERROR: %s fails semi-randomly when pulseaudio is running\n' \
+               "$(basename "$0")"
+    fi
+    return "$exit_status"
+}
+
+trap 'exit_handler $?' EXIT
+
 # SOF CI has a dependency on usb audio
 remove_module snd_usb_audio
 


xiulipan:
@keqiaozhang @aiChaoSONG I think we try to kill pulseaudio in our test case already right?
Any comments about this tool change? Should we add kill pulseaudio also for this tool for independent run or have some notice that kmod remove/insert should be run with pulseaudio killed.

The comment is hard to understand to me. I think it wants to say _check if pulseaudio is alive if kmod remove fail_?

PS: the actual issue with pulseaudio is that it will try to go through all PCM when you probe the SOF, if you do not wait for idle after kmod_insert, DSP will be busy and kmod will fail. Maybe dump the DSP status/use_count will also help here.

keqiaozhang:
@xiulipan ，we remove the pulseaudio completely when setting up the DUT（in depsetup.sh). therefore, we shouldn't meet such situation if we set up the DUT properly

marc-hb:
This patch is not meant for DUTs, it's for developers. This is just zero or one line of additional logging, no logical change.

xiulipan:
I think we'd better show info list **please disable pulseaudio**

marc-hb:
> I think we'd better show info list please disable pulseaudio

OK, I will add a short `printf`. I'm also trying to keep this change relatively small because as you said DUTs should never have pulseaudio running when running this, this is meant to be just a developer convenience. I added this clue because it took me some time to guess why this script was sometimes passing and sometimes failing. I have disabled everything pulseaudio in `systemctl --user` but pulseaudio is always stronger than you!

xiulipan:
@marc-hb Also some tips from https://github.com/thesofproject/sof-test/pull/591
an active logger thread will also block the kmod_remove, as well as any aplay process

I think we may dump full ps -aux here to help debug why kmod could not be removed.

marc-hb:
Unlike pulseaudio, the logger does not magically and secretly spawn after you tried everything to stop it.

xiulipan:
@marc-hb Please add more context info in the output message. We need to mention that `pulseaudio` is protected by system and could not be easily killed. 
Please also update the commit and comment to mention the issue with `pulseaudio`. 

PS: I got into difficult to understand what this sentence means. Could you help to explain or refine?

> So even on success show when it is running cause it could make this script fail the next time.

marc-hb:
> PS: I got into difficult to understand what this sentence means. Could you help to explain or refine?

Rephrased.

xiulipan (approved):

aiChaoSONG (approved):

marc-hb:
https://sof-ci.01.org/softestpr/PR592/build574/devicetest/ is a TGL NVMe storage issue

Issue #514, opened 2020-11-13
Some USB noise need to be filtered out


aiChaoSONG:
check report 732 on CML_OLP_RT711 suspend case
```
[ 7160.094705] usb 2-4.1: Disable of device-initiated U1 failed.
[ 7160.469352] usb 2-4.1: Disable of device-initiated U2 failed.
[ 7171.776533] usb 2-4.1: device not accepting address 3, error -62
```

aiChaoSONG:
@marc-hb I just realized that our filters for some USB noise don't work,  because the `.` only match a single character. maybe change `.` to `*`, or we specify the whole false failure text? 
```
ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U1 failed\.'
ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U2 failed\.'

# CHT devices with USB hub, issues reported by sof-test
# BugLink: https://github.com/thesofproject/sof-test/issues/431
ignore_str="$ignore_str"'|hub 2-.: .'
ignore_str="$ignore_str"'|usb 2-.: .'

# TGL devices with USB 3.1 devices, issues reported by sof-test
# BugLink: https://github.com/thesofproject/sof-test/issues/482
ignore_str="$ignore_str"'|usb 3-.: device descriptor read/64, error .'
ignore_str="$ignore_str"'|usb 3-.: device not accepting address ., error .'
```

marc-hb:
> maybe change . to *

You mean change `.` to `.*`?  Or better, to `[[:digit:]-.]*`

```
ignore_str="$ignore_str"'|hub 2-.: .'
ignore_str="$ignore_str"'|usb 2-.: .'

ignore_str="$ignore_str"'|usb 3-.: device descriptor read/64, error .'
ignore_str="$ignore_str"'|usb 3-.: device not accepting address ., error .'
```
In the lack of a final `$`, a last `.` is pointless and should be removed. It sets a bad example.

Use `fake_kern_error()` to test regular expressions like this. Regular expressions are extremely error-prone, can never be tested too much. You can stuff a single, quick test with dozens of `fake_kern_error`

Is this an issue with the CML system? If yes did we report it somewhere? If not then how about we stop using broken USB peripherals?

aiChaoSONG:
@marc-hb In fact, our USB peripherals work, hard to tell which is broken.  I have see usb error logs with same message but different usb version or error code, same usb version but different message, this make the filter RE hard to write.

[...]

Issue #593, opened 2021-01-26
Failure to remove snd_hwdep module in kmod test on CML_HEL_RT5682


keqiaozhang:
Error message in kmod test: 
```
Removing snd_hwdep
rmmod: ERROR: Module snd_hwdep is in use by: snd_usb_audio
```

No error trace available.
How to reproduce:
```
$ /sof-test/test-case/check-kmod-load-unload.sh -l 50
```

marc-hb:
This happened in nightly run 1807. Thanks to @xiulipan's advice I looked at `dmesg` and this seems to be the first and real issue:
```
kernel: [ 8823.371612] sof-audio-pci 0000:00:1f.3: FW Poll Status: reg=0xf0f successful
kernel: [ 8823.371618] sof-audio-pci 0000:00:1f.3: DSP core(s) enabled? 0 : core_mask 1
kernel: [ 8824.513526] usb 1-2.3: reset full-speed USB device number 6 using xhci_hcd
kernel: [ 8824.591467] usb 1-2.3: device firmware changed
kernel: [ 8824.592274] usb 1-2.3: USB disconnect, device number 6
kernel: [ 8824.677281] usb 1-2.3: new full-speed USB device number 7 using xhci_hcd
kernel: [ 8824.763718] usb 1-2.3: New USB device found, idVendor=0d8c, idProduct=0014, bcdDevice= 1.00
kernel: [ 8824.763728] usb 1-2.3: New USB device strings: Mfr=1, Product=2, SerialNumber=0
kernel: [ 8824.763732] usb 1-2.3: Manufacturer: C-Media Electronics Inc.
kernel: [ 8824.773029] input: C-Media Electronics Inc. as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2.3/1-2.3:1.3/0003:0D8C:0014.0003/input/input40
kernel: [ 8824.825245] hid-generic 0003:0D8C:0014.0003: input,hidraw1: USB HID v1.00 Device [C-Media Electronics Inc.] on usb-0000:00:14.0-2.3/input3
kernel: [ 8824.876500] usbcore: registered new interface driver snd-usb-audio
kernel: [ 8825.867711] usb 1-2.3: reset full-speed USB device number 7 using xhci_hcd
kernel: [ 8825.935271] usb 1-2.3: device descriptor read/64, error 8
kernel: [ 8826.119759] usb 1-2.3: device descriptor read/all, error -32
[...]

PR #595, opened 2021-01-27
kernel-log-check: revise ignore_str for usb errors.

aiChaoSONG:
Fixes: #431 #514

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -164,7 +164,7 @@ ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send\(\): error -5'
 ignore_str="$ignore_str"'|platform regulatory\.0: Direct firmware load for regulatory\.db failed with error -2'
 ignore_str="$ignore_str"'|cfg80211: failed to load regulatory\.db'
 ignore_str="$ignore_str"'|EXT4-fs \(nvme0n1p6\): re-mounted\. Opts: errors=remount-ro'
-ignore_str="$ignore_str"'|usb .+: Enable of device-initiated .+ failed\.'
+ignore_str="$ignore_str"'|usb .+: (Enable|Disable) of device-initiated .+ failed\.'
 ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zone \(-61\)'
 
 # Dell CML-U laptop with SoundWire, issues reported by sof-test
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -164,8 +164,7 @@ ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send\(\): error -5'
 ignore_str="$ignore_str"'|platform regulatory\.0: Direct firmware load for regulatory\.db failed with error -2'
 ignore_str="$ignore_str"'|cfg80211: failed to load regulatory\.db'
 ignore_str="$ignore_str"'|EXT4-fs \(nvme0n1p6\): re-mounted\. Opts: errors=remount-ro'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U1 failed\.'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U2 failed\.'
+ignore_str="$ignore_str"'|usb .+: Enable of device-initiated .+ failed\.'
 ignore_str="$ignore_str"'|thermal thermal_zone.*: failed to read out thermal zone \(-61\)'
 
 # Dell CML-U laptop with SoundWire, issues reported by sof-test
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -164,8 +164,7 @@ ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send\(\): error -5'
 ignore_str="$ignore_str"'|platform regulatory\.0: Direct firmware load for regulatory\.db failed with error -2'
 ignore_str="$ignore_str"'|cfg80211: failed to load regulatory\.db'
 ignore_str="$ignore_str"'|EXT4-fs \(nvme0n1p6\): re-mounted\. Opts: errors=remount-ro'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U1 failed\.'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U2 failed\.'
+ignore_str="$ignore_str"'|usb .-.+: (Enable|Disable) of device-initiated .+ failed\.'
```
xiulipan, line 167:
```suggestion
ignore_str="$ignore_str"'|usb .+: (Enable|Disable) of device-initiated .+ failed\.'
```

xiulipan (changes requested):
Interesting here, what is the `2-` means here? USB 2 or USB bus 2?
Do we want to ignore all USB 2 and 3 issues?

better with 

```
ignore_str="$ignore_str"'|hub [23]-.+:'
ignore_str="$ignore_str"'|usb [23]-.+:'
```
And remove all other duplicate filters

tools/sof-kernel-log-check.sh
aiChaoSONG: we can ignore error messages for usb hub, but not all usb. for hub, I use `hub [23]-.+'`, and for usb, I update to use specific message `usb [23]-.+: Failed to suspend device, error` from #431.

marc-hb:
> Interesting here, what is the 2- means here? USB 2 or USB bus 2?

bus 2, check `lsusb -tv`

> ignore_str="$ignore_str"'|usb [23]+-.+:'

No this would ignore absolutely all USB errors, including from USB audio devices (e.g. #593) and from USB network devices that we use.

tools/sof-kernel-log-check.sh
```diff
@@ -164,8 +164,7 @@ ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send\(\): error -5'
 ignore_str="$ignore_str"'|platform regulatory\.0: Direct firmware load for regulatory\.db failed with error -2'
 ignore_str="$ignore_str"'|cfg80211: failed to load regulatory\.db'
 ignore_str="$ignore_str"'|EXT4-fs \(nvme0n1p6\): re-mounted\. Opts: errors=remount-ro'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U1 failed\.'
-ignore_str="$ignore_str"'|usb 2-3: Enable of device-initiated U2 failed\.'
+ignore_str="$ignore_str"'|usb .+: (Enable|Disable) of device-initiated .+ failed\.'
```
xiulipan, line 167: I will do some clean up later to merge filter by model.

xiulipan (approved):
Then I think we filter all hub issues and some usb issues

marc-hb:
Let's first stabilize the current test results with small, careful and focused changes. Once the filters have not changed a while then we can start from a "steady green base" and look at refactorings, cleanups and other simplifications.

Continuous Integration = small changes spread over time.

xiulipan:
@marc-hb I see nothing with CI, are you good to merge this?

plbossart (dismissed):

marc-hb (approved):

marc-hb:
@aiChaoSONG there is git conflict now, probably because I just merged #624 sorry. Should be small.

aiChaoSONG:
@marc-hb rebased, thanks.

tools/sof-kernel-log-check.sh
```diff
@@ -189,8 +188,8 @@ ignore_str="$ignore_str"'|\[drm:drm_dp_send_link_address \[drm_kms_helper\]\] \*
 
 # CHT devices with USB hub, issues reported by sof-test
 # BugLink: https://github.com/thesofproject/sof-test/issues/431
-ignore_str="$ignore_str"'|hub 2-.: .'
-ignore_str="$ignore_str"'|usb 2-.: .'
+ignore_str="$ignore_str"'|hub [23]-.+'
+ignore_str="$ignore_str"'|usb [23]-.+: Failed to suspend device, error'
```
marc-hb, line 192: I think the first line ignores everything that the second line does, so the second line is pointless.

The first line has two issues: 1. it ignores all USB errors on devices `2-9: ...`, it ignores zero error on devices like `2-10: ...`

> As we want to ignore all hub error but not all usb errors, ...

I don't think there is any concept of "hub error" vs "other errors", all errors are prefixed the same.

Can you please ignore only the errors reported in #431 ?

aiChaoSONG:
The third-party issues is not seen even this is not merged, close.

keqiaozhang (approved):

PR #598, opened 2021-01-29
tplgtool.py: print exception detail for debug

aiChaoSONG:
Remove try..catch block, and let exceptions raise
up for debug purpose.

Signed-off-by: Chao Song <chao.song@intel.com>

### DIFF
--- a/tools/tplgtool.py
+++ b/tools/tplgtool.py
@@ -539,12 +539,8 @@ class TplgParser():
         return block
 
     def parse(self,tplg_file):
-        try:
-            with open(tplg_file,"rb") as fd:
-                self._tplg_binary = fd.read()
-        except:
-            print("File %s open error" %tplg_file)
-            sys.exit(1)
+        with open(tplg_file,"rb") as fd:
+            self._tplg_binary = fd.read()
 
         # here we call a header with its data a block
         parsed_tplg = []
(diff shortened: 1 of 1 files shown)


keqiaozhang (approved):

xiulipan (changes requested):
I think this PR should be 2 commits to add try catch but not print exception:
1. add try catch for file open
2. add try catch for import

aiChaoSONG:
> I think this PR should be 2 commits to add try catch but not print exception:

@xiulipan I think this is redundant work. What we want to do is to print error and exit. without try..catch, this is done. why bother to add them.

marc-hb (approved):
The first `except` did look silly, glad to see it go.

Both blocks caught any exception and used `sys.exit()` which hides backtraces, glad to see that go too.

However, the second `except` had some information: the name of the `python3-graphviz` package (which is the same in Fedora, I just checked). Have you considered something like this:
```python
try
  from graphviz import Digraph
except ModuleNotFoundError as mnfe:
  print("graphviz package not found, try `sudo apt install python3-graphviz`")	
  raise mnfe
```

aiChaoSONG:
> Have you considered something like this:

@marc-hb Yes, but I think the trace message is already good to tell us that `graphviz` is not installed. and graphviz can be installed with `sudo apt install python3-graphviz` or `pip3 install graphviz`, let's not limit how the user will install this package.

xiulipan (changes requested):
> > I think this PR should be 2 commits to add try catch but not print exception:
> 
> @xiulipan I think this is redundant work. What we want to do is to print error and exit. without try..catch, this is done. why bother to add them.

Sorry, I disagree here. The try catch is not use only for debug. This PR to me is also user experience improvement for 2 different use case here:

1. file issue: (permission, file path and other)
2. import lib missing.

so I insist to spilt the commit and create more appropriate commit for each change.

xiulipan (approved):
It seem I made a mistake about deleting or adding try catch.

aiChaoSONG:
The failures have nothing to do with this PR, merging

PR #600, opened 2021-02-01
userspace: add verify-ucm-bootsequence

libinyang:
This test case will check the UCM2 codec initialization setting
on the tested cards.

Signed-off-by: Libin Yang <libin.yang@intel.com>

### DIFF
--- a/case-lib/user.sh
+++ b/case-lib/user.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# function dpkg_version_ge uses dpkg to get the package version
+# and compares it with the specified version. If the package version
+# is less than the specified version, it means the version dependency
+# is not met and return 1.
+# Args: $1: package name
+#       $2: specified version
+dpkg_version_ge()
+{
+    local ver
+    ver=$(dpkg -l "$1" |grep "$1" | awk '{print $3}')
+    [[ -z "$ver" ]] && die "failed to get $1 version"
+    dlogi "find $1 version  $ver"
+    version_ge "$ver" "$2"
+}
+
+# function version_ge compares the 2 versions specified in the parameters.
+version_ge()
+{
+    printf '%s\n' "$2" "$1" | sort -V -C
+}
--- a/test-case/verify-bootsequence.sh
+++ b/test-case/verify-bootsequence.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+##
+## Case Name: verify-bootsequence.sh
+## Preconditions:
+##    UCMv2 is installed in /usr/share/alsa/ucm2 folder
+## Description:
+##    Run alsactl init to test the initialization settings are correct or not
+## Case step:
+##    1. check the alsa-lib, alsa-utils version
+##    2. save the original amixer settings
+##    3. run alsactl init
+##    4. restore the original amixer settings
+## Expect result:
+##    alsactl init runs successfully
+##
+
+set -e
+
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+# shellcheck source=case-lib/user.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/user.sh
+
+# BootSequence is introduced in alsa-utils 1.2.2-16
+ALSAUTILS_VER=1.2.2-16
+# BootSequence is introduced in alsa-lib 1.2.2-44
+ALSALIB_VER=1.2.2-44
+# BootSequence is introduced in alsa-ucm-conf 1.2.3-15
+ALSAUCM_VER=1.2.3-15
+
+func_opt_parse_option "$@"
+
+main()
+{
+    local tmp
+
+    dpkg_version_ge "libasound2" "$ALSALIB_VER"
+    dpkg_version_ge "alsa-utils" "$ALSAUTILS_VER"
+    dpkg_version_ge "alsa-ucm-conf" "$ALSAUCM_VER" || dlogi "alsa-ucm-conf version is too low"
+
+    # alsactl init only applies to SOF card
+    [[ -n "$SOFCARD" ]] || die "No SOF card"
+    cardname=$(sof-dump-status.py -s "$SOFCARD") || exit 1
+    [[ -n "$cardname" ]] || die "Failed to get card $SOFCARD short name"
+
+    tmp=$(mktemp -d) || die "Failed to create tmp folder."
+    cd "$tmp"
+    dlogi "Indirectly checking the card $cardname UCM2 initial settings (if any) with alsactl init"
+    alsactl store -f tmp.conf || die "Failed to save the amixer setting"
+    dlogc "alsactl init $SOFCARD"
+    alsactl init "$SOFCARD" || die "Failed to run alsactl init"
+    alsactl restore -f tmp.conf || die "Failed to restore the amixer setting"
+}
+
+main "$@"


test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+##
+## Case Name: verify-ucm-bootsequence.sh
+## Preconditions:
+##    N/A
```
xiulipan, line 6: Is UCM file needed here?

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,71 @@
...
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+
+# BootSequence is introduced in alsa-utils 1.2.2-16
+ALSAUTILS_VER=1.2.2-16
+# BootSequence is introduced in alsa-lib 1.2.2-44
+ALSALIB_VER=1.2.2-44
+
+func_opt_parse_option "$@"
+
+check_alsa_lib_ver()
```
xiulipan, line 30: Can we make this into some lib? As I suppose we will have more test case with ALSA version requirement.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,71 @@
...
+{
+    alsalib_ver=$(dpkg -l libasound2 | grep libasound2 |awk '{print $3}')
```
xiulipan, line 32: will all OS use dpkg? What about user build bianry?

xiulipan (changes requested):
Some question about the alsa version check. Let's try to remove OS dependency here.

test-case/verify-ucm-bootsequence.sh
libinyang: Yes, I will update it.

test-case/verify-ucm-bootsequence.sh
libinyang: OK

test-case/verify-ucm-bootsequence.sh
libinyang: This is not supported because we can't accurate versions of alsa-lib if user uses build binary.

marc-hb (commented):
Some codestyle suggestions.

I don't know anything about UCM: is there nothing in `/proc`, `/sys` or in some log file that can be displayed to prove that something related to UCM happened?

libinyang:
> Some codestyle suggestions.
> 
> I don't know anything about UCM: is there nothing in `/proc`, `/sys` or in some log file that can be displayed to prove that something related to UCM happened?

@marc-hb UCMv2 is located in /usr/share/alsa/ucm2. It is a configuration file set, which is used by other applications, such as pulseaudio. This patch is  to check the BootSequence is correct or not. UCMv2 introduced a BootSequence, which will be used at the first system boot up to set the codecs kcontrols to a proper value.

case-lib/user.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# function func_check_version uses dpkg to get the package version
+# and compares it with the specified version. If the package version
+# is less than the specified version, it means the version dependency
+# is not met.
+# Args: $1: package name
+#       $2: specifiedversion
+func_check_version()
```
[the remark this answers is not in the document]
libinyang: > To prepare the code for other distributions you could use `sort -V` https://unix.stackexchange.com/questions/285924/how-to-compare-a-programs-version-in-a-shell-script

Do you mean something like:
```
    if [ "$(printf '%s\n' "$2" "$ver" | sort -V | head -n1)" = "$ver" ]; then
        die "$1 version $ver doesn't meet the requirement >= $2"
    fi
```

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if tmp=$(mktemp -d); then
+    cd "$tmp"
+    dlogi "Checking the card $cardname UCM2 initial settings"
+    alsactl store -f tmp.conf || die "Failed to save the amixer setting"
+    dlogc "alsactl init $SOFCARD"
+    alsactl init "$SOFCARD" || die "Failed to run alsactl init"
+    alsactl restore -f tmp.conf || die "Failed to restore the amixer setting"
+else
+    die "Failed to create tmp folder."
```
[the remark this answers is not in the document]
libinyang: ACK. I will update it.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,52 @@
...
+# alsactl init only applies to SOF card
+[[ -z "$SOFCARD" ]] && die "No SOF card"
+if cardname=$(sof-dump-status.py -s "$SOFCARD"); then
+    [[ -z "$cardname" ]] && die "Failed to get card $SOFCARD short name"
+else
+    exit 1
+fi
+
+if tmp=$(mktemp -d); then
+    cd "$tmp"
```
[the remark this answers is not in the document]
libinyang: ACK. I will update it.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,52 @@
...
+# BootSequence is introduced in alsa-lib 1.2.2-44
+ALSALIB_VER=1.2.2-44
+
+func_opt_parse_option "$@"
+
+func_check_version "libasound2" "$ALSALIB_VER" || exit 1
+func_check_version "alsa-utils" "$ALSAUTILS_VER" || exit 1
+
+# alsactl init only applies to SOF card
+[[ -z "$SOFCARD" ]] && die "No SOF card"
```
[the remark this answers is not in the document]
libinyang: ACK. I will update it.

case-lib/user.sh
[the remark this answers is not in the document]
marc-hb: Yes I meant something looking like this.

I recommend a new "version_greater_or_equal()" or "version_ge()" function that would not `die` but just `test` and that would be OS-independent. This would make it easier to port to ChromeOS or other later.

marc-hb (commented):
Thanks @libinyang , rephrasing my question: is it possible for this test to PASS even on a system that does not use any UCM at all? I'm asking because none of the `alsactl` command seem to depend on UCM.

libinyang:
> Thanks @libinyang , rephrasing my question: is it possible for this test to PASS even on a system that does not use any UCM at all? I'm asking because none of the `alsactl` command seem to depend on UCM.

@marc-hb UCM is installed automatically when installing the OS. But the old UCM may not support `BootSequence`. In this case, I think we need also add the UCM version check and if the UCM version doesn't meet the requirement, it should SKIP this test case with `dlogi` info. In some situations, for example, codec RT1308 doesn't have `BootSequence`, `alsactl init` can also handle this situation smoothly. `BootSequence` is not a must feature. But if it is set, `alsactl init` can detect whether the setting can be successfully applied or not.

I will add the code to check the UCM version in the update patch.

case-lib/user.sh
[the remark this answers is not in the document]
libinyang: > Yes I meant something looking like this.
> 
> I recommend a new "version_greater_or_equal()" or "version_ge()" function that would not `die` but just `test` and that would be OS-independent. This would make it easier to port to ChromeOS or other later.

@marc-hb But we still need `dpkg -l` to get the package version. So it seems to be hard be OS-independent. `aplay --version` can get the aplay version. But the version is not accurate enough. And it seems hard to get the alsa-lib version without `dpkg`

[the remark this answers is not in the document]
marc-hb: I wrote "easier to make OS-independent LATER", not now not in this PR. 

Never duplicate any code, that's why we have functions. The suggestion is: `main() -> check_version() -> compare_version()`

The 3rd one is OS-independent and does not die. The 2nd one is `dpkg` specific (for now) and it may die.

case-lib/user.sh
```diff
@@ -0,0 +1,33 @@
...
+# is not met and return 1.
+# Args: $1: package name
+#       $2: specifiedversion
+func_version_ge()
+{
+    ver=$(dpkg -l "$1" |grep "$1" | awk '{print $3}')
+    [[ -z "$ver" ]] && die "failed to get $1 version"
+    dlogi "find $1 version  $ver"
+    if [ "$(printf '%s\n' "$2" "$ver" | sort -V | head -n1)" = "$ver" ]; then
+	return 1
```
marc-hb, line 31: You don't need `return`, the return value of a function is the return value of the last command. Just run `[ ]` (or `test` ) last.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,55 @@
...
+
+func_check_version "libasound2" "$ALSALIB_VER" || exit 1
+func_check_version "alsa-utils" "$ALSAUTILS_VER" || exit 1
+
+# alsactl init only applies to SOF card
+[[ -n "$SOFCARD" ]] || die "No SOF card"
+if cardname=$(sof-dump-status.py -s "$SOFCARD"); then
+    [[ -z "$cardname" ]] && die "Failed to get card $SOFCARD short name"
+else
+    exit 1
```
marc-hb, line 40: The nested logic does not seem necessary here, can you flatten it? Something flat like this:
```shell
cardname=$(sof-dump-status.py -s "$SOFCARD") || die/exit/
[ -n "$cardname" ] || die/exit/
```

Thanks to `set -e` the first `die` may not even be necessary.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,55 @@
...
+    tmp=$(mktemp -d) || die "Failed to create tmp folder."
+    cd "$tmp"
+    dlogi "Checking the card $cardname UCM2 initial settings"
+    alsactl store -f tmp.conf || die "Failed to save the amixer setting"
+    dlogc "alsactl init $SOFCARD"
+    alsactl init "$SOFCARD" || die "Failed to run alsactl init"
+    alsactl restore -f tmp.conf || die "Failed to restore the amixer setting"
+}
+
+main "#@"
```
marc-hb, line 55: This won't work, sorry for the earlier typo. I meant: `main "$@"`

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,55 @@
...
+# BootSequence is introduced in alsa-utils 1.2.2-16
+ALSAUTILS_VER=1.2.2-16
+# BootSequence is introduced in alsa-lib 1.2.2-44
+ALSALIB_VER=1.2.2-44
+
+func_opt_parse_option "$@"
+
+func_check_version "libasound2" "$ALSALIB_VER" || exit 1
+func_check_version "alsa-utils" "$ALSAUTILS_VER" || exit 1
+
```
marc-hb, line 34: Now that there is a `main()` function, can you move all the code to it? Or to some other new function. The suggestion was to not have any code at the top-level.

test-case/verify-ucm-bootsequence.sh
```diff
@@ -0,0 +1,55 @@
...
+# BootSequence is introduced in alsa-utils 1.2.2-16
+ALSAUTILS_VER=1.2.2-16
+# BootSequence is introduced in alsa-lib 1.2.2-44
+ALSALIB_VER=1.2.2-44
+
+func_opt_parse_option "$@"
+
+func_check_version "libasound2" "$ALSALIB_VER" || exit 1
+func_check_version "alsa-utils" "$ALSAUTILS_VER" || exit 1
```
marc-hb, line 33: There's no need to `exit` here if `func_check_version` already dies, it's one or the other.

marc-hb (changes requested):
> @marc-hb UCM is installed automatically when installing the OS.

OK but I think a test with "UCM" in the name should verify that some UCM installation and other UCM things actually happened, otherwise it's confusing.

> I think we need also add the UCM version check and if the UCM version doesn't meet the requirement, it should SKIP this test case with dlogi info

That seems complicated. How about this instead: rename the test to `verify-bootsequence.sh`, I mean just drop "UCM" from the name. If UCM is there, then it tests UCM. If UCM is not there then it does not test UCM but runs exactly the same. Simple.

test-case/verify-ucm-bootsequence.sh
libinyang: > This won't work, sorry for the earlier typo. I meant: `main "$@"`

Right. I will update it.

test-case/verify-ucm-bootsequence.sh
libinyang: > Now that there is a `main()` function, can you move all the code to it? Or to some other new function. The suggestion was to not have any code at the top-level.

Make sense. I will update it.

test-case/verify-ucm-bootsequence.sh
libinyang: ACK

test-case/verify-ucm-bootsequence.sh
libinyang: ACK

case-lib/user.sh
libinyang: ACK

libinyang:
> OK but I think a test with "UCM" in the name should verify that some UCM installation and other UCM things actually happened, otherwise it's confusing.
> 
> That seems complicated. How about this instead: rename the test to `verify-bootsequence.sh`, I mean just drop "UCM" from the name. If UCM is there, then it tests UCM. If UCM is not there then it does not test UCM but runs exactly the same. Simple.

OK, I will update the script name.

> Never duplicate any code, that's why we have functions. The suggestion is: main() -> check_version() -> compare_version()
Oh, I misunderstand your mean before. I will update the code.

libinyang:
> OK but I think a test with "UCM" in the name should verify that some UCM installation and other UCM things actually happened, otherwise it's confusing.
> That seems complicated. How about this instead: rename the test to `verify-bootsequence.sh`, I mean just drop "UCM" from the name. If UCM is there, then it tests UCM. If UCM is not there then it does not test UCM but runs exactly the same. Simple.

 OK, I will update the script name.

> Never duplicate any code, that's why we have functions. The suggestion is: main() -> check_version() -> compare_version()
>

 Oh, I misunderstand your mean before. I will update the code.

xiulipan (approved):

case-lib/user.sh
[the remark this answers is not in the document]
marc-hb: The code in https://github.com/thesofproject/sof-test/pull/600/commits/c626edba79f1985c is basically the opposite of what I just recommended. Can you please not "resolve conversation"?

test-case/verify-bootsequence.sh
```diff
@@ -0,0 +1,56 @@
...
+    func_version_ge "alsa-ucm-conf" "$ALSAUCM_VER" || dlogi "alsa-ucm-conf version is too low"
+
+    # alsactl init only applies to SOF card
+    [[ -n "$SOFCARD" ]] || die "No SOF card"
+    cardname=$(sof-dump-status.py -s "$SOFCARD") || exit 1
+    [[ -n "$cardname" ]] || die "Failed to get card $SOFCARD short name"
+
+    tmp=$(mktemp -d) || die "Failed to create tmp folder."
+    cd "$tmp"
+    dlogi "Checking the card $cardname UCM2 initial settings"
```
marc-hb, line 49: This log message is misleading because it looks like the test does something UCM-specific but it does not. Change to something like:

> dlogi "Indirectly checking the card $cardname UCM2 initial settings (if any) with alsactl store"

marc-hb (changes requested):
Code looks generally good except for:
- misleading `dlogi UCM`
- two version check functions that does not achieve the goal of having one dpkg-independent functions

I think there are also some unnecessary `die` calls thanks to `set -e` but not really a problem

test-case/verify-bootsequence.sh
libinyang: > This log message is misleading because it looks like the test does something UCM-specific but it does not. Change to something like:
> 
> > dlogi "Indirectly checking the card $cardname UCM2 initial settings (if any) with alsactl store"

@marc-hb Yes. Your suggested dlogi is more reasonable. I will use it expect "with alsactl init" not "with alsactl store"

>> two version check functions that does not achieve the goal of having one dpkg-independent functions

Yes. Currently I don't have other better solutions to check the version. I tried commands such as `aplay --version`, but the version information is not enough for us. And what's more, I can't get the alsa-lib and alsa-ucm-conf version with such command. 
If we found a better solution, I will update it.

marc-hb: I mean this (untested)
```shell
dpkg_version_ge()
{
    ver=$(dpkg -l "$1" |grep "$1" | awk '{print $3}')
    [[ -z "$ver" ]] && die "failed to get $1 version"
    dlogi "find $1 version  $ver"
    version_ge "$1" "$2"
}
 
version_ge()
{   
    [ "$(printf '%s\n' "$2" "$ver" | sort -V | head -n1)" = "$2" ]
}
```

libinyang:
patches are updated

case-lib/user.sh
```diff
@@ -0,0 +1,22 @@
...
+    ver=$(dpkg -l "$1" |grep "$1" | awk '{print $3}')
+    [[ -z "$ver" ]] && die "failed to get $1 version"
+    dlogi "find $1 version  $ver"
+    version_ge "$ver" "$2"
+}
+
+# function version_ge compares the 2 versions specified in the parameters.
+version_ge()
+{
+    [ "$(printf '%s\n' "$2" "$ver" | sort -V | head -n1)" = "$2" ]
```
marc-hb, line 21: There is no `$ver` in this function. You did not test it. Testing this interactively is easy: just copy/paste it in your current shell.

There's no need for `[   ... | head -n1 =  ]` because sort has a `-V` option. See my answer to https://unix.stackexchange.com/questions/285924/how-to-compare-a-programs-version-in-a-shell-script

marc-hb (changes requested):

libinyang:
patches are updated based on @marc-hb 's comment

marc-hb (approved):
Thanks , I hope the functions were tested this time.

xiulipan (approved):

PR #602, opened 2021-02-02
userspace: update to latest opt fmt

libinyang:
The OPT_xxx in sof-test has been updated. Let's update to the latest fmt
for check-userspace-paplay.sh and check-userspace-parecord.sh.

Signed-off-by: Libin Yang <libin.yang@intel.com>

### DIFF
--- a/test-case/check-userspace-paplay.sh
+++ b/test-case/check-userspace-paplay.sh
@@ -23,32 +23,32 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 # check pulseaudio runs properly or not
 func_lib_check_pa || die "Please check whether pulseaudio runs correctly or not"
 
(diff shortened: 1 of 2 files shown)


marc-hb (approved):
I guess this happened because  d8665e0f61 and 8384a03095d39 were submitted concurrently + these are not run in automation yet?

xiulipan:
@marc-hb Looks like a good topic for some UNIT TEST (But that would not be easy as we do not have a reference ENV now) for out test case.

libinyang:
> I guess this happened because [d8665e0](https://github.com/thesofproject/sof-test/commit/d8665e0f6129e98d4577ac59ab34f649e827dc0f) and [8384a03](https://github.com/thesofproject/sof-test/commit/8384a03095d3928a5536550fa32af99f0a0281f7) were submitted concurrently + these are not run in automation yet?

These userspace test cases are not in the daily automation test. So they are missed.

PR #615, opened 2021-02-13
lib.sh: add 2nd .ldc fallback in /lib/firmware/intel/sof/sof-abc.ldc

marc-hb:
From a developer perspective this makes easier to deploy the .ldc
dictionary at the same time than the firmware.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -79,15 +79,22 @@ fake_kern_error()
 SOF_LOG_COLLECT=0
 func_lib_start_log_collect()
 {
-    local is_etrace=${1:-0} ldcFile="$SOFLDC"
+    local is_etrace=${1:-0} ldcFile
     # if user doesn't specify file path of sof-*.ldc, fall back to
     # /etc/sof/sof-PLATFORM.ldc, which is the default path used by CI.
-    [[ -n "$ldcFile" ]] || {
-        ldcFile=/etc/sof/sof-$(sof-dump-status.py -p).ldc || {
+    # and then on the standard location.
+    if [ -n "$SOFLDC" ]; then
+        ldcFile="$SOFLDC"
+    else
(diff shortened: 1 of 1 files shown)


marc-hb:
One ALSABAT test failed in https://sof-ci.01.org/softestpr/PR615/build588/devicetest/?model=CML_RVP_SDW&testcase=check-alsabat-headset-playback and BSW failed to boot - nothing related to this.

marc-hb:
This is required for sof-test to use `.ldc` files installed by https://github.com/thesofproject/sof/pull/3845

aiChaoSONG (approved):
LGTM.
FYI， once we transfer to zephyr SOF, we will use tools here: https://github.com/zephyrproject-rtos/zephyr/tree/master/boards/xtensa/intel_adsp_cavs15/tools, which don't require LDC file like sof-logger does.

marc-hb:
Thanks @aiChaoSONG . Does Zephyr support some other kind of dictionary yet to save memory? If yes where is the dictionary found?

kv2019i (approved):
Thanks @marc-hb , this is useful!

fredoh9 (approved):
LGTM, thanks!

aiChaoSONG:
> Thanks @aiChaoSONG . Does Zephyr support some other kind of dictionary yet to save memory? If yes where is the dictionary found?

The tools there can be used directly, don't need to specify some kind of LDC file. but I am not sure if there 'is' some kind of dict or not.

marc-hb:
> but I am not sure if there 'is' some kind of dict or not.

Searched some archives and found this: https://github.com/zephyrproject-rtos/zephyr/pull/30539

aiChaoSONG:
> For now the SOF logger should work with SOF+Zephyr?

@marc-hb sof-logger requires the `*.ldc` file to run, but SOF+Zephyr don't generate it.

PR #627, opened 2021-03-09
lib: pick up card ID only from the line that starts with a number

ranj063:
If DMI info is missing and the card long name is replaced with
the card name set in the machine driver, there will be 2 lines
like this:

 0 [sofhdadsp      ]: sof-hda-dsp - sof-hda-dsp
                      sof-hda-dsp

And the test would pick up the card ID from the second line and
result in a failure.

So, ignore lines that dont start with a number to avoid this.

Signed-off-by: Ranjani Sridharan <ranjani.sridharan@linux.intel.com>

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -35,7 +35,9 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
-    SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{print $1;}')
+	# $1=$1 strips whitespaces
+	SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards |
+		awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1; exit 0;}')
 fi
 
 func_lib_setup_kernel_checkpoint()


marc-hb (commented):
Can you explain what `awk '{$1=$1};1'` does?

awk has built-in `grep` features, can you try shortening the code a bit by embedding `| grep 'sof-[a-z]'` into awk, should be something like: `awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ {print $1;}')`

It should be possible to embed the `-v sof-probes` too but that on the other hand would probably be more confusing.

`info gawk`  https://www.gnu.org/software/gawk/manual/html_node/Expression-Patterns.html

ranj063:
> Can you explain what `awk '{$1=$1};1'` does?
> 
> awk has built-in `grep` features, can you try shortening the code a bit by embedding `| grep 'sof-[a-z]'` into awk, should be something like: `awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ {print $1;}')`
> 
> It should be possible to embed the `-v sof-probes` too but that on the other hand would probably be more confusing.
> 
> `info gawk` https://www.gnu.org/software/gawk/manual/html_node/Expression-Patterns.html

done!

case-lib/lib.sh
```diff
@@ -35,7 +35,8 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
-    SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{print $1;}')
+	# remove leading spaces and check if the line starts with a number
+	SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{$1=$1};1 && $1 ~ /^[0-9]+$/ {print $1;}')
```
marc-hb, line 39: OK, this is not what I expected. I understand `{$1=$1};1` now but it's very cryptic with 2 actions where the second one is implicit. I'm not even sure whether the current line has 2 or 3 actions in total, that's quite confusing. I think you can do the same with a single `{ }`action (untested)
```shell
# $1=$1 strips whitespace
SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards |
   awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1;} ' 
```
```

File: case-lib/lib.sh [hunk 2]
```diff
@@ -35,7 +35,8 @@ fi
 
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
-    SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | awk '{print $1;}')
+	# remove leading spaces and check if the line starts with a number
+	SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards | awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1;}')
```
marc-hb, line 39: Nit: you can insert a newline after the `|`, it should be ignored inside `$( )` no need to escape it. In general whenever inserting a newline somewhere in a script passes a simple test then it will pass all tests. There are many traps in shell scripts but not with _unescaped_ newlines: they either work or don't.

marc-hb (commented):
I'm not familiar enough with `/proc/asound/cards` to approve the logic but the codestyle LGTM. Note SOFCARD can have multiple lines but that's not new.

case-lib/lib.sh
ranj063: # $1=$1 strips whitespace
SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards |
   awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1;}) '  works!

marc-hb: > grep -v 'sof-probes' /proc/asound/cards | grep 'sof-[a-z]' | grep '^[0-9]' | awk '{print $1;}'

This does not make sure $1 is a number and does not strip whitespace (BTW `grep -e 'sof-[a-z]' -e '^[0-9]` would as clear and shorter and faster)

@fredoh9 `'/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1;}')` is not much more complex than the https://www.gnu.org/software/gawk/manual/html_node/Very-Simple.html example, unlike previous versions of this PR it has the same basic `expression { action }` structure documented in https://www.gnu.org/software/gawk/manual/html_node/Getting-Started.html: 

While it's often possible to read some Python or shell code without even knowing either, I agree that reading even basic `awk` syntax requires a little bit more upfront knowledge. If you think `expression { action }` is too complicated then you're basically saying not to use awk all - then what would be the alternative in this particular case?

marc-hb (approved):

fredoh9 (approved):
looks good, fancy matching

PR #630, opened 2021-03-13
[skip ci] .github: apt-get -y update

marc-hb:
Because Ubuntu VMs on github have a package index that is out of date by
design:

https://github.com/actions/virtual-environments/issues/ 1757

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -33,13 +33,16 @@ jobs:
       # :-( https://github.community/t/support-for-yaml-anchors/16128
       - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
       - uses: actions/setup-python@v2
-      - name: temporary python3-pil HACK because github is out of date
-        run: sudo apt-get -y install libimagequant0 libwebpdemux2 &&
-           wget http://security.ubuntu.com/ubuntu/ubuntu/pool/main/p/pillow/python3-pil_7.0.0-4ubuntu0.2_amd64.deb &&
-           sudo dpkg -i python3-pil_*.deb
+
+      # Package index is "out of date by design"
+      # https://github.com/actions/virtual-environments/issues/1757
+      - name: apt-get -y update
+        run: sudo apt-get -y update
+
       - name: get python libs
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -33,13 +33,16 @@ jobs:
...
-           wget http://security.ubuntu.com/ubuntu/ubuntu/pool/main/p/pillow/python3-pil_7.0.0-4ubuntu0.2_amd64.deb &&
-           sudo dpkg -i python3-pil_*.deb
+
+      # Package index is "out of date by design"
+      # https://github.com/actions/virtual-environments/issues/1757
+      - name: apt-get -y update
+        run: sudo apt-get -y update
+
       - name: get python libs
         # FIXME: apt-get succeeds but 'import numpy' still fails!?
```
fredoh9, line 43: We still have failure with "import numpy" after installing python3-numpy?
I installed with pip3 but it should be the same, right?

.github/workflows/pull_request.yml
```diff
@@ -46,3 +46,11 @@ jobs:
...
         run: ./tools/CI/check-gitrange.bash origin/${BASE_REF}...HEAD
                text/x-python pylint --disable=C
+
+
+  yamllint:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: yamllint ourselves
+        run: yamllint .github/workflows/pull_request.yml
```
fredoh9, line 59: Sorry for being silly. Is this calling itself?

.github/workflows/pull_request.yml
marc-hb: What happened here is `pylint` complaining that it cannot find `python3-numpy`. Maybe it was specific to pylint and maybe it was specific to github and maybe it's not happening anymore. Anyway it's off-topic in this PR :-)

.github/workflows/pull_request.yml
marc-hb: Yes this file checks itself. This will obviously not work if it's totally broken but even in that most extreme case it's still a good hint to run this locally.

fredoh9 (approved):
looks good, thank you for removing HACK

Issue #617, opened 2021-02-16
[BUG] sof-kernel-log-check: filter out IGB issues


plbossart:
One more issue on UpExtreme, solved with following string:

````
+ignore_str="$ignore_str"'|igb 0000:0.:00\.0: exceed max 2 second'
````

Maybe we need to ignore all IGB problems really.

marc-hb:
`igb` = Intel Gigabit Ethernet.

I searched this message in the source code and on the Internet and it could be a cable issue, can you try a different cable?

Ignoring bad cables and similar network issues sounds like a bad idea.

plbossart:
It's not a cable issue, I use the same with other interfaces (RTL or USB dongles) and they work fine.

marc-hb:
Gigabit Ethernet uses all 4 pairs both ways. 100Mb/s uses only 2 pairs one-way each.

plbossart:
I have only Gigabit stuff and 935Mbs/s all the way to my devices over Google Fiber. Cables are not the problem

plbossart:
PING:
6.0 ms
DOWNLOAD
936 Mbps
UPLOAD
816 Mbps

marc-hb:
Looking at `drivers/net/ethernet/intel/igb/igb_main.c` this error message is printed when and only when a watchdog found `SR_1000T_REMOTE_RX_STATUS` a.k.a `Remote receiver OK` to be repeatedly wrong 20 times over 2 seconds. Really not good.

How often do you get this message and what other `igb` messages are around it? It could have been just a temporary glitch because of some bad contact or bent cable or temporary interference.

I think we really need this custom sof-test config file where anyone can add any particular error they have. Of course CI would never use such custom additions but only the default and common set of ignored errors. Using such a file would print a warning at the start of the test. It's one thing to ignore network or other glitches at home when you can just stand up, give the system a kick, re-run the tests without even looking at what went wrong and forget about it. It's a totally different story to ignore network glitches in a distant lab, especially during COVID. @fredoh9 and others have already spent _many days_ in the lab troubleshooting various storage and network issues and hiding warning signs is really not a good idea. Ignoring storage and network errors is not like ignoring wifi or keyboard errors. 

(even funnier, ignoring audio errors: https://github.com/thesofproject/sof/pull/2686)

plbossart:
I get this with a different board and different cables

````
===========================>>
[  936.832321] kernel: igb 0000:01:00.0 enp1s0: Reset adapter
[...]

PR #631, opened 2021-03-15
sof-kernel-log-check: filter out IGB messages

plbossart:
sof-test is unusable on Up Extreme due to various IGB messages.
Filter out all messages since they seem to have no functional impact.

BugLink: https://github.com/thesofproject/sof-test/issues/617


marc-hb (approved):
This is sad.

We should have some device-specific filtering ability, maybe something in /etc. Should be easy but I don't have time right now

plbossart:
> This is sad.
> 
> We should have some device-specific filtering ability, maybe something in /etc. Should be easy but I don't have time right now

IGB can only be found on UpExtreme, so there's no point in spending too much time on this.

aiChaoSONG (approved):

fredoh9 (approved):

PR #634, opened 2021-03-23
kmod: sof_remove: add github URL to date snd_sof_nocodec re-ordering

marc-hb:
This will be useful in a future "spring cleaning" in a year or two.


fredoh9 (changes requested):
Eventually we will need to remove "|| true" and redundant one also. So add FIXME.

marc-hb:
FIXME sounded a bit extreme because we want backward compatibility. Added TODO.

fredoh9 (approved):
Looks good.

PR #635, opened 2021-03-23
kmod/tools: tabulate log output, stricter lsmod | grep

marc-hb:
3 commits, most visible one: kmod/tools: tabulate log output

Sample new output:
```
SKIP	snd_hda_codec_realtek  	not loaded
SKIP	snd_hda_codec_generic  	not loaded
RMMOD	snd_soc_acpi
SKIP	snd_hda_ext_core  	not loaded
RMMOD	snd_intel_dspcfg
SKIP	soundwire_intel_init  	not loaded
SKIP	soundwire_intel  	not loaded
SKIP	soundwire_cadence  	not loaded
SKIP	soundwire_generic_allocation  	not loaded
SKIP	regmap_sdw  	not loaded
SKIP	regmap_sdw_mbq  	not loaded
SKIP	soundwire_bus  	not loaded
RMMOD	snd_soc_core
SKIP	snd_hda_codec  	not loaded
SKIP	snd_hda_core  	not loaded
RMMOD	snd_hwdep
SKIP	snd_compress  	not loaded
RMMOD	snd_pcm

SKIP    	snd_sof_pci_intel_tng 	not in tree
SKIP    	snd_sof_pci_intel_apl 	not in tree
SKIP    	snd_sof_pci_intel_cnl 	not in tree
SKIP    	snd_sof_pci_intel_icl 	not in tree
SKIP    	snd_sof_pci_intel_tgl 	not in tree
MODPROBE	snd_sof_acpi
MODPROBE	snd_sof_pci
MODPROBE	snd_usb_audio
```

PR #636, opened 2021-03-23
tools/kmod/remove: more specific regex, remove &> /dev/null

marc-hb:
Remove unnecessary &>/dev/null redirection, replace with grep -q.

grep lsmod output with a more specific regex

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -6,7 +6,7 @@ remove_module() {
 
     local MODULE="$1"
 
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR636/build618/devicetest/?model=CML_HEL_RT5682&testcase=check-suspend-resume-5 is the usual suspend/resume TIMEOUT. No other failure.

marc-hb:
Folded this into #635, sorry for the noise

PR #639, opened 2021-03-29
Use journalctl for log collection and remove dependency on /var/log/kern.log

aiChaoSONG:


case-lib/lib.sh
```diff
@@ -40,10 +40,26 @@ if [ ! "$SOFCARD" ]; then
 		awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1; exit 0;}')
 fi
 
-func_lib_setup_kernel_checkpoint()
+setup_kernel_check_point()
 {
-    # shellcheck disable=SC2034 # external script will use it
-    KERNEL_CHECKPOINT=$(date +%s)
+    local check_point
+    check_point="$1"
```
marc-hb, line 46: This can be a single line because `"$1"` is not a command and has no exit status so SC2155 does NOT apply

case-lib/lib.sh
```diff
@@ -40,10 +40,26 @@ if [ ! "$SOFCARD" ]; then
...
+    case "$check_point" in
+        "boot")
+            # shellcheck disable=SC2034 # external script will use it
+            KERNEL_CHECKPOINT="boot"
+            ;;
+        "now")
+            # There is a 5s delay between two cases by sof-framework,
+            # make the check point 6s earlier to avoid log loss.
```
marc-hb, line 54: This is too brittle. The simpler and safer way to solve this problem is to set checkpoints only at (almost) the same time as when collecting logs and never at any other time.

More importantly: I think this is an important but different problem, not directly related to this Pull Request, so best left aside for now.

marc-hb (changes requested):
Thanks @chao for looking at this, really appreciated.

Please don't hijack the checkpoint function(s) and KERNEL_CHECKPOINT variable for this. The entire purpose of the checkpoints is to NOT scan the logs from boot, so having a special checkpoint case that is not actually a checkpoint is quite confusing.

The few special tests that want to scan all the logs from boot should ideally invoke `journalctl_cmd --boot` directly and specifically without a date, then it will be more obvious what they do. If that's too complicated for some reason then as an alternative they could define some new, `IGNORE_CHECKPOINTS=true` or `SCAN_LOGS_FROM_BOOT=true` global variable. 


Not changing the nature of checkpoints should also reduce the number of lines changed considerably.

case-lib/lib.sh
plbossart: @marc-hb I was not able to understand your feedback. 

a) please describe when checkpoints should be set.

b) why can it be done in a separate PR, since it's a key functionality to enable journalctl.

aiChaoSONG: @marc-hb Before implementing this, I was thinking to use a new Global Variable to collect boot logs, too. But then I found that this can be integrated in the same function, `setup_kernel_check_point`,  to provide a unified API for test case. 

A test case designer can decide whether they want to collect logs from boot, or from now. When you say another check point, I think you mean `setup_kernel_check_point "boot"`, Actually this is only used in the first test case currently.  `Moving forward` principle is not influenced, because we give test case the flexibility setup check point. For this reason, we take "boot" as a virtual check point.

With current code, the command for collecting log from boot is: `journalctl -k -q --no-pager --utc --output=short-monotonic 
--no-hostname --boot `,  the command for collecting log from check point is: `journalctl -k -q --no-pager --utc --output=short-monotonic --no-hostname --boot  --since=@CHECK_POINT`, already good enough to handle both conditions.

By the way, the `--boot` option means collect log of the latest boot, it doesn't mean collect boot logs.

@plbossart @marc-hb **On the potential log loss.**

sof-framework will add a 5s delay between two test cases, we are not sure there will be new logs in this 5s period, so I make the checkpoint 6s earlier with KERNEL_CHECKPOINT=$(($(date +%s) - 6)).  Actually, no case is running on DUT in this 5s period, very low possibility to have kernel logs, I think. The point here becomes, should we care and collect very likely unrelated and sometime may be empty logs between test cases? With 6 here, we may have 1s log duplicated. We can choose 0, 1, 2, 3, 4, 5 or 6  to decrease log loss. In my opinion now, 5 may be the best with very low possibility of log loss and no log duplication.

marc-hb: >  in the same function, setup_kernel_check_point, to provide a unified API for test case.

I don't think it's good to have a unified API for two different things, especially not when the API is named after only one of the two things. 

> Actually this is only used in the first test case currently.

So it's not great either to change and require almost every invocation to over-specify "now" and introduce the abstract concept of a "virtual" checkpoints that are not checkpoints, all this for just one special case. BTW it should be possible to run tests in any order.

In fact the only test that uses `setup_kernel_check_point "boot"` in this PR is apparently `verify-firmware-presence.sh` That test is super short and simple and merely making sure that `/lib/firmware/intel/sof/platform.ri` exists, so it should not even check kernel logs at all, kernel logs are not relevant for such a test. Is it really the only test that needed `kern.log`?

In the future we probably want a test that does nothing and checks boot logs but that would be a different test, it shouldn't be confused with checking `/lib/firmware`.

> By the way, the --boot option means collect log of the latest boot, it doesn't mean collect boot logs.

Sorry I used both somewhat interchangeably.

marc-hb: > The point here becomes, should we care and collect very likely unrelated and sometime may be empty logs between test cases?

Good question but IMHO not new and independent from this PR. One problem at a time.

case-lib/lib.sh
```diff
@@ -327,13 +327,6 @@ die()
...
 }
 
-# force ask buffer data write into file system
-sudo sync -f || true
-# catch kern.log last line as current case start line
-if [ ! "$DMESG_LOG_START_LINE" ]; then
-    DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
-fi
-
 is_sof_used()
```
marc-hb, line 330: In an ideal world every commit should be usable for bisecatability, cherry-picking, reverting, etc. reasons. I guess this implies the entire `DMESG_LOG_START_LINE` removal should be a single commit?

case-lib/hijack.sh
```diff
@@ -50,17 +50,14 @@ function func_exit_handler()
...
-    # May 15 21:28:38 MachineName kernel: [    6.469255] sof-audio-pci 0000:00:0e.0: ipc rx: 0x90020000: GLB_TRACE_MSG
-    # May 15 21:28:38 MachineName kernel: [    6.469268] sof-audio-pci 0000:00:0e.0: ipc rx done: 0x90020000: GLB_TRACE_MSG
-    if [[ -n "$DMESG_LOG_START_LINE" && "$DMESG_LOG_START_LINE" -ne 0 ]]; then
-        tail -n +"$DMESG_LOG_START_LINE" /var/log/kern.log |cut -f5- -d ' ' > "$LOG_ROOT/dmesg.txt"
+
+    if [[ "$KERNEL_CHECKPOINT" =~ ^[0-9]{10} ]]; then
+        journalctl_cmd --since=@"$KERNEL_CHECKPOINT" > "$LOG_ROOT/dmesg.txt"
     else
-        cut -f5- -d ' ' /var/log/kern.log > "$LOG_ROOT/dmesg.txt"
+        die "Kernel check point is not properly setup"
```
marc-hb, line 57:
```
die "KERNEL_CHECKPOINT is not properly setup"
```
Much faster `git grep` when this happens.

test-case/check-reboot.sh
```diff
@@ -150,7 +150,8 @@ $old_content
 
 $boot_once_flag
 export LOG_ROOT='$(realpath $LOG_ROOT)'
-export DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
+setup_kernel_check_point
+export KERNEL_CHECKPOINT
```
marc-hb, line 154: Can you add a comment explaining why this one needs the export when the others don't?

Wouldn't it be simpler to export KERNEL_CHECKPOINT always? If not why not?

marc-hb (changes requested):
I have two "big" questions, maybe related to each other:

- If one test wants to scan the entire boot log with this new design then how it is supposed to do it? Do we even have any such test right now?
- I see many more checkpoint additions than  `DMESG_LOG_START_LINE` removals. In other words some tests are now setting a checkpoint and they were doing nothing before. What part of the kernel log were they scanning before?

test-case/check-reboot.sh
aiChaoSONG: Sorry, this is a mistake from previously version, the `export KERNEL_CHECKPOINT` should be removed.

case-lib/hijack.sh
aiChaoSONG: normally, this should not happen, I just want to use this to tell the case owner that he/she should setup up check point.

aiChaoSONG:
> If one test wants to scan the entire boot log with this new design then how it is supposed to do it? Do we even have any such test right now?

This is in my next PR which add a new function `disable_kernel_check_point`, which set `KERNEL_CHECKPOINT` to "disabled",  and  add one more check, once it is disabled, collect log from boot, and also a new test case to collect boot log.

> I see many more checkpoint additions than DMESG_LOG_START_LINE removals. In other words some tests are now setting a checkpoint and they were doing nothing before. What part of the kernel log were they scanning before?

Some cases don't have DMESG_LOG_START_LINE because it is set implicitly in lib.sh, and now I want test case writer to explicitly set a check point. Why are some other cases have DMESG_LOG_START_LINE? actually I don't know, ask the owner!
```
DMESG_LOG_START_LINE=$(wc -l /var/log/kern.log|awk '{print $1;}')
```

case-lib/hijack.sh
marc-hb: When someone sees the message with `KERNEL_CHECKPOINT` then they can immediately `git grep` that and find all relevant locations in sof-test.

marc-hb:
> Why are some other cases have DMESG_LOG_START_LINE? actually I don't know, ask the owner!

We must make 100% sure every single test catch errors when they happen and especially when refactoring this most critical part of sof-test. Otherwise when end up with "green failures" like this most recent one  and many others before it:
https://github.com/thesofproject/linux/pull/2810#issuecomment-818172849

That requires significant time and effort but taking shortcuts and running  and maintaining tests that catch errors "maybe" is pointless, better not spend time at all if you can't afford to fully understand test this particular refactoring thoroughly. It affects our entire test suite and product quality.

aiChaoSONG:
this PR is separated to two other PRs #643, #649, and they are merged, close this.

PR #642, opened 2021-04-07
sof-dump-status : Add ADL platform info  ;  verify-pcm-list : fix 'set -e' incompatibility   

XiaoyunWu6666:
Add support for ADL-P in tools/sof-dump-status.py since it is a new platform.

Also fix some 'set -e incompatibilities' in test-cases/verify-pcm-list.sh to avoid silent exit.

### DIFF
--- a/test-case/verify-pcm-list.sh
+++ b/test-case/verify-pcm-list.sh
@@ -26,8 +26,8 @@ OPT_HAS_ARG['t']=1         OPT_VAL['t']="${TPLG:-}"
 func_opt_parse_option "$@"
 tplg=${OPT_VAL['t']}
 
-tplg_path=`func_lib_get_tplg_path "$tplg"`
-[[ "$?" != "0" ]] && die "No available topology for this test case"
+tplg_path=$(func_lib_get_tplg_path "$tplg") ||
+       	die "No available topology for this test case"
 
 # hijack DMESG_LOG_START_LINE which refer dump kernel log in exit function
 DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
@@ -35,7 +35,7 @@ DMESG_LOG_START_LINE=$(sof-get-kernel-line.sh|tail -n 1 |awk '{print $1;}')
 tplg_str="$(sof-tplgreader.py $tplg_path -d id pcm type -o)"
 pcm_str="$(sof-dump-status.py -i ${SOFCARD:-0})"
 
-dlogc "sof-tplgreader.py $tplg_file -d id pcm type -o"
+dlogc "sof-tplgreader.py $tplg_path -d id pcm type -o"
 dlogi "Pipeline(s) from topology file:"
 echo "$tplg_str"
 dlogc "sof-dump-status.py -i ${SOFCARD:-0}"
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -10,14 +10,22 @@ class clsSYSCardInfo():
         self.pci_lst=[]
         self.acpi_lst=[]
         self.proc_card={}
-        # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c
-        self._pci_ids={"0x119a":"tng","0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
-            "0x9dc8":"cnl", "0xa348":"cfl", "0x9d71":"kbl", "0x9d70":"skl", "0x34c8":"icl", "0x38c8":"jsl", "0x4dc8":"jsl",
-            "0x02c8":"cml", "0x06c8":"cml", "0xa0c8":"tgl", "0x43c8":"tgl-h", "0x4b55":"ehl", "0x4b58":"ehl", "0x7ad0": "adl-s",
-            # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-acpi-dev.c
-            "0x3438":"bdw", "0x33c8":"hsw", "0x0f04":"byt", "0x2284":"cht",
-            # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/pci/hda/hda_intel.c
-            "0x160c":"bdw" }
+        # https://github.com/thesofproject/linux/tree/topic/sof-dev/sound/soc/sof/intel/
+        self._pci_ids = {
+            # pci-tng.c
+            "0x119a":"tng",
+            # pci-apl.c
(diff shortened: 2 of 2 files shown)


test-case/verify-pcm-list.sh
```diff
@@ -26,16 +26,15 @@ OPT_HAS_ARG['t']=1         OPT_VAL['t']="${TPLG:-}"
 func_opt_parse_option "$@"
 tplg=${OPT_VAL['t']}
 
-tplg_path=`func_lib_get_tplg_path "$tplg"`
-[[ "$?" != "0" ]] && die "No available topology for this test case"
+tplg_path=`func_lib_get_tplg_path "$tplg"`|| die "No available topology for this test case"
```
marc-hb, line 29: You can have a new line after `||` if you prefer shorter lines. No need to escape it.

marc-hb (changes requested):
Please fix the shellcheck and pylint warnings on the lines you're changing anyway. Not everywhere but at least on lines modified by this PR so we get the numbers down, not up. Otherwise the change looks good.

aiChaoSONG:
SOFCI TEST

tools/sof-dump-status.py
```diff
@@ -11,7 +11,7 @@ def __init__(self):
         self.acpi_lst=[]
         self.proc_card={}
         # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c
-        self._pci_ids={"0x119a":"tng","0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
+        self._pci_ids={"0x51c8":"adl-p","0x119a":"tng","0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
```
fredoh9, line 14: Not a big deal but we already have adl-s, Better to move after adl-s. If that line is too long, then may be new line after that?

plbossart: agree, this should be broken in different lines, one per generation, as we did in the kernel. Good suggestion.

test-case/verify-pcm-list.sh
XiaoyunWu6666: OK , I will try shellcheck for better code style :)

tools/sof-dump-status.py
XiaoyunWu6666: gotcha

test-case/verify-pcm-list.sh
```diff
@@ -26,16 +26,16 @@ OPT_HAS_ARG['t']=1         OPT_VAL['t']="${TPLG:-}"
 func_opt_parse_option "$@"
 tplg=${OPT_VAL['t']}
 
-tplg_path=`func_lib_get_tplg_path "$tplg"`
-[[ "$?" != "0" ]] && die "No available topology for this test case"
+tplg_path=$(func_lib_get_tplg_path "$tplg") \
+	|| die "No available topology for this test case"
```
marc-hb, line 30: You don't need to escape the new line when you do this:
```shell
tplg_path=$(func_lib_get_tplg_path "$tplg") ||
	die "No available topology for this test case"
```
In many cases when you leave something incomplete the shell knows it should keep reading and then you don't need to escape the new line.

marc-hb (approved):
> OK , I will try shellcheck for better code style :)

It's not just style, shellcheck can also catch some bugs.

tools/sof-dump-status.py
```diff
@@ -11,13 +11,23 @@ def __init__(self):
...
+                #/linux/sound/soc/sof/intel/pci-tng.c
+                "0x119a":"tng",
+                #/linux/sound/soc/sof/intel/pci-apl.c
+                "0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
+                #/linux/sound/soc/sof/intel/pci-cnl.c
+                "0x9dc8":"cnl", "0xa348":"cfl", "0x02c8":"cml", "0x06c8":"cml",
+                #/linux/sound/soc/sof/intel/pci-icl.c
+                "0x34c8":"icl", "0x38c8":"jsl","0x4dc8":"jsl",
+                #/linux/sound/soc/sof/intel/pci-tgl.c
+                "0xa0c8":"tgl", "0x43c8":"tgl-h", "0x4b55":"ehl", "0x4b58":"ehl","0x7ad0": "adl-s","0x51c8":"adl-p",
```
aiChaoSONG, line 24: The ADL-P platform is using `tgl_desc`, which define the FW name as `sof-tgl.ri`, so you should add `0x51c8` as tgl platform, rather than adl-p. @plbossart Do we have plan to add a new adlp_desc for adl-p platform?
@XiaoyunWu6666 Before you make any code change, let's wait  Pierre's feedback.
```
	{ PCI_DEVICE(0x8086, 0x51c8), /* ADL-P */
		.driver_data = (unsigned long)&tgl_desc},
```

aiChaoSONG: Please revise the comment,
`#/linux/sound/soc/sof/intel/pci-tgl.c` -> `https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/pci-tgl.c`

Then we can view the linux file by a mouse click.

marc-hb: I find this very verbose and repetitive. I think having a full URL a single time at the top is enough,  two clicks does not take much time and code is read much more often than clicked on.

plbossart: new descriptors were added today for ADL in the topic/sof-dev branch

tools/sof-dump-status.py
```diff
@@ -10,14 +10,22 @@ def __init__(self):
...
+                # pci-tng.c
+                "0x119a":"tng",
+                # pci-apl.c
+                "0x5a98":"apl", "0x1a98":"apl", "0x3198":"glk",
+                # pci-cnl.c
+                "0x9dc8":"cnl", "0xa348":"cfl", "0x02c8":"cml", "0x06c8":"cml",
+                # pci-icl.c
+                "0x34c8":"icl", "0x38c8":"jsl","0x4dc8":"jsl",
+                # pci-tgl.c
+                "0xa0c8":"tgl", "0x43c8":"tgl-h", "0x4b55":"ehl", "0x4b58":"ehl","0x7ad0": "adl-s","0x51c8":"adl-p",
```
libinyang, line 24: @XiaoyunWu6666 adl-p is using sof-adl.ri. Please check the commit https://github.com/thesofproject/linux/pull/2786/commits/d09bc51b68678d5108013b0c33304f644cd0cce9

So it should be "0x51c8":"adl"

XiaoyunWu6666: fixed already . Thanks for reminding me  . Also sorry for too many 'push'es this afternoon

aiChaoSONG (approved):

PR #643, opened 2021-04-08
sof-test: include test interval in kernel check point

aiChaoSONG:
The sof-framework will add a 5-second test interval
between two test cases. This patch makes the kernel
check point include this test interval to avoid log loss.

And also rename func_lib_setup_kernel_check_point
to setup_kernel_check_point, because the prefix
doesn't provide much information.

As PR #639 makes very big change, I am going to split it to several relatively small PRs, this is the first one.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -56,3 +56,8 @@ SUDO_LEVEL=${SUDO_LEVEL:-}
 # pipelines will be filtered out. Example:
 #
 # NO_HDMI_MODE=true
+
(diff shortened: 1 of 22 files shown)


aiChaoSONG:
@marc-hb This is the seperated PR as you suggested here: https://github.com/thesofproject/sof-test/pull/639#discussion_r609121518, please add you comment.

marc-hb (approved):
Sorry for the delay. I wished github advertised the _size_ of PRs!

XiaoyunWu6666 (approved):

PR #644, opened 2021-04-08
case-lib: add support for no HDMI mode

aiChaoSONG:
The i915 driver sometimes is not ready on a new platform,
sof-test will fail because broken HDMI pipelines are
present.

This patch adds support for no HDMI mode, once we are
under no HDMI mode, HDMI pipelines will be filtered
out, thus sof-test can run normally.

This patch has to be used with the change here: https://github.com/thesofproject/sof-test/pull/642

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -49,3 +49,10 @@ SOFCARD=${SOFCARD:-}
 # 2: run cmd with sudo, but needs sudo password
 # example: echo $SUDO_PASSWD | sudo -S ls /sys/kernel/debug/
 SUDO_LEVEL=${SUDO_LEVEL:-}
+
+# The i915 driver sometimes is not ready on a new platform, sof-test
+# will fail because broken HDMI pipelines are present.
+# The NO_HDMI_MODE option controls the pipeline filter, if true, HDMI
+# pipelines will be filtered out. Example:
+#
+# NO_HDMI_MODE=true
--- a/case-lib/pipeline.sh
+++ b/case-lib/pipeline.sh
@@ -45,9 +45,11 @@ func_pipeline_export()
         done
     fi
 
-    local opt=""
-    # acquire filter option
-    [[ "$2" ]] && opt="-f '$2'"
+    local opt="$2"
+    # In no HDMI mode, exclude HDMI pipelines
+    [ -z "$NO_HDMI_MODE" ] || opt="$opt & ~pcm:HDMI"
+    opt="-f '${opt}'"
+
     [[ "$ignore" ]] && opt="$opt -b '$ignore'"
     [[ "$SOFCARD" ]] && opt="$opt -s $SOFCARD"
 


case-lib/pipeline.sh
```diff
@@ -47,7 +47,13 @@ func_pipeline_export()
 
     local opt=""
     # acquire filter option
-    [[ "$2" ]] && opt="-f '$2'"
+    [[ -z "$2" ]] || {
+        if [ "$NO_HDMI_MODE" ]; then
+            opt="-f '~pcm:HDMI & $2'"
```
plbossart, line 52: what does this do? I

marc-hb (changes requested):

case-lib/config.sh
```diff
@@ -49,3 +49,9 @@ SOFCARD=${SOFCARD:-}
 # 2: run cmd with sudo, but needs sudo password
 # example: echo $SUDO_PASSWD | sudo -S ls /sys/kernel/debug/
 SUDO_LEVEL=${SUDO_LEVEL:-}
+
+# I915 sometimes is not ready on a new platform, sof-test will fail because
+# broken HDMI pipelins are present.
+# This option controls the pipeline filter, if true, HDMI pipelines will be
+# filtered out.
+NO_HDMI_MODE=${NO_HDMI_MODE:-}
```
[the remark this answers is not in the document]
aiChaoSONG: Yes, but there is no bad to keep this config.sh, this file can help to tell what option(s) can be override by command line.
and for some of the variable, this syntax gives default value if it is not overriden.

case-lib/pipeline.sh
```diff
@@ -46,8 +46,10 @@ func_pipeline_export()
     fi
 
     local opt=""
-    # acquire filter option
-    [[ "$2" ]] && opt="-f '$2'"
+    [ -z "$2" ] || opt="$2"
+    # In no HDMI mode, exclude HDMI pipelines
+    [ -z "$NO_HDMI_MODE" ] || opt="$opt & ~pcm:HDMI"
+    opt="-f '${opt}'"
```
marc-hb, line 52: Did you test `-f & ~pcm:HDMI` and does it work?

case-lib/config.sh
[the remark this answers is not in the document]
marc-hb: You're right, default values like `TPLG_ROOT=${TPLG_ROOT:-/lib/firmware/intel/sof-tplg}` are good.

Comments are good.

When there is no default value, an example is good:
```
# broken HDMI pipelins are present.
# This option controls the pipeline filter, if true, HDMI pipelines will be
# filtered out. Example:
# NO_HDMI_MODE=yes
```

`LOG_ROOT=${LOG_ROOT:-}` is misleading and really bad. Everyone who is not familiar with shell will bang their head wondering what it does (nothing).

There are other unrelated design issues with this config.sh file.

case-lib/pipeline.sh
aiChaoSONG: `-f & ~pcm:HDMI` doens't work, but it is not possible to happen, because https://github.com/thesofproject/sof-test/blob/5af711fcab5f94ff430a5a11f9f742f071129833/case-lib/pipeline.sh#L14

$2 always gets a value.

marc-hb: Great, then remove the misleading `[ -z "$2" ]` test.

aiChaoSONG: All review addressed, thanks @marc-hb. I will submit a cleanup PR for config.sh next.

aiChaoSONG:
The failure on BYT is not related

keqiaozhang (approved):

marc-hb (approved):

kv2019i:
@aiChaoSONG Thanks, this is pretty nice! :)

PR #649, opened 2021-04-14
sof-test: use journalctl for log collection

aiChaoSONG:
Previously, we will use /var/log/kern.log for log collcetion.
At the beginning of a test, we set up DMESG_LOG_START_LINE,
and at the end of the test, we collect kernel log from the
exact point we set up.

However, /var/log/kern.log is not exist on other distro. This
patch discards the existing log collection mechanism, and use
journalctl for log collection.

The changes include:
  - lib.sh: remove implicit setup of DMESG_LOG_START_LINE
  - hijack.sh: use journalctl to replace /var/log/kern.log
    for log collection.
  - test-case: test case fix as we need explicitly set up
    kernel check point for journalctl.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -50,17 +50,14 @@ function func_exit_handler()
         wcLog=$(wc -l "$logfile")
         dlogi "nlines=$wcLog"
     fi
-    # when case ends, store kernel log
-    # /var/log/kern.log format:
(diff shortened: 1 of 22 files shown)


aiChaoSONG:
SOFCI TEST

aiChaoSONG:
SOFCI TEST

keqiaozhang (approved):

aiChaoSONG:
With all marc's review addressed, and failure not related.

PR #652, opened 2021-04-20
check-kwd: set up kernel check point earlier

aiChaoSONG:
Previously, the call of func_pipeline_export is
earlier than setup_kernel_check_point, if we have
failure in export pipeline, the kernel check point
will not be properly set up. So bring kernel check
point set up earlier.


keqiaozhang (approved):

aiChaoSONG:
not big change, and the change is required in daily test, merge

PR #653, opened 2021-04-20
Add a new test case to check kernel boot log

aiChaoSONG:
tested on several DUTs, works fine.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -53,6 +53,8 @@ function func_exit_handler()
 
     if [[ "$KERNEL_CHECKPOINT" =~ ^[0-9]{10} ]]; then
         journalctl_cmd --since=@"$KERNEL_CHECKPOINT" > "$LOG_ROOT/dmesg.txt"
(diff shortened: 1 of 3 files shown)


marc-hb:
Why not just check the logs from boot when no checkpoint is set?

The worse that can happen is: some test forgets to set the checkpoint, then it might fail because of earlier tests. That's a very easy lesson to learn and fix.

aiChaoSONG:
> Why not just check the logs from boot when no checkpoint is set?

Previously, we have experience an issue which the test log is too big in stress test, that's why we only collect log output within the period of running test case, or under some condition, only collect logs from the latest iteration.
Now, I force every test case to explicit setup_kernel_check_point or disable_kernel_check_point in my journalctl PRs. `die` is not good, I agree with you, so https://github.com/thesofproject/sof-test/pull/661 is a good fix.

marc-hb (approved):
> we have experience an issue which the test log is too big in stress test,

I didn't think about this, thanks for the info.

fredoh9 (approved):
looks good. Next step, we need to add this new test  to first test case of all(or one of first)

PR #654, opened 2021-04-22
check-pm: set the kernel check point before runtime PM support check

keqiaozhang:
runtime PM is not supported on BDW and BYT platforms, so we need
to set the kernel check point before the runtime PM support check
or the kernel check point will not be properly set up and cause
runtime PM check TIMEOUT on unsupported platforms.

Signed-off-by: Zhang Keqiao <keqiao.zhang@intel.com>

### DIFF
--- a/test-case/check-runtime-pm-status.sh
+++ b/test-case/check-runtime-pm-status.sh
@@ -50,6 +50,8 @@ func_check_dsp_status()
 }
 
 func_opt_parse_option "$@"
+setup_kernel_check_point
+
 tplg=${OPT_VAL['t']}
 loop_count=${OPT_VAL['l']}
 [[ -z $tplg ]] && die "Miss tplg file to run"


test-case/check-runtime-pm-status.sh
```diff
@@ -82,8 +84,6 @@ do
 
     for i in $(seq 1 $loop_count)
     do
-        # set up checkpoint for each iteration
-        setup_kernel_check_point
```
aiChaoSONG, line 86: please don't remove the setup_kernel_check_point in the loop, because we only want to collect logs from the latest (success or failure) iteration, to reduce log size.

We are OK with only the changes in line 53

aiChaoSONG (commented):
minor change required

aiChaoSONG (approved):
Thanks for the fix

aiChaoSONG:
Merge to unlock daily test

marc-hb:
Can you explain how this was causing a timeout?

marc-hb:
See also related #653

aiChaoSONG:
> Can you explain how this was causing a timeout?

sof-framework will grep keywords `Pass/Fail/NA`, these keywords are logged out in the hijack.sh, if we die earlier, these keywords never appear, sof-framework cannot find such keyword within timeout, so it thinks the test case timeout.

marc-hb:
Ha yes, just like #661

PR #655, opened 2021-04-22
check-kmod: set the kernel check point before starting a test

fredoh9:
In error case, no kernel check point causes TIMEOUT in SOF CI. The below
line will be last log without verdict, then ends with TIMEOUT.

[ERROR] Kernel check point "KERNEL_CHECKPOINT" is not properly set

### DIFF
--- a/test-case/check-kmod-load-unload-after-playback.sh
+++ b/test-case/check-kmod-load-unload-after-playback.sh
@@ -48,7 +48,9 @@ OPT_HAS_ARG['d']=1         OPT_VAL['d']=3
 OPT_NAME['p']='pulseaudio'   OPT_DESC['p']='disable pulseaudio on the test process'
 OPT_HAS_ARG['p']=0             OPT_VAL['p']=1
 
-func_opt_parse_option "$@"
+fxunc_opt_parse_option "$@"
+setup_kernel_check_point
+
 tplg=${OPT_VAL['t']}
 loop_cnt=${OPT_VAL['l']}
 pb_duration=${OPT_VAL['d']}


fredoh9:
I was able to test with TGLU_RVP_NOCODEC. With this PR, there is verdict at the end. Without this PR, last error was `[ERROR] Kernel check point "KERNEL_CHECKPOINT" is not properly set`

```
2021-04-22 20:18:47 UTC Sub-Test: [INFO] ===== Testing: (Round: 1/1) (PCM: smart-nocodec [hw:0,2]<playback>) (Loop: 1/1) =====
2021-04-22 20:18:47 UTC Sub-Test: [COMMAND] aplay   -Dhw:0,2 -r 48000 -c 2 -f S16_LE -d 3 /dev/zero -v -q
aplay: main:830: audio open error: Invalid argument
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/120/gvfs
      Output information may be incomplete.
2021-04-22 20:18:47 UTC Sub-Test: [ERROR] aplay on PCM hw:0,2 failed at 1/1.
2021-04-22 20:18:47 UTC Sub-Test: [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2021-04-22 20:18:47 UTC Sub-Test: [ERROR]  die()  @  /home/ubuntu/sof-test/test-case/../case-lib/lib.sh
2021-04-22 20:18:47 UTC Sub-Test: [ERROR]  main()  @  /home/ubuntu/sof-test/test-case/check-playback.sh:99
2021-04-22 20:18:47 UTC Sub-Test: [INFO] Starting /usr/bin/sof-logger  -l /etc/sof/sof-tgl.ldc -o /home/ubuntu/sof-test/logs/check-kmod-load-unload-after-playback/2021-04-22-20:18:47-11971/etrace.txt
2021-04-22 20:18:48 UTC Sub-Test: [INFO] pkill -TERM sof-logger
Terminated
2021-04-22 20:18:49 UTC Sub-Test: [INFO] nlines=3 /home/ubuntu/sof-test/logs/check-kmod-load-unload-after-playback/2021-04-22-20:18:47-11971/etrace.txt
2021-04-22 20:18:49 UTC Sub-Test: [INFO] Test Result: FAIL!
2021-04-22 20:18:49 UTC [ERROR] aplay check failed
2021-04-22 20:18:49 UTC [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2021-04-22 20:18:49 UTC [ERROR]  die()  @  /home/ubuntu/sof-test/test-case/../case-lib/lib.sh
2021-04-22 20:18:49 UTC [ERROR]  main()  @  /home/ubuntu/sof-test/test-case/check-kmod-load-unload-after-playback.sh:67
2021-04-22 20:18:49 UTC [INFO] Test Result: FAIL!
```

test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -48,7 +48,9 @@ OPT_HAS_ARG['d']=1         OPT_VAL['d']=3
 OPT_NAME['p']='pulseaudio'   OPT_DESC['p']='disable pulseaudio on the test process'
 OPT_HAS_ARG['p']=0             OPT_VAL['p']=1
 
-func_opt_parse_option "$@"
+fxunc_opt_parse_option "$@"
+setup_kernel_check_point
```
plbossart, line 52: @fredoh9 @marc-hb maybe a stupid question, but there was a similar change from @aiChaoSONG for pm_runtime tests yesterday, so should we add this in ALL tests?

marc-hb: This is more or less what I asked in https://github.com/thesofproject/sof-test/pull/639#issuecomment-818270484

Some (future?) tests may want to scan the entire log since boot though. For instance we discussed one "empty" test only for the purpose of checking SOF modules loaded at boot time. It's still not clear what such a test should do wrt checkpoints, @aiChaoSONG 

After excluding aliases and non-audio tests I found only 2 tests that don't invoke `setup_kernel_check_point()` directly:

```
+test-case/check-audio-equalizer.sh
+test-case/check-volume-levels.sh
```

Would these two tests also timeout if they fail?

marc-hb:
@fredoh9 , @aiChaoSONG can you say more about this TIMEOUT, why does it happen, where and how long it is?

Can we please change this checkpoint failure so it fails _immediately_?

marc-hb:
The modified test failed on all platforms with: `check-kmod-load-unload-after-playback.sh: line 51: fxunc_opt_parse_option: command not found`

It's very good that it failed, no green failure for a change.

It's not so good that it timed-out. Why did the exit_handler not run?

marc-hb:
> Some (future?) tests may want to scan the entire log since boot though. For instance we discussed one "empty" test only for the purpose of checking SOF modules loaded at boot time. It's still not clear what such a test should do wrt checkpoints, @aiChaoSONG

Already answered 3 days ago  by @aiChaoSONG in #653, sorry I've been running behind with code reviews.

marc-hb:
> It's not so good that it timed-out. Why did the exit_handler not run? (it ran). Can we just print some magic string at the end?

All checkpoint timeouts fixed in new #661 "don't die in the die handler"

test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -48,7 +48,9 @@ OPT_HAS_ARG['d']=1         OPT_VAL['d']=3
 OPT_NAME['p']='pulseaudio'   OPT_DESC['p']='disable pulseaudio on the test process'
 OPT_HAS_ARG['p']=0             OPT_VAL['p']=1
 
-func_opt_parse_option "$@"
+fxunc_opt_parse_option "$@"
```
marc-hb, line 51: While fixing this typo, can you also add `set -e` if testing is OK?

If `set -e` requires some other changes then forget it, does not really belong to this PR.

marc-hb (changes requested):

test-case/check-kmod-load-unload-after-playback.sh
aiChaoSONG: @plbossart @marc-hb Previously, when we were using DMESG_KERNEL_LAST_LINE which acted like kernel check point, this variable is implicitly set up in lib.sh. But I think it is better to set up check point explicitly in every test case, this will tell where the log collect is started.

aiChaoSONG: >so should we add this in ALL tests

I have a round of clean up on this within the journalctl PR, only minor number of test case(s) which is not run in PR test or daily test may need this. And also, when this happens now, some boundary condition is triggered.

fredoh9:
https://github.com/thesofproject/sof-test/pull/661 is merged

marc-hb:
I'm confused @fredoh9 , @aiChaoSONG : don't we still want this?

aiChaoSONG:
> I'm confused @fredoh9 , @aiChaoSONG : don't we still want this?

I think we still need this, we don't die but only print error messages, without this, we may see the error messages printed out.

marc-hb:
SOFCI TEST

marc-hb:
New KERNEL_CHECKPOINT discussion in
- #918

PR #658, opened 2021-04-22
[SKIP CI] .github: unhardcode from "master" branch that does not exist anymore

marc-hb:
Funny how no one missed shellcheck and pylint...

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #660, opened 2021-04-22
[SKIP CI] .github: don't run git range checks on mere pushes

marc-hb:
... because there's no ${BASE_REF} and no range to check.

Stops spurious failures like:

 https://github.com/marc-hb/sof-test/runs/2414768508

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,10 +7,14 @@
 # yaml merge-expand pull_request.yml exp.yml &&
 #    diff -b -u pull_request.yml exp.yml
 
+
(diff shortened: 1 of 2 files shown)


marc-hb:
Some background info: https://github.community/t/stop-github-actions-running-on-a-fork/17965

aiChaoSONG (approved):

marc-hb:
Thanks again @aiChaoSONG!  Fast tracking this because of the spam. Pure github actions change, does not affect "real" validation.

Note until a few hours ago when #658 was merged there were no github actions at all, so the spam lasted for only a few hours.

PR #661, opened 2021-04-23
func_exit_handler: don't die when KERNEL_CHECKPOINT is missing

marc-hb:
One should never die in the "die handler", this means the final status
is not printed and causes a Jenkins timeout.

Fixes commit 18c7e6155053 ("sof-test: use journalctl for log
collection")


marc-hb:
Informally approved by @aiChaoSONG in https://github.com/thesofproject/sof-test/pull/653#issuecomment-825328554

Can @fredoh9 or someone else  approve? It's a very small but important fix that avoids TIMEOUTs blocking CI resources for a long time.

fredoh9 (approved):
good point!

PR #629, opened 2021-03-13
check-sof-logger: fail on empty logs

marc-hb:
1st commit:
```
fix some quoting and rephrase some errors
Get the number of shellcheck warnings down a bit.
```
2nd commit:
```
fail when there's only the header
Found by chance while testing Zephyr.
```

PR #662, opened 2021-04-23
hijack.sh: fail if the logger trace is empty

### LINKED ISSUE #297: Check for a valid trace for ALL tests

marc-hb:
_Resubmit #383 to new `main` branch_

Does not completely fixes #297 but goes a long way.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -48,6 +48,12 @@ function func_exit_handler()
         # shellcheck disable=SC2154
         wcLog=$(wc -l "$logfile")
         dlogi "nlines=$wcLog"
+        local nlines; nlines=$(wc -l < "$logfile")
+        # Line 1 is the header
+        if [ "$nlines" -le 1 ]; then
+            dloge "Empty logger trace"
+            exit_status=1
+        fi
     fi
     # when case ends, store kernel log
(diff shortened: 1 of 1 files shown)


marc-hb:
The less ambitious #629 passed except on BYT where it fails consistently

marc-hb:
https://sof-ci.01.org/softestpr/PR662/build672/devicetest/ is all green!

aiChaoSONG (approved):
Note that sof-logger is started by `func_lib_start_log_collect` in lib.sh,  and some cases which only do kernel log check don't call this.

marc-hb:
> and some cases which only do kernel log check don't call this.

`if [ "X$SOF_LOG_COLLECT" == "X1" ]; then` on line 27 should take care of that case. These tests would have failed otherwise.

Merging, if BYT (or other?) starts failing this test again then the #663 workaround should help.

PR #663, opened 2021-04-23
hijack.sh: make late logger config changes to nudge stuck DMA traces

marc-hb:
There are bugs that cause DMA drops and the last few lines to get stuck
somewhere. Changing the logger configuration and waiting for 1 second
seems to be enough to nudge the system and force it to DMA a bit more
logs: hopefully all the logs relevant to the current test.

Running this logger configuration command twice makes it very easy to
observe the stuck lines bug: the "ipc" logs corresponding to this -F
command will appear _only once_ at the end of the slogger.txt DMA trace!

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -31,6 +31,23 @@ function func_exit_handler()
             sleep 1s
         }
 
+        # There are bugs that cause DMA drops and the last few lines to
+        # get stuck somewhere. This hack seems to be enough to nudge the
+        # system and force it to DMA a bit more logs: hopefully all the
+        # logs relevant to the current test.
+        # We must use a component that is available everywhere: pga
+        for i in 1 2; do
+            # Running this twice makes it very easy to observe the stuck
+            # lines bug: the "ipc" logs corresponding to this -F command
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR663/build673/devicetest failed on BYT because it has no "tone" component. The tests timed out because they aborted before printing PASS or FAIL. Both issues should be addressed on v2 just forced-pushed.

Some tests did not fail on BYT, most likely because they don't have `set -e`. See https://github.com/thesofproject/sof-test/issues/312 about `set -e`.

Sampling a few other slogger outputs in build673 shows the desired effect is achieved.

marc-hb:
check-sof-logger failed across all platforms in https://sof-ci.01.org/sofpr/PR4086/build8805/devicetest/ and I don't why yet.

marc-hb:
https://sof-ci.01.org/softestpr/PR663/build674/devicetest/ is all green. Looking at a few random slogger outputs shows that the logs are slightly incomplete as explained in the code comments.

aiChaoSONG (approved):

Issue #321, opened 2020-08-07
check-sof-logger need to check error trace


xiulipan:
Add check to error trace to make sure error trace is working.

Depends on https://github.com/thesofproject/sof/issues/3281

marc-hb:
Implemented.

Issue #298, opened 2020-07-29
check-sof-logger need to report fail is no valid log shows


xiulipan:
From latest test result. logger on legacy platform is not working but the test result of check-sof-logger still show pass. Need to fix the issue in check-sof-logger

```
2020-07-29 02:09:31 UTC [REMOTE_INFO] Found file: /usr/bin/sof-logger 96b4a1605636506fab8cfbac4396f7ce
2020-07-29 02:09:31 UTC [REMOTE_INFO] Checking ldc File: /etc/sof/sof-bdw.ldc ...
2020-07-29 02:09:31 UTC [REMOTE_INFO] Found file: /etc/sof/sof-bdw.ldc 6d96f9315129a366644043bbaab6a96b
2020-07-29 02:09:31 UTC [REMOTE_INFO] Try to dump the dma trace log via sof-logger ...
2020-07-29 02:09:31 UTC [REMOTE_COMMAND] sudo /usr/bin/sof-logger -t -l /etc/sof/sof-bdw.ldc -o /home/ubuntu/sof-test/logs/check-sof-logger/2020-07-29-10:09:31-5838/logger.data.log 2> /home/ubuntu/sof-test/logs/check-sof-logger/2020-07-29-10:09:31-5838/logger.error.log &
2020-07-29 02:09:33 UTC [REMOTE_COMMAND] sudo pkill -9 sof-logger
2020-07-29 02:09:34 UTC [REMOTE_INFO] Log data BEG>>
         TIMESTAMP              DELTA C# COMPONENT          LOCATION                      CONTENT
2020-07-29 02:09:34 UTC [REMOTE_INFO] <
```

mengdonglin:
Same requirement as https://github.com/thesofproject/sof-test/issues/297

marc-hb:
Duplicate of #297, closing

PR #666, opened 2021-04-27
check-sof-logger: rewrite most of it so it can actually find bugs

marc-hb:
Probably the main change is fixing the huge etrace test gaps #321 and
https://github.com/thesofproject/sof/issues/3281

Also fixes DMA trace gaps #297 and #298

I initial tried to preserve some of the existing code but it was just
too bad. PR #161 / commit 7274f490b78 seemed especially bad:

- It tried to ignore a specific `ll drift` error but instead it filtered
out almost every log statement out of... stderr, that does not have show
log statements!! (Just for the record this `ll drift` error has been
downgraded to warning now, see
https://github.com/thesofproject/sof/pull/2686 and
https://github.com/thesofproject/sof/issues/3854)

- That same commit also added code that merely starts the DMA trace with
"there is an error below" (without failing the test) but that's eclipsed
by the entire log that follows. Later, the firmware started printing
ERROR every single time when the ERROR FW ABI prefix was introduced yet
no one ever noticed which proves how useless this prefix is was.

So remove this DMA trace prefix as the purpose of this test is - as
clearly stated in #167 - not to find firmware errors but errors with the
sof-logger itself (even though we never had anything looking at firmware
errors so far)

Don't grep for "error" on stderr: anything on stderr is a logger
failure (not a firmware failure).

Don't require whitespace before the TIMESTAMP header.

Add set -e.

Use shell functions.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -3,24 +3,28 @@
 ##
 ## Case Name: check-sof-logger
 ## Preconditions:
-##    sof-logger install in system path
-##    ldc file is in /etc/sof/
+##    sof-logger installed in system path
+##    ldc file is in /etc/sof/ or /lib/firmware
+##
 ## Description:
-##    Check debug tools sof-logger can success work
+##    Checks basic functionality of the sof-logger itself. Does not test
+##    the firmware, i.e., does NOT fail when errors are found in the
+##    logs.
+##
 ## Case step:
 ##    1. check sof-logger in system
 ##    2. check ldc file in system
 ##    3. run sof-logger
 ## Expect result:
-##    sof-logger already catch some thing
+##    sof-logger produces some output and did not fail
 ##
+
+set -e
 
 # shellcheck source=case-lib/lib.sh
 source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
 func_opt_parse_option "$@"
-
-#TODO: need to add arguments for user to give location for logger and ldc file
 
 setup_kernel_check_point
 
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR666/build677/devicetest is all green!

marc-hb:
@aiChaoSONG , @fredoh9 these "Log $ftype BEG>>" changes seem to break some CI parsing, see for instance the bottom of https://sof-ci.01.org/softestpr/PR666/build677/devicetest/?model=CML_RVP_SDW&testcase=check-sof-logger

Any idea what exactly breaks and maybe how?

aiChaoSONG:
@marc-hb drivers.audio.ci.sof-jenkins/blob/main/json2html.py this file will add some style to plain text log, and transform log into html format, in html format, pair of `<any contents>` can not display.

marc-hb:
Thanks for the pointer! BTW how come these tests pass without any TIMEOUT? They don't print PASS at the end! The similar situation in #661 caused TIMEOUTs

aiChaoSONG:
@marc-hb No, there is!!!, if you check the source from your browser, it looks like this:

image

marc-hb:
Got it, thanks! Looks like I just invented the invisible PASS :-)

marc-hb:
@aiChaoSONG , @fredoh9 , @keqiaozhang : I thought this PR just broke some code looking for `Log data BEG>>` and `<<END whatever` but actually:

- I found nothing looking for `BEG` in the repo @aiChaoSONG  pointed at,
- This has been broken long before this PR, see `check-sof-logger` results in for instance older https://sof-ci.01.org/sofpr/PR3995/build8564/devicetest/?model=CML_HEL_RT5682&testcase=check-sof-logger

marc-hb:
https://sof-ci.01.org/softestpr/PR666/build680/devicetest/?model=APL_UP2_NOCODEC&testcase=check-suspend-resume-with-playback is the usual #650 TIMEOUT, everything else is green.

In the same run the check-sof-logger outputs look OK except the 2, now 4 `logger` tabs are all empty but that's a problem that long predates this PR (example at https://sof-ci.01.org/sofpr/PR3995/build8564/devicetest/?model=CML_HEL_RT5682&testcase=check-sof-logger )

Long story short this PR is ready however I'd like much smaller PR #668 to go first to be cautious and make really really sure the `BEG>>` -> `BEG::` change does not break anything. Maybe the reason it does not break anything is because it's already broken?

marc-hb:
https://sof-ci.01.org/softestpr/PR666/build681/devicetest has the usual #650 TIMEOUT and everything else is green and looks good.

The `dmesg` and `logger` tabs are empty for this test (and for this test only) but that's not new at all.

marc-hb:
SOFCI TEST

aiChaoSONG:
Seems this case generated several logger log files.
image

marc-hb:
> Seems this case generated several empty logger log files.

Yes this is known, UI issue, see comments and links above. It's not caused by this PR. You can still download the logs, only the UI is empty.

The only difference this PR makes is that there are now 4 empty tabs instead of 2.

test-case/check-sof-logger.sh
```diff
@@ -41,57 +45,103 @@ ldcFile=$(find_ldc_file) || die ".ldc file not found!"
...
+        test -e "$stderr_file" || die "$stderr_file" not found
+        if test -s "$stderr_file"; then
+            print_logs_exit 1 "stderr $stderr_file is not empty"
+        fi
+    done
+
+    # Search for the log header, should be something like this:
+    # TIMESTAMP  DELTA C# COMPONENT  LOCATION  CONTENT
+    for f in etrace data; do
+        local tracef="$LOG_ROOT/logger.$f.log"
```
aiChaoSONG, line 128: ditto

test-case/check-sof-logger.sh
```diff
@@ -41,57 +45,103 @@ ldcFile=$(find_ldc_file) || die ".ldc file not found!"
...
-    dloge "Error(s) found in firmware log ${data_file}"
-    func_logger_exit 1
-fi
+    local f
 
-if grep -i -q 'error' "$data_file" ; then
-    dlogw "Catch keyword 'ERROR' in firmware log"
-fi
+    for f in etrace_stderr error; do
+        local stderr_file="$LOG_ROOT/logger.$f.log"
```
aiChaoSONG, line 118: ditto

test-case/check-sof-logger.sh
```diff
@@ -41,57 +45,103 @@ ldcFile=$(find_ldc_file) || die ".ldc file not found!"
...
+    # Print $errmsg twice: - once _after_ the (possibly long) logs
+    # because the end is where everyone logically looks atfirst when the
+    # test fails, and; - also now _before_ the logs in case something
+    # goes wrong and we don't make it until the end.
+    test -z "$errmsg" || dloge "$errmsg"
+
+    local bname
+    for ftype in data etrace error etrace_stderr; do
+        printf '\n\n'
+        bname="logger.$ftype.log"
```
aiChaoSONG, line 101: ditto

test-case/check-sof-logger.sh
```diff
@@ -41,57 +45,103 @@ ldcFile=$(find_ldc_file) || die ".ldc file not found!"
...
+# These filenames are kept for backward-compatibility
+# DMA trace
 data_file=$LOG_ROOT/logger.data.log
+# stderr
 error_file=$LOG_ROOT/logger.error.log
 
-func_lib_check_sudo
+# etrace shared memory mailbox, newer feature.
+etrace_file=$LOG_ROOT/logger.etrace.log
+etrace_stderr_file=$LOG_ROOT/logger.etrace_stderr.log
```
aiChaoSONG, line 56: ditto.

test-case/check-sof-logger.sh
```diff
@@ -41,57 +45,103 @@ ldcFile=$(find_ldc_file) || die ".ldc file not found!"
 
 dlogi "Found file: $(md5sum "$ldcFile"|awk '{print $2, $1;}')"
 
+# These filenames are kept for backward-compatibility
+# DMA trace
 data_file=$LOG_ROOT/logger.data.log
+# stderr
 error_file=$LOG_ROOT/logger.error.log
```
aiChaoSONG, line 52: The web cannot render .log file, so it automatically download it for you, that's what you see in the internal web, please rename it to *.txt.

aiChaoSONG (changes requested):
Once the log file are renamed to *.txt, it is good to merge. thank you @marc-hb

marc-hb:
The `[ 1034.664354] kernel: IO error in rt700_jack_detect_handler, ret -16` in https://sof-ci.01.org/softestpr/PR666/build721/devicetest/?model=CML_RVP_SDW&testcase=check-suspend-resume-5 is the well known  https://github.com/thesofproject/linux/issues/2943

In the same run, the logger tabs are visible and not auto-downloaded like the .log files were. Thanks @aiChaoSONG ! It's still confusing to have all 4 of them named the same and to have the stderr tabs empty but that can be fixed later.

aiChaoSONG (approved):

marc-hb:
Thanks @aiChaoSONG . This had been in review for months and  badly needed for tracing changes like thesofproject/sof#4327 and others. Merging.

PR #668, opened 2021-04-27
check-sof-logger: change BEG>> to BEG:: to avoid HTML conflict

marc-hb:
Also change <<END to ::END and print both files always.

<< and >> seem to corrupt the generated HTML and I could not find
anything parsing these anyway as of now.

See discussion in (much larger) PR #666 for more details.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -76,7 +76,7 @@ fw_log_err=$(grep -i 'error' "$data_file" | grep -v '\.c\:[1-9]')
 
 # '[[:blank:]]TIMESTAMP.*CONTENT$' to filter the log header:
 # TIMESTAMP  DELTA C# COMPONENT  LOCATION  CONTENT
-if [[ ! $(sed -n '/[[:blank:]]TIMESTAMP.*CONTENT$/p' "${data_file}") ]]; then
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR668/build678/devicetest/?model=CML_RVP_SDW&testcase=check-sof-logger (nor any other platform) did not seem to mind the change from `>>` to `::`.

marc-hb:
check-sof-logger test outputs in https://sof-ci.01.org/softestpr/PR668/build679/devicetest/ look fine except for the empty "logger" tabs but these tabs have been empty / broken for a long time, long before this PR. For instance they were already empty/broken in old and unrelated https://sof-ci.01.org/sofpr/PR3995/build8564/devicetest/?model=CML_HEL_RT5682&testcase=check-sof-logger

Log collection seems very complicated, see old discussion on that topic in internal gitlab issue 641

aiChaoSONG (approved):
Thanks, colon is good

marc-hb:
Merged during the CI meeting. According to @aiChaoSONG there is no software expecting `BEG`, it looks like it but it's actually just for humans.

If incorrect then please revert quickly the `BEG` commit but please don't revert the other, regex and unrelated commit that is needed for https://github.com/thesofproject/sof/pull/4105

PR #671, opened 2021-04-29
multiple-pipeline: simplify sleeps

marc-hb:
Having the same -w parameter for two different sleeps in the same
iteration made no sense.

It was not clear what the purpose of this first sleep was, it has been in
multiple-pipeline-capture/playback.sh since the dawn of time (= commit
39cd964) For sure this first sleep was helping hide
"uninterruptible state" bug #472 by giving the DSP more time to wake up!

More recently it was found after merging and reverting an earlier
version of this (PR #543 / commit f93a3c8) that this first sleep
was (accidentally?) giving more time for processes to actually
disappeared after being killed at the end of a test round and not
pollute the next test iteration. Make that clearer by moving the first
sleep to the end of the iteration, right after the kills and hardcode to
one second.

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -47,7 +47,7 @@ OPT_NAME['f']='first'
 OPT_DESC['f']='Fill either playback (p) or capture (c) first or any (a) for all pipelines'
 OPT_HAS_ARG['f']=1         OPT_VAL['f']='p'
 
-OPT_NAME['w']='wait'     OPT_DESC['w']='perpare wait time by sleep'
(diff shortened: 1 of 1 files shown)


marc-hb:
In https://sof-ci.01.org/softestpr/PR671/build685/devicetest/ I hardcoded the loop count to 50 and did not see any pollution of one iteration to the next like  in https://github.com/thesofproject/sof-test/pull/543#issuecomment-828885948

keqiaozhang:
Verified this PR on several TGL platform,  sleep 1 is enough for process to clean up completely.

keqiaozhang (approved):

marc-hb:
https://sof-ci.01.org/softestpr/PR671/build686/devicetest is all green

Issue #672, opened 2021-05-03
[BUG] multiple-pipeline.sh fails due to number of PID mismatch


fredoh9:
Found in GLK but not GLK related. Possibly slow platform might run into same problem.

Ran 7 of aplay/arecord, but detected 6/7. Error~. Then killed 7 of PIDs.

Internal daily test 3669 with GLK_BOB_DA7219.

```
2021-05-02 00:06:25 UTC [REMOTE_ERROR] Running process count is 6, but 7 is expected

19298 aplay -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/zero -q
19307 aplay -D hw:0,1 -c 2 -r 48000 -f S16_LE /dev/zero -q
19317 aplay -D hw:0,5 -c 2 -r 48000 -f S16_LE /dev/zero -q
19325 aplay -D hw:0,6 -c 2 -r 48000 -f S16_LE /dev/zero -q
19332 aplay -D hw:0,7 -c 2 -r 48000 -f S16_LE /dev/zero -q
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 312: 19298 Killed                  aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 312: 19307 Killed                  aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 312: 19332 Killed                  aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 312: 19325 Killed                  aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 312: 19317 Killed                  aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
19354 arecord -D hw:0,1 -c 2 -r 48000 -f S16_LE /dev/null -q
19364 arecord -D hw:0,99 -c 4 -r 48000 -f S16_LE /dev/null -q
/home/ubuntu/sof-test/test-case/../case-lib/lib.sh: line 318: 19354 Killed                  arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
[...]

PR #674, opened 2021-05-05
multiple-pipeline: add delay before ps_checks

fredoh9:
Sometimes last aplay or arecord process might delay to start. This causes
the test case to fail due to number of PIDs mistmatch. This is not 100%
reproducible but slower platform has more chance to fail.

fixes: https://github.com/thesofproject/sof-test/issues/672

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -163,6 +163,10 @@ do
             die "Wrong -f argument $f_arg, see -h"
     esac
 
+    # FIXME: Last aplay or arecord process might delay to start.
+    #        Without this sleep, ps_checks returns failure in some cases
+    sleep 0.5
+
     dlogi "checking pipeline status"
     ps_checks
 


marc-hb (changes requested):
I think we unfortunately need this because of my tentative explanation in #672. If you agree with it then please summarize it in the comment here and maybe add a link to it.

fredoh9:
@marc-hb left a detailed explanation about the problem. We will find better management of background processes... later.
https://github.com/thesofproject/sof-test/issues/672#issuecomment-832403391

aiChaoSONG:
@keqiaozhang @fredoh9 @marc-hb  As I observed from Keqiao's demo, aplay/arecord process from last iteration may not be fully killed at the current iteration, so are you sure it is due to `delay to start`, rather than `not fully killed`?

marc-hb:
BTW I would like to close this PR, superseded by PR #676. So can we keep the discussion in bug #672?

aiChaoSONG:
> Which demo? "Not fully killed" sounds like another issue.

When Keqiao found the issue days before, he ran this case once, and I checked the log.

Without the sleep here, for example, in iteration 0, four aplay/arecord processes are started and killed, PIDs = [0 1 2 3], In the second iteration, we start four aplay/arecord again, PIDs=[4 5 6 7], but in the first ps check of the second iteration, six aplay/arecord processes are detected, PIDs=[0 1 4 5 6 7], two are from iteration 0, in the second ps check of the second iteration, only four pipelines are detected(PIDs=[ 4 5 6 7]), and test case thinks it started 6 process, but only 4 were alive at the end, so it failed.

That's why I said `not fully killed`, resources are released (we can run aplay on the same pcm), but process not fully cleaned.  Hope I have made me clear to you.

fredoh9:
https://github.com/thesofproject/sof-test/pull/676 is merged

marc-hb:
Thanks @aiChaoSONG , that really sounds like a different issue to me. Let's keep an eye on it and if it happens again let's please discuss it in a better place  than in this abandoned PR :-)

fredoh9:
@marc-hb @keqiaozhang @aiChaoSONG Yes that is different issue. If you have seen that problem, please open a new issue. I remember the same error in long time agon. I think that will be not so easy to reproduce the problem.

PR #677, opened 2021-05-11
test-speaker: add argument for loop count

keyonjie:
The existed '-l' is for nloops of speaker-test, which is used to
configure the loop number of left/right .wav files in each playback, it
is not suitable for the loop count for the speaker-test stress.

Change the old speaker-test one to '-L' and add new '-l' for stress test
purpose, to make it consistent with stress test of other cases.

Signed-off-by: Keyon Jie <yang.jie@linux.intel.com>


fredoh9 (approved):
Good catch and improvement. Thank you for bring back '-l' option for the test!

marc-hb (changes requested):
Please don't add new shellcheck warnings. You don't have to fix all warnings if you want to keep your PR small and focused but please don't increase the number of warnings.

sofci:

PR #679, opened 2021-05-19
check-sof-logger: fail when there are multiple sof-logger in PATH

marc-hb:
We don't want to test random versions.

For instance some old version was left in /usr/local/bin here:
http://sof-ci.sh.intel.com/#/result/planresultdetail/4007?model=ADLS_RVP_HDA&testcase=check-sof-logger

You can tell it's old version from the TIMESTAMP header missing `(us)` unit.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -29,9 +29,21 @@ type -a sof-logger ||
     die "sof-logger Not Installed!"
 
 
-if type -a sof-logger | tail -n +2 | grep -q . ; then
-    dlogw "There are multiple sof-loggers in system"
-    dlogw "using $(type -p sof-logger)"
+# Checksum a list of files, one filename per stdin line.
+# Whitespace-safe and shellcheck-approved.
+md5list()
+{
+    while read -r; do md5sum "$REPLY"; done
+}
+
+# Recent Ubuntu versions symlink the entire /bin -> /usr/bin so we
+# cannot just count the number of filenames we found. Count the
+# number of different _checksums_ we found in PATH.
+if type -a -p sof-logger | md5list | awk '{ print $1 }' |
+        sort -u | tail -n +2 | grep -q . ; then
+    dloge "There are different sof-logger in PATH on the system $(hostname)!"
+    type -a -p sof-logger | md5list
+    die "Not testing a random sof-logger version"
 fi
 loggerBin=$(type -p sof-logger)
 dlogi "Found file: $(md5sum "$loggerBin" | awk '{print $2, $1;}')"


marc-hb:
https://sof-ci.01.org/softestpr/PR679/build693/devicetest/ found a spurious `/bin/sof-logger` on the following systems:

sh-tglu-rvp-sdw-03, sh-tglh-rvp-hda-05, sh-cml-sku0955-hda-03

The numerous `arecord: set_params:1349: Channels count non available` failures on  jf-cml-hel-rt5682-01 seem unrelated.

marc-hb:
See also (much) bigger sof-logger test changes in #666. They don't conflict with each other.

aiChaoSONG:
the `/bin` is actually a link to `/usr/bin`, so we may see two logger in the PATH, a third logger should be an issue. how about we compare the md5 checksum of each logger we found, and print warnings only when the checksums are not identical.

fredoh9:
> the `/bin` is actually a link to `/usr/bin`, so we may see two logger in the PATH, a third logger should be an issue. how about we compare the md5 checksum of each logger we found, and print warnings only when the checksums are not identical.

This is good point and good idea also.
One thing I didn't like to install to /usr/bin. /usr/local/bin might be better place to install our own tools and utilities.

marc-hb:
https://sof-ci.01.org/softestpr/PR679/build694/devicetest/ has the same `src/audio/kpb.c:154` panic, everything is green otherwise.

https://sof-ci.01.org/softestpr/PR679/build694/devicetest/?model=TGLU_RVP_SDW&testcase=check-sof-logger has identical /bin/sof-logger and /usr/bin/sof-logger and the test does not fail anymore.


```
kernel: [    3.534923] sof-audio-pci-intel-cnl 0000:00:1f.3: error : DSP panic!
kernel: [    3.534928] sof-audio-pci-intel-cnl 0000:00:1f.3: panic: dsp_oops_offset 788480 offset 788480
kernel: [    3.534932] sof-audio-pci-intel-cnl 0000:00:1f.3: status: fw entered - code 00000005
kernel: [    3.535085] sof-audio-pci-intel-cnl 0000:00:1f.3: error: assertion failed
kernel: [    3.535087] sof-audio-pci-intel-cnl 0000:00:1f.3: error: trace point 00004000
kernel: [    3.535090] sof-audio-pci-intel-cnl 0000:00:1f.3: error: panic at src/audio/kpb.c:154
```

marc-hb:
> > ... compare the md5 checksum of each logger we found, and print warnings only when the checksums are not identical.

> This is good point and good idea also.

Done!

fredoh9 (approved):
Wow, this works! Thank you for implementing md5 check!

PR #680, opened 2021-05-21
tools: sof-kernel-log-check: filter irrelated log brought by verify-kernel-boot-log

XiaoyunWu6666:

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -313,6 +313,11 @@ ignore_str="$ignore_str"'|genirq: Flags mismatch irq .'
 # origin logs seen on BDW platforms
 # DMAR: [Firmware Bug]: No firmware reserved region can cover this RMRR [0x00000000ad000000-0x00000000af7fffff], contact BIOS vendor for fixes
 ignore_str="$ignore_str"'|DMAR: \[Firmware Bug\]: No firmware reserved region can cover this RMRR .'
+# origin logs seen on TGL platforms
+# DMAR: DRHD: handling fault status reg 3
+# DMAR: [DMA Read] Request device [00:02.0] PASID ffffffff fault addr 70ad5000 [fault reason 07] Next page table ptr is invalid
+ignore_str="$ignore_str"'|DMAR: DRHD: handling fault status reg 3'
+ignore_str="$ignore_str"'|DMAR: [DMA Read] Request device'
 
 # dw_dmac logs can be ignored
 # origin logs seen on BDW/BYT/CHT platforms


tools/sof-kernel-log-check.sh
```diff
@@ -345,6 +345,15 @@ ignore_str="$ignore_str"'|iwlwifi [[:digit:].:]+: '
 # SDW related logs
 #
 
+#Irrelated log brought in verify-kernel-boot-log.sh
+#http://sof-ci.sh.intel.com/#/result/planresultdetail/4120?model=CML_RVP_HDA&testcase=verify-kernel-boot-log
+ignore_str="$ignore_str"'|Failed to mount NFSD configuration filesystem'
```
marc-hb, line 350: There is something wrong with this particular system, we shouldn't trying to setup anything NFS related on test systems. Please try `sudo apt remove nfs-common` first and see if this resolves this issue.

tools/sof-kernel-log-check.sh
```diff
@@ -345,6 +345,15 @@ ignore_str="$ignore_str"'|iwlwifi [[:digit:].:]+: '
...
+#http://sof-ci.sh.intel.com/#/result/planresultdetail/3937?model=TGLH_0A70_HDA&testcase=verify-kernel-boot-log
+ignore_str="$ignore_str"'|DMAR: DRHD: handling fault status reg '
+#http://sof-ci.sh.intel.com/#/result/planresultdetail/3937?model=CML_SKU0983_SDW&testcase=verify-kernel-boot-log
+ignore_str="$ignore_str"'|tpm tpm0: tpm_try_transmit: send(): error'
```
marc-hb, line 354: Please first try to:
- disable TPM in the BIOS
- blacklisting `tpm_crb` and `tpm_tis`

tools/sof-kernel-log-check.sh
```diff
@@ -345,6 +345,15 @@ ignore_str="$ignore_str"'|iwlwifi [[:digit:].:]+: '
 # SDW related logs
 #
 
+#Irrelated log brought in verify-kernel-boot-log.sh
+#http://sof-ci.sh.intel.com/#/result/planresultdetail/4120?model=CML_RVP_HDA&testcase=verify-kernel-boot-log
+ignore_str="$ignore_str"'|Failed to mount NFSD configuration filesystem'
+#http://sof-ci.sh.intel.com/#/result/planresultdetail/3937?model=TGLH_0A70_HDA&testcase=verify-kernel-boot-log
+ignore_str="$ignore_str"'|DMAR: DRHD: handling fault status reg '
```
marc-hb, line 352: @kv2019i any comment on this one? I googled it and it seems to be related to some IOMMU (mis)configuration in ACPI... confirm? Any workaround?

marc-hb (changes requested):
Please guard platform-specific additions with `case "$platform" in` as on line 263 so we can keep track of where these errors happen.

tools/sof-kernel-log-check.sh
XiaoyunWu6666: yes , sudo apt remove nfs-common first helps resolve this issue. :)

tools/sof-kernel-log-check.sh
fredoh9: sure, let me check the BIOS setting

fredoh9: I disabled TPM from the BIOS. this error goes away.
Security->TPM 2.0 Security
20210526_231229_tpm

tools/sof-kernel-log-check.sh
kv2019i: @marc-hb Sorry, missed the question, so bit of a delay. This is most probably a BIOS issue. I think it is safe to ignore especially in the initial boot. Otherwise we'd need to disable IOMMU, but that's explicitly enabled in our kconfig, so probably not an option here.

tools/sof-kernel-log-check.sh
XiaoyunWu6666: @fredoh9 Thanks so much fred :)

PR #681, opened 2021-05-24
logger: disable sof-logger for zephyr

keqiaozhang:
since zephyr doesnot support sof-logger, so add a function
to disable logger trace for zephyr during the test.

Signed-off-by: Zhang Keqiao <keqiao.zhang@intel.com>

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -343,3 +343,24 @@ disable_kernel_check_point()
 {
     KERNEL_CHECKPOINT="disabled"
 }
+
+is_zephyr()
+{
+    # check if jq is installed, will remove this part
+    # after all DUTs have jq installed.
+    type -p jq || sudo apt install jq -y
+
+    local manifest=/etc/sof/manifest.txt
+    test -e "$manifest" || return 1
+    jq '.version.firmwareType' "$manifest" | grep "zephyr"
+}
+
+logger_disabled()
+{
+    # disable logger when firmware type is zephyr or -s is set to 0
+    if is_zephyr || [[ ${OPT_VAL['s']} -ne 1 ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -58,7 +58,7 @@ then
 	exit 2
 fi
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 function __upload_wav_file
 {
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -57,7 +57,7 @@ loop_cnt=${OPT_VAL['l']}
 out_dir=${OPT_VAL['o']}
 file_prefix=${OPT_VAL['f']}
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 setup_kernel_check_point
 func_lib_check_sudo
--- a/test-case/check-fw-echo-reference.sh
+++ b/test-case/check-fw-echo-reference.sh
@@ -41,7 +41,7 @@ loop_cnt=${OPT_VAL['l']}
 frames=${OPT_VAL['n']}
 frequency=${OPT_VAL['f']}
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 func_pipeline_export "$tplg" "echo:any"
 setup_kernel_check_point
--- a/test-case/check-keyword-detection.sh
+++ b/test-case/check-keyword-detection.sh
@@ -53,7 +53,7 @@ history_depth=${OPT_VAL['s']}
 preamble_time=${OPT_VAL['p']}
 duration=${OPT_VAL['d']}
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 func_pipeline_export "$tplg" "kpbm:any"
 
--- a/test-case/check-pause-release-suspend-resume.sh
+++ b/test-case/check-pause-release-suspend-resume.sh
@@ -108,7 +108,7 @@ esac
 [[ -z $file_name ]] && file_name=$dummy_file
 
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 setup_kernel_check_point
 
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -74,7 +74,7 @@ case $test_mode in
     ;;
 esac
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 [[ -z $file_name ]] && file_name=$dummy_file
 
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -54,7 +54,7 @@ loop_cnt=${OPT_VAL['l']}
 file=${OPT_VAL['f']}
 
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 # checking if source file exists
 if [[ -z $file ]]; then
--- a/test-case/check-runtime-pm-double-active.sh
+++ b/test-case/check-runtime-pm-double-active.sh
@@ -69,7 +69,7 @@ DEV_LST['playback']='/dev/zero'
 APP_LST['capture']='arecord'
 DEV_LST['capture']='/dev/null'
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 func_pipeline_export "$tplg" "type:any"
 
 for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
--- a/test-case/check-runtime-pm-status.sh
+++ b/test-case/check-runtime-pm-status.sh
@@ -65,7 +65,7 @@ DEV_LST['playback']='/dev/zero'
 APP_LST['capture']='arecord'
 DEV_LST['capture']='/dev/null'
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 func_pipeline_export "$tplg" "type:any"
 
 for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
--- a/test-case/check-signal-stop-start.sh
+++ b/test-case/check-signal-stop-start.sh
@@ -40,7 +40,7 @@ count=${OPT_VAL['c']}
 interval=${OPT_VAL['i']}
 test_mode=${OPT_VAL['m']}
 
-[[ ${OPT_VAL['s']} -eq 1 ]] && func_lib_start_log_collect
+logger_disabled || func_lib_start_log_collect
 
 case $test_mode in
     "playback")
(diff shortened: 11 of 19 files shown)


case-lib/lib.sh
```diff
@@ -343,3 +343,19 @@ disable_kernel_check_point()
 {
     KERNEL_CHECKPOINT="disabled"
 }
+
+is_sof_logger_disabled()
```
marc-hb, line 347: I think `logger_disabled()` is long enough name.

case-lib/lib.sh
```diff
@@ -343,3 +343,19 @@ disable_kernel_check_point()
...
+{
+    # get the firmware type
+    manifest=/etc/sof/manifest.txt
+    FIRMWARE_TYPE=$(grep "firmwaretype" < "$manifest" | awk -F ':' '{print $2}' | sed -e 's/^[ ]*//g;s/\"//g')
+    # disable sof-logger for Zephyr built firmware
+    if [ "$FIRMWARE_TYPE" == "zephyr" ]; then
```
marc-hb, line 353: This  above seems a bit long and over complicated, could this be shorten into something like this:
```shell
if grep `firmwaretype.*zephyr` /etc/sof/manifest.txt; then
```
Just a really simplified example: you may want a slightly more complex regular expression.
```

File: case-lib/lib.sh [hunk 3]
```diff
@@ -343,3 +343,19 @@ disable_kernel_check_point()
 {
     KERNEL_CHECKPOINT="disabled"
 }
+
+is_sof_logger_disabled()
+{
+    # get the firmware type
+    manifest=/etc/sof/manifest.txt
+    FIRMWARE_TYPE=$(grep "firmwaretype" < "$manifest" | awk -F ':' '{print $2}' | sed -e 's/^[ ]*//g;s/\"//g')
```
marc-hb, line 351: No need for the `<`.

No need for the `grep`, `awk` can already do `grep` (and many other things): `awk -F ':' '/firmwaretype/ {print $2 }'`

Regular expressions with only special characters are time-consuming to read and deserve a comment, for instance `# remove whitespace`.

marc-hb (changes requested):
Can you please add one more `is_zephyr()` function? I mean `logger_disabled()` invokes `is_zephyr()` like this:
```shell
if is_zephyr; then
   return 0
else ...
```

Can you please give an example of this `/etc/sof/manifest.txt` file and a link to what code creates and deploys it?

This `/etc/sof/manifest.txt` file may be good enough for CI in the very short term but I don't think it will work for most developers. I compile with and without Zephyr many times every day and I also use `sof-test` yet my `/etc/sof/` has been empty since forever.


@andyross , @kv2019i , @lyakh : any other, simple, automatable and reliable way to tell the difference between plain SOF versus SOF+Zephyr?

PS: I've been using a `.tarball-version` hack that shows up in dmesg but this won't work for CI

case-lib/lib.sh
keqiaozhang: @marc-hb ， /etc/sof/manifest.txt has below information:
```
{
    "date" : "2021-05-21 07:35:42 UTC",
    "version" : {
        "kernelBranch" : "topic/sof-dev",
        "kernelCommit" : "dcea8052",
        "sofBranch" : "main",
        "sofCommit" : "72a51a462fb8"
        "firmwaretype" : "zephyr"
    },
    "filepath" : {
        "kernel" : "http://sof-ci-storage.sh.intel.com/release/linux/topic/sof-dev/dcea8052/linux-image-5.13.0-rc1-daily-nocodec-20210520-0_5.13.0-rc1-topic-sof-dev-dcea8052_amd64.deb",
        "firmware" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/fw/zephyr/sof-apl.ri",
        "ldc" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/fw/zephyr/sof-apl.ldc",
        "lst" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/fw/zephyr/sof-apl.lst",
        "topology" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/tplg/sof-apl-nocodec.tplg",
        "logger" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/tools/sof-logger",
        "ctl" : "http://sof-ci-storage.sh.intel.com/release/sof/main/72a51a462fb8/tools/sof-ctl"
    }
}
```
The firmwaretype is added in the deployment stage. related code is in sof-sh/deploy.sh:
```
# generate manifest file
save_manifest() {
    # save Manifest file to DUT
    cat << END > $MANIFEST_FILE
{
    "date" : "$TODAY",
    "version" : {
        "kernelBranch" : "$KERNEL_BRANCH",
        "kernelCommit" : "$LINUX_SHA",
        "sofBranch" : "$SOF_BRANCH",
        "sofCommit" : "$SOF_SHA"
        "firmwaretype" : "$FW_TYPE"
    },
    "filepath" : {
        "kernel" : "$KERNEL_URL",
        "firmware" : "$FW_URL",
        "ldc" : "$LDC_URL",
        "lst" : "$LST_URL",
        "topology" : "$TPLG_URL",
        "logger" : "$LOGGER_URL",
        "ctl" : "$CTL_URL"
    }
}
END
```
so we will generate the manifest.txt every time when deploy sof environment.

case-lib/lib.sh
keqiaozhang: Good point, I will refine my patch.

marc-hb: I didn't know `manifest.txt` was actually a `manifest.json`. The standard, command-line parser for JSON is `jq`, please drop awk, grep etc. and use `jq`:

```
apt-get install jq
jq '.version.firmwaretype' manifest.txt
 => "zephyr"
```

case-lib/lib.sh
aiChaoSONG: @keqiaozhang Please align variable naming style, `fwType` or `firmwareType`

case-lib/lib.sh
keqiaozhang: Updated. Thanks!

case-lib/lib.sh
kv2019i: @keqiaozhang To facilitate running sof-test standalone, the manifest file should be documented somewhere within sof-test repository (an example manifest file perhaps).

case-lib/lib.sh
```diff
@@ -343,3 +343,26 @@ disable_kernel_check_point()
...
 }
+
+is_zephyr()
+{
+    # check if jq is installed, will remove this part
+    # after all DUTs have jq installed.
+    hash jq 2>/dev/null || sudo apt install jq -y &>/dev/null
+
+    manifest=/etc/sof/manifest.txt
+    jq '.version.firmwareType' "$manifest" | grep "zephyr"
```
lyakh, line 354: how about adding the OS (XTOS / Zephyr) information to firmware extended manifest?

kv2019i (commented):
I tested the patch on a system with Zephyr firmware and it did work as expected. We should make the manifest-based configuration a bit cleaner so it makes sense for standalone usage of sof-test, but otherwise this looks good.

case-lib/lib.sh
```diff
@@ -343,3 +343,26 @@ disable_kernel_check_point()
 {
     KERNEL_CHECKPOINT="disabled"
 }
+
+is_zephyr()
+{
+    # check if jq is installed, will remove this part
+    # after all DUTs have jq installed.
+    hash jq 2>/dev/null || sudo apt install jq -y &>/dev/null
```
marc-hb, line 351: Nit: with `type -p` instead of `hash` you can redirect only stdout and not redirect stderr which is "safer".

There's no need to redirect the output of `apt install`, it's not very verbose, it's good to see when it happened and it's even more important to see any error like proxy configuration or network glitch or dependency problem - many things could go wrong. This is test code, the output does not have to be very "tidy", seeing what's going on is much more important.

case-lib/lib.sh
```diff
@@ -343,3 +343,26 @@ disable_kernel_check_point()
...
+
+    manifest=/etc/sof/manifest.txt
```
marc-hb, line 353: make the variable `local` to avoid clashes.

case-lib/lib.sh
```diff
@@ -343,3 +343,26 @@ disable_kernel_check_point()
...
+{
+    # disable logger for zephyr
+    if is_zephyr; then
+        return 0
+    # disable sof-logger when -s is set to 0
+    elif [[ ${OPT_VAL['s']} -ne 1 ]]; then
+        return 0
+    else
+        return 1
+    fi
```
marc-hb, line 367: Style nit/personal preference:
```shell
if is_zephyr || [[ ${OPT_VAL['s']} -ne 1 ]]; then
   return 0
fi

return 1

```

case-lib/lib.sh
marc-hb: `grep -q`.

I think there's also a `test -e "$manifest" || return 1` missing before the `jq`.

marc-hb (changes requested):
I suggest we merge this with the manifest.txt first to get CI started and then try to find a better test later. Only the `is_zephyr()` function will have to change.

case-lib/lib.sh
keqiaozhang: Concise and reasonable. Updated

case-lib/lib.sh
keqiaozhang: > how about adding the OS (XTOS / Zephyr) information to firmware extended manifest?

I'm OK with that, if we all agree, I can add it in deploy.sh.

case-lib/lib.sh
keqiaozhang: @kv2019i , most of our test scripts have "-s" option, you can set it as 0 to skip the trace collection. for example:
```
./check-playback.sh -d 10 -l 1 -r 1 -s 0
```
if you want to run "run-all-tests.sh" without logger, we can figure out a new solution for it next.

aiChaoSONG:
> I think there's also a test -e "$manifest" || return 1 missing before the jq.

@marc-hb @keqiaozhang  For CI devices, $manifest must exist, otherwise, the deployment failed, so the test is not started.
For personal device, this file may not exist. so I think we should regard '$manifest not exist' as 'zephyr not exist', and continue to check other options. this is for CI and personal device compatibility

case-lib/lib.sh
marc-hb: >  If you want to run "run-all-tests.sh" without logger, we can figure out a new solution for it next.

I have a hack that does this here: https://github.com/marc-hb/sof-test/commit/run-all-nologger

marc-hb (approved):
LGTM but probably not for some others yet.

> For personal device, this file may not exist. so I think we should regard '$manifest not exist' as 'zephyr not exist', 

I think the majority of people who reviewed this PR use Zephyr without a manifest.txt and at least one had a good idea that he will submit immediately after this PR is merged :-)

> Edit: If we just want to test if the FW is zephyr, a grep is enough, we can eliminate the dependency on jq

jq is awesome, it's a great dependency and in this case it makes the manifest-based test very reliable. Now if the manifest.txt test goes away soon then I agree the `jq` dependency will be very short lived in this case...

kv2019i (approved):
Fair enough @marc-hb , let's go with this.

keyonjie:
why does zephyr not support sof-logger? without FW logs, how can we debug the SOF issues?

marc-hb:
We could make sof-logger compatible with Zephyr but the plan is to migrate to https://docs.zephyrproject.org/latest/reference/logging/

For now logging is provided by one of these: https://github.com/zephyrproject-rtos/zephyr/pull/35186

fredoh9 (approved):
didn't know sof-logger is not supported for Zephyr

marc-hb:
3 approvals, 8 days in review, simple and easy to maintain change: merging.

Some very good suggestions and objections: please submit incremental changes.

case-lib/lib.sh
marc-hb: Actually, this can probably be just:
```shell
logger_disabled()
{
   is_zephyr || [[ ${OPT_VAL['s']} -ne 1 ]
}
```

marc-hb:
SKIP check-sof-logger if is_zephyr #724

marc-hb:
#809 just enabled sof-logger with Zephyr. Seems to work fine.

Issue #576, opened 2021-01-13
[RFC] enable sof-test on ChromeOS


aiChaoSONG:
I can make sof-test/test-case/check-playback.sh run on ChromeOS, but there are some issues.
[terminal.txt](https://github.com/thesofproject/sof-test/files/5807075/terminal.txt)

1. the visudo configuration seems only takes effect once.
2. there is no /var/log/kern.log file.
3. the sudo funciton in hijack.sh report a library issue, but the file does exist on the system.
```
dmesg: error while loading shared libraries: libtinfow.so.6: cannot open shared object file: No such file or directory
```
4. no sudo group on ChromeOS
5.  we have to specify bash or python3 when run shell/python scripts, check https://chromium.googlesource.com/chromiumos/docs/+/master/security/noexec_shell_scripts.md
6. journalctl require root permission

**How dependencies are installed?**
- I found a package management tool named chromebrew, this tool can help to install dependencies as apt did for us.
- python3 dependencies should be installed with pip3
- this tool can not run with root.

marc-hb:
https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py instead of chromebrew?

Running everything as root should avoid sudo, at least to get things started.

fredoh9:
For python problems, installing miniconda is the solution. It will resolve python version and packages. I will summarize the procedures.

Need package manager, I found chromebrew useful. expect, graphviz pacakage can be installed
https://medium.com/@dihuta/chromebrew-crew-chromeos-package-manager-a91e56a3d898
```
wget -q -O - https://raw.github.com/skycocker/chromebrew/master/install.sh | bash
```

Need to remount to add executable permission.
```
sudo mount -o remount exec /home
sudo mount -o remount exec /home/chronos
sudo mount -o remount exec /home/chronos/user
```

/var/log/kern.log error, I just "touch /var/log/kern.log

sof-logger is not found. Need to look for same version with current SOF binary

/etc/sof is missing, hence sof-{PLATFORM}.ldc is missing. need to find matching file with current SOF binary

fredoh9:
for miniconda, there is a shell script installer.
```
curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
bash "./Miniconda3-latest-Linux-x86_64.sh" -b -p /usr/local/miniconda3

conda -V
conda init

conda update conda -y
conda create -y -n sof-test python=3.7
conda activate sof-test
[...]

PR #684, opened 2021-06-03
test-case: improve Chrome OS compatibility

aiChaoSONG:
These two patches make check-playback/capture test cases can run on Chrome smoothly. suppose sof-test is downloaded to /usr/local, and other dependencies are installed with [chromebrew](http://skycocker.github.io/chromebrew)

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -335,7 +335,7 @@ is_sof_used()
 # a wrapper to journalctl with required style
 journalctl_cmd()
 {
-   journalctl -k -q --no-pager --utc --output=short-monotonic \
+   sudo journalctl -k -q --no-pager --utc --output=short-monotonic \
      --no-hostname "$@"
 }
 
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -78,8 +78,7 @@ do
         if [ ${OPT_VAL['F']} = '1' ]; then
             fmt=$(func_pipeline_parse_value $idx fmts)
         fi
-        # clean up dmesg
-        sudo dmesg -C
+
         for fmt_elem in $(echo $fmt)
         do
             for i in $(seq 1 $loop_cnt)
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -86,8 +86,7 @@ do
         if [ ${OPT_VAL['F']} = '1' ]; then
             fmts=$(func_pipeline_parse_value "$idx" fmts)
         fi
-        # clean up dmesg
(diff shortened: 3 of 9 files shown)


paulstelian97 (approved):

plbossart (dismissed):
Thanks @aiChaoSONG.
maybe that's a stretch but does this help with Fedora compatibility as well? What's missing now?

aiChaoSONG:
> Thanks @aiChaoSONG.
> maybe that's a stretch but does this help with Fedora compatibility as well? What's missing now?

Some commands in sof-test need root privilege, so they are executed as `sudo cmd`, for example, sudo sof-logger. On Ubuntu, we add current user to sudo group, then we don't need to input password when run cmd with sudo.

The same thing needs to be done for Fedora and Chrome, but there is not a sudo group on these two distros as Ubuntu, what I do is to
- add current user to audio group: `sudo usermod --append --groups audio $USER`
- config audio group in `visudo`, add `audio  ALL=(ALL) NOPASSWD: ALL`

After this, you can run sof-test on fedora, just as you run it on ubuntu.

marc-hb (dismissed):

marc-hb (approved):
`[  833.106967] kernel: IO error in rt700_jack_detect_handler, ret -16` in 
https://sof-ci.01.org/softestpr/PR684/build716/devicetest/?model=CML_RVP_SDW&testcase=check-suspend-resume-5 is unrelated https://github.com/thesofproject/linux/issues/2943

case-lib/lib.sh
```diff
@@ -335,7 +335,7 @@ is_sof_used()
 # a wrapper to journalctl with required style
 journalctl_cmd()
 {
-   journalctl -k -q --no-pager --utc --output=short-monotonic \
+   sudo journalctl -k -q --no-pager --utc --output=short-monotonic \
```
marc-hb, line 338: @aiChaoSONG I just had a look a recent ChromeOS image (less than a couple weeks old) and I found zero trace of `journalctl`, only `rsyslogd`... can you explain?
- https://github.com/thesofproject/sof-test/issues/576#issuecomment-1295641285

aiChaoSONG: At the time I tried sof-test on chrome, there was `journalctl`. I have no idea why they remove journalctl in later chrome image, but there is a new tool for log, `craslog`, which works just like `journalctl`.

marc-hb: You're right: there was journalctl but only for a short time. With your help I found a bit more info, see #576

PR #690, opened 2021-06-07
[WIP][SKIP CI] env-setup: add a script to set up test environment

aiChaoSONG:
Only Fedora is supported now, Ubuntu and ChromeOS is WIP.

### DIFF
--- a/env-setup/env-setup.sh
+++ b/env-setup/env-setup.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+
+set -e
+
+sof_test_dir=$(cd "$(dirname "$0")"; cd ..; pwd)
+
+function get_os_distro()
+{
+    grep -w ID /etc/os-release | awk -F= '{print $2}'
+}
+
+function die()
+{
+    printf "\e[31m%s\n\e[0m" "$*"
+    exit 1
+}
+
+function common_env_setup()
+{
+    printf "Add user %s to audio group\n" "$SUDO_USER"
+    if grep -q audio /etc/group | grep -q "$SUDO_USER"; then
+        printf "User \e[32m%s\e[0m is already in audio group\n" "$SUDO_USER"
+    else
+        usermod --append --groups audio "$SUDO_USER"
+    fi
+
+    printf "Enable NOPASSWD mode for user %s\n" "$SUDO_USER"
+    if grep -q "$SUDO_USER" /etc/sudoers; then
+        printf "NOPASSWD mode is already enabled for %s\n" "$SUDO_USER"
+    else
+        echo "$SUDO_USER ALL=(ALL:ALL) NOPASSWD:ALL" >> /etc/sudoers
+    fi
+
+    printf "Copy dynamic debug file 'sof-dyndbg.conf' to /etc/modprobe.d\n"
+    # Always use the latest file
+    cp sof-dyndbg.conf /etc/modprobe.d
+}
+
+function env_setup_ubuntu()
+{
+    printf "TODO"
+}
+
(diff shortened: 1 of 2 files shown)


aiChaoSONG:
@plbossart If you want to run sof-test on Fedora, please try this script.

env-setup/env-setup.sh
```diff
@@ -0,0 +1,111 @@
...
+        printf "User \e[32m%s\e[0m is already in audio group\n" "$SUDO_USER"
+    else
+        usermod --append --groups audio "$SUDO_USER"
+    fi
+
+    printf "Enable NOPASSWD mode for user %s\n" "$SUDO_USER"
+    if grep -q "$SUDO_USER" /etc/sudoers; then
+        printf "NOPASSWD mode is already enabled for %s\n" "$SUDO_USER"
+    else
+        echo "$SUDO_USER ALL=(ALL:ALL) NOPASSWD:ALL" >> /etc/sudoers
```
marc-hb, line 31: For an interactive system cached credentials are enough, no need for NOPASSWD.

env-setup/env-setup.sh
```diff
@@ -0,0 +1,111 @@
...
+    else
+        printf "\e[31mPlease manually download sof-ctl to /usr/local/bin\e[0m\n"
+    fi
+
+    local platform
+    platform=$("$sof_test_dir"/tools/sof-dump-status.py -p)
+    if test -e "/etc/sof/sof-$platform.ldc"; then
+        printf "%s is properly installed\n" "sof-$platform.ldc"
+    else
+        printf "\e[31mPlease manually download %s file to /etc/sof\e[0m\n" "sof-$platform.ldc"
```
marc-hb, line 79: /lib/firmware/intel/sof is a better place - and supported by sof-test now.

env-setup/sof-dyndbg.conf
```diff
@@ -0,0 +1,78 @@
+# ACPI
```
marc-hb, line 1: Is this file copied from some other place?

env-setup/env-setup.sh
```diff
@@ -0,0 +1,111 @@
...
+
+function env_setup_ubuntu()
+{
+    printf "TODO"
+}
+
+function env_setup_fedora()
+{
+    local PKGS="jq expect alsa-utils-alsabat python3-graphviz python3-numpy \
+python3-scipy"
```
marc-hb, line 47: The list of packages should be in a separate file so anyone can run a simple `dnf install $(cat fedora-requirements)` and ignore the rest of this script. I would not recommend to run this script because of some choices it makes and the silent modifications made to `/etc/modprobe.d/` and `/etc/sudoers`.

On the other hand the list of Fedora dependencies is not really debatable and it's very easy to uninstall some package if undesired.

marc-hb (changes requested):
There are two very different types of test devices: centrally managed DUTs in a farm versus developer systems. About half of this script applies to the former and the other half applies to the latter which means it's suitable for neither.

- The configuration of centrally managed DUTs should be very strictly controlled, enforced and re-enforced to avoid variations and unexpected and time-consuming behaviours. In the future this should be implemented not just with shell scripts only but also with higher level and industry standard tools like Ansible, kickstart and others.

- The configuration of developer systems can be checked and recommended but it should not be "brutally" modified like this. Some people may want NOPASSWD, other developers may prefer more secure options. Some people may already have a carefully crafted `sof-dyndbg.conf` and the reference one should be offered but not enforced and it should certainly not overwrite / destroy the previous file. Etc.

PR #693, opened 2021-06-07
sof-kernel-log-check: filter TME warnings

XiaoyunWu6666:
This is to filter out warning shown below
```
kernel: x86/mktme: No known encryption algorithm is supported: 0x4
```

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -314,6 +314,11 @@ ignore_str="$ignore_str"'|genirq: Flags mismatch irq .'
 # DMAR: [Firmware Bug]: No firmware reserved region can cover this RMRR [0x00000000ad000000-0x00000000af7fffff], contact BIOS vendor for fixes
 ignore_str="$ignore_str"'|DMAR: \[Firmware Bug\]: No firmware reserved region can cover this RMRR .'
 
+
+# TME related warnings can be ignored
+# x86/mktme: No known encryption algorithm is supported: 0x4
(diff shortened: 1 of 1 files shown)


marc-hb:
Can we turn off MKTME in Kconfig instead?

marc-hb:
@XiaoyunWu6666 is this error recent? Or maybe on some new hardware/BIOS? The corresponding code is 3 years old
https://github.com/thesofproject/linux/blame/topic/sof-dev/arch/x86/kernel/cpu/intel.c#L513

XiaoyunWu6666:
> @XiaoyunWu6666 is this error recent? Or maybe on some new hardware/BIOS? The corresponding code is 3 years old
> https://github.com/thesofproject/linux/blame/topic/sof-dev/arch/x86/kernel/cpu/intel.c#L513

Hi @marc-hb . This is quite recent [eg. daily-ADL yesterday , id 4417 , on adlp-rvp-sdw]
They seem to happen on ADL platforms , according to dailytest results

keqiaozhang (approved):
LGTM

kv2019i (approved):

plbossart:
let's merge and make this go away, this is a distraction from more important stuff.

marc-hb:
Quoting Kirill for the record:
> Enumeration code only knows about XTS_128. 0x4 indicates that only XTS_256 is supported by the machine.
> We can make the enumeration code aware about XTS_256 or drop the check as MKTME is not going to be enabled upstream. We only care about how many keys there is for TDX.

In simpler words the code is old and needs an update.

PR #696, opened 2021-06-09
Fix check-kmod-load-unload.sh green failure. Other cleanups

marc-hb:
3 commits, see commit messages.

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -97,15 +97,6 @@ do
     # successful remove/insert module pass
     dlogi "==== firmware boot complete: $idx of $loop_cnt ===="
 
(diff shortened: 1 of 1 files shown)


marc-hb:
The test that failed in https://sof-ci.01.org/softestpr/PR696/build720/devicetest/?model=TGLH_RVP_HDA&testcase=check-suspend-resume-with-playback is not the test that is modified here, so it's not related.

fredoh9 (approved):
the pulseaudio workaround is not the solution, there was some effort to work with pulseaudio, but not so successful any way. Deleting old dead code is good idea for now. Later we need to come up with complete solution.
For the other two commits, nothing to argue.

marc-hb:
10 days in review, fairly small fixes: merging.

PR #706, opened 2021-06-11
multiple-pause-resume: make running all pipelines default

aiChaoSONG:
If there are too many pipelines on a platform, the
test case runs very long, because it will iterate
each combination of two out of all pipelines.

This patch make running all pipelines the default
while still maintaining the ability to catch bugs
under multiple pause resume condition

This patch actually makes the test much more strict because we are running all pipelines and pause/resume them 

Without this patch: multiple-pause-resume.sh runs 2m11.594s on GLK.
With this patch: multiple-pause-resume.sh runs 10.171s on GLK

### DIFF
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -30,7 +30,7 @@ OPT_NAME['l']='loop'     OPT_DESC['l']='loop count'
 OPT_HAS_ARG['l']=1         OPT_VAL['l']=3
 
 OPT_NAME['c']='count'    OPT_DESC['c']='combine test pipeline count'
-OPT_HAS_ARG['c']=1         OPT_VAL['c']=2
+OPT_HAS_ARG['c']=1         OPT_VAL['c']=1
 
 OPT_NAME['r']='loop'     OPT_DESC['r']='pause resume repeat count'
 OPT_HAS_ARG['r']=1         OPT_VAL['r']=3
@@ -86,9 +86,10 @@ do
     fi
 done
 
-# get the min value of TPLG:'pipeline count' with Case:'pipeline count'
-[[ ${#pipeline_idx_lst[*]} -gt ${OPT_VAL['c']} ]] && max_count=${OPT_VAL['c']} || max_count=${#pipeline_idx_lst[*]}
-[[ $max_count -eq 1 ]] && dlogw "pipeline count is 1, don't need to run this case" && exit 2
+max_count="${OPT_VAL['c']}"
+if [ "$max_count" -gt "$PIPELINE_COUNT" ] || [ "$max_count" == "1" ]; then
+    max_count="$PIPELINE_COUNT"
+fi
 
 # create combination list
 declare -a pipeline_combine_lst
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -3,19 +3,31 @@
 set -e
 
 ##
-## Case Name: Run multiple pipeline for pause resume
+## Case Name: multiple-pause-resume
 ## Preconditions:
 ##    N/A
 ## Description:
-##    pickup multiple pipline to do pause resume
-##    fake pause/resume with expect
-##    expect sleep for sleep time then mocks spacebar keypresses ' ' to
-##    cause resume action
+##    Run pipelines in parallel and pause-resume them.
+##
+##    We can use '-c' option to control the number of pipelines
+##    that runs in parallel, the default value for '-c' is 1,
+##    with which all pipelines will run in parallel and get pause-
+##    resumed. Manually specified value should be greater than 1,
+##    and less than the maximum pipeline count.
+##
+##    If pipeline count N (1 < N < Max) is specified, each combination
+##    of N pipelines out of all will be tested. For example,
+##    supposed we have 4 pipelines in topology with ID 0, 1, 2, 3, and
+##    '-c 2' is used, below combination will be iterated.
+##    0,1  0,2  0,3  1,2  1,3  2,3
+##
+##    We use 'expect' to send SPACE to each aplay/arecord process,
+##    to simulate the real pause-resume action.
 ## Case step:
-##    1. run 1st pipeline
-##    2. pickup any other pipeline
-##    3. use expect to fake pause/resume in each pipeline
-##    4. go through with tplg file
+##    1. Get the number of pipelines that will run in parallel from '-c' option
+##    2. Get all the combination of N pipelines out of all.
+##    3. Start pipelines and use expect to send SPACE to pause-resume them
+##    4. Goes to next pipeline combination
 ## Expect result:
 ##    no errors occur for either process
 ##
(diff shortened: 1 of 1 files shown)


keyonjie (approved):
nice change, a little bit scare the stricter test may report more errors to us.

test-case/multiple-pause-resume.sh
```diff
@@ -86,9 +86,9 @@ do
     fi
 done
 
-# get the min value of TPLG:'pipeline count' with Case:'pipeline count'
-[[ ${#pipeline_idx_lst[*]} -gt ${OPT_VAL['c']} ]] && max_count=${OPT_VAL['c']} || max_count=${#pipeline_idx_lst[*]}
-[[ $max_count -eq 1 ]] && dlogw "pipeline count is 1, don't need to run this case" && exit 2
+max_count="${OPT_VAL['c']}"
+[ "$max_count" -lt "${#pipeline_idx_lst[*]}" ] && [ "$max_count" != "1" ] ||
+    max_count=${#pipeline_idx_lst[*]}
```
marc-hb, line 91: This is confusing because it looks like if/then/else but it is not equivalent to if/then/else, compare these two:
```shell
if true; then false; else echo REAL_ELSE;     fi    # does not print anything
   true  &&   false   ||  echo NOT_LIKE_ELSE        # prints NOT_LIKE_ELSE !
```

Avoid using `&&` alone like this, it can also cause problems with `set -e`:
```shell
 set -e ;   false && true;   echo HELLO )    # Prints HELLO
( set -e ; ( false && true) ; echo HELLO )    # Does NOT print HELLO!
```

Explanations:

https://www.shellcheck.net/wiki/SC2015
https://mywiki.wooledge.org/BashGuide/TestsAndConditionals#Control_Operators_.28.26.26_and_.7C.7C.29
http://redsymbol.net/articles/unofficial-bash-strict-mode/#short-circuiting


`||` is great for very simple things like `do_this || die` but for anything more complex it's better to fallback on the "English" and usual if/then/else.

marc-hb (changes requested):

test-case/multiple-pause-resume.sh
aiChaoSONG: let's use if.

marc-hb:
```
In test-case/multiple-pause-resume.sh line 90:
if [ "$max_count" -gt "$PIPELINE_COUNT" -o "$max_count" == "1" ]; then
                                        ^-- SC2166: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
```
https://github.com/koalaman/shellcheck/wiki/SC2166

aiChaoSONG:
SOFCI TEST

test-case/multiple-pause-resume.sh
```diff
@@ -30,7 +30,7 @@ OPT_NAME['l']='loop'     OPT_DESC['l']='loop count'
 OPT_HAS_ARG['l']=1         OPT_VAL['l']=3
 
 OPT_NAME['c']='count'    OPT_DESC['c']='combine test pipeline count'
-OPT_HAS_ARG['c']=1         OPT_VAL['c']=2
+OPT_HAS_ARG['c']=1         OPT_VAL['c']=1
```
marc-hb, line 33: This value shows in the `-h`elp message.  But now it's not the real default value anymore. So the help message needs to be fixed.

test-case/multiple-pause-resume.sh
```diff
@@ -86,9 +86,10 @@ do
     fi
 done
 
-# get the min value of TPLG:'pipeline count' with Case:'pipeline count'
-[[ ${#pipeline_idx_lst[*]} -gt ${OPT_VAL['c']} ]] && max_count=${OPT_VAL['c']} || max_count=${#pipeline_idx_lst[*]}
-[[ $max_count -eq 1 ]] && dlogw "pipeline count is 1, don't need to run this case" && exit 2
+max_count="${OPT_VAL['c']}"
+if [ "$max_count" -gt "$PIPELINE_COUNT" ] || [ "$max_count" == "1" ]; then
```
marc-hb, line 90: Why would the user give a parameter bigger than `$PIPELINE_COUNT`, is this a valid use case? If yes it should be explained in the help and/or comments. If not then it should fail.

marc-hb: What happens now when `${#pipeline_idx_lst[*]}` is 1, does the test run? It didn't before this PR.

marc-hb (changes requested):
I find the pipeline logic of this test was difficult to understand before this commit because the English was quite confusing and some of the help and code comments seemed out of sync with the actual code. After this PR the help and comments are still confusing (unchanged) but now they seem even more out of sync with the new code. Please review and fix the `-h` usage and code comments related to the pipeline logic before changing that logic.


> If there are too many pipelines on a platform, the
test case runs very long, because it will iterate
each combination of two out of all pipelines.

That's not what the -h usage said.


> `# merge all pipeline to the 1 group`

What is "the 1 group"?

marc-hb (changes requested):

aiChaoSONG:
SOFCI TEST

fredoh9:
Thank you @aiChaoSONG and @marc-hb. Indeed it will save a lot of time. But by default it run with all pipelines, this is meaningful for stress test but too extreme test case by default, isn't it? We don't want to just saving testing time but want to use the time wisely.

Here are some inputs/comments from core developers, 

1. Reduce random min/max values to 20-50
2. Find right combinations, no need to find all combinations. for example, SSP (or SDW) + DMIC + HDMI. rather than SSP + SSP, HDMI + HDMI. Test case need to understand the pipelines. :)
3. Rather than just random pause/resume, in addition pause + start/stop or resume + start/stop is better. May be new test case is required for this.

marc-hb:
Even if this code change is not accepted we absolutely want these very valuable documentation fixes (with some corrections).

keyonjie:
@marc-hb @fredoh9 @aiChaoSONG we want this improvement in urgent to help on the pause/release xrun issue reproduction, but 13 days has passed, can we speed up this please? I don't expect perfect documentation coverage in one shot, but a significant improvement with no obvious regression should really go much faster IMHO.

keyonjie:
Sorry, with this change applied, I can't reproduce the TGL_RVP_NOCODEC xrun issue anymore, looks this change remove randomness of the pipelines combination and pause/release sequence.

Can we add randomness of the start/stop/pause/release sequence of each pipeline @aiChaoSONG ? It is in fixed pattern with your change here?

keyonjie:
let's don't change the default behavior of this test case at the moment to avoid missing the hot xrun issues capture by multiple-pause-resume case.

marc-hb:
> but a significant improvement with no obvious regression should really go much faster IMHO.

Everyone wants significant speed improvements, as you found later the important question here now is about the potentially significant _test coverage_ difference. Does not seem like an easy question, for sure it's not easy for me.

The meaning of "No obvious regression" is very different for _test_ code, in fact a test more likely to fail is a better test if it catches more bugs = what saves much, much more time in the end https://deepsource.io/blog/exponential-cost-of-fixing-bugs/

This is a good example of why sof-test is not where it should be. Good test coverage in the shortest amount of time is a very difficult trade-off, especially with limited and unreliable hardware resources[*]. So it requires hard work and expert discussions that take time (and BTW thank you @fredoh9  for leading these discussions right now). Yet many of our experts either never look at sof-test, or when they do they don't spend enough time, misunderstand test code and changes like this one. Inaccurate test documentation contributes to that problem of course.

Look at the lack of "diversity" at https://github.com/thesofproject/sof-test/graphs/contributors

[*] our emulation is not where it should be either.

plbossart:
The test is called 'multiple-pause-resume' for a reason, the multiple referred to "multiple devices".
NAK.

aiChaoSONG:
we still have xrun on multicore, change the default behavior may hide the issue, let's close this for now.

Issue #683, opened 2021-06-02
Irrelevant logs in testcase 'verify-kernel-log'  on multiple platforms


XiaoyunWu6666:
**Describe the bug**
verify-kernel-log is a new-added testcase which is used to detect whether SOF card is boot successfully or not.
However , this case usually fails with some irrelevant logs.
This issue is created so that bugs can be identified in CI testresult.
Also we will add support to filter irrelevant logs later.

**Example of Irrelevant Logs **
【verify-kernel-log】
```
kernel: x86/mktme: No known encryption algorithm is supported: 0x4
```
```
kernel: i8042: Can't read CTR while initializing i8042
```
```
[    0.930571] kernel: DMAR: DRHD: handling fault status reg 3
[    0.930578] kernel: DMAR: [DMA Read] Request device [00:02.0] PASID ffffffff fault addr 709c3000 [fault reason 07] Next page table ptr is invalid
```
_________________________________________________________________________________

【check-suspend-resume-with-capture-25】
```
kernel: r8152 3-8.1:1.0 enx34298f909f0b: can't resubmit intr, status -1
```

XiaoyunWu6666:
[addressed by BIOS configuration]
```
kernel: i8042: Can't read CTR while initializing i8042
```

[addressed by filtering out keywords]
```
kernel: x86/mktme: No known encryption algorithm is supported: 0x4
```
solved by https://github.com/thesofproject/sof-test/pull/693
```
kernel: r8152 3-8.1:1.0 enx34298f909f0b: can't resubmit intr, status -1
```
solved by https://github.com/thesofproject/sof-test/pull/692
```
[    0.930571] kernel: DMAR: DRHD: handling fault status reg 3
[    0.930578] kernel: DMAR: [DMA Read] Request device [00:02.0] PASID ffffffff fault addr 709c3000 [fault reason 07] Next page table ptr is invalid
```
solved by https://github.com/thesofproject/sof-test/pull/691

marc-hb:
Thanks @XiaoyunWu6666 , for merely adding one liners like these I don't think we need github issues, the Pull Requests are enough. Issues should be focused on a single problem to avoid interleaving unrelated discussions and in this case issues would just duplicate Pull Requests.

XiaoyunWu6666:
[...]

PR #708, opened 2021-06-14
sof-kernel-log-check: filter out init error from i8042

kv2019i:
Failing PS2/i8042 driver init has no impact to audio test cases.

Seen on Intel daily test #4567


kv2019i:
@XiaoyunWu6666 @marc-hb It's mentioned (in https://github.com/thesofproject/sof-test/issues/683 ) this is addressed by BIOS configuration, but I'm still seeing on some platforms. I'd prefer to just filter out if this pops up with default settings.

marc-hb (approved):

PR #711, opened 2021-06-19
check-sof-logger: fix bug that did not kill/clean sof-logger -t

marc-hb:
2 commits. Main fix:

Killing sudo does NOT kill the sof-logger child! Use timeout as root
instead.

We probably got away with it because of the various "pkill sof-logger"
out there.

Fixes 8732f67 ("check-sof-logger: rewrite most of it so it can
actually find bugs ")

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -111,9 +111,9 @@ print_logs_exit()
     for ftype in data etrace error etrace_stderr; do
         printf '\n\n'
         bname="logger.$ftype.txt"
-        dlogi "Log $ftype BEG::"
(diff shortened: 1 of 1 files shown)


marc-hb:
The only (alsabat) failure in https://sof-ci.01.org/softestpr/PR711/build746/devicetest/?model=CML_SKU0955_HDA&testcase=check-alsabat-headset-playback is not a test modified by this PR

fredoh9 (approved):
printing filename in empty file is good idea.
I like timeout...wait change also.

marc-hb:
Thanks for the prompt review.

Pure bug fix blocking other stuff, expediting this sorry.

PR #710, opened 2021-06-18
check-sof-logger: reload drivers first

marc-hb:
As we look for the FW ABI banner we must start from a clean state

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #712, opened 2021-06-19
TEST IGNORE check-sof-logger: reload drivers first

marc-hb:
As we look for the FW ABI banner we must start from a clean state


marc-hb:
Test for #710 

https://sof-ci.01.org/softestpr/PR712/build747/devicetest shows BDW and BYT failing systematically.

PR #713, opened 2021-06-21
sof-tplgreader: don't use RE for pipeline filtering

aiChaoSONG:
If we have two pipelines with id=1 and id=10, current
RE is not able to filter these two pipelines from each
other.

This patch deprecates RE and use string comparison for
pipeline filtering.

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -172,8 +172,11 @@ class clsTPLGReader:
                 if 'any' in value or value == ['']:
                     check = True if key in line.keys() else False
                 else:
-                    # match for 'keyword'/'keyword [0-9]' target line
-                    check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', str(line[key]), re.I)]) > 0
+                    # check if current pipeline is the one we want
+                    if type(value) == list:
+                        check = str(line[key]) in value
+                    else:
+                        check = str(line[key]) == value
                 if check is bIn:
                     break
             else:


plbossart:
SOFCI TEST

plbossart:
Timeout of APL_NOCODEC is gone with retest.

I also tested this PR locally on UP2-nocodec and UpExtreme-nocodec, no issues reported.

No idea what the pylint error means or if there's a better way to write this in Python but the concept is sound. Thanks @aiChaoSONG for the fix.

marc-hb (changes requested):
I had to "reverse-engineer" way too much code to review this small fix :-( #714 is a pure doc fix to avoid anyone having to go through the same pain again. Among other things it warns about the missing boolean precedence.

> if type(value) == list:

The previous code supported only  `type(value) == list` so please don't introduce this new "feature". Certainly not silently when fixing something unrelated and... probably never.

tools/sof-tplgreader.py
```diff
@@ -172,8 +172,11 @@ def _filterOutput(self, target_lst, filter_dict, bIn):
                 if 'any' in value or value == ['']:
                     check = True if key in line.keys() else False
                 else:
-                    # match for 'keyword'/'keyword [0-9]' target line
-                    check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', str(line[key]), re.I)]) > 0
```
fredoh9, line 176: Some of Pylint warnings are making sense.

```
tools/sof-tplgreader.py:3:0: W0611: Unused import subprocess (unused-import)
tools/sof-tplgreader.py:4:0: W0611: Unused import os (unused-import)
tools/sof-tplgreader.py:5:0: W0611: Unused import re (unused-import)
```

marc-hb: pylint warnings fixed in no functional change PR #714, please help review

marc-hb (changes requested):
> The previous code supported only type(value) == list so please don't introduce this new "feature". Certainly not silently when fixing something unrelated and... probably never.

@plbossart told me he needs this fix urgently so I applied my own suggestion above and resubmitted #715 to supersede this. Apologies.

aiChaoSONG:
@marc-hb  it is correct to exclude the non-list support, actually I spend some time to understand `check = len ([em for em in value if re.match(em + '$|' + em + '[^a-zA-Z]', str(line[key]), re.I)]) > 0` too.

marc-hb:
First, I'm sorry I rushed #715  before we could have this discussion, that's bad. I blame @plbossart, he made me do it! He's always pushing everyone to rush things out... :'-D

Second, you shouldn't have approved my doc fix #714 if you think the code was supposed to support both lists and non-lists! Because I'm loudly saying the opposite in doc fixes #714

Now back to the main question. This function is invoked in only one other function: `_filter_by_dict()` that consumes the output of `parse_dict()`. In `parse_dict()`, `value` is produced by the expression value.strip.split(',')` **which always produces a list** even when there is a single element in the user input. I traced and verified that.

I did spent a little bit of time looking at the removed regex code and I don't understand how it would make sense if value was not a list but only a string. In that case `em` would iterate on every character of the input string, so for instance for `{ 'type' : 'capture' }` then `em` would iterate like this: `c, a, p, t,...`... how would that make sense?

PR #719, opened 2021-06-28
[DON'T MERGE] Enable DEVRES debug log for kmod test

fredoh9:
Warning: this is very verbose.
Intentionally made first commit to ignore DEVRES log.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -105,6 +105,11 @@ function func_exit_handler()
         done
     fi
 
+    # Currently devres debug log is enabled only for kmod test.
+    # When the test case is failed, devres debug level may not be clear.
+    # So exit handler should reset log level.
+    disable_devres_debug_log
+
     # check if function already defined.
     # on exit check whether pulseaudio is disabled.
     ret=0
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -374,3 +374,26 @@ print_module_params()
     grep -H ^ /sys/module/*sof*/parameters/*
     echo "----------------------------------------"
 }
+
+devres_debug_set()
+{
+    local new="$1"
+    local old
+    old=$(cat /sys/module/devres/parameters/log)
+    if [ "$old" -eq "$new" ]; then
+        dlogi "DEVRES log is already $new"
+        return 0
+    fi
+
+    echo "$new" | sudo tee /sys/module/devres/parameters/log > /dev/null
+}
+
+enable_devres_debug_log()
+{
+    devres_debug_set 1
+}
+
+disable_devres_debug_log()
+{
+    devres_debug_set 0
+}
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -44,6 +44,10 @@ if [ ${OPT_VAL['p']} -eq 1 ];then
     func_lib_disable_pulseaudio
 fi
 
+# Enable DEVRES debug log dynamically, this is very verbose
+# it should be disabled at the end of this test
+enable_devres_debug_log
+
 for idx in $(seq 1 $loop_cnt)
 do
     dlogi "===== Starting iteration $idx of $loop_cnt ====="
@@ -110,3 +114,5 @@ do
         sleep 1
     done
 done
+
+disable_devres_debug_log
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -365,6 +365,12 @@ ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 # iwlwifi 0000:00:14.3: No beacon heard and the time event is over already...
 ignore_str="$ignore_str"'|iwlwifi [[:digit:].:]+: '
 
+# Ignore DEVRES debug log which is ERROR by default
+# DEVRES debug log will be enabled in kmod test
+# There are four OPs from DEVRES, ADD, REP, REM, REL
+# for example, [  182.317913] cml_rt1011_rt5682 cml_rt1011_rt5682: DEVRES REL 00000000618e4fb0 devm_card_release (8 bytes)
+# issue link: https://github.com/thesofproject/linux/issues/3008
+ignore_str="$ignore_str"'| DEVRES [A-Z]{3} '
 
 #
 # SDW related logs


tools/sof-kernel-log-check.sh
```diff
@@ -365,6 +365,11 @@ ignore_str="$ignore_str"'|elan_i2c i2c-ELAN0000:.*: invalid report id data'
 # iwlwifi 0000:00:14.3: No beacon heard and the time event is over already...
 ignore_str="$ignore_str"'|iwlwifi [[:digit:].:]+: '
 
+# Ignore DEVRES debug log which is ERROR by default
+# DEVRES debug log will be enabled in kmod test
+# There are four OPs from DEVRES, ADD, REP, REM, REL
+# for example, [  182.317913] cml_rt1011_rt5682 cml_rt1011_rt5682: DEVRES REL 00000000618e4fb0 devm_card_release (8 bytes)
+ignore_str="$ignore_str"'|DEVRES '
```
marc-hb, line 372: I think we can start by being a bit more specific:
```suggestion
ignore_str="$ignore_str"'| DEVRES [A-Z]{3,}'
```

case-lib/lib.sh
```diff
@@ -364,3 +364,31 @@ logger_disabled()
...
+{
+    local devres_log
+    devres_log=$(cat /sys/module/devres/parameters/log)
+    if [ "$devres_log" -eq 0 ]; then
+        dlogi "DEVRES log is already disabled"
+        return 0
+    fi
+
+    #TODO: below command didn't work, wrongly overidden and cause permission issue
+    #sudo bash -c "echo 0 > /sys/module/devres/parameters/log"
```
marc-hb, line 392: Strange, I prefer the version with `tee` but I tried the other version and it works for me. What does "wrongly overridden" mean?

I don't find this TODO very useful. It doesn't explain why the first command did not work for you and I don't think there's really anything TODO either. If any I'd rather have a comment that explains why the naive `sudo cmd > /sys` does not work.

marc-hb (commented):
Please remove the duplication with something like:
```shell

enable_devres_debug_log()
{
    devres_debug_set 1
}

devres_debug_set()
{
    local new="$1"
    local old
    old=$(cat /sys/module/devres/parameters/log)
    if [ "$old" -eq "$new" ]; then
        dlogi "DEVRES log is already $new"
        return 0
    fi

    echo "$new" | sudo tee /sys/module/devres/parameters/log > /dev/null
}


```

tools/sof-kernel-log-check.sh
fredoh9: OP prints 3 characters only. 
```
dev_err(dev, "DEVRES %3s %p %s (%zu bytes)\n",
			op, node, node->name, node->size);
```
I updated like this,
`ignore_str="$ignore_str"'| DEVRES [A-Z]{3} '`

marc-hb: There is also REPLACE, 7 characters long. This won't catch it.

fredoh9: no, for replace, it print REP. OP prints only 3 characters.

fredoh9: [dmesg.txt.gz](https://github.com/thesofproject/sof-test/files/6735086/dmesg.txt.gz)
I attached dmesg output from a kmod test with DEVRES.

marc-hb (approved):

plbossart (approved):

plbossart:
@fredoh9 why did you merge the main branch in this topic branch? shouldn't this be the other way around?

fredoh9:
> @fredoh9 why did you merge the main branch in this topic branch? shouldn't this be the other way around?

Right, I messed up! Github suggested online merge, that I messed up. I need to do manual rebase.

fredoh9:
Messed up with online rebase. I had to to rebase and force update.

test-case/check-kmod-load-unload.sh
```diff
@@ -110,3 +114,5 @@ do
         sleep 1
     done
 done
+
+disable_devres_debug_log
```
marc-hb, line 118: I just realized that this will stay enabled after this test when it fails. This needs to be called in the exit hander too.

marc-hb (changes requested):

test-case/check-kmod-load-unload.sh
fredoh9: Good point, addressed this issue.

case-lib/lib.sh
```diff
@@ -324,6 +324,10 @@ arecord_opts()
 die()
 {
     dloge "$@"
+
+    # Currently devres debug log is enabled only for kmod test
+    # but this is no harm
+    disable_devres_debug_log
```
marc-hb, line 330: I don't think this is the right place sorry. Tests can fail without calling `die()`, either because of `set -e` or because or bad design or any other reason. I think this should be in `func_exit_handler()`

marc-hb (changes requested):

case-lib/lib.sh
fredoh9: thank you for double checking, agreed!

case-lib/hijack.sh
```diff
@@ -105,6 +105,11 @@ function func_exit_handler()
         done
     fi
 
+    # Currently devres debug log is enabled only for kmod test.
+    # When the test case is failed, devres debug level may not be clear.
+    # So exit handler should reset log level.
+    disable_devres_debug_log
```
marc-hb, line 111: I would have put this earlier and/or next to some other kernel related code like the kernel checkpoint stuff but good enough. You confirm none of the commands above will cause more verbose logs, correct? These verbose logs happen only on driver (un)loading?

marc-hb (approved):

case-lib/hijack.sh
fredoh9: I tested with playback test with devres log enabled. No devres log. But devres may be called in any time. Yes, kmod tests have most devres verbose log.

plbossart: let's use traces as suggested by Andy Shevchenko here; https://github.com/thesofproject/linux/issues/3008#issuecomment-871444928

I also wonder if we might enable more traces that just devres.

fredoh9: Good to know we can dump ftrace dump on oops. 
And also looking at traces, what to enable is not so intuitive.

marc-hb: > let's use traces as suggested by Andy Shevchenko here; thesofproject/linux#3008 (comment)

Does that mean we should not merge this? Or can we merge it and change it or add to it later?

Sorry I don't have the time to take a crash course on devres  events right now.

plbossart: not merge and rethink in light of new information.

fredoh9:
close for now

PR #720, opened 2021-06-28
print module parameter values

fredoh9:
Collect module parameters when kernel boot log is saved. Some of
parameters are very useful to debug potential issue.

This is a output example from CML Helios,
```
---------> Printing module parameters <----------
/sys/module/snd_intel_dspcfg/parameters/dsp_driver = 3
/sys/module/snd_sof/parameters/disable_pause = N
/sys/module/snd_sof/parameters/sof_debug = 0
/sys/module/snd_sof_acpi/parameters/fw_path = (null)
/sys/module/snd_sof_acpi/parameters/sof_acpi_debug = 0
/sys/module/snd_sof_acpi/parameters/tplg_path = (null)
/sys/module/snd_sof_intel_hda_common/parameters/always_enable_dmi_l1 = N
/sys/module/snd_sof_intel_hda_common/parameters/codec_mask = -1
/sys/module/snd_sof_intel_hda_common/parameters/disable_rewinds = N
/sys/module/snd_sof_intel_hda_common/parameters/dmic_num = -1
/sys/module/snd_sof_intel_hda_common/parameters/enable_trace_D0I3_S0 = N
/sys/module/snd_sof_intel_hda_common/parameters/hda_model = (null)
/sys/module/snd_sof_intel_hda_common/parameters/sdw_clock_stop_quirks = 8
/sys/module/snd_sof_intel_hda_common/parameters/use_common_hdmi = Y
/sys/module/snd_sof_intel_hda_common/parameters/use_msi = Y
/sys/module/snd_sof_pci/parameters/fw_path = (null)
/sys/module/snd_sof_pci/parameters/sof_pci_debug = 0
/sys/module/snd_sof_pci/parameters/tplg_path = (null)
----------------------------------------
```

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -364,3 +364,13 @@ logger_disabled()
         return 1
     fi
 }
+
+print_module_params()
+{
+    echo "--------- Printing module parameters ----------"
+    grep -H ^ /sys/module/snd_intel_dspcfg/parameters/*
+
+    # for all the *sof* modules
+    grep -H ^ /sys/module/*sof*/parameters/*
+    echo "----------------------------------------"
+}
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -21,4 +21,6 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 func_opt_parse_option "$@"
 disable_kernel_check_point
 
+print_module_params
+
 sof-kernel-log-check.sh


case-lib/lib.sh
```diff
@@ -364,3 +364,25 @@ logger_disabled()
...
+        fi
+    done
+
+    # for all the *sof* modules
+    for param in /sys/module/*sof*/parameters/*
+    do
+        if [[ -f $param ]]; then
+           printf "$param = "'%s\n' "$(cat $param)"
+        fi
+    done
```
marc-hb, line 385: I think you can replace this loop with just one line:
```shell
grep ^ /sys/module/*sof*/parameters/*
```

test-case/verify-kernel-boot-log.sh
```diff
@@ -21,4 +21,6 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 func_opt_parse_option "$@"
 disable_kernel_check_point
 
+print_module_params
+
```
marc-hb, line 25: Considering the output is short enough I think it's tempting to run this somewhere in `case-lib/lib.sh` so it runs for every test.

marc-hb (changes requested):

case-lib/lib.sh
fredoh9: this is very short and great! Thank you for the recommendation

test-case/verify-kernel-boot-log.sh
fredoh9: For SOF CI purpose, we don't have any use case to change any module parameter. I think once in first test case is good enough.

case-lib/lib.sh
```diff
@@ -364,3 +364,13 @@ logger_disabled()
         return 1
     fi
 }
+
+print_module_params()
+{
+    echo "---------> Printing module parameters <----------"
```
aiChaoSONG, line 370: `<` is very special in html,  please don't use it in our log, because it may cause problem in web page.

marc-hb (approved):

PR #723, opened 2021-06-30
multiple-pause-resume: change default values and fix serious loop index bug

fredoh9:
@ranj063 reported this problem. Occasionally it timeout too early.
```
2021-06-30 17:20:20 UTC Sub-Test: [INFO] pipeline: Port1 with aplay
2021-06-30 17:20:20 UTC Sub-Test: [INFO] pipeline: Port2 with aplay
2021-06-30 17:20:20 UTC Sub-Test: [INFO] Check expect exit status
2021-06-30 17:20:20 UTC Sub-Test: [INFO] Port1 to command: aplay -D hw:0,1 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
spawn aplay -D hw:0,1 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q^M
2021-06-30 17:20:20 UTC Sub-Test: [INFO] Port2 to command: arecord -D hw:0,2 -r 48000 -c 2 -f S16_LE -vv -i /dev/null -q
2021-06-30 17:20:20 UTC Sub-Test: [INFO] wait for expect process finished
===> Fred: only after 5 second, it timeout for 250 second!
2021-06-30 17:20:25 UTC Sub-Test: [ERROR] Still have expect process not finished after wait for 250
root        7937    7931 34 17:20 pts/1    00:00:01 aplay -D hw:0,1 -r 48000 -c 2 -f S16_LE -vv -i /dev/zero -q
root        7946    7942 36 17:20 pts/2    00:00:01 arecord -D hw:0,2 -r 48000 -c 2 -f S16_LE -vv -i /dev/null -q
root        7962    6577  0 17:20 pts/0    00:00:00 grep -E aplay|arecord
2021-06-30 17:20:25 UTC Sub-Test: [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2021-06-30 17:20:25 UTC Sub-Test: [ERROR]  main()  @  /home/ubuntu/sof-test/test-case/multiple-pause-resume.sh
```

PR #727, opened 2021-07-02
multiple-pause-resume: reduce min/max value for faster pause/resume

fredoh9:
Faster pause/resume transition is more effective to find bugs. Default
loop/repeat count are slightly increased. For thorough test, use -l/-r
option to specify loop/repeat count.
run-all-test.sh has -l $small_loop to specify loop count so more iteration
can be run.

Made a separate improvement patch from https://github.com/thesofproject/sof-test/pull/723.

### DIFF
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -27,19 +27,20 @@ OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $T
 OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
 
 OPT_NAME['l']='loop'     OPT_DESC['l']='loop count'
-OPT_HAS_ARG['l']=1         OPT_VAL['l']=3
+OPT_HAS_ARG['l']=1         OPT_VAL['l']=5
 
 OPT_NAME['c']='count'    OPT_DESC['c']='combine test pipeline count'
 OPT_HAS_ARG['c']=1         OPT_VAL['c']=2
 
 OPT_NAME['r']='loop'     OPT_DESC['r']='pause resume repeat count'
-OPT_HAS_ARG['r']=1         OPT_VAL['r']=3
+OPT_HAS_ARG['r']=1         OPT_VAL['r']=5
 
-OPT_NAME['i']='min'      OPT_DESC['i']='random range min value, unit is ms'
-OPT_HAS_ARG['i']=1         OPT_VAL['i']='100'
+# pause/resume interval will be a random value bounded by the min and max values below
+OPT_NAME['i']='min'      OPT_DESC['i']='pause/resume transition min value, unit is ms'
+OPT_HAS_ARG['i']=1         OPT_VAL['i']='20'
 
-OPT_NAME['a']='max'      OPT_DESC['a']='random range max value, unit is ms'
-OPT_HAS_ARG['a']=1         OPT_VAL['a']='200'
+OPT_NAME['a']='max'      OPT_DESC['a']='pause/resume transition max value, unit is ms'
+OPT_HAS_ARG['a']=1         OPT_VAL['a']='50'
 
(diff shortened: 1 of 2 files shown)


fredoh9:
Only SC2206 Shellcheck warings left, not clear how to handle this.
https://github.com/koalaman/shellcheck/wiki/SC2206

marc-hb:
SC2206 is simply saying that (as often) `file_idx_lst=(${file_idx_lst[*]} "/dev/zero")` fails if one of the elements has whitespace. On the next iteration, elements with whitespace will be split into multiple elements. For instance,  `file_idx_lst=(${file_idx_lst[*]} "/home/white space")` will be split into `/home/white` and `space` on the next iteration.

There are (at least) two ways to silence the warning (and support whitespace), the first one:
```shell
file_idx_lst[$i]="/dev/zero"
```
The second one, quoting properly with `@`

```shell
file_idx_lst=( "${file_idx_lst[@]}"  "/dev/zero")
```

The `@` character in `"${myarray[@]}"` automagically quotes whitespace exactly like `"$@"` does.

https://www.gnu.org/software/bash/manual/html_node/Arrays.html

fredoh9:
Last change didn't work. Need to debug.

https://github.com/thesofproject/sof-test/compare/ba8a2d629dfbb0cdcac06ff3ebc5c1797c89fdab..600b1427523fdd809d25925939191e7eec98ecbc

fredoh9:
SOF CI PR test doesn't include multiple-pause-resume yet. SKIP CI for now

test-case/multiple-pause-resume.sh
```diff
@@ -93,10 +93,10 @@ done
 
 # create combination list
 declare -a pipeline_combine_lst
-for i in $(sof-combinatoric.py -n ${#pipeline_idx_lst[*]} -p $max_count)
+for i in $(sof-combinatoric.py -n ${#pipeline_idx_lst[*]} -p "$max_count")
 do
     # convert combine string to combine element
-    pipeline_combine_str="$(echo $i|sed 's/,/ /g')"
+    pipeline_combine_str="${i//,/ }"
```
marc-hb, line 99: Could you extend the comment and add something like "... by replacing commas with spaces for the 'for' loop below". This shell syntax is less obvious and  probably less well known that the `sed` one before.

test-case/multiple-pause-resume.sh
```diff
@@ -170,16 +170,17 @@ do
         echo
         if [ "$(pidof expect)" ]; then
             dloge "Still have expect process not finished after wait for $max_wait_time"
-            # now dump process
-            ps -ef |grep -E 'aplay|arecord' || true
+            # list aplay/arecord process and kill them
+            pgrep -a aplay && pkill -9 aplay
+            pgrep -a arecord && pkill -9 arecord
```
marc-hb, line 175: This is really not just "fixing a trivial shellcheck warning"? This may or may not be a good change but it does not solve a problem I'm aware of and this PR is already pretty long so I would leave the `kill` out for now.

```suggestion
            pgrep -a -f arecord || true
```

marc-hb (approved):
The style fixes are great.
I'd like someone more experienced to review the parameter changes in commit `reduce min/max value for faster pause/resume`. @plbossart , @ranj063 ?

ranj063:
> I'd like someone more experienced to review the parameter changes in commit `reduce min/max value for faster pause/resume`. @plbossart , @ranj063 ?

@fredoh9 have you run any device tests with the modified pause/release wait times?

fredoh9:
@ranj063 I ran this PR on APL UP2 and CML Helios for my test but we haven't run official CI test yet.
@keyonjie said he used to run with this min/max values.

fredoh9:
SOFCI TEST

fredoh9:
Latest device test link is broken although the device test was run fine. I checked the result from internal server, test result 5991 is the one.

I will merge this and open an issue for device test result.

PR #730, opened 2021-07-05
sof_remove.sh: kill sof-loggers first as they hang rmmod forever

marc-hb:
For some reason (bug?) using /sys/kernel/debug/sof/trace hangs rmmod
forever. Note playing audio is not an issue, for instance speaker-test
-s 1 -l 0 is interrupted when unloading the drivers.

This is especially a problem when invoking sof-logger over ssh where
Ctrl-C interrupts ssh but not the sof-logger

Will hopefully avoid TIMEOUTs like
https://github.com/thesofproject/sof/issues/4443 which was likely
introduced by commit 6a8871fb849a ("check-sof-logger: reload drivers
first")

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -36,6 +36,22 @@ exit_handler()
     return "$exit_status"
 }
 
+# Always return 0 because if a lingering sof-logger is an error, it's
+# not _our_ error.
+kill_trace_users()
+{
+    local dma_trace=/sys/kernel/debug/sof/trace
+
+    sudo fuser "$dma_trace" || return 0
+
+    ( set -x
+      sudo fuser --kill -TERM "$dma_trace" || true
+      sudo fuser "$dma_trace" || return 0
+      sleep 1
+      sudo fuser --kill -KILL "$dma_trace" || true
+    )
+}
+
 trap 'exit_handler $?' EXIT
 
 # Breaks systemctl --user and "double sudo" is not great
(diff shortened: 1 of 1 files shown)


marc-hb:
@plbossart , @ranj063 , @kv2019i is it expected that using `using /sys/kernel/debug/sof/trace` hangs `rmmod` forever?

marc-hb:
The alsabat failure in https://sof-ci.01.org/softestpr/PR730/build780/devicetest/?model=APL_UP2_NOCODEC&testcase=check-alsabat-headset-playback has unfortunately become routine recently - and unrelated to this. Everything else green. I checked the check-sof-logger logs and I didn't even see any lingering sof-logger. Any way this  PR does not hurt: TIMEOUTs are bad.

plbossart:
@marc-hb if "someone" in userspace opens a file that's exposed by a driver, it's not illogical to have some refcount that would prevent the module from being unloaded.

plbossart (approved):

fredoh9 (approved):

marc-hb:
> @marc-hb if "someone" in userspace opens a file that's exposed by a driver, it's not illogical to have some refcount that would prevent the module from being unloaded.

I would indeed expect that, but then shouldn't `rmmod` fail immediately instead of hanging forever? I can't think of any specific example right now but in the past I experienced `rmmod` failing immediately many times with an error like "X is in use".

Funny enough some (old?) copies of `man rmmod` have a `--wait` option that gives the user the choice: https://linux.die.net/man/8/rmmod Has this option been removed recently, now it is the mandatory default?

marc-hb:
2 approvals, low-risk, small and easily revertable, "fixing" issue sof issue 4443: merging.

PR #731, opened 2021-07-05
[SKIP CI] README.md: add instructions to turn off sudo logs in journalctl

marc-hb:
I got tired of using grep -v and losing color and being afraid of
buffering

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,3 +146,32 @@ apl
 
 * tplgtool.py
 <br> Dumps info from tplg binary file.
+
+
+## System configuration tips
+
+### sudo noise
+
+sof-test uses sudo a lot which creates a lot of noise in `journalctl`
+output which is especially a problem when testing. To turn off that
+noise first run `sudo visudo` and add the following line:
+
+```
+Defaults:USER_LOGIN,root !log_allowed
+```
+For more see `man sudoers`.
+
+Then add the following line at the _top_ of `/etc/pam.d/sudo`
+
+```
+# Be "done" when the pam_succeed_if.so arguments are matched; don't
+# process other lines. If not matched then "ignore" this line.
+# Also support "double sudo" :-(
+session [success=done default=ignore] pam_succeed_if.so quiet         uid = 0 ruser = root
+session [success=done default=ignore] pam_succeed_if.so quiet_success uid = 0 ruser = USER_LOGIN
+```
+
+Note PAM security configuration is complex and
+distribution-specific. This was tested only on Ubuntu 20.04. See `man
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,3 +146,32 @@ apl
...
+For more see `man sudoers`.
+
+Then add the following line at the _top_ of `/etc/pam.d/sudo`
+
+```
+# Be "done" when the pam_succeed_if.so arguments are matched; don't
+# process other lines. If not matched then "ignore" this line.
+# Also support "double sudo" :-(
+session [success=done default=ignore] pam_succeed_if.so quiet         uid = 0 ruser = root
+session [success=done default=ignore] pam_succeed_if.so quiet_success uid = 0 ruser = your_id
```
plbossart, line 171: @marc-hb @fredoh9 is this enabled in deploy scripts?

fredoh9: No, I don't think so

marc-hb:
@plbossart , @fredoh9  any objection?

README.md
```diff
@@ -146,3 +146,32 @@ apl
...
+## System configuration tips
+
+### sudo noise
+
+sof-test uses sudo a lot which creates a lot of noise in `journalctl`
+output which is especially a problem when testing. To turn off that
+noise first run `sudo visudo` and add the following line:
+
+```
+Defaults:your_unix_id,root !log_allowed
```
fredoh9, line 160: your_unix_id and below your_id, can you make them as same? and all CAPITAL? Like YOUR_ID. I was about to copy as it is. :)
Thank you for the tip.

fredoh9: And we have env-check.sh. Practical version of README. Can you add this description to it?(only help text)
Finally can we run this by a script? I would want to deploy this by default for SOF CI devices.

marc-hb: > your_unix_id and below your_id, can you make them as same? and all CAPITAL? Like YOUR_ID.

Done.

> And we have env-check.sh. Practical version of README.

There's some overlap but I don't think these have the same role.

> Can you add this description to it?(only help text)

I would assume that someone who runs `env-check.sh` has already read the `README.md`. I don't think we should cater for people who don't read the `README.md` first, otherwise we may end up with a `README.1st` that says please read the `README.md` file?

> Finally can we run this by a script? 

Yes, that would be awesome but that's a lot more work and figuring out and testing this already took me more than 1 hour. BTW we should also stop re-inventing Ansible.

> I would want to deploy this by default for SOF CI devices.

Only after the corresponding has been locally and successfully tested  by developers for some time of course :-)

fredoh9 (approved):

PR #732, opened 2021-07-05
sof-tplgreader: improve filter for pipeline group

aiChaoSONG:
Previous fix ce58aeda removes unclear RE for pipeline
filtering, but it also removes the ability to filter
a group of pipelines.

For example, pcm names for HDMI pipelines are HDMIn,
where n is a number, with previous RE, we are able
to filter out all HDMI pipelines with '~pcm:HDMI'.
With the fix ce58aeda, we can only filter out explicit
pipeline, for example HDMI1.

This patch brings group filter feature back, while
keeping the code clear.

### DIFF
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -182,8 +182,12 @@ class clsTPLGReader:
                 if 'any' in value or value == ['']:
                     check = True if key in line.keys() else False
                 else:
-                    # check if current pipeline is among the ones we want
-                    check = str(line[key]) in value
+                    # Use full match for numerical values, and use substring
+                    # match for string values
+                    if str(line[key]).isdigit():
+                        check = str(line[key]) in value
+                    else:
+                        check = any([em in str(line[key]) for em in value])
                 if check is bIn:
                     break
             else:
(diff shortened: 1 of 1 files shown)


keqiaozhang (approved):
LGTM

tools/sof-tplgreader.py
```diff
@@ -183,7 +183,7 @@ def _filterOutput(self, pipelines, filter_dict, bIn):
                     check = True if key in line.keys() else False
                 else:
                     # check if current pipeline is among the ones we want
-                    check = str(line[key]) in value
+                    check = any([em in str(line[key]) for em in value])
```
marc-hb, line 186: Looks like we're going from an exact match to substring match without a single comment, documentation or `--help` change. People should be able to use these filters without having to find line number 186 and know enough Python to know  what `str1 in str2` does.

marc-hb (changes requested):
> This patch brings group filter feature back,

This does not explain what the new feature is (plus it's "hidden" in the commit message)

tools/sof-tplgreader.py
aiChaoSONG: updated

marc-hb (changes requested):
> NOTE: numerical values are filtered by full match for exact value match,
and string values are filtered by substring match for filtering a group of
pipelines, for example, HDMI pipelines.

So what happens if there is an "HDMI12" pipeline and I try to match it with '12'? Is it full string  match or substring match ? This needs to be clearer sorry. Don't hesitate to add examples.

tools/sof-tplgreader.py
```diff
@@ -182,8 +182,12 @@ def _filterOutput(self, pipelines, filter_dict, bIn):
                 if 'any' in value or value == ['']:
                     check = True if key in line.keys() else False
                 else:
-                    # check if current pipeline is among the ones we want
-                    check = str(line[key]) in value
+                    # Use full match for numerical values, and use substring
+                    # match for string values
+                    if str(line[key]).isdigit():
+                        check = str(line[key]) in value
```
marc-hb, line 188: So  let's take my  previous exemple of an "HDMI42" pipeline and "42" filter. When I read this in the help below:
```
 -f "type:playback"` -> playback pipelines (value substring match)
 -f "id:3"` -> pipeline whose id is 3 (value full match)
```
I understand that it will be a full match because `'42'.isdigit()` is `True`. Now that's very clear with the examples, thanks! But here you seem to test `"HDMI42".isdigit()`?

aiChaoSONG: for a pipeline, there are id, name, and other key:value pairs. some value is numerical (id, rate) and others are string(pcm, eq, pga), but the problem is that there is no way to know the value is numerical or not. 

the previous RE is not good, because we cannot tell the different between `id:1` and `id:10`, ce58aed remove RE and use full match, which is not good either, because sometimes, we need to exclude a type of pipelines, for example, HDMI pipelines. the example, `HDMI42` is in pcm kv pair, which should be regarded as string value, btw, 42 is not an id for the pipeline, if we want to match id, use `id:42,43,44`

The 'if all bit in a string is digit, it is numerical value, others are string value' is just a strategy, which is good for value type identification.

marc-hb: I don't think you understood my (relatively simple) comment:
- According to the help page, `isdigit()` is run _on the filter_
- When I read the code, I understand that `isdigit()` is run _on the pipelines_

Which is it?

aiChaoSONG: Finally, I get the point, thanks. help message updated.

marc-hb (approved):

aiChaoSONG:
on device test passed, and let's merge this, we need this on some device that HDMI is not ready

Issue #725, opened 2021-07-01
[BUG] sof-test only works when SOF card is the first index, fails with sof-probe as first index


plbossart:
**Describe the bug**

With https://github.com/thesofproject/linux/pull/3007, it can happen that the new 'probe' card is at index0 and the SOF card at index1.

This can also be forced with a module option such as
````
options snd slots=,,,snd_soc_sof_sdw
````

````
root:/proc/asound# more cards 
 0 [sofprobe       ]: sof-probe - sof-probe
       
 3 [sofsoundwire   ]: sof-soundwire - sof-soundwire
````

In both cases, sof-test fails to identify the sof card and generates a lot of issues.

The errors seem to be predominantly due to sof-tplgreader.py and using card0 when card3 should be used.

It could also be that the sof-probe is parsed as a bona fide card, maybe a different name is required.

[log.txt](https://github.com/thesofproject/sof-test/files/6750691/log.txt)

**To Reproduce**
use https://github.com/thesofproject/linux/pull/3007

**Expected behavior**
sof-test behaves the same with and without probes.

**Comment by plbossart** [comment]
@ujfalusi @fredoh9 @marc-hb @aiChaoSONG FYI

**Comment by aiChaoSONG** [comment]
@plbossart The card number is dinamically acquired from /proc/asound/cards like this,
```
if [ ! "$SOFCARD" ]; then
	# $1=$1 strips whitespaces
	SOFCARD=$(grep -v 'sof-probes' /proc/asound/cards |
		awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1; exit 0;}')
fi
```

**Comment by ujfalusi** [comment]
@aiChaoSONG, this is what I have on my test device:
```
# cat /proc/asound/cards 
 0 [sofprobe       ]: sof-probe - sof-probe
                      DellInc.-Latitude5421--020BN2
 1 [sofhdadsp      ]: sof-hda-dsp - sof-hda-dsp
                      DellInc.-Latitude5421--020BN2
```
and the SOFCARD would be:
```
# grep -v 'sof-probes' /proc/asound/cards | awk '/sof-[a-z]/ && $1 ~ /^[0-9]+$/ { $1=$1; print $1; exit 0;}'
0
```
Which is because the card name is `sof-probe` and _not_ `sof-probes`.
But this is still not safe as we could in theory can have other sound card than sof-probes before the SOF card.

[...]

PR #733, opened 2021-07-05
lib.sh: exclude sof-probe in detecting card number

aiChaoSONG:
The card name for probe feature is changed to
sof-probe, change it in sof-test accordingly.

Fixes: #725

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -36,7 +36,7 @@ fi
 # setup SOFCARD id
 if [ ! "$SOFCARD" ]; then
 	# $1=$1 strips whitespaces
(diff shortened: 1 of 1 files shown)


ujfalusi:
@marc-hb, @aiChaoSONG, I'm naming it back the probes card to `sof-probes` as it supposed to be, I did not dropped the `s` consciously for sure, I in fact changed the prefixes in the final driver to sof-probes...
This patch will pick exclude both the correct `sof-probes` and the incorrect `sof-probe` card, I don't have any objections.

ujfalusi (approved):

plbossart (approved):

PR #736, opened 2021-07-09
[SKIP CI] migrate alsa_settings from internal sof-framework repo

fredoh9:
alsa settings per platform is required before running check-alsabat.sh.
These scripts have proper gain settings, required switches on etc.


marc-hb:
I assume these files are strictly identical to the original ones?

Can you add a second commit that starts each file with `set -e`? I think it's a great opportunity.

Different commit to keep the first commit easy to compare with the original.

marc-hb:
https://sof-ci.01.org/softestpr/PR736/build788/devicetest/?model=ADLP_RVP_NOCODEC&testcase=verify-kernel-boot-log looks like a misconfiguration.

fredoh9:
I put "SKIP CI" this is nothing to do with CI yet

marc-hb (dismissed):
I'm approving the idea, I did not compare the files with the originals

fredoh9:
There are some changes in past a few days. :) I will re-sync and update this.

marc-hb:
> There are some changes in past a few days. :) I will re-sync and update this.

Actually: please provide the internal SHA1 you copy from when you do so.

marc-hb (approved):
Again, approving the logic but I didn't compare this to the internal copy.

PR #737, opened 2021-07-13
[SKIP CI] check-volume-levels: fix KERNEL_CHECKPOINT not set error

fredoh9:
KERNEL_CHECKPOINT is mandatory to pass the test case.

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -8,7 +8,6 @@ mydir=$(cd "$(dirname "$0")"; pwd)
 DYNDBG="/etc/modprobe.d/sof-dyndbg.conf"
 
 # check for the system package
-out_str=""
 func_check_pkg(){
     if command -v "$1" >/dev/null; then
         return
@@ -46,7 +45,7 @@ func_check_exec_binary() {
     fi
 }
 
-check_res=0
+out_str="" check_res=0
 printf "Checking for some OS packages:\t\t"
 func_check_pkg expect
 func_check_pkg aplay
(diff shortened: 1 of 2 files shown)


env-check.sh
```diff
@@ -67,6 +67,23 @@ else
     printf "$out_str"
 fi
 
+# octave packages are required only for check-volume-levles.sh
```
marc-hb, line 70: typo

env-check.sh
```diff
@@ -67,6 +67,23 @@ else
...
+# Good to check upfront but this can be optional requirement
+out_str="" check_res=0
```
marc-hb, line 72: Missing semi-colon, empty `out_str`: that does not seem right.

test-case/check-volume-levels.sh
```diff
@@ -35,6 +35,7 @@ OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $'
 OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
 
 func_opt_parse_option "$@"
+setup_kernel_check_point
```
marc-hb, line 38: This code looks like it was missing but the commit message does not really make sense to me.
One optional but useful thing for the commit message would be: how you found that this was missing.

marc-hb (changes requested):

env-check.sh
fredoh9: You mean `out_str=""; check_res=0` but it worked fine and no shellcheck error/warning for this.

any harm empty `out_str` by assigning ""?

env-check.sh
fredoh9: thanks

test-case/check-volume-levels.sh
fredoh9: I updated again, can you review?

env-check.sh
marc-hb: My bad, `out_str="" check_res=0` is perfectly correct without a semi-colon: https://www.gnu.org/software/bash/manual/html_node/Simple-Command-Expansion.html

`out_str` is already defined at the top of the script, redefining it here is confusing because this is not done for any other check. I think this is the contract: if a function sets `check_res` then it MUST set `out_str` too.

marc-hb (changes requested):
Besides the duplicate `out_str=""` everything looks good.

singalsu (approved):

marc-hb (approved):
The out_str + check_res confused me, thanks for your patience

singalsu (approved):

env-check.sh
```diff
@@ -67,6 +66,22 @@ else
     printf "$out_str"
 fi
 
+# octave packages are required only for check-volume-levels.sh
+# Good to check upfront but this can be optional requirement
+out_str="" check_res=0
+func_check_pkg octave
+func_check_pkg octave-signal
+func_check_pkg octave-io
```
marc-hb, line 74: This does not work, that function name is confusing. Fix submitted in #846

PR #738, opened 2021-07-14
kernel-log-check: filter i915 supsending crtc logs

XiaoyunWu6666:
on EHL platforms , when check suspend resume , 'kernel: i915 0000:00:02.0: [drm] *ERROR* Suspending crtc's failed with -22' error happens and leads testcase to fail .
Since this is irrelevant to audio , filter it so that the testcase can run successfully 

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -283,6 +283,11 @@ case "$platform" in
         # i915 0000:00:02.0: [drm] *ERROR* AUX A/DDI A/PHY A: not done (status 0xad4003ff)
         ignore_str="$ignore_str"'|i915 [[:digit:].:]+: \[drm\] \*ERROR\* AUX .+'
         ;;
+    ehl)
+	# i915 crtc logs can be ignored
+	# origin logs seen on EHL_RVP_I2S platforms
+	# i915 0000:00:02.0: [drm] *ERROR* Suspending crtc's failed with -22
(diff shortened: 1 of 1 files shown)


tools/sof-kernel-log-check.sh
```diff
@@ -283,6 +283,11 @@ case "$platform" in
         # i915 0000:00:02.0: [drm] *ERROR* AUX A/DDI A/PHY A: not done (status 0xad4003ff)
         ignore_str="$ignore_str"'|i915 [[:digit:].:]+: \[drm\] \*ERROR\* AUX .+'
         ;;
+    ehl)
+	# i915 crtc logs can be ignored
+	# origin logs seen on EHL_RVP_I2S platforms
+	# i915 0000:00:02.0: [drm] *ERROR* Suspending crtc's failed with -22
+	ignore_str="$ignore_str"'|i915 [[:digit:].:]+: \[drm\] \*ERROR\* Suspending .+'
```
fredoh9, line 290: Instead '.+' just specify the rest of the log, "crtc's failed with".
We don't want random matching unless there are some error patterns.

marc-hb:
> Since this is irrelevant to audio 

Not every i915 error is irrelevant to audio, all display connectors carry audio. I would like @kv2019i to confirm this one.

fredoh9 (approved):
Looks good, Thanks!

@XiaoyunWu6666 for next time, please add description to commit message as well. I see good description in github description but commit message is empty.

marc-hb:
Here's an example of a display problem impossible to ignore: thesofproject/linux#3042

kv2019i:
Out last week, but this seems fine (especially when limited to ehl platform).

PR #746, opened 2021-07-21
Add volume_levels into run-all-tests.sh

singalsu:
This test step checks with audio level measurement that the
correct volume gain and mute commands are applied to each
audio channel. The test script requires Octave.

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -65,7 +65,7 @@ sof-logger
 '
 
 # Requires Octave
-# testlist="$testlist volume_levels"
+testlist="$testlist volume_levels"
 
 main()
 {


singalsu:
Not useful, random manual testers likely do not have octave installed.

marc-hb:
Is octave difficult to install?

singalsu:
> Is octave difficult to install?

No, the needed things can be installed in Debian flavors with "apt-get install octave octave-signal octave-io". That gives core Octave, signal processing, and functions to import/export data from various formats.

marc-hb:
OK then I see no reason not to enable this. Moreover run-all-tests.sh is not used in automation, so we don't even need to worry about deploying it to the test farm (I mean at least not in this PR). @fredoh9 even added the packages in `env-check.sh` already, see commit f9587e073aa6

marc-hb:
SOFCI TEST

marc-hb:
Script not used by CI, merging.

marc-hb (approved):

PR #748, opened 2021-07-22
check-alsabat: run alsa settings before running alsabat test

fredoh9:
If alsa control settings are off, alsabat test tends to fail. Need to run
recommended alsa settings before running alsabat test. To determine right
script, DEVICE0_MODEL need to be defined.


case-lib/lib.sh
```diff
@@ -412,3 +412,26 @@ print_module_params()
...
 }
+
+# check-alsabat.sh need to run optimum alsa control settings
+# param1: platform name
+set_alsa_settings()
+{
+    local PNAME="$1"
+    dlogi "alsa setting for $PNAME"
+    case $PNAME in
+        APL_UP2_NOCODEC | APL_UP2_NOCODEC_ZEPHYR | CML_RVP_NOCODEC | JSL_RVP_NOCODEC | JSL_RVP_NOCODEC_ZEPHYR)
```
XiaoyunWu6666, line 423: hi @fredoh9  , will TGLU_RVP_NOCODEC_CI use CAVS_NOCDEC.sh here ? I just found 'amixer -c sofnocodec cset name='PGA1.1 1 Master Playback Volume' 32' too high for TGLU_RVP_NOCODEC _CI

test-case/check-alsabat.sh
```diff
@@ -58,6 +58,15 @@ then
 	exit 2
 fi
 
+# If DEVICE0_MODEL is defined, set proper gain for the platform
+if [ -z "$DEVICE0_MODEL" ]; then
```
aiChaoSONG, line 62: `DEVICE0_MODEL` is what we set in DUT env setup. But now a device may be selected under multiple MODEL condition, so it may not be correct. for example, we are running a apl_nocodec_zephyr test, but the value for `DEVICE0_MODEL` is apl_nocodec

fredoh9: Good point, we don't need to check ZEPHYR separately. ZEPHYR platforms use same tplg, I clean up ZEPHYR

fredoh9:
I checked the log from device test, looks all good.

aiChaoSONG:
Question: should we run alsa settings for all test cases or only for specific test cases? I think some other case also need this to work properly, or change some specific setting based on this?

If we want to run alsa settings before all test cases, then I suggest to make 
```
# If MODEL is defined, set proper gain for the platform
if [ -z "$MODEL" ]; then
    # treat as warning only
    dlogw "NO MODEL is defined. Please define MODEL to run alsa_settings/MODEL.sh"
else
    #dlogi "apply alsa settings for alsa_settings/MODEL.sh"
    set_alsa_settings "$MODEL"
fi
```
a function, or just integrate into the `set_alsa_settings function`. whichever case want to run alsa settings, it just calls a function or this piece of code will be duplicated elsewhere

PR #750, opened 2021-08-05
[SKIP CI] check-alsabat: dump amixer contents when fails

fredoh9:
Most of alsabat test failures are due to wrong alsa settings. To debug
the failure, amixer contents dumps are very useful.

One of example, recently TGLH_0A5E_SDW failed on alsabat test. By the captured wav, it had overflow problem.
5685?model=TGLH_0A5E_SDW&testcase=check-alsabat-headset-playback

After I sshed to the device and checked the settings, there is no alsa setting issue. It might be changed after that. We need the alsa contents dump when it fails.


aiChaoSONG (approved):
The name alsa is too general, may be `amixer_setting.txt`?, any way, let's dump it.

fredoh9:
Thanks, good idea. I renamed to amixer_settings.txt

aiChaoSONG (approved):

Issue #471, opened 2020-10-24
Remove sof-test dependency on topology file, use /proc instead


marc-hb:
From @ranj063 
> @plbossart @aiChaoSONG @marc-hb this PR (#430) is in the right direction. I think we should decouple sof-test from the topology file even when testing with the SOF driver and use the procfs info for the card/pcm info and even for the topology graph using the asoc dapm info. This will also make sure that we dont have assumptions about the default card number. What do you think?

marc-hb:
From @plbossart 
> Agree that if we can use the /proc info then we have all the information needed for tests with PCM devices.
For probes it's actually the only solution since we expose a compressed interface that will never be shown in the topology but will be shown in the card properties (if it's not show it's a bug). The probes are like 'blue wires', they are not set in stone and not shown in the topology.

> Having the topology would only help to show the graph as an information but would not have an functional input into the tests.

marc-hb:
> Agree that if we can use the /proc info then we have all the information needed for tests with PCM devices.

PR #430 has the following code:
```python
for pcm in card_info['pcm']:
                # There are limited pipeline parameters in the proc file system,
                # add some default parameters to make use of sof-test for legacy HDA test
                pcm['fmt'] = 'S16_LE'
                pcm['fmts'] = 'S16_LE S24_LE S32_LE'
                pcm['rate'] = '48000'
                pcm['channel'] = '2'
                pcm['dev'] = 'hw:{},{}'.format(card_id, pcm['id'])
```
@ranj063 , @plbossart  Is this not actually needed for testing?

BTW `Auxiliary devices in SOF break SOFCARD` #466 seems related.

aiChaoSONG:
@marc-hb For SOF test, we have to depend on topology. Because we cannot get the information from proc, eg:
- We can't identify WoV pipeline/Echo Reference pipeline/DSM pipeline from proc, which need special test case for the feature.
- We can't get component parameters from proc, eg: `SOF_TKN_VOLUME_RAMP_STEP_TYPE 
 SOF_TKN_VOLUME_RAMP_STEP_MS`

xiulipan:
> > Agree that if we can use the /proc info then we have all the information needed for tests with PCM devices.
> 
> PR #430 has the following code:
> 
> ```python
> for pcm in card_info['pcm']:
>                 # There are limited pipeline parameters in the proc file system,
>                 # add some default parameters to make use of sof-test for legacy HDA test
[...]

PR #755, opened 2021-08-23
Rewrite topology tools with Construct to improve the readability and maintainability

YongAnLu00:
I'd like to improve the topology parser to get more info like the widget core, but the legacy code contains too many magic number which is hard to understand and maintain. Here I will rewrite the topology tools with [Construct](https://construct.readthedocs.io) which is a powerful declarative and symmetrical parser and builder for binary data to make the code more maintainable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@
 ## Environment set up
 ### requirements
 #### apt packages
-expect alsa-utils python3 python3-graphviz
+expect alsa-utils python3 python3-construct python3-graphviz
 ```
-sudo apt install expect alsa-utils python3 python3-graphviz
+sudo apt install expect alsa-utils python3 python3-construct python3-graphviz
 ```
 #### user group
 sudo adm audio
@@ -144,8 +144,11 @@ apl
 * sof-tplgreader.py
 <br> tplgtool.py wrapper, it reads info from tplgtool.py to analyze topologies.
 
-* tplgtool.py
+* tplgtool2.py
 <br> Dumps info from tplg binary file.
+
+* tplgtool.py
+<br> Dumps info from tplg binary file. (deprecated)
 
 
 ## System configuration tips
--- a/env-check.sh
+++ b/env-check.sh
@@ -52,6 +52,7 @@ func_check_pkg aplay
 func_check_pkg python3
 # jq is command-line json parser
 func_check_pkg jq
+func_check_python_pkg construct
 func_check_python_pkg graphviz
 func_check_python_pkg numpy
 func_check_python_pkg scipy
--- a/python-deps.txt
+++ b/python-deps.txt
@@ -1,3 +1,5 @@
 python3-numpy
 python3-scipy
 pylint
+python3-graphviz
+python3-construct=2.8.*
--- a/tools/tplgtool.py
+++ b/tools/tplgtool.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import warnings
 import os
 import sys
 import argparse
@@ -856,6 +857,7 @@ class TplgFormatter:
         return interweaved_dict
 
 if __name__ == "__main__":
+    warnings.warn("tplgtool.py is deprecated, use tplgtool2.py instead.", DeprecationWarning)
 
     def parse_cmdline():
         parser = argparse.ArgumentParser(add_help=True, formatter_class=argparse.RawTextHelpFormatter,
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -1,0 +1,910 @@
+#!/usr/bin/env python3
+
+import argparse
+import construct
+import enum
+import os
+import re
+import sys
+import typing
+from collections import defaultdict
+from construct import this, Container, ListContainer, Struct, Const, Switch, Array, Bytes, GreedyRange, FocusedSeq, Pass, Padded, Padding, String, Flag, Byte, Int16ul, Int32ul, Int64ul, Terminated
+from dataclasses import dataclass
+from functools import partial
+
+def __enum_dict(enumtype) -> dict:
+    return dict((e.name, e.value) for e in enumtype)
+
+# here is a workaround to make Enum/FlagsEnum accept enum type, like construct 2.9
+def Enum(subcon, enumtype):
+    return construct.Enum(subcon, default=Pass, **__enum_dict(enumtype))
+
+def FlagsEnum(subcon, enumtype):
+    return construct.FlagsEnum(subcon, **__enum_dict(enumtype))
+
+def get_flags(flagsenum: Container) -> "list[str]":
+    "Get flags for FlagsEnum container."
+    return [name for (name, value) in flagsenum.items() if value is True and not name.startswith("_")]
+
+class TplgType(enum.IntEnum):
+    r"""File and Block header data types.
+
+    `SND_SOC_TPLG_TYPE_`
+    """
+    MIXER = 1
+    BYTES = 2
+    ENUM = 3
+    DAPM_GRAPH = 4
+    DAPM_WIDGET = 5
+    DAI_LINK = 6
+    PCM = 7
+    MANIFEST = 8
+    CODEC_LINK = 9
+    BACKEND_LINK = 10
+    PDATA = 11
+    DAI = 12
+
+class DaiFormat(enum.IntEnum):
+    r"""DAI physical PCM data formats.
+
+    `SND_SOC_DAI_FORMAT_`
+    """
+    I2S = 1
+    RIGHT_J = 2
+    LEFT_J = 3
+    DSP_A = 4
+    DSP_B = 5
+    AC97 = 6
+    PDM = 7
+
+class DaiLnkFlag(enum.IntFlag):
+    r"""DAI link flags.
+
+    `SND_SOC_TPLG_LNK_FLGBIT_`
+    """
+    SYMMETRIC_RATES = 0b0001
+    SYMMETRIC_CHANNELS = 0b0010
+    SYMMETRIC_SAMPLEBITS = 0b0100
+    VOICE_WAKEUP = 0b1000
+
+class DaiMClk(enum.IntEnum):
+    r"""DAI mclk_direction.
+
+    `SND_SOC_TPLG_MCLK_`
+    """
+    CO = 0
+    CI = 1
+
+class DaiBClk(enum.IntEnum):
+    r"""DAI topology BCLK parameter.
+
+    `SND_SOC_TPLG_BCLK_`
+    """
+    CP = 0
+    CC = 1
+
+class DaiFSync(enum.IntEnum):
+    r"""DAI topology FSYNC parameter.
+
+    `SND_SOC_TPLG_FSYNC_`
+    """
+    CP = 0
+    CC = 1
+
+class DaiClockGate(enum.IntEnum):
+    r"""DAI clock gating.
+
+    `SND_SOC_TPLG_DAI_CLK_GATE_`
+    """
+    UNDEFINED = 0
+    GATED = 1
+    CONT = 2
+
+class CtlTlvt(enum.IntEnum):
+    "`SNDRV_CTL_TLVT_`"
+    CONTAINER = 0
+    DB_SCALE = 1
+    DB_LINEAR = 2
+    DB_RANGE = 3
+    DB_MINMAX = 4
+    DB_MINMAX_MUTE = 5
+
+    CHMAP_FIXED = 0x101
+    CHMAP_VAR = 0x102
+    CHMAP_PAIRED = 0x103
+
+class DapmType(enum.IntEnum):
+    "`SND_SOC_TPLG_DAPM_`"
+    INPUT = 0
+    OUTPUT = 1
+    MUX = 2
+    MIXER = 3
+    PGA = 4
+    DRV = 5
+    ADC = 6
+    DAC = 7
+    SWITCH = 8
+    PRE = 9
+    POST = 10
+    AIF_IN = 11
+    AIF_OUT = 12
+    DAI_IN = 13
+    DAI_OUT = 14
+    DAI_LINK = 15
+    BUFFER = 16
+    HEDULER = 17
+    EFFECT = 18
+    SIGGEN = 19
+    SRC = 20
+    ASRC = 21
+    ENCODER = 22
+    DECODER = 23
+
+class PcmFormatsFlag(enum.IntFlag):
+    r"""PCM sample formats.
+
+    `SND_PCM_FMTBIT_LINEAR` and `SND_PCM_FMTBIT_FLOAT`
+    """
+    S8 = 1 << 0
+    U8 = 1 << 1
+    S16_LE = 1 << 2
+    S16_BE = 1 << 3
+    U16_LE = 1 << 4
+    U16_BE = 1 << 5
+    S24_LE = 1 << 6
+    S24_BE = 1 << 7
+    U24_LE = 1 << 8
+    U24_BE = 1 << 9
+    S32_LE = 1 << 10
+    S32_BE = 1 << 11
+    U32_LE = 1 << 12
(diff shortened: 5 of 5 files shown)


plbossart:
@marc-hb @mengdonglin are you good with the added dependency on a package?

marc-hb:
I agree one should never re-invent the wheel and implement a binary parser without relying on some library like Construct. On the other hand:
- Can you explain why Construct is the best choice as opposed to for instance Kaitai? As noted by @plbossart , tying our fate to Construct's is not a decision that should be made lightly. How popular and actively maintained is Construct?
- Issue #471 says we should do _less_ topology parsing!?
- Why do we maintain a parser in sof-test? Doesn't that belong to https://github.com/alsa-project/alsa-utils ?

ranj063:
> * Why do we maintain a parser in sof-test? Doesn't that belong to https://github.com/alsa-project/alsa-utils ?

There's a decoder in alsa-utils (-d switch) IIRC. We should look into using it instead of depending on an external package

marc-hb:
> There's a decoder in alsa-utils (-d switch with alsatplg IIRC).

Is it in C? This PR proposes to "upgrade" our manual Python implementation to a higher level approach in order to reduce code size and maintenance. Downgrading to a low-level, memory management/corruption language would go the opposite direction. Python is slow but I don't think parsing topologies is in any critical path for us.

Is alsa-utils a C-only project? If yes then that would justify keeping an implementation in a higher level language here for the lack of a better place.

YongAnLu00:
Hi @plbossart, I'm trying to use pip to install the dependencies in GitHub Actions because apt seems to install the packages for Python whose path is different than [setup-python](https://github.com/marketplace/actions/setup-python) use. You can see my modification in <https://github.com/thesofproject/sof-test/pull/755/commits/abf226de3a860aca44fdb3907a91d6ac302cf217>.

marc-hb:
`pip` versus `apt` is an important question but it's a premature question. A much more important question is: "Does SOF validation goes down when Construct goes down?"

marc-hb:
> but the legacy code contains too many magic number which is hard to understand and maintain. 

Naming magic numbers does not require a new library. BTW I see the existing parser already uses https://docs.python.org/3/library/struct.html extensively (which will never "go down")

> Here I will rewrite the topology tools with Construct which is a powerful declarative and symmetrical parser and builder for binary data to make the code more maintainable.

How much more maintainable? Can you please share some expected numbers like lines of code and examples of repetitive code not needed any more, number of classes and functions, cyclomatic complexity, potential test coverage and any other metric of your choice. No precise number, just expected orders of magnitude (if there's no order of magnitude difference then it's probably not worth it)

YongAnLu00:
Hi @marc-hb, thanks for your feedback, and sorry for my late reply because I spent some time collecting information.

> * Can you explain why Construct is the best choice as opposed to for instance Kaitai? As noted by @plbossart, tying our fate to Construct's is not a decision that should be made lightly. How popular and actively maintained is Construct?

I think Kaitai could be a good choice if we need to reuse the parser in different languages. But what we need here is just a topology parser in Python and Kaitai have more heavy dependency and complex workflow (we describe the binary format in a `ksy` file and need the compiler to translate Kaitai Struct language to Python). Construct works like an embedded DSL, so we write everything in Python and we can just run `pip install construct` for the requirement. I only find Construct as a stable, maintained actively, self-contained, declarative binary data parser in PyPI, so I believe this is the best choice. I think Construct is the most popular declarative binary data parser in Python ([libraries.io](https://libraries.io/pypi/construct/dependents)) and it's still actively maintained this year.

> * Issue #471 says we should do _less_ topology parsing!?
> * Why do we maintain a parser in sof-test? Doesn't that belong to https://github.com/alsa-project/alsa-utils ?

Sorry for my misleading words, what we actually make is a topology tool for testing and debuging, and parsing topology is just the first step. With this tool, we can not only dump some topology information but also render a topology graph. As https://github.com/thesofproject/sof-test/issues/471#issuecomment-767340442 said, some parameters can only be extracted from topology, and what I want to do is just a replacement for the origin topology tool. Moreover, writing the parser ourselves allow us to parse some private data like schedule core, so it's a more flexible way than an official tool. Also, `alsatplg -d` can decode a binary topology file to the configuration file, but it just moves the problem from parsing a binary file to parsing a configuration file.

marc-hb:
Thanks for the detailed explanations, very useful and convincing.

> but it just moves the problem from parsing a binary file to parsing a configuration file.

OK but parsing text is much lower maintenance than binaries and there is probably a wider choice of libraries to help with that.

YongAnLu00:
> How much more maintainable? Can you please share some expected numbers like lines of code and examples of repetitive code not needed any more, number of classes and functions, cyclomatic complexity, ease of unit testing, error message clarity and any other metric of your choice. No precise number, just expected orders of magnitude (if there's no order of magnitude difference then it's probably not worth it)

I'm not familiar with those metrics, so I'd like to give an example in an intuitive way:

For parsing PCM struct, the legacy code: 
<https://github.com/thesofproject/sof-test/blob/5bb66798f8fa0a5cea44d959070eda9c19545226/tools/tplgtool.py#L335-L379> 
it contains many magic numbers as field offsets. If there is any change, we need to reevaluate those offsets and hope we won't miscalculate them in some steps.

My new code with Construct:
<https://github.com/thesofproject/sof-test/blob/5d11bc754ba428edc467086f1b9a6c831bdceccf/tools/tplgtool-ng/tplgtool.py#L313-L328>
it is simply corresponding to [`struct snd_soc_tplg_pcm` in alsa-lib](https://github.com/thesofproject/alsa-lib/blob/706192341d1d0bbb906d690b227b9dee5c1fc4b5/include/sound/uapi/asoc.h#L516-L531). If something changed, just keep the mapping rule and everything will work fine.

> OK but parsing text is much lower maintenance than binaries and there is probably a wider choice of libraries to help with that.
Well, I don't think parsing binary is **much** higher maintenance than text if we accept Construct. And I find the decoded configuration file seems lost some information or may need another step to retrieve them. So I think it could be a choice but I do not appreciate it.

marc-hb:
Thanks for the example. I don't think it's a very fair example though: the reason why there are hardcoded offsets everywhere right now is not really because of the difference between `struct` and `construct` but because the author didn't really understand how to use `struct` and used extremely short formats, often just one character long!

However I agree that construct code looks more readable in general. Also, if someone (you) makes the effort to fix the currently pretty bad code then he might as well go the extra mile and switch to a different and more readable solution.

So +1 from me for the Construct dependency. It looks like a good project. (I briefly looked at but haven't reviewed the code in this PR)

.github/workflows/pull_request.yml
```diff
@@ -36,14 +36,13 @@ jobs:
       - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
       - uses: actions/setup-python@v2
 
-      # Package index is "out of date by design"
-      # https://github.com/actions/virtual-environments/issues/1757
-      - name: apt-get -y update
-        run: sudo apt-get -y update
+      # Use pip to install packages because apt seems to install
+      # the packages for Python whose path is different than setup-python use.
```
marc-hb, line 40: Please be more specific, what exact error do you get with `apt` packages?

YongAnLu00: <https://github.com/thesofproject/sof-test/blob/40b7f17a42be55d3480e209b13d5a3577ab34a4d/.github/workflows/pull_request.yml#L41-L42>
Is this OK?

marc-hb: "Python import error" is much more specific than "path is different", now I understand thank you.

.github/workflows/pull_request.yml
```diff
@@ -36,14 +36,15 @@ jobs:
...
+      # Pylint will report import errors with apt packages
+      # since apt packages can't be found by current Python.
+      - name: update pip
+        run: python -m pip install --upgrade pip
 
       - name: get python libs
-        # FIXME: apt-get succeeds but 'import numpy' still fails!?
-        run: sudo apt-get -y install python3-numpy python3-scipy pylint
+        run: pip install numpy scipy pylint construct graphviz
```
marc-hb, line 47: Please leave the following line commented out in case someone prefers to use `apt` packages and help Ubuntu fix them. These package names can be difficult to find, also helps other distributions that often use the same names.
```
# sudo apt-get -y install python3-numpy python3-scipy pylint
```

.github/workflows/pull_request.yml
marc-hb: Timely: https://lwn.net/Articles/867657/ Cooperative package management for Python

YongAnLu00: I haven't subscribed to LWN.net, so I don't know what the article said. But I think maybe my words still mislead you. The actual problem is: `apt` always installs packages for Python in the system path (typically `/usr/bin`), while [setup-python](https://github.com/marketplace/actions/setup-python) setup Python in the other location (for us, this is `/opt/hostedtoolcache/Python/3.9.6/x64`, as you can see in the logs), so the packages installed by `apt` can't be found by Python we use. But `pip`  is distributed with Python, so `pip` we use is for Python we use.

.github/workflows/pull_request.yml
YongAnLu00: Well, should I include the all the packages? like:
```yaml
# sudo apt-get -y install python3-numpy python3-scipy pylint python3-construct python3-graphviz
```

.github/workflows/pull_request.yml
marc-hb: "We" don't install anything  `/opt/hostedtoolcache`, github does. That's not supposed to be a visible change. As the previous comment says, in the past it was only a problem for `apt-get install python3-numpy; import numpy` but it was not a problem for the others like `apt-get install python3-scipy; import scipy`. So this looked like a numpy-specific bug, either from Ubuntu or from github.

As this is a workaround for that bug, it would be great to give the exact list of the failing imports. Or even better: report this issue to Ubuntu or github and link to the bug.

YongAnLu00: I tried to perform a simple test: <https://github.com/thesofproject/sof-test/pull/755/checks?check_run_id=3481619127>.
As you can see in the log, `import scipy` also failed with `apt-get`.

As my previous comments said, **all** the packages installed by `apt` can't be imported by Python located in `/opt/hostedtoolcache`.

.github/workflows/pull_request.yml
YongAnLu00: Should we just put the dependencies info in [README.md](https://github.com/YongAnLu00/sof-test/blob/tplgtool-ng/README.md#apt-packages)? I think only a few people will check GitHub Actions to find dependencies.

.github/workflows/pull_request.yml
marc-hb: I just tested that again and now both numpi and scipy imported just fine:

https://github.com/thesofproject/sof-test/pull/759/checks?check_run_id=3487197768

Maybe github fixed these issues.

.github/workflows/pull_request.yml
marc-hb: If someone tries to report an error reported by CI without looking at the exact commands run by CI then they have a big problem. The commands can often be found in the logs but not always, so in general it's always recommended to look at the actual commands instead of guessing.

On the other hand it shouldn't be required to look at github actions to set up a system (before even running any test).

The ideal is to always have as little code as possible in .github/. Here for instance: `pip3 install -r pip-requirements.txt` or `apt-get install $(cat python-deps.txt)` is ideal. If apt is not possible and pip must be used (not sure about that) then the list of apt packages can be a comment in the `requirements.txt` file.

marc-hb (approved):

.github/workflows/pull_request.yml
YongAnLu00: Well, I find the difference between our tests: in your test, everything is installed via `apt`, including pylint, so it works fine; in my test, the packages are installed via `apt`, but use Python located in `/opt/hostedtoolcache`, so the packages can't be found. If we just use `apt`, we even don't need [setup-python](https://github.com/marketplace/actions/setup-python), just keep the Python preinstalled in Ubuntu.

marc-hb: Thanks for solving this mystery! I didn't pay attention to https://github.com/actions/setup-python, my bad. It looks powerful and flexible but more complicated, simpler if we can avoid it.

YongAnLu00:
All of the pylint warnings for sof-tplgreader.py aren't introduced by me. I'm focusing on tplgtool.py, so I only make minimal changes to sof-tplgreader.py for compatibility.

.github/workflows/pull_request.yml
```diff
@@ -34,16 +34,15 @@ jobs:
     steps:
       # :-( https://github.community/t/support-for-yaml-anchors/16128
       - {uses: actions/checkout@v2, with: {fetch-depth: 0}}
-      - uses: actions/setup-python@v2
+      - run: python --version
```
marc-hb, line 37: Thanks for solving that mystery. Could you submit this (and maybe others?) separately so we expedite it?

.github/workflows/pull_request.yml
```diff
@@ -34,16 +34,15 @@ jobs:
...
 
       # Package index is "out of date by design"
       # https://github.com/actions/virtual-environments/issues/1757
       - name: apt-get -y update
         run: sudo apt-get -y update
 
       - name: get python libs
-        # FIXME: apt-get succeeds but 'import numpy' still fails!?
-        run: sudo apt-get -y install python3-numpy python3-scipy pylint
+        run: sudo apt-get -y install $(cat python-deps.txt)
```
marc-hb, line 45: Same here, let's merge this before even looking at the rest and get them out of the way.

marc-hb (changes requested):
Thanks, the new code looks much more readable! Great effort. I'm surprised that it is not that much shorter but that's probably because of all the pydocs (thanks!) and also a lot more whitespace (now I'm a bit curious about the characters count)

My main issue is with the almost complete rewrite in place and "Big Bang" migration. Instead of replacing the existing files, can you please create two new files `sof-tplgreader2.py` and `tplgtool2.py`? Or any other better name you can think of. This will provide a smooth transition period where anyone or anything stuck with a bug or small difference in the new tool can report the bug or incompatibility and then very quickly, locally and temporarily revert back to the old one to unblock themselves.

I checked and there are fewer than 15 references to the old names in sof-test. They don't need to be switched all at the same time: _Continuous_ Integration.

The old scripts can be deleted later. This will break any references outside sof-test that we failed to spot but by that time the new scripts will have been well tested and all new bugs and most incompatibilities squashed.

To not lose the git history of `sof-tplgreader.py` you can have a very first commit that merely copies to `sof-tplgreader2.py` without changing it yet, this helps git follow the history.

.github/workflows/pull_request.yml
YongAnLu00: I have cherry-picked these commits to <https://github.com/thesofproject/sof-test/pull/765>.

marc-hb: Thanks! Approved.

Here's a git trick I use all the time:
```shell
git push my_github_fork HEAD~2:refs/heads/actions-python
```
The trick is `HEAD~2`, so you don't need to switch branch or cherry pick anything.

YongAnLu00:
@marc-hb Creating new files for smooth migration is a good idea. Because I'm focusing on new tplgtool in this PR and @aiChaoSONG encourage me to rewrite sof-tplgreader.py too in the future, I think we can leave sof-tplgreader.py unchanged this time and postpone the migration for sof-tplgreader.py until rewriting it. For a smooth transition, we can start to draw topology graphs with the new tplgtool2.py in our CI.

aiChaoSONG (dismissed):
Minor changes or explanations required, let's use duplex instead of `both`

aiChaoSONG (approved):

aiChaoSONG:
Thanks, @YongAnLu00 for the tool, let's merge the tool, let's do bug fixes and features in further PR

Issue #740, opened 2021-07-14
clean-up: move all shell scripts code to a function


marc-hb:
This a very generic, clean-up task. Its main purpose is to document this good practice: almost all the code in a shell script should be in a function. In other words, the smallest script should look like this:
```shell
main()
{
  do suff
}

main "$@"
```

Rationales:

- More variables can be `local`. This reduces the risk of accidentally overwriting some user environment variables (side effects)
- Functions can be defined in any order, they don't need to be defined before being used
- It provides more line numbers in the FUNCNAME stack on failure 
- moving code around functions does not change the indentation level which helps with git diff, rebase, cherry-pick, merge...
- it's very easy to comment out the last, `main "$@"` line and `source` the entire file without running it to test individual functions interactively

marc-hb:
Simon Tatham agrees :-)
https://hachyderm.io/@simontatham/114511220670677410

marc-hb:
https://google.github.io/styleguide/shellguide.html#function-location

>  Don’t hide executable code between functions. Doing so makes the code difficult to follow and results in nasty surprises when debugging.

PR #758, opened 2021-08-31
Optimize this case to support all the PCM

1994lwz:
Rename the case check-pause-suspend-resume-release.sh to check-pause-suspend-resume-release.sh.

And optimize this case to support all the PCM.

Signed-off-by: VickyX_Intel_Ubuntu <vickyx.li@intel.com>


fredoh9 (changes requested):
commit text has typos. Need more information about this change

test-case/check-pause-suspend-resume-release.sh
```diff
@@ -146,7 +145,7 @@ expect {
         }
 
         #enter suspend-resume cycle once per pause instance
-        set retval [catch { exec bash check-suspend-resume.sh -l 1 } msg]
+        set retval [catch { exec bash /home/ubuntu/sof-test/test-case/check-suspend-resume.sh -l 1 } msg]
```
fredoh9, line 148: Instead of hard coded path, can we use BASH_SOURCE[0] in expect?

fredoh9 (changes requested):

fredoh9:
Since you are touching this file, some of shellcheck errors/warnings are easy to fix.

test-case/check-pause-suspend-resume-release.sh
marc-hb: Agreed, `/home/ubuntu/` is not going to work for most people. It worked with a relative path before, what changed?

test-case/check-pause-suspend-resume-release.sh
```diff
@@ -107,29 +99,36 @@ esac
...
-# expect is tcl language script
-#   catch: Evaluate script and trap exceptional returns
-#   after ms: Ms must be an integer giving a time in milliseconds.
-#       The command sleeps for ms milliseconds and then returns.
-expect <<AUDIO
-spawn $cmd -D $pcm -r $rate -c $channel -f $fmt -vv -i $dummy_file -q
+    # expect is tcl language script
+    #   catch: Evaluate script and trap exceptional returns
+    #   after ms: Ms must be an integer giving a time in milliseconds.
+    #       The command sleeps for ms milliseconds and then returns.
```
marc-hb, line 122: If you change the indentation then take that opportunity to use more shell functions, see why in #740 . One of the function should be something like `test_single_pcm()`

marc-hb (changes requested):
> Rename the case check-pause-suspend-resume-release.sh to check-pause-suspend-resume-release.sh.

(same name)

File renames are very painful because they can endless version control issues. Is this rename really necessary? I see no justification for it in the commit message.

If the rename is absolutely necessary, do the rename in a separate commit that only renames the file, this helps git identify the rename and reduces version control issues a little bit. 

> Optimize this case to support all the PCM #758

This is not an optimization. An optimization does the same thing faster or better in some way. This is adding a new feature: iterating no all PCMs (the same way as before).

test-case/check-pause-suspend-resume-release.sh
1994lwz: I will change the code and use the BASH_SOURCE[0] to get the dirname.

test-case/check-pause-suspend-resume-release.sh
```diff
@@ -107,29 +99,36 @@ esac
...
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    snd=$(func_pipeline_parse_value "$idx" snd)
 
-dlogi "Entering audio stream expect script with: $cmd -D $pcm -r $rate -c $channel -f $fmt -vv -i $dummy_file -q"
-dlogi "Will enter suspend-resume cycle during paused period of audio stream process"
+    dlogi "Entering audio stream expect script with: $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q"
+    dlogi "Will enter suspend-resume cycle during paused period of audio stream process"
 
-rm -rf /tmp/sof-test.lock
+    rm -rf /tmp/sof-test.lock
```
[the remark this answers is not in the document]
1994lwz: After verification, this code is useless, I will delete it.

test-case/check-pause-suspend-resume-release.sh
```diff
@@ -172,13 +171,14 @@ expect {
...
-#flush the output
-echo
-if [ $ret -ne 0 ]; then
-    sof-process-kill.sh
-    [[ $? -ne 0 ]] && dlogw "Kill process catch error"
-    exit $ret
-fi
-sof-kernel-log-check.sh "$KERNEL_CHECKPOINT"
-exit $?
+    ret=$?
```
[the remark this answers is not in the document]
1994lwz: About the rename, because the original name is check-pause-release-suspend-resume.sh, it does not represent the test that our case will be done, so I changed it to check-pause-suspend-resume-release.sh. I will raise a PR to change the case name.

test-case/check-pause-suspend-resume-release.sh
```diff
@@ -172,13 +171,14 @@ expect {
...
+    #flush the output
+    echo
+    if [ $ret -ne 0 ]; then
+        sof-process-kill.sh
+        [[ $? -ne 0 ]] && dlogw "Kill process catch error"
+        exit $ret
+    fi
+    sof-kernel-log-check.sh "$KERNEL_CHECKPOINT" || die "Caught error in kernel log"
+    #exit $?
```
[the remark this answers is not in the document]
1994lwz: delete the #exit $?

marc-hb (changes requested):
This PR is completely broken now, abandon this PR and start a new one.

The first commit still has a rename and code change in a single commit.
The second commit has a bit of code but does NOT add the for loop contrary to its commit message
The third commit does a second rename.

This is actually a good demonstration that renames always make version control much harder. A rename will also break any other script or CI engine that relies on the old name. I bet the new name is better but it's not so much better that it's worth these hassles, so forget about the rename for now.

Please also fix your `git config` to have your real name in the Signed-off field.

Screenshot 2021-09-01 at 13 14 04

PR #760, opened 2021-09-02
Add a new feature: let the case can automatic testing of all PCMs

1994lwz:
Add a new feature: let the case can automatic testing of all PCMs which named check-pause-release-suspend-resume.sh

Signed-off-by: Wanzhuo.Li <wanzhuox.li@intel.com>


test-case/check-pause-release-suspend-resume.sh
```diff
@@ -107,29 +99,34 @@ esac
...
-spawn $cmd -D $pcm -r $rate -c $channel -f $fmt -vv -i $dummy_file -q
+func_pipeline_export "$tplg" "type:$test_mode & ${OPT_VAL['p']}"
+
+for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+do
+    channel=$(func_pipeline_parse_value "$idx" channel)
+    rate=$(func_pipeline_parse_value "$idx" rate)
+    fmt=$(func_pipeline_parse_value "$idx" fmt)
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    snd=$(func_pipeline_parse_value "$idx" snd)
```
marc-hb, line 112: Before the user could pass these on the command line. Not anymore, isn't that also removing a feature? Please explain in the commit message.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -146,7 +143,7 @@ expect {
         }
 
         #enter suspend-resume cycle once per pause instance
-        set retval [catch { exec bash check-suspend-resume.sh -l 1 } msg]
+        set retval [catch { exec bash $(dirname "${BASH_SOURCE[0]}")/check-suspend-resume.sh -l 1 } msg]
```
marc-hb, line 146: Please define a MYDIR or TOPDIR or TESTDIR variable at the top and do the `dirname` dance only once at the top. See examples in other scripts.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +66,22 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
...
 
 OPT_NAME['i']='sleep-period' OPT_DESC['i']='sleep period of aplay, unit is ms'
-OPT_HAS_ARG['i']=1             OPT_VAL['i']='100'
+OPT_HAS_ARG['i']=1             OPT_VAL['i']='200'
 
 OPT_NAME['s']='sof-logger'   OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
+OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
```
marc-hb, line 75: Please explain in the commit message why this new parameter is needed.

marc-hb (changes requested):
The commit message body is empty and the subject line is too long.

Please follow:
https://wiki.openstack.org/wiki/GitCommitMessages#Summary_of_Git_commit_message_structure
... and most other advice on that page.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -52,13 +52,14 @@ set -e
...
+CASEDIR=$(dirname "${BASH_SOURCE[0]}")
+source $CASEDIR/../case-lib/lib.sh
 
 OPT_NAME['m']='mode'         OPT_DESC['m']='test mode. Example: playback; capture'
 OPT_HAS_ARG['m']=1             OPT_VAL['m']='playback'
 
-OPT_NAME['p']='pcm'          OPT_DESC['p']='run test case on specified pipelines'
-OPT_HAS_ARG['p']=1             OPT_VAL['p']='id:any'
+OPT_NAME['S']='pcm'          OPT_DESC['S']='run test case on specified pipelines'
+OPT_HAS_ARG['S']=1             OPT_VAL['S']='id:any'
```
marc-hb, line 62: You cannot keep the same help message for a new option that works very differently! Maybe: "run test case on pipelines matching the filter argument"

marc-hb (changes requested):
The split into 3 commits is good, much more readable and maintainable, thanks!

One one hand you're removing the need to specify compatible parameters which looks good. On the other hand you're also removing the flexibility to choose any parameter in a supported range. I guess this is probably OK for this test but removing a feature should be at least mentioned in the commit messages.

The test case description in the help also needs to be updated to match the changes.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -51,22 +52,14 @@ set -e
 ##  * No unexpected errors should be present in dmesg during or after test
 ##      completion.
 
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+CASEDIR=$(dirname "${BASH_SOURCE[0]}")
+source $CASEDIR/../case-lib/lib.sh
```
marc-hb, line 56: Fix the quoting issues reported by shellcheck. Not all the issues in this script but at least all the ones in new or modified code. Don't add new shellcheck warnings.

marc-hb:
```shell
# shellcheck source=...
```
See how other scripts do it.

mengdonglin (approved):

mengdonglin:
Reviewed the commit messages

test-case/check-pause-release-suspend-resume.sh
YongAnLu00: Maybe you can refer to check-pause-resume.sh to keep your options consistent with it.

test-case/check-pause-release-suspend-resume.sh
1994lwz: I don't remove a feature, now all these parameters can be parsed from the specified or default tplg file automatically. I explained in the commit message "This patch allows user to specify an input tplg file. Then the test case will call func_pipeline_parse_value to automatically obtain pipeline parameters from the tplg file.

User no longer needs to use parameter to specify each pipeline paremeters."

test-case/check-pause-release-suspend-resume.sh
1994lwz: @marc-hb The -S is used to specify the same PCM ID parameter as -p, so I copy the the help info. I change the option from -p to -S to align with case check-pause-resume.sh that uses -S to specify PCM ID.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +70,21 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
 OPT_HAS_ARG['l']=1             OPT_VAL['l']=5
 
 OPT_NAME['i']='sleep-period' OPT_DESC['i']='sleep period of aplay, unit is ms'
-OPT_HAS_ARG['i']=1             OPT_VAL['i']='100'
+OPT_HAS_ARG['i']=1             OPT_VAL['i']='200'
```
YongAnLu00, line 73: I've just tested this in my TGLU_SKU0A32_SDCA device and found that 200ms seems still too short for my device while 300ms will work fine. Maybe we need more investigation to determine a proper period, or just left it unchanged and specify a proper period for each device in our CI?

YongAnLu00 (changes requested):

test-case/check-pause-release-suspend-resume.sh
marc-hb: OK the user can edit the topology file if needed, I did not consider that. Thanks for the explanation.

test-case/check-pause-release-suspend-resume.sh
marc-hb: How about something much longer but still quick like 1s?

marc-hb (changes requested):
It can be painful to transfer code from one commit to another but sorry, each commit must be standalone, usable and run better than all the previous ones; no regression. You also shouldn't add code with warnings in one commit and then fix the warnings in a later commit. Amend the earlier commit and never have warnings in any commit.

Looking at intermediate commit 2b9ab9e59815 right now I see new shell script code in the middle of TCL code which almost certainly shows that it does not work without a  later commit that fixes it. That sort of git history is not acceptable, sorry.

To fix git history make sure you're very familiar with interactive rebase 
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
https://www.thinktecture.com/en/tools/git-interactive-rebase/ 
https://stackoverflow.com/questions/15213814/how-should-i-move-changes-from-one-commit-to-another
(first guides I found, there are probably better ones).
I know there are github projects [that don't care about git history](https://github.com/isaacs/github/issues/959#issuecomment-592978474) and not require any interactive rebase/rewrite skill, we do require in this project that every commit works and is better than the previous ones for many maintenance reasons: bisecting, branching, cherry-picking, reverting etc. This is also required by many other projects so this is a very useful skill to have.

Also make sure you use the git plugin in your editor, many editors have git plugins that make git much faster to use (fugitive, magit, etc.). If your editor does not have a good git plugin you should consider a different editor

marc-hb:
> It can be painful to transfer code from one commit to another 

Another thing to make life easier is to expedite small, unrelated fixes easier to review in a different, parallel Pull Request to get them merged faster and quickly out of the way. For this and other reasons I rarely submit HEAD into a pull request, almost always HEAD~n.

YongAnLu00 (changes requested):
I found that we have to force the suspend/resume type for the GLK platform. For check-suspend-resume.sh, it is controlled by `-T` option, so you should add the same option and forward its value to check-suspend-resume.sh.

<https://github.com/thesofproject/sof-test/blob/0bb4b66656416a249af50ce64aba49ad7c42db1c/test-case/check-suspend-resume.sh#L30-L31>

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -145,8 +150,15 @@ expect {
             exit 1
         }
 
+        #get the type of platform
+        set retval [catch { exec uname -a | grep -o glk } platform]
+
```
YongAnLu00, line 155: Get platform from `uname` is a bad, unreliable workaround. What you see in our CI device is just from its hostname following our naming convention, so it's unportable and unreliable.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +72,21 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
...
 OPT_NAME['i']='sleep-period' OPT_DESC['i']='sleep period of aplay, unit is ms'
-OPT_HAS_ARG['i']=1             OPT_VAL['i']='100'
+OPT_HAS_ARG['i']=1             OPT_VAL['i']='500'
 
 OPT_NAME['s']='sof-logger'   OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
+OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg path, default is environment variable: TPLG'
+OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
+
```
YongAnLu00, line 82: First, add `-T` option:
```suggestion

 OPT_NAME['T']='type'    OPT_DESC['T']="suspend/resume type from /sys/power/mem_sleep" 
 OPT_HAS_ARG['T']=1         OPT_VAL['T']="" 
```

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +72,21 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
...
 
-pcm=${OPT_VAL['p']}
-fmt=${OPT_VAL['f']}
-channel=${OPT_VAL['c']}
-rate=${OPT_VAL['r']}
 repeat_count=${OPT_VAL['l']}
 sleep_period=${OPT_VAL['i']}
 test_mode=${OPT_VAL['m']}
 file_name=${OPT_VAL['F']}
+tplg=${OPT_VAL['t']}
```
YongAnLu00, line 90: Then, get its value for forwarding:
```suggestion
type=${OPT_VAL['T']}

```

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -145,8 +150,15 @@ expect {
...
+        #get the type of platform
+        set retval [catch { exec uname -a | grep -o glk } platform]
+
         #enter suspend-resume cycle once per pause instance
-        set retval [catch { exec bash check-suspend-resume.sh -l 1 } msg]
+        if {[string equal \$platform "glk"] != 0} {
+            set retval [catch { exec bash $SUSPEND_RESUME -T s2idle} msg]
+        } else {
+            set retval [catch { exec bash $SUSPEND_RESUME} msg]
+        }
```
YongAnLu00, line 161: Finally, forward it to check-suspend-resume:
```suggestion
        set retval [catch { exec bash $CASEDIR/check-suspend-resume.sh -l 1 -T $type } msg]
```

YongAnLu00 (changes requested):
You really misunderstand my words. I mean we should add an option and forward it instead of a hard-coded, unreliable workaround for GLK. Here I give an example in my review comments.

YongAnLu00 (approved):
I've just triggered a test (6539) for this on all models in the daily test, and all results are green.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +71,30 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
...
-fmt=${OPT_VAL['f']}
-channel=${OPT_VAL['c']}
-rate=${OPT_VAL['r']}
 repeat_count=${OPT_VAL['l']}
 sleep_period=${OPT_VAL['i']}
 test_mode=${OPT_VAL['m']}
 file_name=${OPT_VAL['F']}
+tplg=${OPT_VAL['t']}
+
+if [ "${OPT_VAL['T']}" ]; then
```
marc-hb, line 93:
```suggestion
if [ -n "${OPT_VAL['T']}" ]; then
```
otherwise an incorrect `"${OPT_VAL['T']}"` could generate very confusing behaviors.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,21 +71,30 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
...
+    opt="-l 1 -T ${OPT_VAL['T']}"
+else
+    opt="-l 1"
```
marc-hb, line 96: Please change `opt` to something longer and clearer like for instance `suspend_opts`

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -178,8 +180,7 @@ AUDIO
     #flush the output
     echo
     if [ $ret -ne 0 ]; then
-        sof-process-kill.sh
-        [[ $? -ne 0 ]] && dlogw "Kill process catch error"
+        sof-process-kill.sh || dlogw "Kill process catch error"
```
marc-hb, line 192:
```shell
sof-process-kill.sh || dlogw "sof-process-kill.sh failed"
```

marc-hb (changes requested):
The final code changes look mostly OK but the intermediate commits are still not working. Commit https://github.com/thesofproject/sof-test/pull/760/commits/c4e5cc6c1875a67 still has shell script in the middle of the TCL expect code, so it is broken (that's fixed by next commit https://github.com/thesofproject/sof-test/pull/760/commits/d1628af60b32)

Same thing for the shellcheck warnings: don't fix them in another commit; instead don't add any new warnings at all in the first place.

You must do an interactive rebase and make sure the test can run at each commit. It does not have to PASS on every platform at every commit but it must be able to run (and run better than the previous commit). To make conflict solving easier, perform one complete interactive rebase for each piece of code to move, I mean fix a single problem each time and perform as many interactive rebases as needed.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -75,7 +75,7 @@ OPT_NAME['l']='loop'         OPT_DESC['l']='loop count'
 OPT_HAS_ARG['l']=1             OPT_VAL['l']=5
 
 OPT_NAME['i']='sleep-period' OPT_DESC['i']='sleep period of aplay, unit is ms'
-OPT_HAS_ARG['i']=1             OPT_VAL['i']='100'
+OPT_HAS_ARG['i']=1             OPT_VAL['i']='500'
```
plbossart, line 74: can I ask what platforms? This is a 5x increase, but the interval between pause_release and pause_push are rather small.

I also don't get why this is different from the regular pause-resume test.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -13,6 +13,8 @@ set -e
 ##    have system enter suspend state for 5 secs
 ##    resume from suspend state
 ##    release audio stream from paused state
+##    repeat to all pipelines
```
plbossart, line 16: the test description is not clear

line 15: is the stream always paused prior to suspend? Or can this be random/variable?
line 16: can we have multiple streams in parallel? This will be required for our tests, in fact if you see https://github.com/thesofproject/linux/issues/3151 I am precisely having issues with suspend-resume with 2 paused streams.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -74,7 +76,7 @@ OPT_HAS_ARG['i']=1             OPT_VAL['i']='500'
 OPT_NAME['s']='sof-logger'   OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
-OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg path, default is environment variable: TPLG'
```
plbossart: TPLG is the path+filename, not just the path. This doesn't seem like a valid correction.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -76,8 +76,11 @@ OPT_HAS_ARG['i']=1             OPT_VAL['i']='500'
 OPT_NAME['s']='sof-logger'   OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
-OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg path, default is environment variable: TPLG'
-OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
+OPT_NAME['t']='tplg'         OPT_DESC['t']='tplg path, default is environment variable: TPLG'
+OPT_HAS_ARG['t']=1             OPT_VAL['t']="$TPLG"
+
+OPT_NAME['T']='type'         OPT_DESC['T']="suspend/resume type from /sys/power/mem_sleep"
+OPT_HAS_ARG['T']=1             OPT_VAL['T']=""
```
plbossart, line 83: I was not able to understand what the option means. the 'from' part threw me off, did you mean instead that it would set the suspend/resume type by writing to sysfs?

plbossart (changes requested):
thanks for updating this script, see comments/suggestions below.

test-case/check-pause-release-suspend-resume.sh
marc-hb: This option is merely passed  passed-through to `test-case/check-suspend-resume.sh` and the description is merely copied from there (which does not mean it cannot be improved)

test-case/check-pause-release-suspend-resume.sh
marc-hb: Moreover this is a correction of a previous commit in this same PR. The git history is still a mess, don't waste your time reviewing each commit. In addition to having commits that can't run, history that does not bisect, reverts or cherry-picks and confused reviewers, this is also confusing github: so you can't answer some threads unless you select the right commit.

Back to the description: "path" is ambiguous, for some people path the `dirname` but for many others like https://docs.python.org/3/library/pathlib.html path = dirname + filename.

"file" is vague but at least every knows it's vague, so no one makes wrong assumptions about it. So I agree "file" is better.

TPLG supports both complete paths and filenames.

marc-hb (changes requested):

test-case/check-pause-release-suspend-resume.sh
1994lwz: @YongAnLu00 Yongan test the script on his TGLU_SKU0A32_SDCA device and found that 200ms seems still too short, and Marc suggest change the sleep time much longer but still quick like 1se

test-case/check-pause-release-suspend-resume.sh
1994lwz: In this PR, we will first implement a single pipeline test of pause-release-suspend-resume, and then we will re-prove a PR to implement multiple pipeline tests.

PR #763, opened 2021-09-10
run-all-tests.sh: print partial results when interrupted with Ctrl-C

marc-hb:
Convenient when seeing a failure. Sample output:

```
2021-09-10 22:27:37 UTC Sub-Test: [INFO] Test Result: PASS!
^C

PASS: firmware-presence; speaker; firmware-load;
FAIL: sof-logger;

 test end with 1 failed tests

Testing was INTERRUPTED, results are incomplete!
```

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -87,6 +87,9 @@ main()
 	local failures=()
 	local passed=()
 
+	# On Ctrl-C
+	trap interrupted_results INT
+
 	local tests_length=10 time_delay=3
 
 	while getopts "l:T:h" OPTION; do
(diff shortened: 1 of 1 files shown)


marc-hb:
The "real" title of this PR is `run-all-tests.sh: print partial results when interrupted with Ctrl-C` which should sound much more appealing. Sorry for the noise.

fredoh9 (approved):
I like this change. I used to scroll up to find the results when I need to do Ctrl+C in between.

marc-hb:
Thanks @fredoh9 ! Merging now because this PR has been in review for more than 2 weeks and this script is not used in any automation.

PR #769, opened 2021-09-17
hijack.sh: print FAIL on missing logfile to avoid long CI timeout

marc-hb:
For reasons beyond the scope of this commit, our Jenkins CI does not
know when a test script has finished to run. Instead of checking the
exit status like a normal test runner it waits for a "FAIL" or "PASS" or
"N/A" string to be printed. When the test aborts without printing one of
these, Jenkins falls back on a fairly long timeout - per test.

This unusual approach has been causing issues with `set -e` (#312)

In some relatively rare but possible failures the DSP or the drivers do
not load at all, examples:

https://sof-ci.01.org/sofpr/PR4782/build10387/devicetest/
https://sof-ci.01.org/sofpr/PR4766/build10353/devicetest/
https://sof-ci.01.org/sofpr/PR4768/build10348/devicetest/

In that case the $logFile is not found and many tests abort without
printing FAIL.

This commit prints FAIL when $logfile is missing and avoids the many
per-test Jenkins timeouts.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -62,16 +62,25 @@ function func_exit_handler()
             sudo pkill -KILL "$loggerBin"
             exit_status=1
         fi
-        # logfile is set in a different file: lib.sh
+
+        # $logfile is defined in a different file (lib.sh)
         # shellcheck disable=SC2154
-        wcLog=$(wc -l "$logfile")
-        dlogi "nlines=$wcLog"
-        local nlines; nlines=$(wc -l < "$logfile")
-        # Line 1 is the header
-        if [ "$nlines" -le 1 ]; then
-            dloge "Empty logger trace"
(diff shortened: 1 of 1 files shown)


marc-hb:
Only one TIMEOUT in only one test on only one platform but the logs show PASS?
https://sof-ci.01.org/softestpr/PR769/build842/devicetest/

Everything else is green.

fredoh9 (approved):
Looks good to me. Thank you for checking for this. And nice comment in addition.

In minor comment, New line at line 83 is your intention? Indentation is good, so I wouldn't add new line there.

marc-hb:
Thanks for the review!

> In minor comment, New line at line 83 is your intention? Indentation is good, so I wouldn't add new line there.

Yes it was intentional and to mirror the new line at the start of the block, so the block is more visible.

PR #771, opened 2021-09-21
sof_remove.sh: Update SOF client driver names

ujfalusi:
A new client is added for message injection: 'snd_sof_ipc_msg_injector'
the snd_sof_ipc_test is renamed to 'snd_sof_ipc_flood_test'

In the latest version of the SOF client series.

Signed-off-by: Peter Ujfalusi <peter.ujfalusi@linux.intel.com>

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -157,6 +157,8 @@ remove_module snd_soc_intel_sof_maxim_common
 #-------------------------------------------
 remove_module snd_sof_probes
 remove_module snd_sof_ipc_test
+remove_module snd_sof_ipc_flood_test
+remove_module snd_sof_ipc_msg_injector
 remove_module snd_sof_dma_trace
 
 # snd_sof_nocodec dependencies re-ordered


ujfalusi:
I always forget to update the sof-tools first :(
Draft SOF client PR's results does not look good w/o this patch: https://sof-ci.01.org/linuxpr/PR3136/build6374/devicetest/

marc-hb (approved):

tools/kmod/sof_remove.sh
```diff
@@ -156,7 +156,8 @@ remove_module snd_soc_intel_sof_maxim_common
 # SOF client drivers
 #-------------------------------------------
 remove_module snd_sof_probes
-remove_module snd_sof_ipc_test
+remove_module snd_sof_ipc_flood_test
```
marc-hb, line 159: Please remove both the old and new name for the transition. Explain with a comment.

marc-hb (changes requested):

marc-hb:
> I always forget to update the sof-tools first :(

https://en.wikipedia.org/wiki/Test-driven_development

tools/kmod/sof_remove.sh
ujfalusi: I'm not following you, sorry.
`snd_sof_ipc_test` is no longer exist in my SOF client series as it was the flood test and it got renamed to `snd_sof_ipc_flood_test`

Since we have a new debug feature for injecting messages, we also have a new client driver: `snd_sof_ipc_msg_injector`

The commit message tells this also.

marc-hb: > snd_sof_ipc_test is no longer exist in my SOF client series

sof-test is used to test older sof branches.

Assuming we'd even want that[*], we don't have a way to "atomically" upgrade both sof and sof-test at the same time for everyone.

[*] good luck bisecting a test failure then.

ujfalusi: OK, I'll keep the retired module name.
Just a sec.

ujfalusi:
Changes since v1:
- keep the old `snd_sof_ipc_test` in the list of modules

tools/kmod/sof_remove.sh
marc-hb: I'm happy with the code change now but the comment does not seem to hint at backward compatibility at all. So I'm afraid someone not well informed and trigger-happy could just go and remove that line

ujfalusi: Or just drop the comment ;)

Old PRs with the `snd_sof_ipc_test` do not apply to sof-dev anymore due to extensive changes and it is not present in my updated PR, but it might re-surface.

ujfalusi:
Changes since v2:
- Remove the needles comment after the `snd_sof_ipc_test` module name.

tools/kmod/sof_remove.sh
marc-hb: Again, why not just add this latest and greatest explanation as a comment? It's not like this script is drowning under too many comments and having to lookup and go to the code review to find this sort of information is tedious and uncertain.

ujfalusi: I don't think it is good to have multi-line comment for a module name which was used during the development cycle of a big feature set.
I figured that instead of going word by word changes in a comment, it is cleaner to have no comment at all.

marc-hb: > which was used during the development cycle of a big feature set.

Sorry I missed the "old" one was never merged in the first place.

marc-hb (approved):

PR #768, opened 2021-09-17
[SKIP CI] tplgtool2.py: Show component core in graph 

YongAnLu00:
add an option to allow tplgtool2.py dump graph with components
core, so people can understand and analyze problems of multicore
pipelines more concisely.

Here is an example for `sof-tgl-nocodec-ci.tplg`:

sof-tgl-nocodec-ci

This is based on #755.

Signed-off-by: Yongan Lu <yongan.lu@intel.com>

PR #773, opened 2021-09-28
tplgtool2.py: accept absolute topology file path

YongAnLu00:
`Path.glob()` didn't implement for absolute path (will raise NotImplementedError).
Here is a workaround for this limitation from Python.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -868,7 +868,7 @@ if __name__ == "__main__":
         parser.add_argument('-t', '--tplgroot', type=str, help="load tplg file from tplg_root folder")
         parser.add_argument('-d', '--dump', type=str, help='dump specified topology information, '
             'if multiple information types are wanted, use "," to separate them, eg, `-d pcm,graph`')
-        parser.add_argument('filename', type=str, help='topology file name(s), ' \
+        parser.add_argument('filename', type=str, help='topology file name pattern(s), ' \
             'if multiple topology file names are specified, please use "," to separate them')
         # The below options are used to control generated graph
         parser.add_argument('-D', '--directory', type=str, default=".", help="output directory for generated graph")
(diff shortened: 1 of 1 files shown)


paulstelian97 (dismissed):

tools/tplgtool2.py
```diff
@@ -882,14 +882,23 @@ def arg_split(value: str) -> "list[str]":
         "Split comma-delimited arguements value."
         return [v.strip() for v in value.split(',')]
 
+    def glob_or_abs(rootpath: Path, filename: str) -> Path:
+        # this is a workround for Path.glob(): it didn't implement for absolute path,
+        # even though the document said the path can be absolute:
+        # https://docs.python.org/3.8/library/glob.html#glob.glob
```
marc-hb, line 888: You're confusing https://docs.python.org/3.8/library/glob.html#glob.glob with https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

marc-hb (changes requested):
This seems like a lot of code for something with a name as simple as `get_tplg_paths()`. Can you please add short pydoc to that function with the expected inputs and outputs? Otherwise it's too time consuming to review.

PS: Python is used by millions. If it had a bug that simple then either the code or the documentation would have been fixed a long time ago.

tools/tplgtool2.py
marc-hb: This works:
```python
[ x for x in pathlib.Path("/").glob("usr/*/ls*") ]
```
This does not work:
```python
[ x for x in pathlib.Path("/").glob("/usr/*/ls*") ]

 NotImplementedError("Non-relative patterns are unsupported")
```
I presume the latter is what you ran into, confirm?

If I guessed correctly then the question is why do you need to look at `"/"` + `"/usr/*/ls*"`? How about not getting in this situation in the first place?

YongAnLu00: @marc-hb Thanks for your correction. We have many ways to give topology paths, for example:
* `tplgtool2.py -t /lib/firmware/intel/sof-tplg sof-tgl-nocodec.tplg`: find `/lib/firmware/intel/sof-tplg/sof-tgl-nocodec.tplg` 
* `tplgtool2.py sof-tgl-nocodec.tplg`: find `sof-tgl-nocodec.tplg` in current working directory
* `tplgtool2.py /lib/firmware/intel/sof-tplg/sof-tgl-nocodec.tplg`: absolute path
* `tplgtool2.py -t /lib/firmware/intel/sof-tplg sof-tgl-nocodec.tplg,sof-tgl-rt5682.tplg`: find `/lib/firmware/intel/sof-tplg/sof-tgl-nocodec.tplg`  and `/lib/firmware/intel/sof-tplg/sof-tgl-rt5682.tplg`
* `tplgtool2.py -t /lib/firmware/intel/sof-tplg sof-tgl-*.tplg`: find all files match `sof-tgl-*.tplg` in `/lib/firmware/intel/sof-tplg`
*  and so on...

So this is not for searching `/usr/*/ls` in `/`, but for searching `/usr/*/ls` when root path is unspecified.

Most of these are designed to keep compatible with the legacy tool:
<https://github.com/thesofproject/sof-test/blob/530d8eb24abdf69794733c75658681082be7bd51/tools/tplgtool.py#L881-L900>
The 'wild-card' style filename could be a replacement for the special filename `all` in legacy code.

Or maybe you have a better design?

tools/tplgtool2.py
```diff
@@ -882,14 +882,23 @@ def arg_split(value: str) -> "list[str]":
...
+
     def get_tplg_paths(tplgroot: typing.Union[str, None], filenames: str) -> "list[Path]":
         rootpath = Path.cwd() if tplgroot is None else Path(tplgroot.strip())
         if not rootpath.is_dir():
             raise NotADirectoryError(rootpath)
         return [
             fullpath 
             for filename in arg_split(filenames) 
-            for fullpath in rootpath.glob(filename) 
+            for fullpath in glob_or_abs(rootpath, filename)
```
marc-hb, line 901: It finally makes sense but I found the 2-level list comprehension hard to follow. Returning a generator called `glob_or_abs` is not very intuitive either. The "or_abs" is an internal implementation detail, users of that function don't need to know about it.


Maybe this:
```python
return [  ​topology
   ​for pattern in patterns.split(',')
   for topology in do_glob(basepath, pattern)
]
```

marc-hb (changes requested):

marc-hb (approved):
Thanks much clearer now

YongAnLu00:
I have triggered internal tests (6971 and 6972) with `tplgtool2.py` from this PR and #768 as the graph drawing tool, and the test results look good.

Issue #761, opened 2021-09-07
[BUG] Check SOF soundcard is enumerated first


fredoh9:
**Describe the bug**
If SOF soundcard is not enumerated, no need to test further, that's wasting time and resources.

This is an example. It has a kernel oops during the rt5682 probe which prevents the ASOC component from being registered so there will never be a card created.
https://sof-ci.01.org/linuxpr/PR3130/build6275/devicetest/?model=CML_HEL_RT5682&testcase=verify-pcm-list

Test results also misleading. Playback test are done with USB soundcard. They must be reported as failures.
https://sof-ci.01.org/linuxpr/PR3130/build6275/devicetest/?model=CML_HEL_RT5682&testcase=check-playback-10sec

**Expected behavior**
Simple way, check SOF soundcard is available before starting any test.

Better way,
There are first group of `verify-*.sh`. Any of failure of` verify-*.sh` test group, we can safely abort entire test suite.

marc-hb:
That's not really a job for sof-test because sof-test has no test runner or idea of what a test suite or test plan is. Maybe it should but right now it hasn't.

plbossart:
@marc-hb sof-test relies on the external TPLG variable to know what to do, so we could either detect that the topology and card name are aligned, or pass a SOF_CARD external variable to check the expected card name.

marc-hb:
> so we could either detect that the topology and card name are aligned, 

Isn't that more or less happening already? In https://sof-ci.01.org/linuxpr/PR3130/build6275/devicetest/?model=CML_HEL_RT5682&testcase=verify-pcm-list
```
2021-09-01 06:30:37 UTC [REMOTE_ERROR] Pipeline(s) from topology don't match pipeline(s) from system
```
https://github.com/thesofproject/sof-test/blob/main/test-case/verify-pcm-list.sh

BTW: #762

marc-hb:
> If SOF soundcard is not enumerated, no need to test further, that's wasting time and resources.

I want  to close this because:
- every sof-test should fail IMMEDIATELY when there is no sound card. If not, please re-open with the specific test that does not and will fix. So running all the tests should waste very little time when the sound card is missing.
- sof-test has (unfortunately) no concept of test plan or test runner. So aborting a test plan does (unfortunately) not belong to sof-test, it's something that has to be outside sof-test = wherever the test plan is implemented.

plbossart:
[...]

PR #777, opened 2021-10-02
verify-sof-firmware-load: invoke systemctl with --no-pager

marc-hb:
Otherwise this script stops run-all-tests.sh when it fails.


marc-hb:
Found while testing #761

fredoh9 (approved):
I saw the problem too, thank you for the fix

PR #790, opened 2021-10-27
tplgtool2: Hide core info when unspecified in topology

YongAnLu00:
Hide core info instead of fallback to "core:0" when it is unspecified in
topology, because the default "0" could be misleading for some
components like Buffer which are under the pipeline with the non-0 core.

example for `sof-tgl-nocodec-ci`:
sof-tgl-nocodec-ci


aiChaoSONG (approved):

YongAnLu00:
I have triggered the internal test 7747 for this PR, everything looks good.

PR #794, opened 2021-10-27
opt.sh: catch duplicate --long options

marc-hb:
As found by Pierre in fix #791

Until #791 is merged it should fail like this:
```
23:29:31 [ERROR] Duplicate '--loop' option: loop count
23:29:31 [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
23:29:31 [ERROR]  die()  @  ./test-case/../case-lib/lib.sh
23:29:31 [ERROR]  _fill_opt_vars()  @  sof-test/case-lib/opt.sh:59
23:29:31 [ERROR]  func_opt_parse_option()  @  sof-test/case-lib/opt.sh:120
23:29:31 [ERROR]  main()  @  ./test-case/multiple-pause-resume.sh:48
```

### DIFF
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -10,9 +10,9 @@
 #    1: this option requires an extra argument
 #    0: this option behaves like boolean, and requires no argument
 # OPT_VAL:
-#    - the extra argument required if OPT_HAS_ARG=1
-#    - default boolean value 0/1 if OPT_HAS_ARG=0, if boolean option
-#      is specified in command line, default OPT_VAL will be flipped.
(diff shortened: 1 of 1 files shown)


greg-intel:
I'm curious why the duplication of an argument occurs.  It's not common to have to check that an arg only appears once, so I'm wondering if there's a root cause we can fix?

marc-hb:
> I'm curious why the duplication of an argument occurs.

Because of hasty copy/paste/diverge, already fixed by #791. This one is so it never happens again.

greg-intel (approved):

PR #797, opened 2021-10-29
logging: add new SOF_LOGS_BASE feature 

marc-hb:
\+ 2 other minor fixes

Using root to write inside a regular user directory (sof-test) is not
great. It creates files difficult to remove and it does not work across
network filesystems. Don't change the default location but add a new
SOF_LOGS_BASE feature.


marc-hb:
alsabat failure totally unrelated to this https://sof-ci.01.org/softestpr/PR797/build912/devicetest/. ADL was not available at the time.

fredoh9 (approved):
This is really good idea to add an option to change log path by defining SOF_LOGS_BASE.

PR #799, opened 2021-11-01
check-sof-logger: catch ERROR in firmware traces and fail

marc-hb:
Add function find_logger_error() in case-lib so it can re-used by other
tests later.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -256,8 +256,21 @@ main()
         }
     done
 
+    local OK=true
+
+    for f in "${stdout_files[@]}"; do
+        local tracef="$LOG_ROOT/logger.$f.txt"
+        check_error_in_file "$tracef" || {
+            OK=false; printf '\n'
+        }
+    done
+
     # Show all outputs even when everything went OK
-    print_logs_exit 0
+    if $OK; then
+        print_logs_exit 0
+    else
+        print_logs_exit 1 "^^ ERROR(s) found in firmware logs ^^"
+    fi
 }
 
 main "$@"
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -356,6 +356,30 @@ func_lib_start_log_collect()
     sudo "${loggerCmd[@]}" > "$logfile" &
 }
 
+# -B 2 shows the header line when the first etrace message is an ERROR
+# -A 1 shows whether the ERROR is last or not.
+check_error_in_file()
+{
+    local platf; platf=$(sof-dump-status.py -p)
+
+    case "$platf" in
+        byt|bdw)
+            # Maybe downgrading this to WARN would be enough, see #799
+            #  src/trace/dma-trace.c:654  ERROR dtrace_add_event(): number of dropped logs = 8
+            dlogw 'not looking for ERROR on BYT/BDW because of known DMA issues #4333 and others'
+            return 0
+            ;;
+    esac
+
+    test -r "$1" || {
+        dloge "file NOT FOUND: '$1'"
+        return 1
+    }
+    if grep -B 2 -A 1 -E 'ERRO?R?' "$1"; then
+       return 1
+    fi
+}
+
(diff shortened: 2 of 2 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR799/build917/devicetest/?model=BDW_WSB_RT286&testcase=check-sof-logger has a very interesting:
```
 TIMESTAMP      (us)           DELTA  C# COMPONENT          LOCATION                      CONTENT	ktime=117.383s  @  2021-11-01 23:27:33 +0000 UTC
[  1852324251.812] (           0.000) c0 dma-trace             src/trace/dma-trace.c:422  INFO SHM: FW ABI 0x3014000 DBG ABI 0x5003000 tag v1.9-rc1-211-gb0bb21dc0ee6 src hash 0xa9c07289 (ldc hash 0xa9c07289)
[      324686.862] (      324686.875) c0 dma-trace             src/trace/dma-trace.c:605  ERROR dtrace_add_event(): number of dropped logs = 1
```

marc-hb:
SOFCI TEST

marc-hb:
> https://sof-ci.01.org/softestpr/PR799/build917/devicetest/?model=BDW_WSB_RT286&testcase=check-sof-logger has a very interesting:

https://sof-ci.01.org/softestpr/PR799/build16/devicetest has something similar on BYT (BDW is not in PR testing anywmore)

```
[   178559514.207] (           0.000) c0 dma-trace             src/trace/dma-trace.c:347  INFO SHM: FW ABI 0x3016000 DBG ABI 0x5003000 tags SOF:v2.0-rc1-813-g5cfc37040bbe src hash 0xb72c3013 (ldc hash 0xb72c3013)
[       27442.082] (       27442.082) c0 dma-trace             src/trace/dma-trace.c:654  ERROR dtrace_add_event(): number of dropped logs = 8
```

ujfalusi:
@marc-hb, is it a recent issue or it has been there?
It is very likely to happen if the firmware is generating 'enough' trace messages up to the point when `sof-logger` starts and opens the sof/trace debugfs file to start consuming the dtrace messages.

If you do not start the `sof-logger` at all then yeah, you are going to drop messages, but is this an error?

marc-hb:
Thanks @ujfalusi for the prompt review!

Interesting. So this reproduced systematically on BYT and BDW... because the buffer is smaller / too small? Or maybe because they have more interfaces and more to log.

So I can just downgrade this ERROR to a WARN, right?

ujfalusi:
@marc-hb, I think it is WARN rather an ERR, yes (if you refer to the fw print about the dropped new entries. It only takes to not start sof-logger to eventually have them, which is fine?

It can explain 'blackouts' in logs

marc-hb:
https://sof-ci.01.org/softestpr/PR799/build112/devicetest is all green. I recommend we merge this "as is" and if this error message become too frequent we can downgrade it.

test-case/check-sof-logger.sh
```diff
@@ -256,8 +256,31 @@ main()
...
+
+    case "$platf" in
+        byt|bdw)
+            dlogw 'not looking for ERROR on BYT/BDW because of known DMA issues #4333 and others'
+            ERROR_scan=false
+            ;;
+    esac
+
+    if $ERROR_scan; then
+    for f in "${stdout_files[@]}"; do
```
keqiaozhang, line 271: Why not indent?

marc-hb: It was supposed to be temporary, I can indent.

marc-hb: You pushed me to make more changes, thanks.

marc-hb:
https://sof-ci.01.org/softestpr/PR799/build113/devicetest/ has one usual WHL UPX suspend/resume TIMEOUT, everything else is green.

keqiaozhang (approved):
LGTM

marc-hb:
Thanks @keqiaozhang

marc-hb:
Pattern matching fix submitted in #946

PR #805, opened 2021-11-13
lib.sh: fix sudo "$loggerCmd" quoting

marc-hb:
sudo "$loggerCmd" works only with our redefined sudo function. Replace
it with an array-based and properly quoted sudo "${loggerCmd[@]}" that
works with any sudo.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -140,10 +140,10 @@ func_lib_start_log_collect()
         return 3
     fi
 
-    local loggerCmd="$SOFLOGGER $logopt -l $ldcFile -o $logfile"
-    dlogi "Starting $loggerCmd"
+    local loggerCmd=("$SOFLOGGER" "$logopt" -l "$ldcFile" -o "$logfile")
+    dlogi "Starting ${loggerCmd[*]}"
     # Cleaned up by func_exit_handler() in hijack.sh
-    sudo "$loggerCmd" &
+    sudo "${loggerCmd[@]}" &
 }
 
 # Calling this function is often a mistake because the error message


marc-hb:
https://sof-ci.01.org/softestpr/PR805/build922/devicetest/?model=ADLP_RVP_NOCODEC&testcase=verify-kernel-boot-log failed because the tests started before the system finished booting. Internal issue #147 (for now)

https://sof-ci.01.org/softestpr/PR805/build922/devicetest/?model=CML_RVP_SDW&testcase=check-suspend-resume-with-playback-5 is the usual rtcwake TIMEOUT

Everything else is green.

case-lib/lib.sh
```diff
@@ -140,10 +140,10 @@ func_lib_start_log_collect()
         return 3
     fi
 
-    local loggerCmd="$SOFLOGGER $logopt -l $ldcFile -o $logfile"
-    dlogi "Starting $loggerCmd"
+    local loggerCmd=("$SOFLOGGER" "$logopt" -l "$ldcFile" -o "$logfile")
```
fredoh9, line 143: Quoting is good, but why we need () here?

case-lib/lib.sh
```diff
@@ -140,10 +140,10 @@ func_lib_start_log_collect()
...
+    dlogi "Starting ${loggerCmd[*]}"
```
fredoh9, line 144: loggerCmd is array? how? below sudo "${loggerCmd[@]}" also same question. Quoting is good

case-lib/lib.sh
greg-intel: The parenthesis make it an array.  I'm just not sure what the difference between [*] and [@] is, as far as I can tell/remember they retrieve all elements, so not sure why we've got two versions. (The globber "*" might be clearer to most)

greg-intel: Found the answer:
` If the word is double-quoted, ${name[*]} expands to a single word with the value of each array member separated by the first character of the IFS special variable, and ${name[@]} expands each element of name to a sep- arate word. When there are no array members, ${name[@]} expands to nothing. If the double-quoted expansion occurs within a word, the expansion of the first parameter is joined with the beginning part of the original word, and the expansion of the last parameter is joined with the last part of the original word.`

marc-hb: > Quoting is good, but why we need () here?

From https://mywiki.wooledge.org/BashGuide/Arrays

> Strings are without a doubt the most used parameter type. But they are also the most misused parameter type. It is important to remember that a string holds just one element. [...] The only safe way to represent multiple string elements in Bash is through the use of arrays. 

Shellcheck is very good at catching this mistake.

Scrolling down to "Expand elements"

> The syntax "${myfiles[@]}" is extremely important. It works just like "$@" does for the positional parameters: it expands to a list of words, with each array element as one word, no matter what it contains. Even if there are spaces, tabs, newlines, quotation marks, or any other kind of characters in one of the array elements, it'll still be passed along as one word to whichever command we're running.

Also in https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html and https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_02

case-lib/lib.sh
marc-hb: `"${name[@]}"` cannot be used inside a larger string, I mean you cannot do `"blabla ${name[@]}"` because that does not make sense. `"${name[@]}"` expands to: `"element 1" "element 2" ...`  so where would `"blabla"` be in that list?

case-lib/lib.sh
fredoh9: Thank you for good answers. I forgot about "()" and lost there.

fredoh9:
More fundamental question. Can you elaborate more why previous code didn't work on certain condition?

marc-hb:
> More fundamental question. Can you elaborate more why previous code didn't work on certain condition?

Very good question... answered in the first line of the commit message. The failure happens only when you do this:

```diff
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -150,7 +150,7 @@ function func_exit_handler()
 
 SUDO_LEVEL=""
 # overwrite the sudo command, sudo in the script can direct using sudo command
-sudo()
+disabled_sudo()
 {
     func_hijack_setup_sudo_level || true
     local cmd
```

I locally disable the sudo "hijack" because it makes `set -x` unreadable.

case-lib/lib.sh
greg-intel: Same, at first I saw it and thought subshell.

marc-hb: Guys, I'm sorry I can't fix 50 years old shell syntax and make it clearer :-) Arrays and subshells are fairly basic bash features, I don't think this requires a comment.

marc-hb:
Any concern left?

fredoh9 (approved):
Thank you for the patience. Took more time to understand the failure.
I had to try this myself. Good finding and the fix.

marc-hb:
I screwed quotation in this and added a regression. Fix submitted in #817, please review quickly thx!

PR #807, opened 2021-11-19
keyword-detection: don't use fixed card number

aiChaoSONG:
The SOF card number is not always 0, because we set up
a USB sound card for loopback.


marc-hb:
When and were is SOFCARD defined?

aiChaoSONG:
> When and where is SOFCARD defined?

https://github.com/thesofproject/sof-test/blob/814b904acfca65dbebb869b0ab96735dc9a51cb0/case-lib/lib.sh#L36

keyonjie (approved):
Looks good, thanks.

marc-hb (approved):

PR #809, opened 2021-11-22
lib.sh: re-enable sof-logger for Zephyr

marc-hb:
check-sof-logger was re-enabled for Zephyr in commit 56db25953967 and it
passed the daily tests. Re-enable logging for all tests when using
Zephyr.

Partial revert of 325e57244e2 ("logger: disable sof-logger for zephyr")


greg-intel (approved):
Looks straightforward.

PR #810, opened 2021-11-23
lib.sh: switch is_zephyr() to look for zephyr strings in .ldc

marc-hb:
Now that the sof-logger is well tested with Zephyr, change the
is_zephyr() test to count the number of "zephyr" strings in the .ldc
dictionary file. The previous test was based on /etc/sof/manifest.txt
which is a CI-only file. This probably removes a dependency on 'jq' too.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -390,9 +390,15 @@ disable_kernel_check_point()
 
 is_zephyr()
 {
-    local manifest=/etc/sof/manifest.txt
-    test -e "$manifest" || return 1
-    jq '.version.firmwareType' "$manifest" | grep "zephyr"
+    local ldcFile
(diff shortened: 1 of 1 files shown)


marc-hb:
Big regressions in https://sof-ci.01.org/softestpr/PR810/build926/devicetest unrelated to this, see https://github.com/thesofproject/linux/pull/3292 for more details.

marc-hb:
SOFCI TEST

keqiaozhang (approved):
This removes the dependency on jq.

greg-intel (approved):
LGTM.  I don't know enough about the ldc file to determine if there's a way to check a specific value directly, but 0 vs 30 is a pretty clear message.

Issue #726, opened 2021-07-01
[FEATURE] sof-test should run without DMA trace - only etrace


plbossart:
**Is your feature request related to a problem? Please describe.**

Facts:
* Most distributions do not enable DMA trace (considered a developer option). This can be overridden with a kernel parameter but doing so changes what users report.
* The DMA trace is known to be invasive (bandwidth, etc) and either creates issues or prevent overflows (https://github.com/thesofproject/sof/issues/3868#issuecomment-850845612)

We should be able to run sof-tests, possibly in a more limited version since less information is available, without *requiring* the DMA trace to be enabled.

**Describe the solution you'd like**
autodetection that only /sys/kernel/debug/sof/etrace is available and skips of all DMA trace parsing/error detection

**Describe alternatives you've considered**
none

@lgirdwood @kv2019i @marc-hb @aiChaoSONG @fredoh9 @ranj063 @lyakh FYI

marc-hb:
We already have a -s option: https://github.com/thesofproject/sof-test/pull/681

What's missing and and what level? In run-all-tests.sh https://github.com/marc-hb/sof-test/commit/run-all-nologger ?

plbossart:
@marc-hb I don't want to disable sof-logger (which will read the etrace which is ALWAYS enabled).


I just want to disable the DMA trace.... (sof-logger -t)

plbossart:
See e.g. https://github.com/thesofproject/sof/issues/3673, how would one run such tests if we cannot disable the DMA trace...

marc-hb:
I just tried `modprobe snd_sof sof_debug=0` as documented at https://thesofproject.github.io/latest/developer_guides/debugability/logger/index.html but the firmware seems to still receive the same `ipc: debug cmd 0x30000` request from the kernel to enable the DMA trace.

plbossart:
@marc-hb try PR https://github.com/thesofproject/linux/pull/3021, that disables the trace for real. The current code can only enable the trace with the module parameter, not disable it.

marc-hb:
> @marc-hb try PR thesofproject/linux#3021, that disables the trace for real. 

[...]

PR #811, opened 2021-11-24
case-lib: add support for no logger mode

aiChaoSONG:
Logger is not supported for firmware without
ldc file, add a new mode to test without
sof-logger.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -67,3 +67,7 @@ SUDO_LEVEL=${SUDO_LEVEL:-}
 # in CI, which is controlled by sof-framework. In manual run, user can
 # override the default value.
 SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-5}
+
+# Logger is not always supported, we want sof-test to run even
+# without logger.
+LOGGER_MODE=${LOGGER_MODE:-true}
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -397,7 +397,7 @@ is_zephyr()
 
 logger_disabled()
 {
-    [[ ${OPT_VAL['s']} -eq 0 ]]
+    [[ ${OPT_VAL['s']} -eq 0 ]] || [[  "$LOGGER_MODE" == "false" ]]
 }
 
 print_module_params()


aiChaoSONG:
To use this feature, run `export NO_LOGGER_MODE=true` before run any test cases, or you can make this line in your .bashrc. if you run test case remotely via ssh, insert `NO_LOGGER_MODE=true` into /etc/environment

keqiaozhang (dismissed):

marc-hb:
How about making the `find_ldc_file` failure not fatal instead?

aiChaoSONG:
> How about downgrading the `find_ldc_file` failure to your `dlogw` warning instead?

I think it can not resolve the problem, even if the function `find_ldc_file` is not fatal, sof-test still try to start sof-logger, and kill it in the end, which causes the case failure. What Pierres want is the `no logger mode` actually, that is when there is no ldc file, or even logger cannot support the FW, we still want the test case run, and get the right exit status.

case-lib/lib.sh
```diff
@@ -107,6 +107,11 @@ find_ldc_file()
...
 func_lib_start_log_collect()
 {
+    [ -z "$NO_LOGGER_MODE" ] || {
+        dlogw "sof-test is running in NO_LOGGER_MODE, firmware log will not be collected"
+        return
+    }
+
     local is_etrace=${1:-0} ldcFile
 
     ldcFile=$(find_ldc_file) || return $?
```
marc-hb, line 117:
```suggestion
    ldcFile=$(find_ldc_file) || {
       dlogw "no .ldc file found, firmware log will not be collected"
       return 0
    }
```

marc-hb:
> even if the function find_ldc_file is not fatal, sof-test still try to start sof-logger, 

No, see suggested code change above.

> What Pierres want is the no logger mode actually, that is when there is no ldc file, or even logger cannot support the FW, we still want the test case run, and get the right exit status.

The user should not have to _manually_ tell sof-test that there is no .ldc file when there is already a function that knows how to look for the .ldc file _automatically_.

aiChaoSONG:
This is kind of use ldc file to hide sof-logger incapability

case-lib/config.sh
```diff
@@ -63,6 +63,11 @@ SUDO_LEVEL=${SUDO_LEVEL:-}
 #
 # NO_BT_MODE=true
 
+# Logger is not supported for firmware without ldc file. With NO_LOGGER_MODE
+# enabled, sof-logger will not be started.
+#
+# NO_LOGGER_MODE=true
+
```
dbaluta, line 70: I am in favor of not using negative logic :). e.g LOGGER_MODE= true (we have logging) LOGGER_MODE=false (we don't have logging).

aiChaoSONG: Because the NO_XXX_MODE is minority, and XXX_MODE is majority. Normally, NO_XXX_MODE is set for special one or two devices, or otherwise we need to set XXX_MODE for all devices. 

Ex, in the early stage of new HW, graphics driver is not ready, HDMI pipelines are broken, so we only need set NO_HDMI_MODE for the new HW.

marc-hb: > Because the NO_XXX_MODE is minority, and XXX_MODE is majority.

So make LOGGER_MODE=true and HDMI_MODE the default values and problem solved. There's never a good reason for negative options.

marc-hb (changes requested):
> This is kind of use ldc file to hide sof-logger incapability

Why would you install an .ldc file when it cannot work?

Now let's say you want to temporarily test without a working sof-logger for some reason like #726. Then it's much simpler to rename the .ldc file than remember the name of an environment variable and pass it to a far away process.

marc-hb (changes requested):
Also, there is already a `logger_disabled()` function in the same `lib.sh` file, so we should adjust it and not duplicate it elsewhere. It was successfully used by Zephyr for a few months. 

See #681, #726 and commit 0a464ff0

aiChaoSONG:
@marc-hb OK, let keep on your first suggestion.

case-lib/lib.sh
```diff
@@ -109,7 +109,10 @@ func_lib_start_log_collect()
 {
     local is_etrace=${1:-0} ldcFile
 
-    ldcFile=$(find_ldc_file) || return $?
+    ldcFile=$(find_ldc_file) || {
+       dlogw "No .ldc file found, firmware log will not be collected"
+       return 0
+    }
```
plbossart, line 115: this test is a bit odd. In itself, it looks fine, but if you scroll down the code there are still errors that are reported if SOFLOGGER is not defined, or not well configured.

The ask is really to completely avoid depending on sof-logger and ONLY use alsa-lib/alsa-bat for test results.

marc-hb:
> Also, there is already a logger_disabled() function in the same lib.sh file, so we should adjust it and not duplicate it elsewhere. It was successfully used by Zephyr for a few months.
> See #681, #726 and commit 0a464ff

Zephyr could not use sof-logger for months. Yet it could pass some tests with zero dependency on sof-logger thanks to the function `logger_disabled()`. As this function `logger_disabled()` has been well tested up to 1 week ago and still exists, please re-use it and modify it. Don't change `func_lib_start_log_collect()` at all.

marc-hb:
v1 commit 7b89481a7f5 used manual `LOGGER_MODE` configuration
v2 commit 6f1c6aa5021 used automatic `find_ldc_file` configuration
v3 commit 161b34e234 is back to manual `LOGGER_MODE` configuration (in a better tested place)

Any objection to automatic configuration in the better tested place?


In general, the fewer settings the better. Fewer settings means everyone including CI runs the same code, it reduces combinatorial explosion.

Here's why I think we can and should avoid a new LOGGER_MODE setting:

- If you have no matching `.ldc` filename on your system, then why would you be required to manually configure sof-test with `LOGGER_MODE=false` to say that you don't want logging? You can't have logging anyway without a matching .ldc file, so it should be obvious that you don't want logging. If the lack of .ldc file was an accident, then it's not the job of every test to detect that system configuration accident and a new variable is a very high price to pay for just that.

- If you do have a matching `.ldc` filename on your system, then
   - either it works and sof-test should definitely be using it
   - or it does not work (leftover from a previous build?) and your system configuration is broken and must be cleaned up: just delete that bogus `.ldc` file. Again it's not the job of the tests to verify your system configuration, it's better and simpler if the tests assume the system is configured correctly and automatically adjust to it. There's an infinite number of ways to misconfigure a system; sof-test should only report the very dangerous / time-consuming misconfigurations. Thanks to the dictionary checksum this misconfiguration has a very good error message already.
   
   
On the other hand, sof-test should always report clearly whether it found some .ldc file or not and what decision it made.

marc-hb:
In a private discussion @plbossart expressed a strong preference for the environment variable type of interface.

My top concern with the environment variable was adding a new configuration step for everyone who does not have or want any .ldc file (valid or not). As mentioned by @plbossart , this includes some Linux distributions.

After some more thought I realized we can have it all: automated when possible and both manual configuration ways. This is what I just submitted in #813 which I hope will supersede this.

Note I changed the environment variable to `SOF_LOGGING=none` so it's ready for a future `SOF_LOGGING=etrace` as requested in #726

marc-hb:
Superseded  by #813

PR #813, opened 2021-12-01
lib.sh: disable logging with SOF_LOGGING=none env or missing .ldc

marc-hb:
Some configurations don't support the logger. Others do but do not want
to run it for various reasons.

Supersedes #811, see earlier discussions there.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -79,6 +79,7 @@ fake_kern_error()
 
 }
 
+# Prints the .ldc file found on stdout, errors on stderr.
 find_ldc_file()
 {
     local ldcFile
@@ -87,6 +88,7 @@ find_ldc_file()
     # and then on the standard location.
     if [ -n "$SOFLDC" ]; then
         ldcFile="$SOFLDC"
+        >&2 dlogi "SOFLDC=${SOFLDC} overriding default locations"
     else
         local platf; platf=$(sof-dump-status.py -p) || {
             >&2 dloge "Failed to query platform with sof-dump-status.py"
@@ -98,7 +100,7 @@ find_ldc_file()
     fi
 
     [[ -e "$ldcFile" ]] || {
-        >&2 dloge "LDC file $ldcFile not found, check the SOFLDC environment variable or copy your sof-*.ldc to /etc/sof"
+        >&2 dlogi "LDC file $ldcFile not found"
         return 1
     }
     printf '%s' "$ldcFile"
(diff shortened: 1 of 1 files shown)


marc-hb:
Need to fix the commit message.

marc-hb:
https://sof-ci.01.org/softestpr/PR813/build934/devicetest has two known issues: ADL NTP sync / too fast boot + very recent (5.6) "preemptible" regression https://github.com/thesofproject/linux/issues/3283

To really test this in CI (in addition to extensive testing locally) I hardcoded `SOF_LOGGING=none` in throw-away PR #814. The results in https://sof-ci.01.org/softestpr/PR814/build935/devicetest/ look good: all tests passed or skipped without any FW logs - except for the (very special/different) sof-logger test which passed with FW logs.

case-lib/lib.sh
```diff
@@ -397,7 +399,27 @@ is_zephyr()
 
 logger_disabled()
 {
-    [[ ${OPT_VAL['s']} -eq 0 ]]
+    local ldcFile
+    # Some firmware/OS configurations do not support logging.
+    ldcFile=$(find_ldc_file) || {
+        dlogi '.ldc dictionary file not found, SOF logs collection disabled'
+        return 0 # 0 is 'true'
```
greg-intel, line 406: Not sure this comment is needed, as the script self documents on the line before.

greg-intel (approved):

case-lib/lib.sh
```diff
@@ -397,7 +399,27 @@ is_zephyr()
...
+    }
+
+    # Disable logging when available...
+    if [ ${OPT_VAL['s']} -eq 0 ]; then
+        return 0
+    fi
+
+    # ... across all tests at once.
+    # In the future we should support SOF_LOGGING=etrace (only), see
```
fredoh9, line 415: Need more clarification. In future we will support only etrace? why? slogger seems useful to me

fredoh9 (approved):
In general, I like this idea. logger_disabled() has 3 options to support
1. no ldc file in the system
2. OPT_VAL['s'] == 0
3. $SOF_LOGGING == none

case-lib/lib.sh
marc-hb: See #726 referenced on the next line.

case-lib/lib.sh
marc-hb: You're right, it's overkill. I will remove but only if someone requests some other changes because it does do not much harm either and uses very little real estate.

marc-hb (commented):
Thanks for the quick reviews. As this (and #811 before it) is eagerly waited, I will merge in 6-7 hours if no one requests any change. Sorry for the short notice (not so short when including earlier discussions in #811)

aiChaoSONG (approved):

PR #814, opened 2021-12-01
HACK: SOF_LOGGING=none for CI testing

marc-hb:


marc-hb:
https://sof-ci.01.org/softestpr/PR814/build935/devicetest is as expected. More details in #813

PR #817, opened 2021-12-03
lib.sh: fix etrace collection on failure (recent regression)

marc-hb:
Fixes failures to collect etrace on failure. It goes like this:
```
UTC [INFO] Starting sof-logger '' -l /lib/firmware/intel/sof/sof-apl.ldc
 -o sof-test/logs/check-runtime-pm-status/2021-12-02-23:49:42-22493/etrace.txt
error: Unused parameter ''
Usage sof-logger <option(s)> <file(s)>
```
This happens when $logopt is empty.

Fixes recent regression introduced by commit 814b904acfca6 ("lib.sh: fix
sudo "$loggerCmd" quoting")

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -142,7 +142,9 @@ func_lib_start_log_collect()
         return 3
     fi
 
-    local loggerCmd=("$SOFLOGGER" "$logopt" -l "$ldcFile" -o "$logfile")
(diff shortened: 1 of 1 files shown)


greg-intel (approved):
Fix makes sense to me.

marc-hb:
Thanks for the review. Expediting this very small change because it fixes a recent regression, sorry for the lack of notice.

PR #821, opened 2021-12-07
SOF_TEST_INTERVAL fixes

marc-hb:
See commit messages

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -19,11 +19,6 @@ declare -A TPLG_IGNORE_LST
 # shellcheck disable=SC2034
 TPLG_IGNORE_LST['pcm']='HDA Digital'
 
-# Will be set by the lib function, don't need to set
-# Catches the last line of /var/log/kern.log, which will be used by
-#   sof-kernel-log-check.
-# KERNEL_CHECKPOINT
-
 # If not set will be automatically set by logging_ctl function
 # Test case log root
 # EXAMPLE: the log for test-case/check-ipc-flood.sh will be stored at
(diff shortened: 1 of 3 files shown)


case-lib/config.sh
```diff
@@ -63,7 +58,17 @@ SUDO_LEVEL=${SUDO_LEVEL:-}
...
-SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-5}
+# SOF_TEST_INTERVAL informs sof-test of how long the external test
+# runner waits between the end of one sof-test and the start of the next
+# sof-test. sof-test uses this value to assign the corresponding kernel
+# logs "no man's land" to the second test, which can be confusing (see
+# for instance https://github.com/thesofproject/sof/issues/5032) but
+# safer: it makes sure no kernel error escapes.
+# See initial review in https://github.com/thesofproject/sof-test/pull/639
+#
+# Note the default value used by Intel's sof-jenkins is 5 seconds, as
```
aiChaoSONG, line 69: Please note that Jenkins doesn't run any test case, it run case by making use of the test runner, which is sof-framework.

The sof-framework never sets this value, it just wait 5 seconds before it run next test case, that's why we make the initial value 5 seconds, only log within the 1 second before running test case will be collected

marc-hb: > The sof-framework never sets this value

Why not? That seems wrong. sof-test should be user friendly, it should not force everyone else and especially not interactive users to configure this because one script in sof-framework is lazy.

>  that's why we make the initial value 5 seconds

You mean the default value?

> only log within the 1 second before running test case will be collected

I fail to parse this part sorry.

aiChaoSONG: here you make `SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-1}`, the default value is 1, so the log collect point is 4 (- SOF_TEST_INTERVAL - 3)  second before running the test case. but the last test end for CI is 5 second before running the test case, so the log in the 1 second after last test end is not collected in CI.

marc-hb:
The 2 timeouts in https://sof-ci.01.org/softestpr/PR821/build941/devicetest/ don't seem related.

I force-pushed something unrelated to this branch by mistake, apologies. I force-pushed and restored the earlier commits.

case-lib/config.sh
marc-hb: It is a bit ridiculous that the most frequent sof-test user and the one that needs SOF_TEST_INTERVAL the most does not define SOF_TEST_INTERVAL. I filed internal bug 158, kept the default value as is, adjusted the comments and resubmitted all the other changes.

marc-hb:
CML was not available in https://sof-ci.01.org/softestpr/PR821/build945/devicetest/ but everything else is green.

aiChaoSONG (approved):

marc-hb:
As reported by @plbossart , just found that SOF_TEST_INTERVAL was also used in test _iterations_. Fix submitted in:
- thesofproject/sof-test/pull/870

PR #822, opened 2021-12-07
kernel-log-check: remove filter that ignores DSP boot retries 

marc-hb:
Now that sof-test is using journalctl log levels and that the levels of
this DSP retries have been fixed by
thesofproject/linux#3250, it should not be
required to filter out these messages anymore. This cleanup removes
almost 30 lines of code.

This is a "logical" revert of the addition in d238c48 and of the
many changes after it; most recently 61b077b. See the code reviews
of all these commits for more details.

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -51,6 +51,9 @@ suspend-resume
 suspend-resume-with-playback
 suspend-resume-with-capture"
 
+# Requires Octave
+testlist="$testlist volume_levels"
+
 # To focus on some particular tests edit and rename this to
 # 'testlist'. Last definition wins.
 # shellcheck disable=SC2034
@@ -63,9 +66,6 @@ capture_d1l100r1
 tplg-binary
 sof-logger
 '
-
-# Requires Octave
-testlist="$testlist volume_levels"
 
 main()
 {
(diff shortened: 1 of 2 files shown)


marc-hb:
In https://sof-ci.01.org/softestpr/PR822/build944/devicetest `sh-adlp-rvp-nocodec-04` is having a really bad day with `input/output` errors and a ton of
```
[      241155.459167] (          20.572916) c2 dai          1.2            src/audio/dai.c:951  WARN dai_copy(): nothing to copy
[      243134.625755] (        1979.166626) c2 dw-dma                 src/drivers/dw/dma.c:1108 INFO dw_dma_free_data_size() size is 0!
[      243155.354921] (          20.729166) c2 dai          1.2            src/audio/dai.c:951  WARN dai_copy(): nothing to copy
[      245134.729843] (        1979.374878) c2 dw-dma                 src/drivers/dw/dma.c:1108 INFO dw_dma_free_data_size() size is 0!
[      245155.250675] (          20.520832) c2 dai          1.2            src/audio/dai.c:951  WARN dai_copy(): nothing to copy
```
however none of this matches the removed filter and everything else is green.

marc-hb:
SOFCI TEST

marc-hb:
@ujfalusi can you confirm we should not see anything at the ERROR level any more and if yes approve?

This PR is only removing error filters.

ujfalusi:
> @ujfalusi can you confirm we should not see anything at the ERROR level any more and if yes approve?
> 
> This PR is only removing error filters.

Non fatal prints should no longer appear in ERROR level, but the firmware panic (which might be non fatal) is still printed in ERROR level.

ujfalusi (approved):

PR #824, opened 2021-12-08
hijack.sh: read from -i /dev/null when changing the log level

marc-hb:
sof-logger inconveniently does not allow changing the log level without
reading some logs at the same time. We've always redirected these logs
to /dev/null but Zephyr uses a different etrace format that can cause
errors on stderr and we don't want to discard stderr because we never
want to ignore ALL errors in test code. We can't switch to the DMA trace
either because reading the DMA trace never stops.

So, read the trace from neither the DMA nor the shared memory but from
/dev/null instead.

Keep redirecting output to /dev/null too because we still don't want to
see the banner either.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -46,7 +46,7 @@ function func_exit_handler()
             # lines bug: the "ipc" logs corresponding to this -F command
             # will appear _only once_ at the end of the slogger.txt DMA
             # trace!
-            sudo "$SOFLOGGER" -l "${ldcf}" -F info=pga > /dev/null ||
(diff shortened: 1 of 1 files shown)


marc-hb:
This is required for fixing Zephyr issue thesofproject/sof/issues/5032

marc-hb:
Unrelated rtcwake timeout in https://sof-ci.01.org/softestpr/PR824/build947/devicetest/?model=APL_UP2_NOCODEC&testcase=check-suspend-resume-with-capture-5, everything else green.

aiChaoSONG (approved):

marc-hb:
Thanks @aiChaoSONG . Small, well tested one-line change up for review for 5 days: merging

PR #826, opened 2021-12-08
print_modules_params: don't SKIP the test when driver is not loaded

marc-hb:
grep return code is '2' when files are missing. '2' is also the sof-test
convention for skipping tests.
When using print_modules_params in a test with set -e, this causes the
test to be SKIPped when the driver is not loaded.

Simply ignore all grep errors in this function; it's not the job of a
"print_" function to fail. Missing audio is already caught in many other
places.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -436,10 +436,10 @@ logger_disabled()
 print_module_params()
 {
     echo "--------- Printing module parameters ----------"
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR826/build951/devicetest/ is unbelievably all green. To celebrate, reviewers must approve without any comment (just kidding)

miRoox (approved):

PR #830, opened 2021-12-24
check-sof-logger: print stderr output when etrace fails to start

marc-hb:
We can't even start we want to provide feedback right away (thanks to
set -e) and not rely on distant and elusive output post-processing.


marc-hb:
Unavailable CML_HEL_RT5682 in https://sof-ci.01.org/softestpr/PR830/build961/devicetest, everything else green.

greg-intel (approved):
LGTM.  Fail fast accomplished.

Issue #834, opened 2022-01-12
[BUG] sof_remove.sh needs to remove snd_soc_max98390


kv2019i:
**Describe the bug**
sof_remove.sh fails with sof-adl-max98390, see Intel daily plan #9293

RMMOD	snd_soc_core
rmmod: ERROR: Module snd_soc_core is in use by: snd_soc_max98390
snd_soc_max98390       32768  0

kv2019i:
Seen in daily test so P1

plbossart:
snd_soc_max98390 is not a machine driver @kv2019i. It's a codec driver and it has to be handled manually *after* you remove the machine driver.

kv2019i:
@plbossart wrote:
> snd_soc_max98390 is not a machine driver @kv2019i. It's a codec driver and it has to be handled manually _after_ you remove the machine driver.

I stand corrected. It still needs to be removed. And wouldn't it be easier to just remove "snd_soc_*" twice, first bringing down the mach drivers and on second round the codecs. Enumerating these separately adds constant churn to the scripts.

plbossart:
I can't claim I am the expert here, but I would think you will need to iterate more than twice?

We never looked at modprobe -r, not sure if this is any different from rmmod

````  -r, --remove
           This option causes modprobe to remove rather than insert a module. If the modules it depends on are also unused, modprobe will try to remove them too. Unlike insertion, more than one
           module can be specified on the command line (it does not make sense to specify module parameters when removing modules).
````

XiaoyunWu6666:
https://github.com/thesofproject/sof-test/pull/837

XiaoyunWu6666:
Another relevant PR for solving similar problem https://github.com/thesofproject/sof-test/pull/865

PR #836, opened 2022-01-12
kmod: add missing codecs

plbossart:
max98390 is missing on remove

We are also missing a number of codecs on insert, reshuffle the list
so that it uses the same order as for remove - the difference being
that we don't insert common modules automagically loaded by others.

BugLink: https://github.com/thesofproject/sof-test/issues/834

### DIFF
--- a/tools/kmod/sof_insert.sh
+++ b/tools/kmod/sof_insert.sh
@@ -34,9 +34,24 @@ sudo true
 
 insert_module snd_soc_da7213
 insert_module snd_soc_da7219
+insert_module snd_soc_pcm512x_i2c
+insert_module snd_soc_wm8804_i2c
+
 insert_module snd_soc_rt274
 insert_module snd_soc_rt286
 insert_module snd_soc_rt298
+
+insert_module snd_soc_rt700
+insert_module snd_soc_rt711
+insert_module snd_soc_rt1308
+insert_module snd_soc_rt1308_sdw
+insert_module snd_soc_rt715
+insert_module snd_soc_rt711_sdca
+insert_module snd_soc_rt1316_sdw
+insert_module snd_soc_rt715_sdca
+insert_module snd_soc_sdw_mockup
+
+insert_module snd_soc_rt1011
 insert_module snd_soc_rt5640
 insert_module snd_soc_rt5645
 insert_module snd_soc_rt5651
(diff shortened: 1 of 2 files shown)


tools/kmod/sof_insert.sh
```diff
@@ -47,20 +62,12 @@ insert_module snd_soc_rt5677_spi
...
 insert_module snd_soc_max98373_sdw
-
-insert_module snd_soc_rt700
-insert_module snd_soc_rt711
-insert_module snd_soc_rt1308
-insert_module snd_soc_rt1308_sdw
-insert_module snd_soc_rt715
-insert_module snd_soc_rt1011
+insert_module snd_soc_max98373_i2c
+insert_module snd_soc_max98390
```
marc-hb, line 70: Missing snd_soc_max98373?

marc-hb (commented):
How come we never needed to insert `snd_soc_rt711_sdca` and one or two others? Maybe for products we don't test routinely?


> so that it uses the same order as for remove 

In the future, would it make sense to have that both scripts iterate on the same array?

I thought the golden rule was to insert and remove in reverse order but apparently that apparently does not matter for codecs, correct? (it would still be possible to iterate two ways on the same array)

tools/kmod/sof_insert.sh
plbossart: it's added by snd_soc_max98373_i2c.

plbossart:
> How come we never needed to insert `snd_soc_rt711_sdca` and one or two others? Maybe for products we don't test routinely?

Not sure. maybe they are automagically re-added by modprobe. It think if we want to be consistent we should have them here.

> 
> > so that it uses the same order as for remove
> 
> In the future, would it make sense to have that both scripts iterate on the same array?

it would be hard to capture the case of common modules that don't need to be explicitly loaded, but do need to be explicitly removed.

> I thought the golden rule was to insert and remove in reverse order but that apparently does not matter for codecs, correct? (it would still be possible to iterate two ways on the same array)

the codecs are independent to a large extent - well except when there are common modules and e.g. 
snd_soc_max98373_i2c loads snd_soc_max98373

marc-hb (approved):
The NTP sync issue https://sof-ci.01.org/softestpr/PR836/build964/devicetest/?model=TGLU_RVP_SDW&testcase=verify-kernel-boot-log was just a network glitch, I checked the logs. Everything else is green.

> it would be hard to capture the case of common modules that don't need to be explicitly loaded, but do need to be explicitly removed.

My assumption was that only "leaves" would be in the array, common modules would not be and would be removed after all leaves. Could that work? It would also make it much clearer which are  shared versus not.

aiChaoSONG (approved):

plbossart:
merging so that it's taken in the daily tests today.

Issue #780, opened 2021-10-05
[BUG] envcheck.sh doesn't detect octave packages on Fedora 34


plbossart:
**Describe the bug**

root@fedora sof-test]# dnf install octave-signal
Last metadata expiration check: 2:14:43 ago on Tue 05 Oct 2021 03:36:58 PM EDT.
Package octave-signal-1.4.1-8.fc34.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[root@fedora sof-test]# dnf install octave-io
Last metadata expiration check: 2:14:49 ago on Tue 05 Oct 2021 03:36:58 PM EDT.
Package octave-io-2.6.3-2.fc34.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[root@fedora sof-test]# ./env-check.sh 
Checking for some OS packages:		Warning
	Executable  sof-ctl  not found, please put sof-ctl in PATH
Optional: Octave packages are required for check-volume-levels.sh
	Please install the  octave-signal  package
	Please install the  octave-io  package

The package detection doesn't seem to work too well?

plbossart:
it's not a package name, it's the detection that fails

```
[root@fedora sof-test]# dnf install octave octave-signal
Last metadata expiration check: 2:48:31 ago on Thu 27 Jan 2022 01:29:37 PM PST.
Package octave-6:6.4.0-1.fc35.x86_64 is already installed.
Package octave-signal-1.4.1-10.fc35.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[root@fedora sof-test]# dnf install octave octave-io
Last metadata expiration check: 2:48:35 ago on Thu 27 Jan 2022 01:29:37 PM PST.
Package octave-6:6.4.0-1.fc35.x86_64 is already installed.
Package octave-io-2.6.3-4.fc35.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[root@fedora sof-test]# ./env-check.sh 
Checking for some OS packages:		pass
Optional: Octave packages are required for check-volume-levels.sh
	Please install the  octave-signal  package
	Please install the  octave-io  package
````

marc-hb:
Oops, thanks for double checking. In fact the octave did not work on Ubuntu either!! Looks like it was never tested. Ubuntu fix submitted in #846. For Fedora, a disclaimer is printed instead.

marc-hb:
[...]

PR #846, opened 2022-01-28
env-check.sh: fix check for octave packages 

marc-hb:
2 commits. Main one:

Fixes (untested?) commit f9587e0 ("env-check: display optional
octave package dependency") that added the octave checks and misused
the (misnamed) func_check_pkg() function. That function works only for
packages with an executable of the same name.

Add new func_check_dpkg() that runs a dpkg check instead.

Fixes: #780

### DIFF
--- a/env-check.sh
+++ b/env-check.sh
@@ -4,6 +4,16 @@ set -e
 
 mydir=$(cd "$(dirname "$0")"; pwd)
 
+ubuntu_warning()
+{
+cat <<EOF
+
+    Disclaimer: this script is tested only on a somewhat recent Ubuntu version.
+    On other distributions your mileage may vary.
+
+EOF
+}
+
 # enable dynamic debug logs for SOF modules
 DYNDBG="/etc/modprobe.d/sof-dyndbg.conf"
 
@@ -44,6 +54,16 @@ func_check_exec_binary() {
         check_res=1
     fi
 }
+
+main()
+{
+    ubuntu_warning
+}
+
+main "$@"
+
+# All code should be moved to a function eventually:
+# https://github.com/thesofproject/sof-test/issues/740
 
 out_str="" check_res=0
 printf "Checking for some OS packages:\t\t"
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR846/build968/devicetest has some NTP failures + rtcwake timeout. This script changed here is not run in automation, totally unrelated.

env-check.sh
```diff
@@ -92,9 +106,9 @@ fi
 # octave packages are required only for check-volume-levels.sh
 # Good to check upfront but this can be optional requirement
 out_str="" check_res=0
-func_check_pkg octave
-func_check_pkg octave-signal
-func_check_pkg octave-io
+func_check_dpkg octave
+func_check_dpkg octave-signal
+func_check_dpkg octave-io
```
plbossart, line 111: can we add something for Fedora? A number of us use it since it 'just works' on newer devices, and it's also in production for Lenovo devices.

marc-hb: We can make this script smarter but is it really worth our time? Right now on Fedora it says "cannot check for octave-io", that's a clear enough message. If we start supporting Fedora here then we need to validate the rest of the script on Fedora too.

marc-hb: Whatever happens with Fedora, this PR fixes the check for Ubuntu. It's broken everywhere right now.

env-check.sh
```diff
@@ -4,10 +4,20 @@ set -e
 
 mydir=$(cd "$(dirname "$0")"; pwd)
 
+ubuntu_warning()
+{
+cat <<EOF
```
greg-intel, line 9: Just curious, why the preference of HEREDOC over a one-line `echo -e`?

```bash
echo -e "\n\n\tDisclaimer: this script is tested only on a somewhat recent Ubuntu version.\n\tOn other distributions your mileage may vary.\n\n"
```
Legibility?

greg-intel (approved):

env-check.sh
marc-hb: Because it's two lines long? :-)

Thanks for the review.

PR #848, opened 2022-02-02
run-test-all.sh: cleanup for IPC4

marc-hb:
2 commits

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -187,6 +187,12 @@ test_pcm_list()
 }
 test_sof-logger()
 {
+	( set +x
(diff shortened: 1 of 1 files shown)


test-case/run-all-tests.sh
```diff
@@ -21,8 +21,6 @@ very_short_tests()
 
 
 testlist="
-firmware-presence
-firmware-load
```
plbossart, line 25: sounds good for now, we are missing some parts in the kernel/firmware interaction. It's ok to remove this firmware-load for now and re-add it later when the IPC4 support actually enables it.

plbossart (approved):
thanks @marc-hb

marc-hb:
Small change easy to revert if any problem: merging.

PR #850, opened 2022-02-04
sof-kernel-log-check: ignore i2c_hid_acpi i2c-GDIX0000:00 errors

marc-hb:
Ignore 'failed to change power setting` and other boot time
failures. See:
https://sof-ci.sh.intel.com/#/result/planresultdetail/9950?model=TGLU_VOLT_SDW&testcase=verify-kernel-boot-log

Note the failure does not seem fatal, `echo i2c-GDIX0000:00 >
/sys/bus/i2c/drivers/i2c_hid_acpi/bind` still seems to work on the
device reporting the error.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #851, opened 2022-02-04
sof-kernel-log-check: pure whitespace fixes

marc-hb:
As requested in #850.

We unfortunately use a mix of tabs and spaces. Let's try to keep each
file consistent with at least itself.

I have a shortcut in my editor to switch, afraid everyone should have
one too.


greg-intel (approved):
Yep, I will not be using that editor again.  I was typing spaces, but I must have missed the "betray your whitespace desires" step.

marc-hb:
Same, unrelated failures as in #850. Will investigate.

PR #852, opened 2022-02-09
sof_insert: load snd-sof-pci/acpi _before_ they're loaded automatically

marc-hb:
Loading modules that are already loaded is pointless.

Now it is possible to pass module parameters.

### DIFF
--- a/tools/kmod/sof_insert.sh
+++ b/tools/kmod/sof_insert.sh
@@ -22,7 +22,10 @@ insert_module() {
 # Test sudo first, not after dozens of SKIP
 sudo true
 
-
+# Modules parameters can be passed here as a quick and purely local
+# hack.  Be conscious of unexpected interactions with "official"
+# parameters located in /etc/modprobe.d/*.conf
+#
 # This is not required unless you want to change the default flags.
 # For the list of flags search:   git -C linux grep SOF_DBG_
 # Warning: the DMA trace can be forced ON in Kconfig and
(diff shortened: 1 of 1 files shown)


tools/kmod/sof_insert.sh
```diff
@@ -72,17 +72,19 @@ insert_module snd_soc_max98390
 # insert top-level ACPI/PCI SOF drivers. They will register SOF components and
 # load machine drivers as needed. Do not insert any other sort of audio module,
 # code dependencies will be used to load the relevant modules.
+insert_module snd_sof_acpi
+
 insert_module snd_sof_acpi_intel_byt
 insert_module snd_sof_acpi_intel_bdw
 
+insert_module snd_sof_pci # ipc_type=1 sof_pci_debug=1 ...
```
plbossart, line 80: we already pass parameters with options snd_sof_pci ipc_type=1 in a /etc/modprobe.d/something.conf

That could conflict or become inconsistent. I am not sure we want to add module parameters here, we only wanted to manage the dependencies.

To your point though it makes no sense to add them last, I think this is a left-over from earlier versions before we flipped the dependencies.

marc-hb: This is only a quick hack useful locally, it's never meant to be committed and used in production. I will add a comment to clarify this.

It's useful locally because it shows up in "git diff" whereas changes in /etc/ are usually forgotten.

We already had this discussion in #734 :-)

marc-hb:
https://sof-ci.01.org/softestpr/PR852/build972/devicetest/?model=APL_UP2_NOCODEC_ZEPHYR&testcase=multiple-pipeline-playback looks very much like crash thesofproject/sof/issues/4414 and the failures happening right after it seem to be the fallout.

plbossart (approved):

PR #853, opened 2022-02-09
verify-firmware-load: add SOF_VERSION_CHECK variable

plbossart:
when using SOF_VERSION_CHECK=none, the verify-sof-version-check.sh
script is skipped.

This enables us to pass the kmod load/unload tests with IPC4, e.g with

SOF_VERSION_CHECK=none SOF_LOGGING=none TPLG=/lib/firmware/intel/avs-tplg/sof-hda-generic.tplg ~/sof-test/test-case/check-kmod-load-unload-after-playback.sh

### DIFF
--- a/test-case/check-kmod-load-unload.sh
+++ b/test-case/check-kmod-load-unload.sh
@@ -90,12 +90,19 @@ do
     sof-kernel-log-check.sh "$KERNEL_CHECKPOINT" ||
         die "Found error(s) in kernel log after module insertion"
 
-    dlogi "checking if firmware is loaded successfully"
-    "$(dirname "${BASH_SOURCE[0]}")"/verify-sof-firmware-load.sh ||
-         die "Failed to load firmware after module insertion"
+    ( set +x
+      if [ "$SOF_VERSION_CHECK" = 'none' ]; then
+	  printf '$''SOF_VERSION_CHECK=none, skipping verify-sof-firmware-load.sh\n'
+	  dlogi "==== firmware boot check skipped: $idx of $loop_cnt ===="
+      else
+	  dlogi "checking if firmware is loaded successfully"
+	  "$(dirname "${BASH_SOURCE[0]}")"/verify-sof-firmware-load.sh ||
+              die "Failed to load firmware after module insertion"
+	  dlogi "==== firmware boot complete: $idx of $loop_cnt ===="
+      fi
+    )
 
     # successful remove/insert module pass
-    dlogi "==== firmware boot complete: $idx of $loop_cnt ===="
 
     # After the last module insertion, it still takes about 10s for 'aplay -l' to show device
     # list. We need to wait before aplay can function. Here, wait dsp status to suspend to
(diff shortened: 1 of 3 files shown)


test-case/verify-sof-firmware-load.sh
```diff
@@ -22,6 +22,13 @@ func_opt_parse_option "$@"
 
 setup_kernel_check_point
 
+( set +x
```
marc-hb, line 25: Left over debug code?

test-case/verify-sof-firmware-load.sh
```diff
@@ -22,6 +22,13 @@ func_opt_parse_option "$@"
...
+    if [ "$SOF_VERSION_CHECK" = 'none' ]; then
+        printf '$''SOF_VERSION_CHECK=none, skipping verify-sof-firmware-load.sh\n'
+        exit 2
```
marc-hb, line 28: There is a much more readable `skip_test` function now.

test-case/check-kmod-load-unload.sh
```diff
@@ -90,12 +90,19 @@ do
     sof-kernel-log-check.sh "$KERNEL_CHECKPOINT" ||
         die "Found error(s) in kernel log after module insertion"
 
-    dlogi "checking if firmware is loaded successfully"
-    "$(dirname "${BASH_SOURCE[0]}")"/verify-sof-firmware-load.sh ||
-         die "Failed to load firmware after module insertion"
+    ( set +x
```
marc-hb, line 93: Left over debug code?

test-case/check-kmod-load-unload.sh
```diff
@@ -90,12 +90,19 @@ do
...
+      if [ "$SOF_VERSION_CHECK" = 'none' ]; then
+	  printf '$''SOF_VERSION_CHECK=none, skipping verify-sof-firmware-load.sh\n'
+	  dlogi "==== firmware boot check skipped: $idx of $loop_cnt ===="
```
marc-hb, line 96: For consistency:
```suggestion
	  dlogi "==== firmware load check skipped: $idx of $loop_cnt ===="
```

marc-hb:
Rather than adding yet another variable I'd rather make this test simply pass with IPC4. This test was buggy anyway!!! See fix submitted in #855

test-case/verify-sof-firmware-load.sh
plbossart: i copy-pasted it from your SOF_LOGGING work :-)

I have no idea what this does.

test-case/check-kmod-load-unload.sh
plbossart: same, you tell me...

test-case/verify-sof-firmware-load.sh
plbossart: Is this what you had in mind?
````
if [ "$SOF_VERSION_CHECK" = 'none' ]; then
        skip_test ''SOF_VERSION_CHECK=none"
fi
````

marc-hb: Yes this should work and be equivalent.

Issue #842, opened 2022-01-28
[FEATURE] IPC4: verify-sof-firmware-load.sh fails


plbossart:
This test is unable to find out if the firmware was successfully loaded with hard-coded dependencies on 'sof

````starting test_firmware-load 
+ test_firmware-load
+ /root/sof-test/test-case/verify-sof-firmware-load.sh
2022-01-27 23:55:28 UTC Sub-Test: [INFO] Checking SOF Firmware load info in kernel log
[   61.608284] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ffffff10 00000001 00000000 00001002
[   61.608287] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000 00000000 00000180 00000001
[   61.608289] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000
[   61.608292] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc4 create module 6 - instance 0 - pipe 1 - core 0 - domain 0 - param size 64
[   61.608788] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000 00000180 00000180 00000000
[   61.608795] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 0000bb80 00000010 ffffff10 00000001
[   61.608799] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000 00001002 ffffffff 7fffffff
[   61.608802] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000 00000000 0000000a 00000000
[   61.608806] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc4 create module 4 - instance 1 - pipe 1 - core 0 - domain 0 - param size 84
[   61.609439] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 0000066f 000000c0 000000c0 00000001
[   61.609446] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 0000bb80 00000010 ffffff10 00000001
[   61.609450] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 00000000 00001002 0000bb80 00000010
[   61.609453] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ffffff10 00000001 00000000 00001002
[...]

Issue #845, opened 2022-01-28
[FEATURE] IPC4: check-kmod-load-unload.sh fails due to sof-logger dependencies


plbossart:
root@fedora ~]# SOF_LOGGING=none TPLG=/lib/firmware/intel/avs-tplg/sof-hda-generic.tplg /root/sof-test/test-case/check-kmod-load-unload.sh -l 1

````
2022-01-28 00:47:42 UTC [INFO] ===== Starting iteration 1 of 1 =====
2022-01-28 00:47:42 UTC [INFO] wait dsp power status to become suspended
2022-01-28 00:47:42 UTC [INFO] run kmod/sof-kmod-remove.sh

WARNING: running as root is not supported

RMMOD	snd_usb_audio
SKIP	snd_hda_intel  	not loaded
RMMOD	snd_sof_pci_intel_tng
SKIP	snd_sof_pci_intel_skl  	not loaded
RMMOD	snd_sof_pci_intel_apl
RMMOD	snd_sof_pci_intel_cnl
RMMOD	snd_sof_pci_intel_icl
RMMOD	snd_sof_pci_intel_tgl
RMMOD	snd_sof_acpi_intel_byt
RMMOD	snd_sof_acpi_intel_bdw
RMMOD	snd_sof_intel_hda_common
RMMOD	snd_sof_acpi
RMMOD	snd_sof_pci
RMMOD	snd_sof_intel_atom
SKIP	snd_soc_catpt  	not loaded
SKIP	snd_intel_sst_acpi  	not loaded
SKIP	snd_intel_sst_core  	not loaded
SKIP	snd_soc_sst_atom_hifi2_platform  	not loaded
SKIP	snd_soc_skl  	not loaded
SKIP	snd_sof_intel_byt  	not loaded
SKIP	snd_sof_intel_bdw  	not loaded
SKIP	snd_sof_intel_hda_common  	not loaded
RMMOD	snd_sof_intel_hda
SKIP	snd_sof_intel_ipc  	not loaded
RMMOD	snd_sof_xtensa_dsp
RMMOD	snd_soc_acpi_intel_match
SKIP	snd_soc_sof_rt5682  	not loaded
SKIP	snd_soc_sof_da7219_max98373  	not loaded
SKIP	snd_soc_sst_bdw_rt5677_mach  	not loaded
SKIP	snd_soc_sst_broadwell  	not loaded
SKIP	snd_soc_sst_bxt_da7219_max98357a  	not loaded
[...]

PR #855, opened 2022-02-09
verify-sof-firmware-load: fix test to actually check the FW is loaded

marc-hb:
Searching the logs for the firmware version was stupid because old logs
were enough to make the test pass!

Check whether `/sys/kernel/debug/sof/fw_version` exists instead.

This also makes the test pass with IPC4 and fixes https://github.com/thesofproject/sof-test/issues/842 and https://github.com/thesofproject/sof-test/issues/845

### DIFF
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -20,6 +20,9 @@ source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
 func_opt_parse_option "$@"
 
+# We don't use KERNEL_CHECKPOINT in this test. FIXME: (re)move this line
+# after checking how this line interferes with other tests invoking us
+# (e.g.: load-unload test)
 setup_kernel_check_point
 
 cmd="journalctl_cmd"
--- a/test-case/verify-sof-firmware-load.sh
+++ b/test-case/verify-sof-firmware-load.sh
@@ -5,11 +5,11 @@ set -e
 ##
 ## Case Name: verify-sof-firmware-load
 ## Description:
-##    Check if the SOF fw loaded successfully in the kernel logs
-## Case step:
-##    Check kernel logs to search fw load info
+##    Check if /sys/kernel/debug/sof/fw_version exists
 ## Expect result:
-##    Get fw version info in dmesg
+##    /sys/kernel/debug/sof/fw_version exists
+##
+##    When found, also show the firmware version(s) found in the kernel logs
 ##    sof-audio-pci 0000:00:0e.0: Firmware info: version 1:1:0-e5fe2
 ##    sof-audio-pci 0000:00:0e.0: Firmware: ABI 3:11:0 Kernel ABI 3:11:0
 ##
@@ -27,18 +27,29 @@ setup_kernel_check_point
 
 cmd="journalctl_cmd"
 
-dlogi "Checking SOF Firmware load info in kernel log"
+# TODO: move this to main()
+dlogi "Showing all SOF Firmware version(s) in kernel log (mind the timestamps!)"
 if $cmd | grep -q " sof-audio.*Firmware.*version"; then
     # dump the version info and ABI info
     $cmd | grep "Firmware info" -A1 | head -n 12
     # dump the debug info
     $cmd | grep "Firmware debug build" -A3 | head -n 12
-    exit 0
+fi
 
-else # failed, show some logs
+main()
+{
+    local fw_version=/sys/kernel/debug/sof/fw_version
 
+    if sudo test -e "$fw_version"; then
+        printf '  ------ %s found  ----\n' "$fw_version"
+        exit 0
+    fi
+
+    # failed, show some logs
     journalctl_cmd --lines 50 || true
     printf ' ------\n  Check journalctl status: \n ---- \n'
     systemctl --no-pager status systemd-journald* || true
-    die "Cannot find the sof audio version"
-fi
+    die "Cannot find ${fw_version}"
+}
+
+main "$@"


plbossart:
````
[root@fedora ~]# SOF_VERSION_CHECK=none SOF_LOGGING=none TPLG=/lib/firmware/intel/avs-tplg/sof-hda-generic.tplg ~/sof-test/test-case/verify-sof-firmware-load.sh 
2022-02-10 00:33:03 UTC [INFO] Showing all SOF Firmware version(s) in kernel log (mind the timestamps!)
  ------ /sys/kernel/debug/sof/fw_version found  ----
2022-02-10 00:33:03 UTC [INFO] Test Result: PASS!
[root@fedora ~]# more /sys/kernel/debug/sof/fw_version

````

doesn't sound very good to me, what does this test in practice? an empty firmware version is a success?

**Comment by marc-hb** [comment]
> an empty firmware version is a success?

The test is named "verify-sof-firmware-load" so yes that looks like a success to me, why not?

Certainly more appropriate than passing based on old and irrelevant logs...

**Comment by plbossart** [comment]
> > an empty /sys/kernel/debug/sof/fw_version version is a success?
> 
> The test is named "verify-sof-firmware-load" so yes that looks like a success to me, why not?
> 
> Certainly more appropriate than passing based on old and irrelevant logs...
> 
> Any better alternative to suggest?

if you want to check that the firmware booted, then we should look for specific strings that tell us just that. Or indirectly find a valid version number. An empty version number if not a good sign.

This is what you want to check for:
```
[    3.279704] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: firmware boot complete
[    3.279705] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: fw_state change: 5 -> 6
````

**Comment by marc-hb** [comment]
Checking the logs does not really work because it involves time. This test currently passes after unloading the drivers, this is ridiculous.

Is there really nothing equivalent in `/sys/kernel/debug/sof` ? I could look at several files, that would still be much easier.

**Comment by marc-hb** [comment]
> An empty version number if not a good sign.

This may mean that the firmware is not a good state (which will be detected by further tests) but can this ever be a sign that the firmware is _not loaded_?

I've been using an sof-logger patch (thesofproject/sof/pull/5027) for a few months locally. This patch makes the sof-logger wait for `/sys/kernel/debug/sof/fw_version` before starting. I've experienced a number of FW boot failures and I've never seen sof-logger start in that case.

**Comment by marc-hb** [comment]
How about `aplay -l`?

**Comment by libinyang** [comment]
> How about `aplay -l` or /proc/asounds/cards?

`aplay -l` and /proc/asound/cards only tell whether audio is probed and pcm is created successfully or not. If `aplay -l` passed it means fw is loaded OK. But if it fails, it doesn't mean FW is loaded unsuccessfully. There are some other reasons to make the probe fail.

**Comment by libinyang** [comment]
> > an empty /sys/kernel/debug/sof/fw_version version is a success?
> 
> The test is named "verify-sof-firmware-load" so yes that looks like a success to me, why not?
> 
> Certainly more appropriate than passing based on old and irrelevant logs...
> 
> Any better alternative to suggest?

If we want to check the fw_version, it will meet the same issue as aplay. If driver loads unsuccessfully in the main thread, the fw_version entry will be removed. In this situation, you can't tell FW is loaded successfully or not. 
You can try below patch to have a test:
```
diff --git a/sound/soc/sof/core.c b/sound/soc/sof/core.c
index 5c5d97df0d99..7572bea1b1ae 100644
--- a/sound/soc/sof/core.c
+++ b/sound/soc/sof/core.c
@@ -302,6 +302,8 @@ static int sof_probe_continue(struct snd_sof_dev *sdev)
                        "error: failed to register machine driver %d\n", ret);
                goto fw_trace_err;
        }
+       dev_err(sdev->dev, "in %s %d fake fail\n", __func__, __LINE__);
+       goto sof_machine_err;

        ret = sof_register_clients(sdev);
        if (ret < 0) {
```
You can find "firmware boot complete" is showed in `dmesg`, which means fw is loaded successfully. But cat fw_version will fail.
So I think @plbossart 's suggestion is the only correct solution.

plbossart:
> Checking the logs does not really work because it involves time. This test currently passes after unloading the drivers, this is ridiculous.
> 
> Is there really nothing equivalent in `/sys/kernel/debug/sof` ? I could look at several files, that would still be much easier.

@marc-hb not able to understand why a log isn't good. that's what we use to report errors, don't we? if you want to double check that a driver is bound, you can add an additional check. I think it's feasible with sysfs or lspci, but that's different to checking if the firmware booted.

plbossart:
> > Why can't we also do this on the initial test?
> 
> We can but it's a different test.
> 
> So, new proposal:
> 
> * Rename `verify-sof-firmware-load` to `verify-fw-loaded-at-least-once`. So this test stops lying about what it does.
> * Fix the unload-reload test so it does NOT use `verify-fw-loaded-at-least-once`. Thanks to the new and correct name the bug is now very obvious. Make the unload-reload test use log kernel checkpoints instead (like other tests)
> * In all cases look for `firmware boot complete` instead of firmware versions to support IPC4.
> * Internally, move common code to `case-lib/` to avoid duplication.

sounds good to me @marc-hb. 

I am not sure if 'at least once' is the right name, you can't load the same firmware twice but that's fine. Maybe 'first boot' or something.

marc-hb:
> So, new proposal:

Submitted in #860 which superseded this. I'm keeping this open for now as reminder because there may still be some stuff to salvage from this PR.

> I am not sure if 'at least once' is the right name, you can't load the same firmware twice but that's fine. Maybe 'first boot' or something.

I haven't included any rename in #860 so it can be reviewed and merged faster. Will do later.

marc-hb:
> This is what you want to check for:
> `[    3.279704] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: firmware boot complete`

I just noticed by chance that even when running `verify-sof-firmware-load` as the very first test this line can appear _multiple_ times. Because of D3.

Should not be a problem.

marc-hb:
> Submitted in https://github.com/thesofproject/sof-test/pull/860 which superseded this. I'm keeping this open for now as reminder because there may still be some stuff to salvage from this PR.

The rest (and a bit more) submitted in
- #873

PR #858, opened 2022-02-14
volume-basic-test: Change the control search expression

ranj063:
IPC4 uses "gain" as the widget name instead of "PGA". So searching for
"PGA" would result in the test getting skipped with IPC4. But both IPC3
and IPC4 have the pga widget names suffixed with
".Pipeline_ID.Instance_ID". So use the regular expression to search for
volume controls instead of searching for "PGA".

Signed-off-by: Ranjani Sridharan <ranjani.sridharan@linux.intel.com>

### DIFF
--- a/test-case/volume-basic-test.sh
+++ b/test-case/volume-basic-test.sh
@@ -58,7 +58,7 @@ sleep 1
 [[ ! $(pidof aplay) ]] && die "$pid process is terminated too early"
 
 sofcard=${SOFCARD:-0}
-pgalist=($(amixer -c"$sofcard" controls | grep -i PGA | sed 's/ /_/g;' | awk -Fname= '{print $2}'))
+pgalist=($(amixer -c"$sofcard" controls | grep "[0-9]\.[0-9]" | sed 's/ /_/g;' | awk -Fname= '{print $2}'))
 dlogi "pgalist number = ${#pgalist[@]}"
 [[ ${#pgalist[@]} -ne 0 ]] || skip_test "No PGA control is available"
 


test-case/volume-basic-test.sh
```diff
@@ -58,7 +58,7 @@ sleep 1
 [[ ! $(pidof aplay) ]] && die "$pid process is terminated too early"
 
 sofcard=${SOFCARD:-0}
-pgalist=($(amixer -c"$sofcard" controls | grep -i PGA | sed 's/ /_/g;' | awk -Fname= '{print $2}'))
+pgalist=($(amixer -c"$sofcard" controls | grep "[0-9]\.[0-9]" | sed 's/ /_/g;' | awk -Fname= '{print $2}'))
```
aiChaoSONG, line 61: I am not sure if we have other controls takes this form in IPC4, if no, it is good.
another possible fix can can be `grep -E PGA|gain`

aiChaoSONG (approved):

test-case/volume-basic-test.sh
marc-hb: BTW this code was overly complicated. It can be simplified like this:
```shell
pgalist=($(amixer -c"$sofcard" controls | awk -Fname= '/PGA/ { print $2 }')
```
With `-Fname` there's no reason to convert spaces back and forth.

marc-hb:
Is this required to test IPC4?

aiChaoSONG:
> Is this required to test IPC4?

It is required to test cAVS, because the volume control in cAVS is named as `gain`

ranj063:
im going to close this one for now. this string will include all the other controls as well.

PR #862, opened 2022-02-18
check-kmod-load-unload: wait 5 seconds for firmware to bootNew function poll_wait_for(). 

marc-hb:
check-kmod-load-unload: wait 5 seconds for firmware to boot

Now that the test has been just fixed in commit 1a9ccac4ecfa83b1a601fc, of course
it started failing:

https://sof-ci.01.org/sofpr/PR5383/build12069/devicetest/?model=BYT_MB_NOCODEC&testcase=check-kmod-load-unload-after-playback

This proves the older version wasn't testing anything: classic case of
"green failure"
https://github.com/thesofproject/sof-test/issues?q=+label%3A%22False+Pass+%2F+green+failure%22+

Firmware does not boot instantly. We even have retries:

https://github.com/thesofproject/linux/pull/3255
https://github.com/thesofproject/linux/commit/2a3c9269474c

So, poll journalctl for a maximum of 5 seconds before failing.

  -------

lib.sh: add new poll_wait_for() function

This will immediately be used for waiting until the firmware is loaded
but is generic and will be useful in other places.


marc-hb:
suspend/resume timeout in https://sof-ci.01.org/softestpr/PR862/build983/devicetest/?model=APL_UP2_NOCODEC_ZEPHYR&testcase=check-suspend-resume-with-capture-5, everything else is green

marc-hb:
@plbossart , @ujfalusi , what is the maximum time we should wait between modprobe completion and 'firmware boot complete' message?

plbossart:
> @plbossart , @ujfalusi , what is the maximum time we should wait between modprobe completion and 'firmware boot complete' message?

modprobe completion means we will execute the driver probe, but for SOF it's handled in a workqueue. In addition we will do a request_module() for HDaudio stuff, and i915 IIRC, so in theory it could take a while. I think in those cases it's not uncommon to wait 60s.

The firmware download itself should not take longer than 500ms, even with the worst sand-bagging in the world and retries.

plbossart:
@marc-hb 60s is just one way of expressing a dependency on things that take time - but we don't know how much.

Make that 10-15s for test scripts and that should be good enough

marc-hb:
SOFCI TEST

case-lib/lib.sh
```diff
@@ -40,6 +40,41 @@ if [ ! "$SOFCARD" ]; then
...
+# Arguments:
+#
+#   - poll interval in secs
+#   - timeout in secs, rounded up to the next interval
+#   - command and arguments
+#
+poll_wait_for()
+{
+    test $# -ge 3 ||
+        die "poll_wait_for() invoked with $# arguments"
```
aiChaoSONG, line 52: if the user use it wrong, maybe we should inform the correct usage?

marc-hb: We could but this function is not meant to be used outside sof-test, so the only user using it wrong is either you or me.

More seriously, the only user using it wrong is someone writing sof-test code who can just open the file and look at the header above = what they should have done already before even starting to use the function.

greg-intel (approved):

aiChaoSONG (approved):

PR #863, opened 2022-02-19
multiple-pipeline.sh: temporarily disable capture on BDW 

marc-hb:
Known issue https://github.com/thesofproject/sof/issues/3170 has been
polluting the test results for years.

Signed-off-by: Marc Herbert <marc.herbert@intel.com

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -142,7 +142,18 @@ ps_checks()
         func_error_exit "Caught abnormal process status of aplay"
 }
 
+main()
+{
+    local platf; platf=$("${TESTDIR}"/tools/sof-dump-status.py --platform)
+    if [ "$platf" = bdw ] && [ "$f_arg" != 'p' ] && ! is_zephyr; then
+        skip_test \
+            "multi-capture disabled on BDW https://github.com/thesofproject/sof/issues/3170"
+    fi
+}
 
+main "$@"
+
+# TODO: move this to main() https://github.com/thesofproject/sof-test/issues/740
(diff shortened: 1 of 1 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR863/build986/devicetest/?model=BDW_WSB_RT286&testcase=multiple-pipeline-capture is skipped as expected. The other failures are obviously unrelated.

mengdonglin (commented):
@marc-hb Can we also disable BYT and CHT? All they're legacy platforms of low priority.

keqiaozhang:
Issue 3170 mainly happens on BDW_WSB_RT286 when testing "multiple-pipeline-capture", but we also observed this issue on BYT_MB_NOCODEC when testing "simultaneous-playback-capture", but the reproduction rate is relativity lower than BDW.
Test ID:9950/9924 ,etc.

marc-hb:
I would like to start with BDW first. If we see it regularly on BYT we can very easily add it later.

XiaoyunWu6666:
see alternative PR :https://github.com/intel-innersource/drivers.audio.ci.sof-framework/pull/185

marc-hb:
> see alternative PR sof-framework/pull/185

This is too static, it requires to list the test cases one by one. In daily run 10459 only one test was skipped and 3 others failed. So let's skip in sof-test where we have the exact and up to date information at run-time no matter what exact test suite is run and no matter what test runner is used. Morever this skip runs after test setup.

Skipping in sof-framework can be useful sometimes but here it's not.

greg-intel (approved):

PR #865, opened 2022-02-28
sof_remove: remove module snd_soc_intel_sof_realtek_common

XiaoyunWu6666:
snd_soc_intel_sof_realtek_common is missing on remove so that we need to add it in modules removed.

Related test failure : see inner daily testresult 10553 ADLP_GMB_I2S check-kmod-load-unload-25

With this commit , the failure was gone

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -152,6 +152,7 @@ remove_module snd_soc_sof_sdw
 remove_module snd_soc_ehl_rt5660
 remove_module snd_soc_intel_hda_dsp_common
 remove_module snd_soc_intel_sof_maxim_common
+remove_module snd_soc_intel_sof_realtek_common
 
 #-------------------------------------------
 # SOF client drivers


keqiaozhang (approved):

tools/kmod/sof_remove.sh
```diff
@@ -168,6 +168,7 @@ remove_module snd_sof_dma_trace
 # when we stop testing old branches.
 remove_module snd_sof_nocodec || true
 
+remove_module snd_soc_intel_sof_realtek_common
```
plbossart, line 171: good change but in the wrong place, this should be moved to line 155

plbossart (changes requested):

tools/kmod/sof_remove.sh
XiaoyunWu6666: Thanks for reminding! Already fixed it

keqiaozhang (approved):

aiChaoSONG (commented):
Suggestion: better to show the module name in the commit title, ex: `kmod: remove snd_soc_intel_sof_realtek_common module`, `add missing snd_soc_intel_sof_realtek_common module` the `add missing module` doesn't show much useful information, really.

So in the commit title, you should summarize what you do, and provide as much information as possible, and of course, be concise. Normally, there is a verb, and a informative object with descriptive adjectives

marc-hb:
> the "add missing module" doesn't show much useful information, really.

One simple reason why this title does not have enough information: the same title applies to many other previous commits to this file. This proves it is not specific enough. Imagine that you're running `git lol --oneline -- tools/kmod/sof_remove.sh` like this: https://github.com/thesofproject/sof-test/commits/main/tools/kmod/sof_remove.sh

In ideal world, all commit subjects look different so you can immediately find the one you're looking for or spot the one missing. This is not going to work if half the commits have the same title.

Note how most commits in https://github.com/thesofproject/sof-test/commits/main/tools/kmod/sof_remove.sh start with the name of the file. This helps them stand out in https://github.com/thesofproject/sof-test/commits/main/tools

This fairly short page has a lot of great advice: https://wiki.openstack.org/wiki/GitCommitMessages

plbossart (approved):

marc-hb:
I had a look at the errors in https://sof-ci.01.org/softestpr/PR865/build994/devicetest, none is related.

greg-intel:
> This fairly short page has a lot of great advice: https://wiki.openstack.org/wiki/GitCommitMessages

Fred also hit me with https://cbea.ms/git-commit/

marc-hb:
> Fred also hit me with https://cbea.ms/git-commit/

Nice except for "use only the command line". You should obviously use both.

PR #861, opened 2022-02-17
verify-sof-firmware-load: fix error message according to new test

marc-hb:
Fixes commit cd2d7073337332 that now looks for the "firmware boot
complete" message instead of the firmware version(s) but forgot to
update the error message.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #868, opened 2022-03-01
[SKIP CI] verify-sof-firmware-load: fix error message according to new test

marc-hb:
Fixes commit cd2d7073337332 that now looks for the "firmware boot
complete" message instead of the firmware version(s) but forgot to
update the error message.


marc-hb:
Duplicate of #861, closing sorry for the noise.

PR #870, opened 2022-03-01
checkpoints: do not use SOF_TEST_INTERVAL in test _iterations_

marc-hb:
2 commits, the main one:

The purpose of SOF_TEST_INTERVAL is to avoid log gaps between tests,
_not_ between test iterations of the same test.

Fixes issue where the (recent) check-kmod-load-unload.sh was catching
_multiple_ firmware boots, including boots from previous test
iterations, example:
```
sof-test/test-case/check-mod-load-unload.sh -l 50

Polling 'sof_firmware_boot_complete --since=@1646100584' every 1s for 10s
[ 2968.155103] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: firmware boot complete
[ 2970.706560] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: firmware boot complete
[ 2974.522111] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: firmware boot complete
Completed 'sof_firmware_boot_complete --since=@1646100584' after 0s and 1 attempts
```
This may fix other, similar tests too.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -86,6 +86,8 @@ function func_exit_handler()
     fi
 
     if [[ "$KERNEL_CHECKPOINT" =~ ^[0-9]{10} ]]; then
+        # Do not collect the entire duration of the test but only the
+        # last iteration.
         journalctl_cmd --since=@"$KERNEL_CHECKPOINT" > "$LOG_ROOT/dmesg.txt"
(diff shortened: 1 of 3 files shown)


plbossart (approved):
tested on my TGL device, works fine. Thanks @marc-hb

PR #871, opened 2022-03-01
skip the test on Echo Reference pipeline 

keqiaozhang:
Skip the Echo Reference pipeline test in multiple-pause-resume and multiple-pipeline.

### DIFF
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -56,7 +56,7 @@ rnd_range=$((rnd_max - rnd_min))
 [[ $rnd_range -le 0 ]] && dlogw "Error random range scope [ min:$rnd_min - max:$rnd_max ]" && exit 2
 
 tplg=${OPT_VAL['t']}
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ~pcm:Amplifier Reference"
 
 logger_disabled || func_lib_start_log_collect
 
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -84,10 +84,15 @@ APP_LST['capture']='arecord_opts'
 DEV_LST['capture']='/dev/null'
 
 # define for load pipeline
+# args: $1: playback or capture
+#       $2: optional filter
 func_run_pipeline_with_type()
 {
+    local direction="$1"
+    local opt_filter
+    [ -n "$2" ] && opt_filter="& $2"
     [[ $tmp_count -le 0 ]] && return
-    func_pipeline_export "$tplg" "type:$1"
+    func_pipeline_export "$tplg" "type:$direction $opt_filter"
     local -a idx_lst
     if [ ${OPT_VAL['r']} -eq 0 ]; then
         idx_lst=("$(seq 0 $((PIPELINE_COUNT - 1)))")
@@ -105,7 +110,7 @@ func_run_pipeline_with_type()
 
         dlogi "Testing: $pcm [$dev]"
 
-        "${APP_LST[$1]}" -D "$dev" -c "$channel" -r "$rate" -f "$fmt" "${DEV_LST[$1]}" -q &
+        "${APP_LST[$direction]}" -D "$dev" -c "$channel" -r "$rate" -f "$fmt" "${DEV_LST[$direction]}" -q &
 
         : $((tmp_count--))
         if [ "$tmp_count" -le 0 ]; then return 0; fi
@@ -165,11 +170,11 @@ do
         'p' | 'a')
             tmp_count=$max_count
             func_run_pipeline_with_type "playback"
-            func_run_pipeline_with_type "capture"
+            func_run_pipeline_with_type "capture" "~pcm:Amplifier Reference"
             ;;
         'c')
             tmp_count=$max_count
-            func_run_pipeline_with_type "capture"
+            func_run_pipeline_with_type "capture" "~pcm:Amplifier Reference"
             func_run_pipeline_with_type "playback"
             ;;
         *)


test-case/multiple-pause-resume.sh
```diff
@@ -56,7 +56,7 @@ rnd_range=$((rnd_max - rnd_min))
 [[ $rnd_range -le 0 ]] && dlogw "Error random range scope [ min:$rnd_min - max:$rnd_max ]" && exit 2
 
 tplg=${OPT_VAL['t']}
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ~cap_name:Smart Amplifier Playback 2,Echo Reference Capture 3"
```
XiaoyunWu6666, line 59: are we going to skip the combinations of pipeline 2 and other pipelines

XiaoyunWu6666: I am wondering whether it will be better to skip the combination of pipeline 2 and pipeline 3 when constructing pipeline_combine_lst

keqiaozhang: Yes, we can only skip the combination of smart_amp pipeline, but I wonder if such a test would make sense.
eg. check the echo reference capture pipeline with other normal playback pipelines, this is not a reasonable use case.

test-case/multiple-pause-resume.sh
```diff
@@ -56,7 +56,7 @@ rnd_range=$((rnd_max - rnd_min))
 [[ $rnd_range -le 0 ]] && dlogw "Error random range scope [ min:$rnd_min - max:$rnd_max ]" && exit 2
 
 tplg=${OPT_VAL['t']}
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ~smart_amp"
```
plbossart, line 59: if you are going to filter something, it should be the echo reference. That's a clear case of a useless test. pause on playback makes a lot more sense.

plbossart (changes requested):

test-case/multiple-pipeline.sh
```diff
@@ -105,7 +105,7 @@ func_run_pipeline_with_type()
 
         dlogi "Testing: $pcm [$dev]"
 
-        "${APP_LST[$1]}" -D "$dev" -c "$channel" -r "$rate" -f "$fmt" "${DEV_LST[$1]}" -q &
+        "${APP_LST[${1%% *}]}" -D "$dev" -c "$channel" -r "$rate" -f "$fmt" "${DEV_LST[${1%% *}]}" -q &
```
marc-hb, line 108: Good idea but this is really too cryptic, it took me several minutes to understand what this does. Please use more English like this instead:

```shell
# args: $1: playback or capture
#          $2: optional filter

func_run_pipeline_with_type()
{
     local direction="$1"
     local opt_filter
     if [ -n "$2" ]; then opt_filter="& $2"; fi
     ...
     func_pipeline_export "$tplg" "type:$direction $opt_filter"
     ...
    "${APP_LST[$direction]}" -D "$dev" -c "$channel" -r "$rate" -f "$fmt" "${DEV_LST[$direction]}" -q &
     ...
}


func_run_pipeline_with_type 'capture'  '~pcm:Amplifier Reference'

```

marc-hb (changes requested):

test-case/multiple-pipeline.sh
keqiaozhang: Thanks Marc. it's simple and easy to understand. I have pushed a new version.

test-case/multiple-pause-resume.sh
```diff
@@ -56,7 +56,7 @@ rnd_range=$((rnd_max - rnd_min))
 [[ $rnd_range -le 0 ]] && dlogw "Error random range scope [ min:$rnd_min - max:$rnd_max ]" && exit 2
 
 tplg=${OPT_VAL['t']}
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ~pcm:Amplifier Reference"
```
plbossart, line 59: I would go one step further and remove 'Amplifier'.

We've seen cases where we talk about 'Echo Reference'.

keqiaozhang: @plbossart , so you mean that we should remove both smart amplifier pipeline and echo reference pipeline form the test?

plbossart: "amplifier reference' and 'echo reference' are the same thing, aren't they?

plbossart: I meant use ~pcm:Reference

keqiaozhang: here is the output of these 2 pipeline parsed by sof-tplgreader.py:
```
pcm=Speaker;id=2;type=playback;cap_name=Smart Amplifier Playback 2;fmts=S16_LE S24_LE S32_LE;fmt=S16_LE;rate_min=48000;rate_max=48000;ch_min=2;ch_max=2;rate=48000;channel=2;dev=hw:0,2;snd=/dev/snd/pcmC0D2p;smart_amp=SMART_AMP3.0;
```

```
pcm=Amplifier Reference;id=3;type=capture;cap_name=Echo Reference Capture 3;fmts=S16_LE S24_LE S32_LE;fmt=S16_LE;rate_min=48000;rate_max=48000;ch_min=2;ch_max=2;rate=48000;channel=2;dev=hw:0,3;snd=/dev/snd/pcmC0D3c;smart_amp=SMART_AMP3.0;
```

if we want to filter them all, we can use `~smart_amp`

marc-hb:
Screenshot 2022-03-03 at 20 58 30

keqiaozhang:
I have updated my PR to skip both Smart Amplifier playback pipeline and Echo Reference capture pipeline in the tests.

plbossart:
> I have updated my PR to skip both Smart Amplifier playback pipeline and Echo Reference capture pipeline in the tests.

Have you tried skipping ~Reference only as I mentioned above. I don't think it's necessary to skip all tests.

keqiaozhang:
> > I have updated my PR to skip both Smart Amplifier playback pipeline and Echo Reference capture pipeline in the tests.
> 
> Have you tried skipping ~Reference only as I mentioned above. I don't think it's necessary to skip all tests.

`~pcm:Reference` and `~pcm:Amplifier Reference` are same, these all skip the Echo Reference capture pipeline but keep the Smart Amplifer playback pipeline(speaker) in the tests. if that's want you want,  we can also use `~cap_name:Echo Reference Capture ` to avoid the misunderstanding. So which one do you prefer? The ’sof-tplgreader.py‘ output of  this pipeline:
```
pcm=Amplifier Reference;id=3;type=capture;cap_name=Echo Reference Capture 3;fmts=S16_LE S24_LE S32_LE;fmt=S16_LE;rate_min=48000;rate_max=48000;ch_min=2;ch_max=2;rate=48000;channel=2;dev=hw:0,3;snd=/dev/snd/pcmC0D3c;smart_amp=SMART_AMP3.0;
```

plbossart:
@keqiaozhang I have no preference, as long as you keep testing the playback path I am happy.

keqiaozhang:
> ~pcm:Amplifier Reference

@plbossart , I have updated the PR to skip the echo reference capture pipeline and keep the smart amplifier playback pipeline in the tests.

plbossart (approved):

marc-hb:
This made PIPELINE_COUNT out of sync, fixed submitted by @miroox in #875

PR #872, opened 2022-03-01
multiple-pipeline: skip the Smart Amp playback pipeline when testing

keqiaozhang:
multiple capture pipelines

In this case, if we test all capture pipeline at same time, but
capture pipeline count is less than requested count defined by
-c option, playback pipeline will be started to ensure requested
number of pipelines are running in parallel.

Consider this scenario, the capture pipeline count is less than
requested count and the tplg contains both Smart Amplifier pipeline
and Echo reference capture pipeline. The EC capture pipeline start
first, then smart amp playback pipeline follow, this will cause ipc
tx timed out. This patch skips the smart amp playback pipeline when
testing multiple capture pipelines to avoid such scenario.

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -105,7 +105,7 @@ func_run_pipeline_with_type()
 
         dlogi "Testing: $pcm [$dev]"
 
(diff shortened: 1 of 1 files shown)


marc-hb:
Would it make sense to add this commit to similar PR #871? I mean keep the commits separate if that makes sense but in only one PR.

keqiaozhang:
This PR is merged to https://github.com/thesofproject/sof-test/pull/871.

PR #875, opened 2022-03-09
multiple-pipeline: skip Echo Reference pipeline when getting count

miRoox:
Previously, the Echo Reference pipeline has been skipped when running
the parallel pipeline test, but it wasn't skipped when calculating
PIPELINE_COUNT. Therefore, on all pipelines mode, the expected process
count is inconsistent with the actual running process. To solve this
issue, skip the same pipeline when acquiring the pipeline count for
consistency.

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -67,9 +67,11 @@ tplg=${OPT_VAL['t']}
 f_arg=${OPT_VAL['f']}
 logger_disabled || func_lib_start_log_collect
 
+# skip the Echo Reference pipeline
+MULTI_PIPELINE_FILTER='~pcm:Amplifier Reference'
 max_count=0
 # this line will help to get $PIPELINE_COUNT
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ${MULTI_PIPELINE_FILTER}"
 # acquire pipeline count that will run in parallel, if the number of pipeline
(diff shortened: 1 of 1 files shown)


keqiaozhang (dismissed):
LGTM.
@miRoox ,thanks for the fix, I missed this case.

miRoox:
Tested on ADLP_BRYA_SDW and TGLU_VOLT_SDW, the inner test report 10814 is all green.

test-case/multiple-pipeline.sh
```diff
@@ -69,7 +69,7 @@ logger_disabled || func_lib_start_log_collect
 
 max_count=0
 # this line will help to get $PIPELINE_COUNT
-func_pipeline_export "$tplg" "type:any"
+func_pipeline_export "$tplg" "type:any & ~pcm:Amplifier Reference"
```
marc-hb, line 72: Thanks @miRoox for the quick fix! Can you please go one small step further and remove the duplication to make sure it does not happen again?
```suggestion
MULTI_PIPELINE_FILTER='~pcm:Amplifier Reference'
func_pipeline_export "$tplg" "type:any & ${MULTI_PIPELINE_FILTER}"
```

+ use the same variable everywhere below.

miRoox: done

marc-hb (approved):

keqiaozhang (approved):

PR #877, opened 2022-03-10
 add bpftrace scripts to track performance and behavior of SOF functions in the kernel

nklayman:
Signed-off-by: Noah Klayman <noah.klayman@intel.com>

This PR adds some example [bpftrace](https://github.com/iovisor/bpftrace) scripts to track performance and behavior of various SOF functions in the kernel. To run them, you'll need to [install bpftrace](https://github.com/iovisor/bpftrace/blob/master/INSTALL.md), then run the script with `bpftrace [path to script]`. If you want to use the hda_irq scripts, you'll need to be running [this kernel pr](https://github.com/thesofproject/linux/pull/3482).

Example output from `suspend_resume_time.bt`:
```shell
❯ bpftrace ./kernel_tracing/bpftrace_scripts/suspend_resume_time.bt
Attaching 5 probes...
kprobe:sof_resume (runtime) started
kretprobe:sof_resume (runtime) finished, took 132135788 nsecs
kprobe:sof_suspend (system) started
kretprobe:sof_suspend (system) finished, took 738362 nsecs
kprobe:sof_resume (system) started
kretprobe:sof_resume (system) finished, took 197517950 nsecs
kprobe:sof_suspend (runtime) started
kretprobe:sof_suspend (runtime) finished, took 32807764 nsecs
kprobe:sof_resume (runtime) started
kretprobe:sof_resume (system) finished, took 187189967 nsecs
kprobe:sof_resume (system) started
kretprobe:sof_resume (system) finished, took 180462467 nsecs
kprobe:sof_suspend (runtime) started
kretprobe:sof_suspend (runtime) finished, took 32866837 nsecs
kprobe:sof_resume (runtime) started
kprobe:sof_resume (system) started
kretprobe:sof_resume (system) finished, took 188085072 nsecs
kprobe:sof_resume (runtime) started
kretprobe:sof_resume (system) finished, took 188318490 nsecs
kprobe:sof_suspend (runtime) started
kretprobe:sof_suspend (runtime) finished, took 32849861 nsecs
^C


@usecs[kretprobe:sof_resume, runtime]: 
[128K, 256K)           5 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

@usecs[kretprobe:sof_resume, system]: 
[128K, 256K)           5 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

@usecs[kretprobe:sof_suspend, runtime]: 
[16K, 32K)             2 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                  |
[32K, 64K)             3 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

@usecs[kretprobe:sof_suspend, system]: 
[256, 512)             1 |@@@@@@@@@@@@@                                       |
[512, 1K)              4 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

Additionally, this PR includes a Python script to collect metrics and run conditions on them, see `kernel_tracing/metric_evaluation`. It's intended to be used on the CI machines to prevent performance regressions.

### DIFF
--- a/kernel_tracing/bpftrace_scripts/hda_irq_source_counts.bt
+++ b/kernel_tracing/bpftrace_scripts/hda_irq_source_counts.bt
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+
+// This keeps count of the irqs from each source
+
+kprobe:hda_dsp_interrupt_thread {
+  @total = count();
+}
+
+tracepoint:sof_intel:sof_intel_hda_irq {
+  @counts[str(args->source)] = count();
+}
--- a/kernel_tracing/bpftrace_scripts/suspend_resume_time.bt
+++ b/kernel_tracing/bpftrace_scripts/suspend_resume_time.bt
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+
+
+// This tracks the time it takes for runtime and system suspend/resume
+// It prints out timings as they happen, and will generate a histogram
+// when you stop the script with ctrl-c
+
+// Use the `test-case/check-suspend-resume.sh` script to try this out
+
+kprobe:sof_resume, kprobe:sof_suspend {
+  // When the function enters, we record a timestamp and the type
+  @start = nsecs;
+  @type = arg1 ? "runtime" : "system";
+  printf("%s (%s) started\n", probe, @type);
+}
+
+kretprobe:sof_resume, kretprobe:sof_suspend {
+  // When the function returns, we use the saved timestamp to determine execution time
+  printf("%s (%s) finished, took %d nsecs\n", probe, @type, nsecs - @start);
+  @usecs[probe, @type] = hist((nsecs - @start) / 1000);
+  @avg[probe, @type] = avg((nsecs - @start) / 1000);
+}
+
+END {
+  // Clean up variables that shouldn't be printed
+  delete(@start);
+  delete(@type);
+}
--- a/kernel_tracing/metric_evaluation/conftest.py
+++ b/kernel_tracing/metric_evaluation/conftest.py
@@ -1,0 +1,91 @@
+import argparse
+import os
+import json
+import subprocess
+import re
+
+# This file does all the setup for the "test_bpftrace_conditions" function in test.py
+# It loads the test cases from the json file, and runs the bpftrace script and shell commands
+# It collects the variables from the bpftrace script through stdout,
+# and then uses Pytest Parameterization to run the test function for each condition
+
+
+def pytest_addoption(parser):
+    parser.addoption("--spec_file", type=argparse.FileType('r'),
+                     help="json file with test cases", required=True)
+
+
+def pytest_itemcollected(item):
+    """ Overwrite test name with name defined in json file """
+    item.name = item.name.split('[', 1)[1][:-1]
+    # pylint: disable=protected-access
+    item._nodeid = item.name
+
+
+class BPFTrace:
+    def __init__(self, bpftrace_script):
+        self.script = bpftrace_script
+        self.process = None
+
+    def start(self):
+        self.process = subprocess.Popen(
+            ["bpftrace", self.script], stdout=subprocess.PIPE, encoding="utf-8")
+        # Wait for bpftrace script to load
+        while True:
+            line = self.process.stdout.readline()
+            if "Attaching" in line:
+                break
+
+    def stop(self):
+        # Send ctrl-c
+        self.process.send_signal(subprocess.signal.SIGINT)
+        self.process.wait(10)
+
+        # Collect printed variables for condition evaluation
+        lines = self.process.stdout.read().split("\n")
+        bpftrace_vars = {}
+        for line in lines:
+            # Check if line is outputting a variable from the script
+            # bpftrace output format is `@name[key1, key2, ...]: val` (keys are optional)
+            if re.match("@.*: .*", line):
+                key, val = line.split(": ", 1)
+                bpftrace_vars[key] = val
+        return bpftrace_vars
+
+
+def collect_test_results(test_case, working_dir):
+    """ This runs the bpftrace script and the shell command, and collects the trace output from stdout """
+    # Start bpftrace collection
+    bpftrace_script = os.path.join(working_dir, test_case['bpftrace'])
+    bpf = BPFTrace(bpftrace_script)
+    bpf.start()
+
+    # Execute shell command
+    shell_cmd = test_case['shell']
+    print("$", shell_cmd)
+    subprocess.run(shell_cmd, cwd=working_dir, shell=False, check=True)
+
+    # Stop tracing and return collected output vairables
+    print("BPF tracing finished")
+    bpf_vars = bpf.stop()
+    return bpf_vars
+
+
+def pytest_generate_tests(metafunc):
+    """ Parametrize each JSON test case, once for each of its conditions """
+    if not "bpftrace_condition" in metafunc.fixturenames:
+        raise RuntimeError("Invalid test case.")
+    spec_file = metafunc.config.option.spec_file
+    spec_dir = os.path.dirname(os.path.realpath(spec_file.name))
+    spec = json.load(spec_file)
+    conditions = []
+    # Generate a list of conditions to evaluate
+    for test_case in spec['cases']:
+        bpftrace_vars = collect_test_results(test_case, spec_dir)
+        for condition in test_case['conditions']:
+            conditions.append((test_case['name'], condition, bpftrace_vars))
+
+    # Parameterize the conditions so that the test function gets run for each condition
+    # We also set the ids of the functions to be "name: condition" for better reporting
+    metafunc.parametrize("bpftrace_condition", conditions, ids=map(
+        lambda c: f"{c[0]}: {c[1]}", conditions))
--- a/kernel_tracing/metric_evaluation/no_dependencies.json
+++ b/kernel_tracing/metric_evaluation/no_dependencies.json
@@ -1,0 +1,12 @@
+{
+    "cases": [
+      {
+        "name": "Demo test",
+        "bpftrace": "./simple_bpftrace.bt",
+        "shell": ["sleep", "1"],
+        "conditions": [
+          "@demo_value == 42"
+        ]
+      }
+    ]
+  }
--- a/kernel_tracing/metric_evaluation/readme.md
+++ b/kernel_tracing/metric_evaluation/readme.md
@@ -1,0 +1,34 @@
+# Metric Evaluation
+
+The `test.py` script in this directory executes tests specified in a JSON file.
+The JSON file is formatted as such:
+
+```json
+{
+  "cases": [
+    {
+      "name": "Name to be displayed in output",
+      "bpftrace": "path to the bpftrace script used to collect metrics, relative to spec file",
+      "shell": "shell command to be run alongside bpftrace script, relative to spec file",
+      "conditions": [
+        "list of conditions that will be evaluated against bpftrace output",
+        "these are python expressions",
+        "bpftrace vars can be referenced with @{var name}, ie @avg",
+        "maps are referenced by @{var name}[{key name(s)}], ie @avg[ipc_time]"
+      ]
+    }
+  ]
+}
+```
+See `sample_spec.json` for an example. `no_dependencies.json` is an alternative
+spec file that should be able to run without any special kernel config.
+
+
+To run a test file, pass it as an argument to the `test.py` script:
+
+```bash
+./test.py my_tests.json
+```
+
+If any of the tests fail, the script will exit with code 1. Individual test
+output will also be printed.
--- a/kernel_tracing/metric_evaluation/sample_spec.json
+++ b/kernel_tracing/metric_evaluation/sample_spec.json
@@ -1,0 +1,23 @@
+{
+  "cases": [
+    {
+      "name": "IPC TX Time",
+      "bpftrace": "../bpftrace_scripts/ipc_tx_time.bt",
+      "shell": "./sample_shell.sh",
+      "conditions": [
+        "@avg < 500"
+      ]
+    },
+    {
+      "name": "Suspend/Resume Time",
+      "bpftrace": "../bpftrace_scripts/suspend_resume_time.bt",
+      "shell": "../../test-case/check-suspend-resume.sh",
+      "conditions": [
+        "@avg[kretprobe:sof_resume, runtime] < 150000",
+        "@avg[kretprobe:sof_resume, system] < 200000",
+        "@avg[kretprobe:sof_suspend, runtime] < 50000",
+        "@avg[kretprobe:sof_suspend, system] < 1000"
+      ]
+    }
+  ]
+}
--- a/kernel_tracing/metric_evaluation/simple_bpftrace.bt
+++ b/kernel_tracing/metric_evaluation/simple_bpftrace.bt
@@ -1,0 +1,7 @@
+// This is used to demo the metric evaluation script without requiring any
+// dependencies. Run `./test.py no_dependencies.json` to try it, and change
+// demo_value to see the failure output.
+
+BEGIN {
+    @demo_value = 42;
+}
--- a/kernel_tracing/metric_evaluation/test.py
+++ b/kernel_tracing/metric_evaluation/test.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2022 Intel Corporation. All rights reserved.
+
+# Run with --help for a description of this script
+
+import argparse
+import pytest
+import os
+
+
+def test_bpftrace_conditions(bpftrace_condition):
+    """ This is the test function that is called by pytest """
+    (_name, condition, bpftrace_vars) = bpftrace_condition
+    condition_with_vars = condition
+    for key, val in bpftrace_vars.items():
+        condition_with_vars = condition_with_vars.replace(key, val)
+    has_syntax_error = False
+    try:
+        # Eval does not decrease security as the spec file already embeds shell commands anyway
+        # pylint: disable=eval-used
+        if not eval(condition_with_vars):
+            pytest.fail(
+                f"Failed condition: {condition}. Evaluated to {condition_with_vars}.", pytrace=False)
+    except SyntaxError:
+        has_syntax_error = True
+    # We need pytest.fail to be called outside of the try/except block so the unneeded stack trace is not included
+    if has_syntax_error:
+        pytest.fail(
+            f"Invalid condition: {condition}. Are you sure the bpftrace script outputs necessary variables?", pytrace=False)
+
+
+def main():
+    """ This calls pytest, which runs the test above """
+    parser = argparse.ArgumentParser(
+        description="""
+This script takes as an argument a path to a json file that contains test cases.
+See readme.md for more details.
+"""
+    )
+    parser.add_argument("spec_file", type=argparse.FileType('r'))
+    args = parser.parse_args()
+    os.chdir(os.path.dirname(os.path.abspath(args.spec_file.name)))
+    pytest.main([parser.prog, "-s", "-v", "-rA",
+                "--spec_file", os.path.basename(args.spec_file.name)])
+
+
+if __name__ == "__main__":
+    main()
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 logs/
 tools/__pycache__/
+kernel_tracing/metric_evaluation/__pycache__/
--- a/kernel_tracing/bpftrace_scripts/hda_irq_source_deltas.bt
+++ b/kernel_tracing/bpftrace_scripts/hda_irq_source_deltas.bt
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+
+// This collects the deltas between irqs from each source
+
+tracepoint:sof_intel:sof_intel_hda_irq {
+  if (@times[str(args->source)]) {
+    @delta_usecs[str(args->source)] = hist((nsecs - @times[str(args->source)]) / 1000);
+  }
+  @times[str(args->source)] = nsecs;
+}
+
+END {
+  clear(@times)
+}
--- a/kernel_tracing/bpftrace_scripts/ipc_tx_time.bt
+++ b/kernel_tracing/bpftrace_scripts/ipc_tx_time.bt
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+
+// This tracks the time it takes for ipc communications (sof_ipc_tx_message execution time)
+// It prints out timings as they happen, and will generate a histogram
+// when you stop the script with ctrl-c
+
+kprobe:sof_ipc_tx_message {
+  if (@start != 0 ) {
+    printf("ERROR: overlapping tx");
+    exit();
+  }
+  @start = nsecs;
+  // We could get also the name of the device and other properties from the args,
+  // but we'd need to enable DEBUG_INFO and DEBUG_INFO_BTF to get the types
+  // It would allow us to sort timings by device for example
+  printf("tx started\n");
+}
+
+kretprobe:sof_ipc_tx_message {
+  printf("tx finished, took %d nsecs\n", nsecs - @start);
+  @usecs = hist((nsecs - @start) / 1000);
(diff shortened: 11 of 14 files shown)


tracing/suspend_resume_time.bt
```diff
@@ -0,0 +1,24 @@
+// This tracks the time it takes for runtime and system suspend/resume
```
plbossart, line 1: // SPDX-License-Identifier: BSD-3-Clause
// Copyright(c) 2022 Intel Corporation. All rights reserved.

plbossart, line 1: commit message: you need to have some sort of prefix, then the action, e.g.

tracing: add suspend and resume time tracking with bptrace

Also add some of the explanations in the commit message itself, not the PR.

plbossart (changes requested):
looks quite promising @nklayman, minor nit-picks to improve the commit below.

tracing/suspend_resume_time.bt
nklayman: None of the files in sof-test had this, are you sure it's necessary?

nklayman: Good point, just updated.

plbossart: I prefer to avoid any ambiguity related to license/copyright.

plbossart:
added the 'kernel maintainer' team for feedback.

marc-hb (commented):
Can you rename the directory to something more specific, maybe `sof-test/linux/tracing/` or `sof-test/ktracing`? We have a lot of "tracing" already and `sof-test/tracing/` sounds definitely like firmware tracing.

Can you add a README with some pointers, context/purpose/intro/etc. or if too short for a README then some comments in this new `.bt` file? For now it's very... "raw".

nklayman:
Just added a readme and updated the directory structure to make it more clear what this is.

plbossart:
> Just added a readme and updated the directory structure to make it more clear what this is.

it's better to squash all the commits together and do a git push --force to have a single commit, no?

marc-hb:
> it's better to squash all the commits together and do a git push --force to have a single commit, no?

@nklayman has just used Github the way it was designed to be used, not the way we use it.

Github tolerates but does not really support rewriting history and force pushes. Github is optimized for a single commit per pull request + fixup commits just like @nklayman did. For instance try the "View changes" button above: this time it works! Same for many other Github features.

More info at https://github.com/zephyrproject-rtos/zephyr/issues/39194

Note Github has an option to squash all commits right before merging. Not sure what is the final commit message then.

nklayman:
I prefer to push individually and then squash into one commit before merging, either manually or through the Github UI. Otherwise, it's a lot harder to tell what's changed and see what code review comments refer to. I can do the single commit method if you'd prefer, but as long as we squash before merging the end result is the same, and the process to get there is a bit better.

marc-hb:
_As long as the goal is to have a single commit in the end_, I totally agree the process is much better and the end result is the same. It is the way Github is designed to be used after all!

The problem is purely about **consistency**: as Github PRs do not support reviewing multiple commits "natively", this means learning how to use Github TWICE, in two different ways. One for reviewing a single commit (in the end), and a second workflow for reviewing multiple commits (in the end).

I don't mind at all but I've been the "local git guru" in many teams for many years so I'm not representative. I've seen many, many people struggling with ONE way to use git or Github.

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,112 @@
...
+  def stop(self):
+    # Send ctrl-c
+    self.process.send_signal(subprocess.signal.SIGINT)
+    self.process.wait()
+
+    # Collect printed variables for condition evaluation
+    lines = self.process.stdout.read().split("\n")
+    vars = {}
+    for line in lines:
+      if re.match("@.*: .*", line):
```
marc-hb, line 52: Regular expressions are "write-only" code so please provide one example as a comment.

kernel_tracing/metric_evaluation/spec.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    {
+      "name": "Suspend/Resume Time",
+      "script": "../bpftrace_scripts/suspend_resume_time.bt",
+      "shell": "../../test-case/check-suspend-resume.sh",
+      "conditions": [
+        "@avg[kretprobe:sof_resume, runtime] < 150000",
+        "@avg[kretprobe:sof_resume, system] < 200000",
+        "@avg[kretprobe:sof_suspend, runtime] < 50000",
+        "@avg[kretprobe:sof_suspend, system] < 1000"
```
marc-hb, line 22: Does `@avg` mean "average"? Averages hide intermittent bugs, they're not very interesting. 

https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,112 @@
...
+    shell_commands = [shell_commands]
+
+  # Start bpftrace collection
+  bpf = BPF(bpftrace_script)
+  bpf.start()
+
+  # Execute shell commands
+  for cmd in shell_commands:
+    print_color(colors.BLUE, f"$ {cmd}")
+    subprocess.call(cmd, cwd=working_dir, shell=True)
```
marc-hb, line 76: https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module
> The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle. For more advanced use cases, the underlying [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) interface can be used directly.

https://docs.python.org/3/library/subprocess.html#older-high-level-api
> Prior to Python 3.5, these three functions comprised the high level API to subprocess. You can now use [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) in many cases, but lots of existing code calls these functions.
> `subprocess.call(args, *, ...`

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,112 @@
...
+        key, val = line.split(": ", 1)
+        vars[key] = val
+    return vars
+
+def check_condition(condition, vars):
+  for key, val in vars.items():
+    condition = condition.replace(key, val)
+  return eval(condition)
+
+def run_test(test_case, working_dir):
```
marc-hb, line 62: `check_condition`, `run_test`, `print_color`,... I'm not familiar with `pytest` but a lot of this script looks like re-implementing a test framework. We unfortunately already made that mistake in this sof-test repo + in some internal repos and I'd hate doing it again.

tl;dr: did you take a quick look at `pytest`?

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,112 @@
...
+    print_color(colors.BLUE, f"Running test case: {test_case['name']}")
+    case_failed = run_test(test_case, spec_dir)
+    if case_failed:
+      print_color(colors.RED, "Test case failed")
+      has_failed = True
+    else:
+      print_color(colors.GREEN, "Test case passed")
+
+  if has_failed:
+    sys.exit(1)
```
marc-hb, line 108: Another re-implementation of a basic test framework feature.

marc-hb (changes requested):
Don't forget to run `pylint` now and then. It's not gospel but it's pretty good.

kernel_tracing/metric_evaluation/test.py
nklayman: The reason I did this all manually is so that the tests can be defined in a JSON file, which makes it a lot simpler to write/manage tests. While I could convert this script to a library and have the tests written in Python, that would reduce the simplicity of writing tests. JSON is also easy to manipulate with a program, so it'd be much easier to design a script to generate test suites that are customized per-device. If you're just as happy to write the test cases in Python, then I'll make this a library that works with `pytest`. Otherwise, I'll look into using `pytest` with the current JSON-based test definitions.

marc-hb: > Otherwise, I'll look into using pytest with the current JSON-based test definitions.

Yes.

I'm not very familiar with pytest but I don't see why it would stop you from configuring tests in JSON. I think it's pretty flexible.

nklayman: The issue is that pytest organizes test cases into functions. Therefore, I'd need to create a function for each test case in the JSON which isn't really possible. If each case didn't have it's own function, the error output would be pretty useless as you wouldn't even know which test case is failing. If pytest has some way to dynamically create tests, then it would work great. I'll do some more investigating to see if there is a method for this.

marc-hb: I think pytest supports test parameters. So hopefully you can have a single function for all the tests. Random example (found very quickly, there are probably better ones)
https://docs.pytest.org/en/6.2.x/example/parametrize.html

```
  <Function test_timedistance_v2[20011212-20011211-expected0]>
  <Function test_timedistance_v2[20011211-20011212-expected1]>
```

nklayman: After many hours, I was able to wrangle pytest so that it works with the `spec.json`. Getting the output formatting right took forever, but I'm really happy with the end result:
image

Give it a try and let me know if you have any more suggestions.

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,113 @@
...
+    # Execute shell commands
+    for cmd in shell_commands:
+        print("$", cmd)
+        subprocess.run(cmd, cwd=working_dir, shell=True, check=True)
+
+    print("finished")
+    # Stop tracing and evaluate conditions
+    bpf_vars = bpf.stop()
+    ok = True
+    check_conditions(conditions, bpf_vars)
```
marc-hb, line 89: You can dynamically generate one test instance per condition at run-time thanks to `pytest_generate_tests()`, example at https://stackoverflow.com/questions/49076378/pytest-generate-tests-based-on-arguments

This will save a lot of the code in `check_conditions()` (now `check_condition()`, singular) and produce a much more readable output.

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,113 @@
...
+    """ This gets called by pytest, args are generated in conftest.py """
+    bpftrace_script = os.path.join(working_dir, test_case['script'])
+    shell_commands = test_case['shell']
+    conditions = test_case['conditions']
+    if isinstance(shell_commands, str):
+        shell_commands = [shell_commands]
+
+    # Start bpftrace collection
+    bpf = BPF(bpftrace_script)
+    bpf.start()
```
marc-hb, line 78: Starting and stopping the BPF trace is not part of the test but part of the test "fixture" https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,113 @@
...
+        self.process.wait()
+
+        # Collect printed variables for condition evaluation
+        lines = self.process.stdout.read().split("\n")
+        bpftrace_vars = {}
+        for line in lines:
+            # Check if line is outputting a variable from the script
+            # bpftrace output format is `@name[key1, key2, ...] = val` (keys are optional)
+            if re.match("@.*: .*", line):
+                key, val = line.split(": ", 1)
```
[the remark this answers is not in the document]
nklayman: `name` would be the name of the map, which can be indexed by any number of keys. ie `@max_time[ipc, stream] = 500` would be the map `max_time` indexed by two keys, in this case the value `500` is tied to keys `ipc` and `stream`. It's essentially a flattened nested map.

kernel_tracing/metric_evaluation/test.py
marc-hb: More relevant link sorry:
https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

kernel_tracing/metric_evaluation/test.py
nklayman: I switched to doing it entirely through parameterization (haven't pushed yet), and I think it makes the error output worse. Now, the report lists each condition instead of just the cases, as each condition is an individual parameterized test.

image

Do you want me to go forward with this or revert back to the original setup?

marc-hb: > Now, the report lists each condition instead of just the cases, as each condition is an individual parameterized test.

That was indeed my request: moving PASS / FAIL statuses to pytest instead of manually handling and presenting test results. What's wrong with it?

pytest has the ability to output test results in various formats including the junitXML standard for integration with CI. Handling failed conditions manually means it would not be possible for CI to show which conditions have failed and which ones have passed.

https://docs.pytest.org/en/7.1.x/how-to/output.html
https://docs.pytest.org/en/6.2.x/usage.html#creating-junitxml-format-files

nklayman:
@marc-hb all your comments should be addressed in the latest update. I switched it over to the one-condition-per-test format that you prefer, and applied the small changes. @plbossart I also squashed the commits together into two as you requested, one for the bpftrace scripts themselves and one for the metric evaluation part.

Marking this "ready for review" now as it's at a good state for others to look at. The metric evaluation script will most likely need some changes to make it work in the CI, but that's probably a better task for the CI devs rather than myself.

kernel_tracing/metric_evaluation/spec.json
nklayman: Yes, it means average. However, it's purely a demo (I just made the numbers up), so it doesn't matter much. I figured it'd be better to let those more familiar with SOF figure out the real metrics/conditions to use. I assume we'll end up with some sort of script to generate test cases with device-specific conditions and tests anyways. I've renamed the spec file to `sample_spec.json` to make it more clear that it's just a demo.

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Run with --help for a description of this script
+
+import argparse
+import pytest
+
+
+def test(case):
+    """ This is the test function that is called by pytest """
+    (_, condition, bpftrace_vars) = case
```
marc-hb, line 14:
```suggestion
    (_ignored_something, condition, bpftrace_vars) = case
```
From an "API" perspective it's useful to know what gets discarded.

kernel_tracing/metric_evaluation/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+
+    print("finished")
+    # Stop tracing and evaluate conditions
+    bpf_vars = bpf.stop()
+    return bpf_vars
+
+
+def pytest_generate_tests(metafunc):
+    """ Generate parameter set for each test case in json file """
+    if "case" in metafunc.fixturenames:
```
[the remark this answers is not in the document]
aiChaoSONG: `fixturenames`, I think this is a typo

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,54 @@
...
+    conditionWithVars = condition
```
aiChaoSONG, line 15: let's use snake_case

kernel_tracing/metric_evaluation/test.py
```diff
@@ -0,0 +1,54 @@
...
+    parser = argparse.ArgumentParser(
+        description="""
+This script takes as an argument a path to a json file that contains test cases.
+These cases specify a bpftrace script to collect data while specified shell commands are run.
+When the shell commands finish, the bpftrace script is stopped and variables are collected.
+Python conditions in the spec file are evaluated, replacing @vars with the values from the bpftrace script.
+If any condition fails this script exits with code 1, otherwise it exits with code 0.
+Shell commands are run in the directory of the spec file.
+The bpftrace script path is relative to the spec file as well.
+See sample_spec.json for an example.
```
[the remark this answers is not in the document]
nklayman: I ended up fixing this because it was easier than updating the help message.

kernel_tracing/metric_evaluation/sample_spec.json
```diff
@@ -0,0 +1,26 @@
+{
+  "cases": [
+    {
+      "name": "IPC TX Time",
+      "script": "../bpftrace_scripts/ipc_tx_time.bt",
+      "shell": [
```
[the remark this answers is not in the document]
nklayman: I think using "shell" as the name still makes sense here, but happy to change it if you disagree.

nklayman:
@marc-hb Thanks for the review! Almost all your comments should be addressed now. The no-dependency test option is a bit tricky because the script can't really work without bpftrace. If you're okay with having that a dependency I could make a very simple case, otherwise I could try aliasing `bpftrace` to echo out some sample output so there's truly no dependencies. Let me know which one you prefer.

marc-hb:
> The no-dependency test option is a bit tricky because the script can't really work without bpftrace

The dependency on bpftrace user space is not a problem because it's as simple as `apt install bpftrace`. The annoying part is the kernel part. Reconfiguring and rebooting a kernel before submitting a fix for a minor python warning is not convenient. So I'm just wondering if it would be possible to have some kind of "bpftrace no-op" example / test case for this = anything that does not require a new kernel. If bpftrace has some hard kernel dependencies that are typically not available in stock kernels provided by distributions then too bad forget it. Maybe we could add some sort of "mock bpftrace" later.

(I've not looked at the new version yet)

marc-hb (approved):
I have not tried to run it again but I find the entire code very clear and easy to read and none of this is running in production yet so bug fixes and incremental improvements can be done later: I recommend we merge this to make it easier for others to try.

@plbossart could you have another look, especially the `.bt` files that I'm not familiar with?

marc-hb (dismissed):
Some review comments from a generous `pytest` expert friend who prefers to stay anonymous and not involved:
>  i don't know bpftrace but it looks a little bit like you might want to be using a parameterized fixture to run the test cases instead.
> simple example usage: https://github.com/zephyrproject-rtos/west/blob/a343b7a7982355d0e5521ddcc69f700940d89e50/tests/test_manifest.py#L2885
> with a parameterized fixture, you'll get finer-grained reporting and can customize pytest's notions of what each test is. as-is, it looks like you'll get a global pass/fail on a single function

I think we have parametrization already but maybe `@pytest.mark.parametrize` is more "pythonic", flexible and generally better way to do it. Again this can be a future improvement after merge.

marc-hb:
> I think we have parametrization already but maybe @pytest.mark.parametrize is more "pythonic", flexible and generally better way to do it

I just remembered reading somewhere complaining that `@pytest.mark.parametrize` is static and can't be used after reading a config file. Vague and second-hand information, sorry for any noise.

nklayman:
@marc-hb latest push includes a `no_dependencies.json` that should work with just bpftrace installed. Also fixes edge cases where the spec file arg doesn't start with a `./` and when the bpftrace script doesn't exit in time.

marc-hb:
Nice try but:
```
collecting ... modprobe: FATAL: Module kheaders not found in directory /lib/modules/9.9.9-rc9softest
$ ['sleep', '1']
BPF tracing finished
```

Also, not running as root still hangs:
```
collecting ... ERROR: bpftrace currently only supports running as the root user.
  ... hangs ...
```

```
$ ps f

    987 pts/0    Ss     0:00 -bash
   1161 pts/0    R+     0:39  \_ python3 ./kernel_tracing/metric_evaluation/test.py kernel_tracing/metric_evaluation/no_dependencies.json
   1163 pts/0    Z+     0:00      \_ [bpftrace] <defunct>
```

BTW only bpftrace should run as root, should be started as `sudo bpftrace`.

Again all this can be fixed later.

kernel_tracing/bpftrace_scripts/hda_irq_source_counts.bt
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
plbossart, line 1: typo in commit message: timestamp.

kernel_tracing/metric_evaluation/simple_bpftrace.bt
```diff
@@ -0,0 +1,7 @@
+// This is used to demo the metric evaluation script without requiring any
+// dependencies. Run `./test.py no_dependencies.json` to try it, and change this
+// value to see the failure output.
+
+BEGIN {
+    @val = 42;
+}
```
plbossart, line 7: I don't get what this is for and what '42' might possibly represent - other that "the answer is always 42"

kernel_tracing/metric_evaluation/no_dependencies.json
```diff
@@ -0,0 +1,12 @@
...
+      {
+        "name": "Demo test",
+        "bpftrace": "./simple_bpftrace.bt",
+        "shell": ["sleep", "1"],
+        "conditions": [
+          "@val == 42"
+        ]
+      }
+    ]
+  }
```
plbossart, line 12: no idea what this does?

plbossart (commented):
the code split and what file is used for what isn't very clear to me @nklayman

kernel_tracing/metric_evaluation/simple_bpftrace.bt
nklayman: The answer is always 42 for this script. The purpose is to be able to run the python framework without having any of the kernel dependencies for bpf installed.

kernel_tracing/bpftrace_scripts/suspend_resume_time.bt
```diff
@@ -19,6 +19,7 @@ kretprobe:sof_resume, kretprobe:sof_suspend {
   // When the function returns, we use the saved timestamp to determine execution time
   printf("%s (%s) finished, took %d nsecs\n", probe, @type, nsecs - @start);
   @usecs[probe, @type] = hist((nsecs - @start) / 1000);
+  @avg[probe, @type] = avg((nsecs - @start) / 1000);
```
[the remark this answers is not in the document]
nklayman: I added it as part of that commit since the Python script can't operate on the histogram value, so it needs another value to use for the threshold checking. I think this will end up all squashed together anyways, but I can move it to the previous commit if you'd prefer.

kernel_tracing/metric_evaluation/no_dependencies.json
nklayman: It declares a test case for the python script to run. The help message in the python script explains how to execute it. I could add additional documentation if you think that makes sense.

marc-hb: I don't remember the specific but for a very first example / demo I would generally err on the "too much documentation" side. Don't get me wrong: I mean "too much" as "rephrasing the same thing twice", NOT as "digressions about more advanced stuff".

nklayman: I added a readme in the metric_evaluation folder that explains what this all does.

kernel_tracing/bpftrace_scripts/suspend_resume_time.bt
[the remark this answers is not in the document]
plbossart: I think it'd be clearer if all bpf scripts were moved to the previous patch

kernel_tracing/metric_evaluation/readme.md
```diff
@@ -0,0 +1,34 @@
...
+  "cases": [
+    {
+      "name": "Name to be displayed in output",
+      "bpftrace": "path to the bpftrace script used to collect metrics, relative to spec file",
+      "shell": "shell command to be run alongside bpftrace script, relative to spec file",
+      "conditions": [
+        "list of conditions that will be evaluated against bpftrace output",
+        "these are python expressions",
+        "bpftrace vars can be referenced with @{var name}, ie @avg",
+        "maps are referenced by @{var name}[{key name(s)}], ie @avg[ipc_time]"
```
plbossart, line 17: so in the example above, that does '@val' refer to?

kernel_tracing/metric_evaluation/simple_bpftrace.bt
plbossart: Ah ok, so now I get it.

maybe rename to '@test_check' or something. 'val' threw me off, I didn't understand what this was.

plbossart (commented):
mostly ok @nklayman, just tripped on the 'val' variable that could really mean anything.

@marc-hb you good with this version?

kernel_tracing/metric_evaluation/readme.md
nklayman: That refers to the static val = 42 assignment in the no_dependencies script. I'll be renaming it to test_value so it's more clear.

nklayman:
BPF script changes have been moved to the first patch and `var` has been renamed. @marc-hb @plbossart everything look good now?

plbossart (approved):
LGTM, I'll let @marc-hb comment further on the Python parts. Thanks @nklayman !

marc-hb (approved):
Great start, very clear and readable. None of this is used in production yet so bugs can always be fixed later.

PR #880, opened 2022-03-16
check-sof-logger: disable again dma_nudge() workaround for stuck DMA #4333

marc-hb:
@ujfalusi  has a tentative fix in
https://github.com/thesofproject/linux/pull/3523, remove the
workaround so it can better tested and evaluated.

(cherry picked from commit 4b3f84754ddec8051cf3dac6180ef32950c6fdf0)

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -205,7 +205,12 @@ main()
 
             # Workaround for DMA trace bug
             # https://github.com/thesofproject/sof/issues/4333
-            if [ "$f" = data ]; then
+#            if [ "$f" = data ]; then
+
+            # let's check whether #4333 was finally fixed by sof #4763
+            # + linux #3523
+            if false; then
+
                 dloge "Empty or stuck DMA trace? Let's try to nudge it."
                 dloge '  vv  Workaround for SOF issue 4333  vv'
                 local second_chance="$LOG_ROOT/logger.dma_trace_bug_4333.txt"


marc-hb:
https://sof-ci.01.org/softestpr/PR880/build1015/devicetest/ is empty because CI is completely broken this week.

ujfalusi:
@marc-hb , the stuck part still going to need workaround since the firmware will only do the copy on dma burst boundaries, it is still possible that we have stalled trace down in rhe firmware and we need to poke it to give it to us.

marc-hb:
>  since the firmware will only do the copy on dma burst boundaries,

How far apart are DMA burst boundaries?

When the firmware boots it prints quite a lot of logs, I would say at least 1 kilobytes. Bug #4333 has always been "a lot of logs or nothing". There's a digression in #4333 about missing "lagging" logs but please ignore it, this basic test is nowhere near smart enough to catch that.

marc-hb:
> Bug #4333 has always been "a lot of logs or nothing".

I take that back, sorry. In daily run 10879?model=TGLH_SKU0A70_HDA&testcase=check-sof-logger (2002-03-09) the FW logs were not empty but they stopped after only 3 lines, just before receiving the first IPC.

TGLH_SKU0A70_HDA seems a bit "special", see https://github.com/thesofproject/sof/issues/4333#issuecomment-1069676026

test-case/check-sof-logger.sh
```diff
@@ -205,7 +205,12 @@ main()
 
             # Workaround for DMA trace bug
             # https://github.com/thesofproject/sof/issues/4333
-            if [ "$f" = data ]; then
+#            if [ "$f" = data ]; then
```
greg-intel, line 208: Is this comment intentionally left in?  Also not sure I follow the reasoning for the `if false; then` line.  Guessing that's what the comments are referencing.

greg-intel (approved):

test-case/check-sof-logger.sh
marc-hb: This is just a test DRAFT, don't take it seriously.

The `if false` is a way to comment out 10 lines with a one-line change.

marc-hb:
SOFCI TEST

PR #882, opened 2022-03-22
lib.sh: disable sof-logger when running IPC4 mode

XiaoyunWu6666:
Currently, sof-logger is not supported when running IPC4 mode.

Thus, disable SOF logs collection globally when DUT runs IPC4 mode (this can be telled by system file /sys/module/snd_sof_pci/parameters/ipc_type).

This should be removed after sof-logger support for IPC4 has been provided.

Signed-off-by: Xiaoyun Wu(Iris) <xiaoyun.wu@intel.com>

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -459,6 +459,32 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+is_IPC4()
+{
+    local ipc_type
+
+    ipcFile=/sys/module/snd_sof_pci/parameters/ipc_type
+    # If no /sys/module/snd_sof_pci/parameters/ipc_type exists
+    # the DUT is running IPC3 mode
+    [ -e "$ipcFile" ] || {
+        dlogi 'No /sys/module/snd_sof_pci/parameters/ipc_type, DUT runs IPC3 mode'
+        return 1
+    }
+
+    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
+    # If the value of file ipc_type is:
+    # 0: DUT runs IPC3 mode
+    # 1: DUT runs IPC4 mode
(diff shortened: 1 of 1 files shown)


XiaoyunWu6666:
this can replace PR https://github.com/intel-innersource/drivers.audio.ci.sof-sh/pull/53 and  https://github.com/intel-innersource/drivers.audio.ci.sof-framework/pull/202.
We make sof-test disable sof-logger directly when it detects ipc4 setting in system file , instead of passing the SOF_LOGGING variable to test framework or setting it as the environment variable

case-lib/lib.sh
```diff
@@ -459,6 +459,32 @@ is_zephyr()
...
+        return 1
+    }
+
+    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
+    # If the value of file ipc_type is:
+    # 0: DUT runs IPC3 mode
+    # 1: DUT runs IPC4 mode
+    ipc_type=$(cat $ipcFile)
+    if [ $ipc_type -eq 1 ]; then
+        dlogi "/sys/module/snd_sof_pci/parameters/ipc_type is ${ipc_type}, DUT runs IPC4 mode"
```
greg-intel, line 480: Sorry, by "runs" does it mean "will run in?"  Or does it mean it ran in IPC4 mode?  Current tense is ambiguous here and below.

plbossart:
please submit PRs from your forked repository, not from a branch on the sof-test GIT tree.

case-lib/lib.sh
XiaoyunWu6666: hi @greg-intel  I modified the tense in #884 .
@plbossart Sorry, I mistook the repo . Already prepare a new PR from my own forked repo #884  , thx~

PR #884, opened 2022-03-23
lib.sh: disable sof-logger when running IPC4 mode

XiaoyunWu6666:
Currently, sof-logger is not supported when running IPC4 mode.

Thus, disable SOF logs collection globally when DUT runs IPC4 mode (this can be telled by system file /sys/module/snd_sof_pci/parameters/ipc_type).

This should be removed after sof-logger support for IPC4 has been provided.

Signed-off-by: Xiaoyun Wu(Iris) <xiaoyun.wu@intel.com>

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -459,6 +459,27 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+ipc4_used()
+{
+    local ipc_type
+    ipc_file=/sys/module/snd_sof_pci/parameters/ipc_type
+
+    # If /sys/module/snd_sof_pci/parameters/ipc_type does not exist
+    # the DUT is running IPC3 mode
+    ipc_type=$(cat $ipc_file) || {
+        return 1
+    }
+
+    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
+    # If the value of file ipc_type is:
+    # 0: DUT runs IPC3 mode, ipc_used return 1(false)
+    # 1: DUT runs IPC4 mode, ipc4_used return 0(true)
+    [ $ipc_type -eq 1 ] || {
+        return 1
+    }
+    return 0
+}
+
 logger_disabled()
 {
     local ldcFile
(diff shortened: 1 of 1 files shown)


XiaoyunWu6666:
this can replace PR https://github.com/intel-innersource/drivers.audio.ci.sof-sh/pull/53 and https://github.com/intel-innersource/drivers.audio.ci.sof-framework/pull/202.
We make sof-test disable sof-logger directly when it detects ipc4 setting in system file , instead of passing the SOF_LOGGING variable to test framework or setting it as the environment variable

case-lib/lib.sh
```diff
@@ -459,6 +459,32 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+is_IPC4()
```
aiChaoSONG, line 462: ipc4_used() or ipc3_used()

case-lib/lib.sh
```diff
@@ -459,6 +459,32 @@ is_zephyr()
...
+{
+    local ipc_type
+
+    ipcFile=/sys/module/snd_sof_pci/parameters/ipc_type
```
aiChaoSONG, line 466: snake case: ipc_file

case-lib/lib.sh
```diff
@@ -459,6 +459,32 @@ is_zephyr()
...
+    # If no /sys/module/snd_sof_pci/parameters/ipc_type exists
+    # the DUT is running IPC3 mode
+    [ -e "$ipcFile" ] || {
```
aiChaoSONG, line 469: let's keep the function minimal, and reduce the logs, can you try this?
```
ipc_type = $(cat $ipc_file) || {
     return CODE_FOR_IPC3
}

[ "$ipc_type" -eq 1 ] || {
    return CODE_FOR_IPC3
}

return CODE_FOR_IPC4
```

case-lib/lib.sh
```diff
@@ -481,6 +507,15 @@ logger_disabled()
         return 0
     fi
 
+    is_IPC4 && {
```
aiChaoSONG, line 510: We can add logging here, tell the user we are in IPC4 mode. or IPC3 mode.

XiaoyunWu6666:
test result on DUT is as expected

aiChaoSONG (approved):

marc-hb:
> Currently, sof-logger is not supported when running IPC4 mode.

@plbossart please correct me: I think this is wrong. sof-logger is not supported when running cAVS firmware. It should be supported when running open-source IPC4.

If a test fails with IPC4, how is the bug going to be identified without logs?

aiChaoSONG:
@marc-hb  I think the kernel doesn't setup debugfs for SOF under IPC4, so sof-logger can not work.

case-lib/lib.sh
```diff
@@ -481,6 +502,16 @@ logger_disabled()
         return 0
     fi
 
+    ipc4_used && {
+        # TODO:
+        # Need to remove disabling sof-logger
+        # after sof-logger support for IPC4 has been provided in the future
+        dlogi 'Currenly sof-logger is not supported when running IPC4 mode'
```
greg-intel, line 509: Currenly should be Currently. 

Sorry I didn't respond before the PR was merged.

PR #885, opened 2022-03-25
check-pause-release-suspend-resume: timing fixes + switch to sof-tplgreader

keqiaozhang:


plbossart (changes requested):
can this be added to run-all-tests.sh?

We need all developpers to replicate what CI does.

marc-hb (commented):
Since you're rewriting most of the test, please take the opportunity to move all the code to functions, see why in #740. At least one `main()` function, maybe others too.

keqiaozhang:
> can this be added to run-all-tests.sh?
> 
> We need all developpers to replicate what CI does.

Done.

keqiaozhang:
> Since you're rewriting most of the test, please take the opportunity to move all the code to functions, see why in #740. At least one `main()` function, maybe others too.

PR is updated.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -51,47 +51,50 @@ set -e
...
-fmt=${OPT_VAL['f']}
-channel=${OPT_VAL['c']}
-rate=${OPT_VAL['r']}
-repeat_count=${OPT_VAL['l']}
-sleep_period=${OPT_VAL['i']}
-test_mode=${OPT_VAL['m']}
-file_name=${OPT_VAL['F']}
+parse_param()
+{
+    func_opt_parse_option $@
```
marc-hb, line 88: shellcheck, easy

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -51,45 +51,39 @@ set -e
...
 source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
 
-OPT_NAME['m']='mode'         OPT_DESC['m']='test mode. Example: playback; capture'
-OPT_HAS_ARG['m']=1             OPT_VAL['m']='playback'
+CASEDIR=$(dirname "${BASH_SOURCE[0]}")
 
-OPT_NAME['p']='pcm'          OPT_DESC['p']='audio pcm. Example: hw:0,0'
-OPT_HAS_ARG['p']=1             OPT_VAL['p']='hw:0,0'
+OPT_NAME['m']='mode'             OPT_DESC['m']='test mode. Example: playback; capture'
+OPT_HAS_ARG['m']=1               OPT_VAL['m']='playback'
```
marc-hb, line 60: Please remove all these whitespace changes in this area. They make it impossible to find the real changes.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -103,82 +106,107 @@ case $test_mode in
...
-        incr i
-        if { \$i > $repeat_count } { exit 0 }
-        exp_continue
-    }
+    [[ -z $file_name ]] && file_name=$dummy_file
 }
+
+run_test()
+{
+    for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
```
marc-hb, line 124: You can save one indentation level by moving the `for` loop to the `main()` function.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -112,73 +106,77 @@ logger_disabled || func_lib_start_log_collect
...
-    sof-process-kill.sh
-    [[ $? -ne 0 ]] && dlogw "Kill process catch error"
-    exit $ret
-fi
-sof-kernel-log-check.sh "$KERNEL_CHECKPOINT"
-exit $?
+    ret=$?
+    if [ $ret -ne 0 ]; then
+        sof-process-kill.sh || dlogw "sof-process-kill.sh failed"
+        exit $ret
```
marc-hb: Do we really want to preserve the exit code here? If it's 2 then it will be misinterpreted as a SKIP! We experienced this recently in some different place.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -103,82 +106,107 @@ case $test_mode in
...
-            puts "errorCode was: \$error_code"
-            exit 1
-        }
+    # only run suspend/resume once for each loop.
+    # Use system default value if no sleep type is specified
+    if [ -n "${OPT_VAL['T']}" ]; then
+        sleep_opts="-l 1 -T ${OPT_VAL['T']}"
+    else
+        sleep_opts="-l 1"
+    fi
```
marc-hb, line 117:
```shell
sleep_opts="-l 1"
[ -z "${OPT_VAL['T']}" ] || sleep_opts+="-T ${OPT_VAL['T']}"
```

marc-hb (changes requested):
The breakdown into multiple commits was a nice try but it didn't really work:

- The first commit message talks about sof-tplgreader only but it has tons of other, unrelated changes.
- The move to functions should have come first (but now it's too late).

You can still try to keep the "small" commits as independent commits if possible but please squash the 2 big commits into a single big commit, it's just simpler and more "honest".

Keep the `run-all-tests.sh` commit separate of course.

test-case/check-pause-release-suspend-resume.sh
keqiaozhang: Fixed.

test-case/check-pause-release-suspend-resume.sh
keqiaozhang: Good suggestion. PR is updated.

marc-hb (commented):
Shell script code looks good to me except for a couple shellcheck warnings easy to fix.

@plbossart or @ranj063 can you please review the test / expect logic? I think one of you requested this?

plbossart:
> > can this be added to run-all-tests.sh?
> > We need all developpers to replicate what CI does.
> 
> Done.

I don't see any changes to run-all-tests.sh @keqiaozhang ?

marc-hb:
> I don't see any changes to run-all-tests.sh @keqiaozhang ?

That was moved to #886, not sure why. No big deal.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -103,82 +106,96 @@ case $test_mode in
...
-        puts \$msg
-        puts "Finished suspend-resume test"
-        if { \$retval } {
-            puts "suspend resume cycle has failed."
-            set error_code [lindex \$::errorCode {2}]
-            puts "errorCode was: \$error_code"
+    for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
+    do
+        run_test
+        if [ $? -ne 0 ]; then
```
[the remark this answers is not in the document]
keqiaozhang: Updated.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -51,47 +51,50 @@ set -e
 ##  * No unexpected errors should be present in dmesg during or after test
 ##      completion.
 
+# shellcheck source=case-lib/lib.sh
 source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
```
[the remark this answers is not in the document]
marc-hb: No, not completely. Can you please run shellcheck locally before pushing? Especially after fixing shellcheck issues? It would have saved a lot of time.
```shell
sudo apt install shellcheck
shellcheck -x test-case/check-pause-release-suspend-resume.sh
```

keqiaozhang:
@plbossart @marc-hb @ranj063 
Please help with the final review. If no objections, I will merge this PR by end of this week.

marc-hb:
There's one shellcheck warning very easy to fix. We should not still be looking at trivial shellcheck warnings 1 month after the first version, this is a waste of time. Run shellcheck locally before submitting it takes a split second.

marc-hb:
@plbossart , @ranj063 , @lyakh could you please view the pause/release/suspend/resume logic?

plbossart:
> @plbossart , @ranj063 , @lyakh could you please review the pause/release/suspend/resume logic?

I don't speak shell fluently and I am unable to figure out what the 'logic' is in this heap of shell commands.

I will note however that I don't understand why the pause can only be done after a delay. What possible 'unexpected errors' could happen if we pause immediately after starting? It looks like we are masking a set of possible issues without a clear justification.

marc-hb:
Note the vast majority of this test and PR is not written in shell script but in https://en.wikipedia.org/wiki/Expect

marc-hb:
This test was finally rewritten in 006946cceaac0dcea36878f55e49f081ae98d367 + follow-ups.

Issue #888, opened 2022-04-11
[BUG] Filter out NVME harmless issues


plbossart:
https://sof-ci.01.org/linuxpr/PR3576/build7525/devicetest/

Suspend-resume is broken because of harmless NVME issues discussed here: https://lore.kernel.org/linux-nvme/F2ACCD82-052F-473F-9882-1703147FA662@oracle.com/T/#m14a43e0c73c589876d0cf20680c792e8d2ef5063

We should temporarily filter out this error to proceed with 5.18-rcX support
[  894.167961] kernel: nvme0: Admin Cmd(0x6), I/O Error (sct 0x0 / sc 0x2)

marc-hb:
Merged. Close?

plbossart:
> Merged. Close? Passing now?

no results that I can see. 

Well I can see them if I use the private link https://sof-ci.sh.intel.com/#/result/planresultdetail/11758 but the SOF CI doesn't show the results publicly.

marc-hb:
We have some kind of outage right now because Github is rate-limiting us. Investigation in prgoress.

plbossart:
https://sof-ci.01.org/linuxpr/PR3785/build446/devicetest/?model=TGLU_UP_HDA&testcase=check-suspend-resume-with-playback-5

````
===========================>>
[ 1013.347888] kernel: nvme nvme0: Device shutdown incomplete; abort shutdown
<<===========================
````

marc-hb:
As usual a reminder to please make sure storage error messages are actually harmless before filtering them out. https://github.com/thesofproject/linux/issues/3387 and friends cost us literally months of work, especially @keqiaozhang and @fredoh9's time.  Who knows, maybe some of that time could have been saved by paying more attention to error messages?

marc-hb:
I believe this has been fixed, please re-open if not.

PR #889, opened 2022-04-11
sof-kernel-log-check: temporarily filter out NVME harmless issues

plbossart:
kernel 5.18-rc1 added a set of harmless NVME errors that we should
filter out until removed upstream.

BugLink: https://github.com/thesofproject/sof-test/issues/888


marc-hb (approved):
Thanks for the new bug, upstream report and selective regex!

aiChaoSONG (approved):

Issue #891, opened 2022-04-15
[BUG] filter out ptp_kvm issues


plbossart:
Across the board fails in https://sof-ci.sh.intel.com/#/result/planresultdetail/11848 (public test results not available after two hours of tests).

````
[    1.559736] kernel: fail to initialize ptp_kvm
````

I have no idea what on earth this could be.

marc-hb:
This was fixed a long time ago by kconfig change
- https://github.com/thesofproject/kconfig/pull/70

PR #892, opened 2022-04-15
sof-kernel-log-check: filter ptp_kvm issues

plbossart:
Across the board fails in  https://sof-ci.sh.intel.com/#/result/planresultdetail/11848

BugLink: https://github.com/thesofproject/sof-test/issues/891

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -383,6 +383,10 @@ ignore_str="$ignore_str"'|I/O error, dev loop., sector 0 op 0x0:.READ. flags 0x8
 # https://github.com/thesofproject/sof-test/issues/888
 ignore_str="$ignore_str"'|nvme0: Admin Cmd\(0x[[:digit:]]+\), I/O Error \(sct 0x0 / sc 0x2\)'
 
+# PTP KVM errors
+# https://github.com/thesofproject/sof-test/issues/891
+ignore_str="$ignore_str"'|fail to initialize ptp_kvm'
+
 #
 # SDW related logs
 #


marc-hb (changes requested):
This is an upstream regression; let's not ignore it. I think I have a better fix in https://github.com/thesofproject/kconfig/pull/70


The following upstream commit changed the default values of CONFIG_HYPERVISOR_GUEST, CONFIG_PARAVIRT, CONFIG_KVM_GUEST, CONFIG_PTP_1588_CLOCK_KVM and a couple hundreds others

410ce3dd5055b x86/config: Make the x86 defconfigs a bit more usable


CONFIG_PTP_1588_CLOCK_KVM is a feature where a KVM guest can get very high precision clock from the host. This feature re-uses /dev/ptp devices which were originally designed for PTP network time (higher precision than NTP) but CONFIG_PTP_1588_CLOCK_KVM is unrelated to networking.

CONFIG_PTP_1588_CLOCK_KVM does not really matter, however I think we do care about the other ones and we don't want them. For instance:

```
config PARAVIRT
        bool "Enable paravirtualization code"
        depends on HAVE_STATIC_CALL
        help
          This changes the kernel so it can modify itself when it is run
          under a hypervisor, potentially improving performance significantly
          over full virtualization.  However, when run without a hypervisor
          the kernel is theoretically slower and slightly larger.

config KVM_GUEST
        bool "KVM Guest support (including kvmclock)"
        depends on PARAVIRT
        select PARAVIRT_CLOCK
        select ARCH_CPUIDLE_HALTPOLL
        select X86_HV_CALLBACK_VECTOR
        default y
        help
          This option enables various optimizations for running under the KVM
          hypervisor. It includes a paravirtualized clock, so that instead
          of relying on a PIT (or probably other) emulation by the
          underlying device model, the host provides the guest with
          timing infrastructure such as time of day, and system time
          
```

PR #894, opened 2022-04-25
skip the EDAC pnd2 errors

keqiaozhang:
This error occurs after upgrading Ubuntu to 22.04 LTS
on APL platform.


keqiaozhang:
Confirmed that this is generic 5.15 kernel specific issue, so we can overlook it.

PR #897, opened 2022-04-29
Enable cavstool

keqiaozhang:


case-lib/lib.sh
```diff
@@ -151,6 +151,39 @@ find_ldc_file()
...
+    if [ -z "$CAVSTOOL" ]; then
+        CAVSTOOL=$(command -v cavstool.py) || {
+            dlogw 'No cavstool.py found in PATH'
+            return 1
+        }
+    fi
+
+    test -x "$CAVSTOOL" || {
+        dlogw "$CAVSTOOL not found or not executable"
+        return 2
```
marc-hb, line 168: 2 means SKIP. Did you test this?

case-lib/lib.sh
```diff
@@ -151,6 +151,39 @@ find_ldc_file()
...
+    }
+
+    logfile=$LOG_ROOT/cavstool.txt
+
+    if func_hijack_setup_sudo_level ;then
+        # shellcheck disable=SC2034 # external script will use it
+        ZPH_LOG_COLLECT=1
+    else
+        >&2 dlogw "without sudo permission to run $CAVSTOOL command"
+        return 3
```
marc-hb, line 178: You're returning 3 different error codes but the callers never look at them. I don't think you ever tested all this error handling code; what problem does it solve? sudo, the shell and other system utilities already provide error messages for free, what was wrong with them?

Every line of code you submit should have been run at least once. Write the code you actually need, not the code you think you need.

test-case/check-alsabat.sh
```diff
@@ -42,6 +42,9 @@ OPT_HAS_ARG['n']=1             OPT_VAL['n']=240000
 OPT_NAME['s']='sof-logger'      OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
 OPT_HAS_ARG['s']=0             OPT_VAL['s']=1
 
+OPT_NAME['z']='cavstool'       OPT_DESC['z']="Zephyr native logging tool, the data will store at $LOG_ROOT"
+OPT_HAS_ARG['z']=0             OPT_VAL['z']=1
```
marc-hb, line 46: Why do we need to make this configurable, do we have multiple names to support? Don't submit code that we don't need yet.

https://www.martinfowler.com/bliki/Yagni.html

marc-hb (dismissed):

marc-hb (commented):
I forgot how difficult this code was to understand. I'm adding some comments in #906, please help review.


This looks a good first step, thanks!

Actually, I think this is a good SECOND step: could you try to add the same feature to check-sof-logger.sh first in a separate PR? So if we find issues with cavstool.py then it will impact only the check-sof-logger.sh test whereas this PR is affecting all tests. If the new check-sof-logger.sh works fine for a couple days then we can merge this.

case-lib/hijack.sh
```diff
@@ -76,11 +85,14 @@ function func_exit_handler()
...
         }
         sleep 1s
-        if pgrep "$loggerBin"; then
-            dloge "$loggerBin resisted pkill -TERM, using -KILL"
-            sudo pkill -KILL "$loggerBin"
-            exit_status=1
-        fi
+	for logtool in "$cavstoolBin" "$loggerBin"
+	do
+            if pgrep -f -n "$logtool"; then
```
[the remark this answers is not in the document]
keqiaozhang: For example, when running cavstool.py, it will trigger 2 processes in process list. 
```
root       16635  0.0  0.0  13856  4748 pts/0    S+   08:34   0:00 sudo cavstool.py -l
root       16636  0.4  0.2  28744 16708 pts/0    S+   08:34   0:01 python3 /usr/local/bin/cavstool.py -l
```

With `-f` option, it will list all 2 processes 
```
pgrep -f cavstool.py
16635
16636
```
But it only list the child process with `-n` 
```
pgrep -f -n cavstool.py
16636
```

aiChaoSONG:
> @aiChaoSONG , can we change the error level for above logs?

These I think these logs are from zephyr system, not the ones we can control

marc-hb:
I just found by chance that `cavstool.py` disables power management (forcing `/sys/devices/pci0000\:00/0000\:00\:0e.0/power/control`). So we can't use it unconditionally yet. I'm looking into that right now.

case-lib/hijack.sh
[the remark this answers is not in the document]
keqiaozhang: That makes sense, remove `-n`.

keqiaozhang:
> > @aiChaoSONG , can we change the error level for above logs?
> 
> These I think these logs are from zephyr system, not the ones we can control

@marc-hb , any comments on this?

marc-hb (changes requested):
Blocking this until cavstool is compatible with D3. Working on it.

keqiaozhang:
>I just found by chance that cavstool.py disables power management (forcing
> /sys/devices/pci0000:00/0000:00:0e.0/power/control

@marc-hb please share more information about this, I tried cavstool on APL and TGL UP extreme board. Power control always on when running cavstool.py.

marc-hb:
> Power control always on when running cavstool.py.

Exactly.

marc-hb:
cavstool.py has unrelated stability issues slowing or blocking progress: https://github.com/zephyrproject-rtos/zephyr/issues/46864

There are tentative fixes.

marc-hb:
cavstool support for D3 submitted in:
- https://github.com/zephyrproject-rtos/zephyr/pull/47128

keqiaozhang:
Except for the TIMEOUT of suspend/resume on WHL, all other failures are related to the SOF issue https://github.com/thesofproject/sof/issues/6020.

Also thanks to https://github.com/thesofproject/sof/issues/6020, it helped to prove that this PR really works.
Please refer to `multiple-pause-resume-5.sh` on `WHL_UPEXT_HDA_ZEPHYR`.

marc-hb:
Looking at the internal site I see that https://sof-ci.01.org/softestpr/PR897/build93/devicetest was aborted by... @keqiaozhang ?

case-lib/hijack.sh
```diff
@@ -54,9 +54,18 @@ function func_exit_handler()
         # shellcheck disable=SC2154
         logfile="$logfile"
 
-        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
-        # We need this to avoid the confusion of a "Terminated" message
+        local loggerBin wcLog logtool; loggerBin=$(basename "$SOFLOGGER");
```
marc-hb, line 57: Don't declare `logtool` 30 lines before the first time it's used, it's less readable.

case-lib/lib.sh
```diff
@@ -293,6 +293,29 @@ find_ldc_file()
     printf '%s' "$ldcFile"
 }
 
+func_cavstool_etrace_collect()
```
marc-hb, line 296: Add a comment / header documenting important control flow expectations
- whether this function leaves anything running in the background. Any timeout.
- what the caller is expected to clean up and when

marc-hb (changes requested):
This works but it's not ideal because it runs `cavstool` _only at the end_ of the test. `sof-logger` cannot follow the `etrace` ring buffer, that's why `sof-logger` is reading `etrace` only at the end of the test. BUT `cavstool` can follow a ring buffer in `etrace` which is MUCH better so it should be started at the start of the test and we would not miss any Zephyr logs.

I think the `func_cavstool_etrace_collect()` "API" should be changed like this:

NOW:
- `func_cavstool_etrace_collect 0`: starts sof-logger for **DMA** trace for the whole test duration
- `func_cavstool_etrace_collect 1`: short, one-shot sof-logger **etrace** at the end of the test

NEXT:
- `func_cavstool_etrace_collect 0` : starts BOTH sof-logger DMA trace AND cavstool **for the whole test duration**
- `func_cavstool_etrace_collect 1`: start sof-logger + etrace **only at the end of a test.**

keqiaozhang:
@marc-hb , the PR testing results are not good on Zephyr platforms. It seems that starting BOTH sof-logger DMA trace AND cavstool for the whole test duration will introduce new IPC errors/timeouts. 
I can reproduce it manually(reproduction rate is about 1/5).  For triggering the issue, we need to run one test case for several times or run different cases in sequence.

I've confirmed that `cavstool` will be terminated before the end of each case.

case-lib/hijack.sh
```diff
@@ -54,9 +54,18 @@ function func_exit_handler()
         # shellcheck disable=SC2154
         logfile="$logfile"
 
-        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
-        # We need this to avoid the confusion of a "Terminated" message
+        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER");
```
marc-hb, line 57: Spurious edit
```suggestion
        local loggerBin wcLog; loggerBin=$(basename "$SOFLOGGER")
```

case-lib/lib.sh
```diff
@@ -293,6 +293,29 @@ find_ldc_file()
...
+{
+    local logopt="-l"
```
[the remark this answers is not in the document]
marc-hb: Ping

case-lib/lib.sh
```diff
@@ -334,7 +352,17 @@ func_lib_start_log_collect()
 
     if [ "X$is_etrace" == "X0" ];then
         logfile=$LOG_ROOT/slogger.txt
+
+        # start cavstool in the background to collect the etrace.
+        # Since cavstool can follow a ring buffer in etrace, so it should be
+        # started at the start of the test and we would not miss any Zephyr logs.
+        is_zephyr && func_cavstool_etrace_collect
```
marc-hb, line 359:
```suggestion
       # FIXME (new) sof bug #1234
       # is_zephyr && func_cavstool_etrace_collect
```

marc-hb (changes requested):

case-lib/lib.sh
```diff
@@ -293,6 +293,24 @@ find_ldc_file()
...
+        CAVSTOOL=$(command -v cavstool.py) || {
+            dlogw 'No cavstool.py found in PATH'
+            return 1
+        }
+    fi
+
+    local cavstoolCmd=("$CAVSTOOL" "$logopt")
+    dlogi "Starting ${cavstoolCmd[*]}"
+    # Cleaned up by func_exit_handler() in hijack.sh
+    sudo "${cavstoolCmd[@]}" > "$clogfile" &
```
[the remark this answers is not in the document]
marc-hb: You must use `>&` because we really want BOTH

case-lib/lib.sh
```diff
@@ -293,6 +293,24 @@ find_ldc_file()
     printf '%s' "$ldcFile"
 }
 
+func_cavstool_etrace_collect()
+{
+    local logopt="-l"
+    local clogfile=$LOG_ROOT/cavstool.txt
```
marc-hb, line 299:
```suggestion
    local clogfile=$LOG_ROOT/etrace.txt
```
Please keep the same name for consistency. I'm working on another PR that scans for ERRORs.

marc-hb:
Looks like adding cavstool to the sof-logger test ONLY is already causing some trouble: https://github.com/thesofproject/sof-test/pull/941

keqiaozhang:
I have filed a SOF issue:https://github.com/thesofproject/sof/issues/6039

Since the latest cavstool.py supports verbose log, so I enabled it accordingly `local clogopt="--log-only --verbose"`.

case-lib/lib.sh
```diff
@@ -293,6 +293,25 @@ find_ldc_file()
...
+            dlogw 'No cavstool.py found in PATH'
+            return 1
+        }
+    fi
+
+    local cavstoolCmd=("$CAVSTOOL" "$clogopt")
+    dlogi "Starting ${cavstoolCmd[*]}"
+    # Cleaned up by func_exit_handler() in hijack.sh
+    # shellcheck disable=SC2024
+    sudo "${cavstoolCmd[@]}" >& "$clogfile" &
```
marc-hb, line 312:
```suggestion
    sudo "${cavstoolCmd[@]}" >& --log-only --verbose &
```
This tries to load firmware file named `'--log-only --verbose'`

I just wasted a couple of hours on this... already asked to inline this pointless variable before (when it was a single option and working)

case-lib/lib.sh
```diff
@@ -293,6 +293,25 @@ find_ldc_file()
...
+    local clogfile=$LOG_ROOT/etrace.txt
+
+    if [ -z "$CAVSTOOL" ]; then
+        CAVSTOOL=$(command -v cavstool.py) || {
+            dlogw 'No cavstool.py found in PATH'
+            return 1
+        }
+    fi
+
+    local cavstoolCmd=("$CAVSTOOL" "$clogopt")
```
marc-hb, line 308:
```suggestion
    local cavstoolCmd=("$CAVSTOOL"  --log-only --verbose)
```

Otherwise this tries to load a firmware file named `'--log-only --verbose'`

I just wasted a couple of hours on this... already asked to inline this pointless variable before (when it was a single option and working)

marc-hb (changes requested):

case-lib/lib.sh
marc-hb: OK, I realize only I see this because I disable the sof-test `sudo` function, I'm using the regular `sudo` command.

Should still be fixed.

marc-hb:
SOFCI TEST

case-lib/lib.sh
marc-hb: Please help review and merge this `sudo()` fix first:
- #943

case-lib/lib.sh
```diff
@@ -334,7 +353,18 @@ func_lib_start_log_collect()
...
+        # Since cavstool can follow a ring buffer in etrace, so it should be
+        # started at the start of the test and we would not miss any Zephyr logs.
+        # FIXME (new) sof SOF bug #6039
+	is_zephyr && func_cavstool_etrace_collect
     else
+        # cavstool starts at the beginning of the test, we don't need to start
+        # it again. We only need to start sof-logger to collect the etrace logs
+        # only for the non-zephyr platforms at the end of a test.
+        is_zephyr && return 0
+
```
marc-hb, line 367: The `$logfile` variable in this function is a global variable passed to lib.sh and I just realized it is a complete mess that must be fixed this PR

marc-hb (changes requested):
I just realized the $logfile is a complete mess that must be fixed before this PR in a different and earlier PR

marc-hb:
> func_cavstool_etrace_collect 0 : starts BOTH sof-logger DMA trace AND cavstool for the whole test duration

I'm sorry for this wrong advice. @kv2019i explained me (and you?) that cavstool is "spamming" the SRAM which can have bad interactions with the firmware - as you found in test results above and filed in thesofproject/sof#6039. So we're back to having just a quick cavstool look at the end of the test, really sorry for the back and forth.

cc: @nklayman , @andyross , @jsarha , @enjiamai, @teburd

enjiamai:
Excuse me, @marc-hb , is this caused by the modified cavstool.py (client-server-base)? I am working on another PR to split the pure cavstool.py (original one) and the client-server logic. Does this help? Thanks.

marc-hb:
> Excuse me, @marc-hb , is this caused by the modified cavstool.py (client-server-base)?

No, it's a warning that any castool.py is "spamming" the SRAM which is causing issues.

andyross:
> cavstool is "spamming" the SRAM which can have bad interactions with the firmware

It should only be polling 4 words every 100ms.  Obviously if there's a ton of data to be copied out, that's going to have impact.  But it's intended to be a reasonably good citizen.   Are there existing interactions it's causing?  We should look at those, I suspect.

kv2019i:
@andyross wrote:
>> cavstool is "spamming" the SRAM which can have bad interactions with the firmware
> 
> It should only be polling 4 words every 100ms. Obviously if there's a ton of data to be copied out, that's going to have 
> impact. But it's intended to be a reasonably good citizen. Are there existing interactions it's causing? We should look at those,

Ack, thanks Andy. I actually assumed it was polling more frequently (and gave the above feedback). So yeah, this seems more like a bug (and probably not in cavstool.py).  Given the problems only occur in large test campaigns, this is not so easy to debug. We do have other logging backends coming available very soon (https://github.com/zephyrproject-rtos/zephyr/pull/47058 and https://github.com/thesofproject/sof/pull/6072 ), I'd like to focus energy on getting these working first. Both mtrace (SRAM) and probes (DMA/DDR) backends provide capability to notify the host (with a IPC/DMA irq) when new data is available, eliminating the need for any polling of data by host. Given the possible people who could debug the rare failures with cavstool.py with SOF test campaigns are the same people working on the new backends, I'd prefer to complete the new backends first and prioritize enabling these in the SOF CI.

marc-hb:
> Ack, thanks Andy. I actually assumed it was polling more frequently (and gave the above feedback). So yeah, this seems more like a bug (and probably not in cavstool.py).

So may be the bug WAS in cavstool.py after all? Just fixed by @ujfalusi ?
- https://github.com/zephyrproject-rtos/zephyr/pull/50178

marc-hb:
Closed in the weekly CI meeting because IPC3+Zephyr will never be supported in production. Zephyr will be IPC4 only which means it will always have mtrace support.

PR #899, opened 2022-05-02
Various suspend/resume fixes

marc-hb:
See commit messages.

### DIFF
--- a/test-case/check-suspend-resume.sh
+++ b/test-case/check-suspend-resume.sh
@@ -82,8 +82,8 @@ do
     setup_kernel_check_point
     sleep_count=$(cat /sys/power/wakeup_count)
     dlogc "Run the command: rtcwake -m mem -s ${sleep_lst[$i]}"
-    sudo rtcwake -m mem -s ${sleep_lst[$i]}
-    [[ $? -ne 0 ]] && die "rtcwake return value error"
+    sudo rtcwake -m mem -s "${sleep_lst[$i]}" ||
+        die "rtcwake returned $?"
     dlogc "sleep for ${wait_lst[$i]}"
     sleep ${wait_lst[$i]}
     dlogi "Check for the kernel log status"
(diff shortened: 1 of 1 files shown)


marc-hb:
None of the two errors in https://sof-ci.01.org/softestpr/PR899/build15/devicetest/ are new.

test-case/check-suspend-resume.sh
```diff
@@ -48,7 +48,7 @@ if [ "$type" ]; then
     # check for type value effect
     grep -q "$type" /sys/power/mem_sleep || {
         grep -H '^' /sys/power/mem_sleep
-        skip_test "Unsupported sleep type argument: $type"
+        die "Unsupported sleep type argument: $type"
```
paulstelian97, line 52: Two changes on the same thing? Feels odd but okay...

paulstelian97 (commented):
I like this for the most part, the fact that the new code is more readable (on one of the commits I didn't understand the old code but I kinda do the new one, so I leave it to you to confirm that functionality doesn't really change) is a good thing.

Just one comment that I want to get clarified.

test-case/check-suspend-resume.sh
marc-hb: The first commit here is a no change cleanup that will stay. The second one switches the SKIP to a FAIL, it could be reverted if it is found to break "backwards compatibility" somewhere.

greg-intel: What does the following line do here?
`grep -H '^' /sys/power/mem_sleep`

greg-intel (approved):

test-case/check-suspend-resume.sh
marc-hb: It's just a very fancy `cat`; try it on any file and you will see every line is prefixed by the file name.

marc-hb:
`sudo()` fix submitted in:
- #943

PR #904, opened 2022-05-11
tplgtool2: compatible with construct 2.8-2.10

miRoox:
Conditional import to ensure tplgtool2 compatible with construct 2.8-2.10.
To migrate Ubuntu 20.04 (construct 2.8.16) to 22.04 (construct 2.10.67).

see also:
- https://packages.ubuntu.com/jammy/python3-construct
- https://github.com/construct/construct/blob/master/docs/transition29.rst
- https://github.com/construct/construct/blob/master/docs/transition210.rst

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -8,19 +8,32 @@ import re
 import sys
 import typing
 from collections import defaultdict
-from construct import this, Container, ListContainer, Struct, Const, Switch, Select, Array, Bytes, GreedyBytes, GreedyRange, FocusedSeq, Pass, Padded, Padding, Prefixed, String, Flag, Byte, Int16ul, Int32ul, Int64ul, Terminated
+from construct import this, Container, ListContainer, Struct, Switch, Select, Array, Bytes, GreedyBytes, GreedyRange, FocusedSeq, Pass, Padded, Padding, Prefixed, Flag, Byte, Int16ul, Int32ul, Int64ul, Terminated
 from dataclasses import dataclass
 from functools import cached_property, partial
 
-def __enum_dict(enumtype) -> dict:
-    return dict((e.name, e.value) for e in enumtype)
+# Pylint complain about the missing names even in conditional import.
+# pylint: disable=E0611
+if construct.version < (2, 9, 0):
+    from construct import Const, String
 
-# here is a workaround to make Enum/FlagsEnum accept enum type, like construct 2.9
-def Enum(subcon, enumtype):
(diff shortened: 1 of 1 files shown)


tools/tplgtool2.py
```diff
@@ -8,19 +8,25 @@
...
+    from construct import String
+
+    def __enum_dict(enumtype):
+        return dict((e.name, e.value) for e in enumtype)
+    # here is a workaround to make Enum/FlagsEnum accept enum type, like construct 2.9
+    def Enum(subcon, enumtype):
+        return construct.Enum(subcon, default=Pass, **__enum_dict(enumtype))
+    def FlagsEnum(subcon, enumtype):
+        return construct.FlagsEnum(subcon, **__enum_dict(enumtype))
+if (2, 9, 0) <= construct.version:
```
marc-hb, line 27: skip a line before this?

tools/tplgtool2.py
```diff
@@ -8,19 +8,25 @@
...
-    return dict((e.name, e.value) for e in enumtype)
-
-# here is a workaround to make Enum/FlagsEnum accept enum type, like construct 2.9
-def Enum(subcon, enumtype):
-    return construct.Enum(subcon, default=Pass, **__enum_dict(enumtype))
-
-def FlagsEnum(subcon, enumtype):
-    return construct.FlagsEnum(subcon, **__enum_dict(enumtype))
+# Pylint complain about the missing names even in conditional import.
+# pylint: disable=E0611
```
marc-hb, line 16: Can you move this useful link from the commit message to a comment?
https://github.com/construct/construct/blob/master/docs/transition29.rst

marc-hb (commented):
Why block < 2.8.0 now? Only because it's not tested? It wasn't blocked before.

Whether blocking older versions is the right decision or not, I don't like adding this block AND support for newer version _in the same commit_.

tools/tplgtool2.py
miRoox: I append it after the condition.

miRoox:
Tested http://sof-ci.sh.intel.com/#/result/planresultdetail/12505?model=APL_UP2_NOCODEC&testcase=verify-tplg-binary, it works on Ubuntu 22.04.

marc-hb:
SOFCI TEST

marc-hb:
SOFCI TEST

aiChaoSONG (approved):

aiChaoSONG:
merge for the compatibility with ubuntu 22.04

PR #911, opened 2022-05-25
lib.sh: lower BYT expectations in new storage_checks()

marc-hb:
Storage performance on some BYT devices is incredibly slow, see for
instance failures in
https://sof-ci.01.org/linuxpr/PR3552/build164/devicetest


greg-intel (approved):
So apl falls through to the default settings now?  I'm assuming these settings have been tested, approving.

fredoh9 (approved):
thank you for the update

PR #914, opened 2022-05-27
case-lib: add an option to control fallback behavior

aiChaoSONG:
Previously, if SOF card is not detected from the system,
we will export pipelines from proc file system, this is
good for legacy HDA test, but for SOF test, this causes
false positive because usually under this condition, we
are testing USB sound card or Nvidia HDMI.

This patch adds an option to control the fallback behavior,
It should be set to 'false' for SOF test,  and set to 'true'
for legacy HDA test.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -70,3 +70,14 @@ SUDO_LEVEL=${SUDO_LEVEL:-}
 # waited by the internal test runner used by sof/jenkins CI and that
 # test runner does not define SOF_TEST_INTERVAL (internal bug 158)
 SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-5}
+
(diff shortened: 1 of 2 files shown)


marc-hb (approved):
I would prefer new modes like "TEST_HDA" or "TEST_USB" but it would be much work, this is quick and simple and gets the job done without too much technical debt. Thx for the quick fix!

keqiaozhang (approved):

plbossart (approved):
Ideally we would have a more explicit parameter to e.g. test on USB only or NVIDIA only, but this is good enough for our own tests.

Thanks @aiChaoSONG !

aiChaoSONG:
The new env is setup on our two RKL devices, merging

PR #916, opened 2022-06-07
check-suspend-resume: diff wakeup_sources with --report-identical-files

marc-hb:
When the wakeup count gets totally stuck it's possible for the
wakeup_sources to stay completed unchanged. See example at
https://sof-ci.01.org/sofpr/PR5884/build560/devicetest/?model=TGLU_UP_HDA_ZEPHYR&testcase=check-suspend-resume-5

In that case only the header of wakeup_sources gets printed which gives
the wrong impression that wakeup_sources is completely empty! The
`--report-identical-files` diff option avoids that confusion.

### DIFF
--- a/test-case/check-suspend-resume.sh
+++ b/test-case/check-suspend-resume.sh
@@ -109,8 +109,11 @@ dump_stats()
     printf '\n'
     grep ^ /sys/power/suspend_stats/*
     printf '\n'
+    # Show colunm names
     sudo cat /sys/kernel/debug/wakeup_sources | head -n 1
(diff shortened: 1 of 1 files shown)


marc-hb:
The only suspend/resume failures in https://sof-ci.01.org/softestpr/PR916/build45/devicetest/ are caused by a `sync` lasting more than 5s, which is unrelated to this PR.

Unrelated PR #917 increases the sync duration from 5s to 7s

greg-intel (approved):

marc-hb:
Thanks for the prompt review. Very small and minor change, merging.

PR #917, opened 2022-06-07
lib.sh: increase max `sync` duration from 5s to 7s

marc-hb:
There has been a number of test runs failing with a sync duration just
slightly over 5s. I think 5+ seconds is wrong and should not happen but
we have bigger problems right now, let's revisit and lower this back
later.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -199,7 +199,7 @@ storage_checks()
     case "$platf" in
         # Some cheap and old BYTs with eMMC do 2MB/s write or less!
         byt) megas=4 ; max_sync=25 ;;
-        *) megas=100; max_sync=5 ;;
+        *) megas=100; max_sync=7 ;;
     esac
 
     ( set -x


paulstelian97:
What does this truly do? Potentially slow down the tests to avoid some spurious fails?

marc-hb:
If you "scroll down" (Github shows really too little context by default, not sure why) you can see this is used in `time timeout -s CONT "$max_sync" sudo sync` (and only there). So this does not change the test duration at all, it only changes the PASS/FAIL status when the `sync` takes between 5s and 7s.

plbossart (approved):

greg-intel (approved):

marc-hb:
Interestingly, https://sof-ci.01.org/softestpr/PR917/build46/devicetest/?model=BYT_MB_NOCODEC&testcase=check-playback-3times failed because the `sync` lasted 5.29865 s

PR #918, opened 2022-06-10
check-suspend-resume: add new -u option to test without audio

marc-hb:
This should help identifying suspend/resume issues unrelated to audio

2 commits must be reviewed separately

### DIFF
--- a/test-case/check-suspend-resume.sh
+++ b/test-case/check-suspend-resume.sh
@@ -35,6 +35,9 @@ OPT_HAS_ARG['T']=1         OPT_VAL['T']=""
 OPT_NAME['S']='sleep'    OPT_DESC['S']='suspend/resume command:rtcwake sleep duration'
 OPT_HAS_ARG['S']=1         OPT_VAL['S']=5
 
+OPT_NAME['u']='unload-audio'  OPT_DESC['u']='unload audio modules for the test'
+OPT_HAS_ARG['u']=0            OPT_VAL['u']=0
+
 OPT_NAME['w']='wait'     OPT_DESC['w']='idle time after suspend/resume wakeup'
 OPT_HAS_ARG['w']=1         OPT_VAL['w']=5
 
@@ -132,6 +135,21 @@ main()
 # TODO: remove this after issue fixed.
     sleep 1
 
+    local keep_modules=true already_unloaded=false
+
+    if [ ${OPT_VAL['u']} = 1 ]; then
+        keep_modules=false
+    fi
+
+    sudo lsmod | grep -q snd.sof || {
+        already_unloaded=true
+        $keep_modules ||
+            dlogw 'modules already unloaded, ignoring option -u!'
+    }
+
+    $already_unloaded || $keep_modules || "$TOPDIR"/tools/kmod/sof_remove.sh ||
+        die "Failed to unload audio drivers"
+
     expected_wakeup_count=$(cat /sys/power/wakeup_count)
     expected_stats_success=$(cat /sys/power/suspend_stats/success)
     save_initial_stats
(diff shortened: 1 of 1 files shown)


test-case/check-suspend-resume.sh
```diff
@@ -132,13 +135,25 @@ main()
...
+
     expected_wakeup_count=$(cat /sys/power/wakeup_count)
     expected_stats_success=$(cat /sys/power/suspend_stats/success)
     save_initial_stats
     for i in $(seq 1 $loop_count)
     do
         sleep_once "$i"
     done
+
+    $keep_modules || "$TOPDIR"/tools/kmod/sof_remove.sh ||
```
plbossart, line 154: should this be sof_insert? I am not sure why we use sof_remove twice, unless it's a copy-paste issue

marc-hb: Good catch, thank you! Late Friday code...

Fixed. Also added logic not to do anything when the drivers are already unloaded.

marc-hb:
Unrelated `snap` issue on https://sof-ci.01.org/softestpr/PR918/build49/devicetest/?model=TGLU_UP_HDA_ZEPHYR&testcase=verify-kernel-boot-log, working on it.

test-case/check-suspend-resume.sh
```diff
@@ -132,13 +135,33 @@ main()
...
+        die "Failed to unload audio drivers"
+
     expected_wakeup_count=$(cat /sys/power/wakeup_count)
     expected_stats_success=$(cat /sys/power/suspend_stats/success)
     save_initial_stats
     for i in $(seq 1 $loop_count)
     do
         sleep_once "$i"
     done
+
```
plbossart, line 160: should there be a checkpoint here before reloading audio drivers?

marc-hb: There could be but it's not really necessary because one of the last thing `sleep_once` did is to check the kernel logs and `die` if that failed. So if the execution makes it to this point it means the kernel logs are free of errors so far.

plbossart: I don't disagree, it's more that if you get an error while loading audio drivers it might be a bit difficult to check if the test worked until the point where you start loading driver. You'd get a fail and no intermediate information, that's it.

marc-hb: You will see the difference because of the different `die` error message; one is `dump_and_die "Caught error in kernel log"` whereas the other one below is  `die "Found kernel error after reloading audio drivers"`

Setting the `$KERNEL_CHECKPOINT` is completely quiet so it won't help in that respect.

marc-hb: Think of `$KERNEL_CHECKPOINT` as a way to avoid errors from previous tests but inside the same test looking for errors and dying is actually a "stronger" checkpoint.

plbossart: I will have to trust you on that, I don't have any context here.

plbossart (approved):
LGTM

marc-hb:
There is another, theoretical problem with `$KERNEL_CHECKPOINT`: it's not "atomic". In other words, kernel errors between `sof-kernel-log-check.sh` and setting `$KERNEL_CHECKPOINT` are theoretically missed! This is not a real problem in practice because:
- we usually do nothing in that window
- we "cheat" and actually check KERNEL_CHECKPOINT - N seconds! Ugly workaround.

But still, `journalctl` performance issues and other exceptions aside I think we should generally set KERNEL_CHECKPOINT only once at the start of a test to be really sure not to miss any error.

@aiChaoSONG , @fredoh9 thoughts?

PR #923, opened 2022-06-17
test-case: manual: Simple probes functionality test

jsarha:
Signed-off-by: Jyri Sarha <jyri.sarha@intel.com>

This is very simple and fast put together test for probe capture functionality.

### DIFF
--- a/test-case/manual/check-simple-probe-capture.md
+++ b/test-case/manual/check-simple-probe-capture.md
@@ -1,0 +1,89 @@
+# Probes test
+Simple probes functionality test.
+
+# Preconditions
+1. Linux kernel should have these config options enabled:
+   CONFIG_SND_SOC_SOF_DEBUG_PROBES=y
+   CONFIG_SND_SOC_SOF_HDA_PROBES=y
+
+   These options are not user selectable but are automatically
+   selected by the SOF platform options supporting probes. These
+   platforms select SND_SOC_SOF_INTEL_SKL/APL/CNL/ICL/TGL/MTL. Those
+   options in term select SND_SOC_SOF_HDA_COMMON, which selects
+   SND_SOC_SOF_HDA_PROBES, which selects SND_SOC_SOF_DEBUG_PROBES.
+
+2. Firmware should have these config options enabled:
+   CONFIG_PROBE=y
+
+   The following defaults are Ok:
+   CONFIG_PROBE_POINTS_MAX=16
+   CONFIG_PROBE_DMA_MAX=4
+
+   In fact PROBE_DMA_MAX is ignored as inection is not currently supported
+
+2. Enable the module, by creating file /etc/modprobe.d/sof-probes.conf, with
+   following line as content:
+   options snd_sof_probes enable=1
+
+3. Add following line to /etc/modprobe.d/sof.conf:
+   options snd slots=,,,snd_sof_probes
+
+   to make the sound card for the probes to be locked at card3.
+
+4. Have tiny compress installed:
+   git clone https://github.com/alsa-project/tinycompress.git
+   sudo ./gitcompile --prefix= $OUT
+   sudo make install
+
+5. Have sof-probes from sof/tools/probes installed
+
+
+# Test Description
+* Enable probe capture, start playback, bind the probe capture to a buffer
+  in playback pipeline, convert the capture to wav, verify the wav contains
+  the content played.
+
+  The probes functionality is described in more generic way here:
+  https://github.com/thesofproject/sof-docs/blob/master/developer_guides/debugability/probes/index.rst
+
+# Simple probe capture case
+## Check preconditions
+1. Look for comprY under /proc/asound/cardX/, you should find for example:
+   /proc/asound/card3/compr0/
+2. Check that you have a working tiny compress "crecord" binary available.
+3. Check that you have a working sof tools "sof-probes" binary available.
+
+## Run Instructions
+1. Start a capture on probe alsa device (3 and 0 are from
+   /proc/asound/card3/compr0):
+   ```
+   crecord -c3 -d0 -b8192 -f4 -FS32_LE -R48000 -C4 /tmp/extract.wav
+   ```
+2. Start a playback
+   ```
+   aplay -Dplughw:0,0 test.wav -v -q
+   ```
+
+3. Bind probe capture to a buffer on playback pipeline (as super user):
+   ```
+   echo 36,1,0 >  /sys/kernel/debug/sof/probe_points
+   ```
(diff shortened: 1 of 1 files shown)


kv2019i (commented):
Excellent, thanks @jsarha . I think this is really useful to document the test process.

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,78 @@
+# Probes test
+Simple probes functionality test.
+
+# Preconditions
+1. Linux kernel should have these optios enabled:
```
fredoh9, line 5: Typo: options
but 'option' is confusing to me, should we say CONFIG

fredoh9 (commented):
Overall looks good and thank you for the update.
Almost ready to merge, good clarity for manual test note. When you update it please make it normal PR.

test-case/manual/check-simple-probe-capture.md
jsarha: Thanks, will apply.

aiChaoSONG (approved):
@jsarha Are those configurations in this document all supported by thesofproject/sof and thesofproject/linux? So I can try manually.

jsarha:
> @jsarha Are those configurations in this document all supported by thesofproject/sof and thesofproject/linux? So I can try manually.

Yes, they are. Also the both PRs fixing the problems related to this are now merged. I wrote the test descriptions based on how I got the feature working last week, so I think it should be good.

marc-hb:
The form looks great but I'm ignorant about the content. @jsarha could you get a couple more reviewers?

ranj063 (approved):

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,77 @@
+# Probes test
+Simple probes functionality test.
+
+# Preconditions
+1. Linux kernel should have these config options enabled:
+   CONFIG_SND_SOC_SOF_DEBUG_PROBES=y
+   CONFIG_SND_SOC_SOF_HDA_PROBES=y
```
ujfalusi, line 7: I would note that these options are _not_ user selectable.

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,77 @@
...
+
+2. Firmware should have these config options enabled:
+   CONFIG_PROBE=y
+   CONFIG_PROBE_POINTS_MAX=16
+   CONFIG_PROBE_DMA_MAX=4
```
ujfalusi, line 12: The PROBE_DMA_MAX is for injection which is not supported on Linux and I dare to say that it has never been tested?
I would not mention it for now. The extraction works just fine with the default value of 0.

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,77 @@
...
+
+2. Enable the module, by creating file /etc/modprobe.d/sof-probes.conf, with
+   following line as content:
+   options snd_sof_probes enable=1
+
+3. Add following line to /etc/modprobe.d/sof.conf:
```
ujfalusi, line 18: to make the sound card for the probes to be locked at card3

ujfalusi (commented):
@jsarha, great to have something for the probes!
I have couple candidates for clarification, if you think they stand.

test-case/manual/check-simple-probe-capture.md
jsarha: Should be now fixed.

test-case/manual/check-simple-probe-capture.md
jsarha: Should be now fixed.

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,86 @@
...
+   'make menuconfig'. Edit .config manually instead.
+
+2. Firmware should have these config options enabled:
+   CONFIG_PROBE=y
+
+   The following defaults are Ok:
+   CONFIG_PROBE_POINTS_MAX=16
+   CONFIG_PROBE_DMA_MAX=4
+
+   In fact PROBE_DMA_MAX is ignored as inection is not currently supported
```
ujfalusi, line 19: s/inection/injection
Why mention CONFIG_PROBE_DMA_MAX in the first place?

test-case/manual/check-simple-probe-capture.md
```diff
@@ -0,0 +1,86 @@
+# Probes test
+Simple probes functionality test.
+
+# Preconditions
+1. Linux kernel should have these config options enabled:
+   CONFIG_SND_SOC_SOF_DEBUG_PROBES=y
+   CONFIG_SND_SOC_SOF_HDA_PROBES=y
+
+   These options are not user selectable and can not be enabled with
+   'make menuconfig'. Edit .config manually instead.
```
ujfalusi, line 10: No, you can not edit the .config, these options are config selected:
SND_SOC_SOF_INTEL_SKL/APL/CNL/ICL/TGL/MTL selects SND_SOC_SOF_HDA_COMMON, which selects SND_SOC_SOF_HDA_PROBES, which SND_SOC_SOF_DEBUG_PROBES.
The probes client is only available for HDA platforms.

ujfalusi (changes requested):

test-case/manual/check-simple-probe-capture.md
ujfalusi: For me it comes with 0 as default, but never checked the kconfig file to be honest..

test-case/manual/check-simple-probe-capture.md
marc-hb: I don't know about these CONFIG_ in particular but in general `.config` should never be edited manually. Any "real world" Kconfig system has complex dependency rules and manual edition is likely to break consistency.

https://docs.zephyrproject.org/2.6.0/guides/kconfig/setting.html#visible-and-invisible-kconfig-symbols
> An invisible symbol is a symbol without a prompt. Invisible symbols are not shown in the interactive configuration interfaces, and users have no direct control over their value. They instead get their value from defaults or from other symbols.

If some hack is really required locally and temporarily, it should be a hack in a Kconfig file, not in the .config file.

(yes I know it's the kernel but `kconfig.py` is bug-for-bug compatible and has much better documentation. It's what we use in SOF too)

I see this has been removed in the current version.

kv2019i:
@jsarha @ujfalusi  Would be nice to get a version of this merged. I had to look this up to re-learn how to use probes ... and I mean, I thought this was merged, so it took some time before I realized and I need to look it up here in open PRs....

PR #926, opened 2022-07-07
[RFC]: Add initial patches for sof-test i.MX support

LaurentiuM1234:
We want to use sof-test in our setup which is based on YOCTO distro. Please let us know if you think sof-test can be multiplatform.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -81,3 +81,6 @@ SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-5}
 # Refer to: https://github.com/thesofproject/sof-test/issues/471 and
 # https://github.com/thesofproject/sof-test/issues/913 for more information.
 FALLBACK_TO_PROC=${FALLBACK_TO_PROC:-false}
+
+# Skip the storage checks if they aren't required
+ENABLE_STORAGE_CHECKS=${ENABLE_STORAGE_CHECKS:-1}
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -102,7 +102,9 @@ function func_exit_handler()
 
     stop_test || exit_status=1
 
-    storage_checks || exit_status=1
+    if [ "$ENABLE_STORAGE_CHECKS" = '1' ]; then
+        storage_checks || exit_status=1
+    fi
 
     if [[ "$KERNEL_CHECKPOINT" =~ ^[0-9]{10} ]]; then
         # Do not collect the entire duration of the test but only the
--- a/case-lib/opt.sh
+++ b/case-lib/opt.sh
@@ -176,7 +176,7 @@ func_opt_parse_option()
             [[ "$TPLG" ]] && printf "TPLG=%s" "$TPLG "
             printf "%s %s\n" "$SCRIPT_NAME" "$SCRIPT_PRAM"
             printf 'Commit:\n'
-            git -C "$SCRIPT_HOME" log --oneline --decorate -n 5
+            git -C "$SCRIPT_HOME" log --oneline --decorate -n 5 || true
         } >> "$LOG_ROOT/version.txt"
     fi
 


case-lib/lib.sh
```diff
@@ -222,12 +224,12 @@ storage_checks()
...
       time timeout -s CONT "$max_sync" sudo sync || return $?
       # Spend a few seconds to test and show the current write speed
-      timeout -s CONT 5 dd if=/dev/zero of=~/HD_TEST_DELETE_ME bs=1M count="$megas" conv=fsync ||
+      timeout -s CONT 5 dd if=/dev/zero of=$WRITE_SPEED_CHECK_DUMMY_FILE bs=1M count="$megas" conv=fsync ||
           return $?
       time timeout -s CONT "$max_sync" sudo sync
     ) || return $?
 
-    rm ~/HD_TEST_DELETE_ME
+    rm $WRITE_SPEED_CHECK_DUMMY_FILE
```
plbossart, line 232: Also not sure why this write check was added unconditionally, it was meant to debug some issues with mass storage but I don't think it needs to be a key feature that is run in all cases on all platforms, it should be opt-in. @marc-hb this is likely for you to comment on.

plbossart (commented):
It'd be fantastic if sof-test is run on multiple platforms, thanks @LaurentiuM1234

marc-hb (changes requested):
We should absolutely make the code able to run without git but not like this.

dbaluta:
@LaurentiuM1234 @plbossart thanks for comments. We will go with your suggestions and update this PR.

LaurentiuM1234:
Thanks for all the comments @marc-hb @plbossart ! Following your suggestions, I have made the following adjustments:
1) The `git log` command is now piped to the `true` command so as to allow the tests to run on platforms which don't have the `git` command. This modification avoids all the extra code you have pointed out.
2) The storage checks are now done based on the value of the SKIP_STORAGE_CHECKS env variable. Based on this: 

> We're current replacing the drives and should turn this off soon.
> If you can't wait and really need to turn this off right now then add a conditional around the only place where storage_checks is invoked. It's quite strange to keep running a test that write from /dev/zero to /dev/null and prints some pointless logs.

I have decided to use an env variable as a temporary solution that will allow the tests to run on i.MX until the storage checks are turned off.

case-lib/opt.sh
```diff
@@ -176,7 +176,7 @@ func_opt_parse_option()
             [[ "$TPLG" ]] && printf "TPLG=%s" "$TPLG "
             printf "%s %s\n" "$SCRIPT_NAME" "$SCRIPT_PRAM"
             printf 'Commit:\n'
-            git -C "$SCRIPT_HOME" log --oneline --decorate -n 5
+            git -C "$SCRIPT_HOME" log --oneline --decorate -n 5 | true
```
marc-hb, line 179:
```suggestion
            git -C "$SCRIPT_HOME" log --oneline --decorate -n 5 || true
```
I'm sorry my earlier typo mislead you.

Compare:
```shell
printf foo | true
printf foo || true
```

marc-hb (changes requested):
Almost there

case-lib/hijack.sh
```diff
@@ -102,7 +102,9 @@ function func_exit_handler()
 
     stop_test || exit_status=1
 
-    storage_checks || exit_status=1
+    if [ ! "$SKIP_STORAGE_CHECKS" -eq 1 ]; then
```
marc-hb, line 105: Nit
```suggestion
    [ "$SKIP_STORAGE_CHECKS" -eq 1 ] || storage_checks ||
        exit_status=1
```
The fewer negations the better and `!` can't be copied to the command-line.

marc-hb (changes requested):

marc-hb (approved):

case-lib/config.sh
```diff
@@ -81,3 +81,6 @@ SOF_TEST_INTERVAL=${SOF_TEST_INTERVAL:-5}
 # Refer to: https://github.com/thesofproject/sof-test/issues/471 and
 # https://github.com/thesofproject/sof-test/issues/913 for more information.
 FALLBACK_TO_PROC=${FALLBACK_TO_PROC:-false}
+
+# Skip the storage checks if they aren't required
+SKIP_STORAGE_CHECKS=${SKIP_STORAGE_CHECKS:-0}
```
marc-hb, line 86: Change of mind sorry, can you please remove the negation and make this `ENABLE_STORAGE_CHECKS, defaulting to 1`? Later we can just change the default to zero.

case-lib/hijack.sh
marc-hb: Change of mind sorry
```shell
if [ "$ENABLE_STORAGE_CHECKS" = '1' ]; then
   storage_checks || exit_status=1
fi
```
Zero negation; no need.

LaurentiuM1234:
Thanks again for the suggestions @marc-hb !

1) Regarding the git command: this was a slip on my behalf. I should have ran `shellscript` locally and this could probably have been avoided.
2) Regarding the storage checks: I agree the negation was not necessary. Now the code should be much more clear.

dbaluta (approved):

marc-hb (approved):
This one is small and has been in review for 5 days - merging.

PR #927, opened 2022-07-07
[RFC]: Add i.MX tests for module insert/removal

LaurentiuM1234:
This depends on PR: https://github.com/thesofproject/sof-test/pull/926

### DIFF
--- a/tools/kmod/sof_insert.sh
+++ b/tools/kmod/sof_insert.sh
@@ -75,6 +75,11 @@ insert_module snd_soc_max98390
 insert_module snd_soc_es8316
 insert_module snd_soc_es8326
 
+insert_module snd_soc_wm8960
+
+# core SOF driver
+insert_module snd_sof
+
 # insert top-level ACPI/PCI SOF drivers. They will register SOF components and
 # load machine drivers as needed. Do not insert any other sort of audio module,
 # code dependencies will be used to load the relevant modules.
@@ -95,6 +100,17 @@ insert_module snd_sof_pci_intel_cnl
 insert_module snd_sof_pci_intel_icl
 insert_module snd_sof_pci_intel_tgl
 insert_module snd_sof_pci_intel_mtl
+
+# OF driver
+insert_module snd_sof_of
+
+# helper drivers
+insert_module imx_common
+insert_module snd_sof_xtensa_dsp
+
+# i.MX specific drivers
+insert_module snd_sof_imx8
+insert_module snd_sof_imx8m
 
 # USB
 insert_module snd_usb_audio
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -88,6 +88,13 @@ remove_module snd_sof_pci_intel_mtl
 remove_module snd_sof_acpi_intel_byt
 remove_module snd_sof_acpi_intel_bdw
 
+#-------------------------------------------
+# Top level devices
+# i.MX-specific drivers
+#-------------------------------------------
+remove_module snd_sof_imx8
+remove_module snd_sof_imx8m
+
 #--------------------------------------------------
 # With older kernels this is in use by snd_sof_pci,
 # see https://github.com/thesofproject/linux/pull/2683
@@ -100,6 +107,7 @@ remove_module snd_sof_intel_hda_common || true
 remove_module snd_sof_acpi
 remove_module snd_sof_pci
 remove_module snd_sof_intel_atom
+remove_module imx_common
 
 #-------------------------------------------
 # legacy drivers (not used but loaded)
@@ -164,6 +172,11 @@ remove_module snd_sof_ipc_test
 remove_module snd_sof_ipc_flood_test
 remove_module snd_sof_ipc_msg_injector
 remove_module snd_sof_dma_trace
+
+#-------------------------------------------
+# SOF OF driver
+#-------------------------------------------
+remove_module snd_sof_of
 
 # snd_sof_nocodec dependencies re-ordered
 # in https://github.com/thesofproject/linux/pull/2800
@@ -231,6 +244,8 @@ remove_module snd_soc_dmic
 
 remove_module snd_hda_codec_realtek
 remove_module snd_hda_codec_generic
+
+remove_module snd_soc_wm8960
 
 #-------------------------------------------
 # Remaining core SOF parts


tools/kmod/sof_remove.sh
```diff
@@ -2,206 +2,13 @@
...
-remove_module snd_hda_codec
-remove_module snd_hda_core
-remove_module snd_hwdep
-remove_module snd_compress
-remove_module snd_pcm_dmaengine
-remove_module snd_pcm
-
+TOPDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
+
+case $PLATFORM in
```
plbossart, line 7: how is PLATFORM set?

plbossart (commented):
Sounds good to me, just not clear on the new PLATFORM variable.

tools/kmod/sof_remove.sh
marc-hb: "This depends on PR: https://github.com/thesofproject/sof-test/pull/926"

marc-hb (commented):
~I don't understand why this required~, the current code is already able to skip missing or unloaded modules, it already does that routinely. Can you explain why you can't just append other modules to the existing lists? What problem does this solve?

Also, are all `snd_soc_*` really Intel-specific? If they are, then why are some modules named `snd_soc_intel_*`? Also, where would `amd/snd-soc-acp-rt5645-mach.ko` fit?

marc-hb (changes requested):

tools/kmod/sof_remove.sh
dbaluta: @marc-hb very good suggestion. One single comment about:

```
 for remove_sub_script in "$TOPDIR"/tools/kmod/sof_remove_*.sh; do
         "$remove_sub_script"
    done
```

I'm not sure this is the correct way to run the ```sof_remove* scripts``` as for example it doesn't make any sens to try to remove IMX modules on Intel platform or viceversa.

tools/kmod/sof_insert_intel.sh
```diff
@@ -0,0 +1,88 @@
...
+insert_module snd_sof_pci_intel_cnl
+insert_module snd_sof_pci_intel_icl
+insert_module snd_sof_pci_intel_tgl
+insert_module snd_sof_pci_intel_mtl
+
+# USB
+insert_module snd_usb_audio
+
+# without the status check force quit
+builtin exit 0
```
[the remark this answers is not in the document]
dbaluta: I'm not even sure what does this mean. Anyhow, @LaurentiuM1234 please create a separate commit and remove this.

LaurentiuM1234:
Thanks for the comments @plbossart @marc-hb  ! Based on your suggestions, I have made the following changes:

1) Based on:

>This line is very old and obsolete, feel free to drop it. It predates trap, before trap there was a custom exit function overriding the default one. That's gone.

I have decided to remove said line.

2) Based on:

>No copy/paste/diverge please. Create new functions in lib.sh if needed. Is a function needed though? Why not keep it in the main sof_remove.sh script? Because you don't want it for IMX? Which parts don't you want more specifically and why?

It was a poor choice from my part to copy and paste those lines instead of adding them in one place. Now all of those pre-insert/pre-remove steps have been moved to a setup function which in turn is called by the main function in sof_{remove/insert}.sh.

3) sof_{remove/insert}.sh now contains a main function in which the appropriate module insertion/removal script is called based on the first parameter received from CL. Reason: https://github.com/thesofproject/sof-test/issues/740

4) I have added a new parameter to check-kmod-load-unload{-after-playback}.sh which in turn is used when calling sof_{remove/insert}.sh. The purpose of this parameter is to allow sof_{remove/insert}.sh to call the appropriate module
insertion/removal scripts.

marc-hb:
@LaurentiuM1234 please add a `# shellcheck source=...` line before each `source` command, see examples in other scripts in the same repo, This should fix most of the shellcheck warnings. Then please take a look at the remaining warnings. To use it that line, invoke `shellcheck` with the `-x` option at the repo's top-level. 

(I didn't review the current version yet)

marc-hb:
I don't know why yet but the consistent check-sof-logger failures in https://sof-ci.01.org/softestpr/PR927/build63/devicetest/ really seem caused by this.

marc-hb (changes requested):
First, thanks for the stellar breakdown into multiple commits, it makes the review so much easier, I wish we'd see that more often!

It will also help `git blame` work across renames.

Looks mostly good to me except for the new mandatory "$platform" argument. @plbossart convinced me this new argument is a good thing but I do not agree it should be _mandatory_. It should also not be called "platform", more on that below.

> I'm not sure this is the correct way to run the sof_remove* scripts as for example it doesn't make any sens to try to remove IMX modules on Intel platform or viceversa.

I see a few reasons why it makes sense to have a stupid `for` loop that runs all the scripts for all platforms when no $platform argument is passed
- Backward-compatibility. This script is invoked in a number of places and not just in sof-test.
- "snd_soc_sof_sdw is intel specific but we forgot the prefix. So there will be some interesting cases to deal with." => there will be corner cases and bugs and then a stupid `for` loop will help
- I know Yocto is all about optimizing and keeping only the absolute minimum necessary but "traditional" Linux distributions are not like that.
- Someone could actually want to try to load and unload modules from all vendors even when some don't drive any hardware for increased test coverage
- If you don't want to use the stupid default for IMX/yocto then simply ignore it and don't use it and it will make no difference to you. It will be only 3 lines of sof-test code.

As I wrote above, the new argument must not be called "platform" not because the name "platform" is super vague and sucks (even when it does) but because it's already being used all across sof and sof-test for making the difference between intel product generations. Just `git grep -i platform` in sof and sof-test and you will see. I suggest `$vendor`?

marc-hb:
> > Please make that an optional command line parameter. Environment variables suck, they hide from logs.

> I have added a new [platform->vendor] parameter to check-kmod-load-unload{-after-playback}.sh which in turn is used when calling sof_{remove/insert}.sh

This parameter is also needed in `check-sof-logger.sh` (which could be the reason why https://sof-ci.01.org/softestpr/PR927/build63/devicetest/ failed) and also needed in `check-suspend-resume.sh` (recently added) and maybe others in the future. This does not scale, I didn't fully realize this when I asked you to switch to a CLI parameter, apologies. I still hate environment variables (they hide in CI and don't show in logs) but in this case I think your first shot was right.

Please add a new, optional `$VENDOR` (or some better name) to `config.sh` so you don't have to modify any test. Still default to empty = everything, I did not change mind for this.

marc-hb:
Current status just for the record: modified tests are failing in https://sof-ci.01.org/softestpr/PR927/build69/devicetest/ and there is a `config.sh` conflict with your other PR that I just merged.

dbaluta:
> I see a few reasons why it makes sense to have a stupid `for` loop that runs all the scripts for all platforms when no $platform argument is passed

@marc-hb i have no strong preference on this. Current implementation actually follows your advice. If no parameter is given then we iterate over all insert/remove *sh scripts and we ran them.

The problem is that for example trying to insmod Intel drivers on IMX will cause a fail. Similar happens when you try to insert i.MX ARM compiled drivers on Intel.

marc-hb:
> `rmmod: ERROR: Module snd_sof_xtensa_dsp is in use by: snd_sof_intel_hda_common`

> We have this common module snd_sof_xtensa_dsp which sof_remove_imx.sh tries to remove. But this is nonsense. Because sof_remove_imx.sh doesn't correctly know the order of modules to be removed on Intel. And it doesn't need to know.

I think this just shows a copy/paste/diverge issue. So the rationale for this split is: 

> but in practice I don't think we want to have a giant file with all modules managed by all SOF contributors though. Too many changes left and right and too much churn. It's wise to split the parts that are platform specific and let each contributor deal with their own additions.

If better maintenance is the goal then common modules should be in some `common.sh` file, their maintenance should note be copied/pasted/diverged across multiple vendor scripts.

For instance something like this:

```shell

    if [ -n "$VENDOR" ]; then
        "$TOPDIR"/tools/kmod/sof_remove_"$VENDOR".sh || return $?
    fi

    dlogi "VENDOR not specified. Running all removal subscripts."

    for remove_subscript in "$TOPDIR"/tools/kmod/sof_remove_*.sh; do
        dlogi "Currently running: $remove_subscript"
        "$remove_subscript" || return $?
    done

   "$TOPDIR"/tools/kmod/sof_common_remove.sh

```
Insertion should NOT have that problem thanks to dependencies, common modules should not be loaded explicitly.

BTW I'm still curious what stopped you from simply adding IMX modules to the current "giant files".

Speaking of `return $?`, while you're making all these changes and testing them please try to add `set -e` here and there. It's not perfect but it helps a lot (#312)

LaurentiuM1234:
> BTW I'm still curious what stopped you from simply adding IMX modules to the current "giant files".

Because I believe it wouldn't scale very well:

1) 

> You're right @marc-hb but in practice I don't think we want to have a giant file with all modules managed by all SOF contributors though. Too many changes left and right and too much churn. It's wise to split the parts that are platform specific and let each contributor deal with their own additions.

2)

Let's assume we go for the huge script and let's assume sof_remove.sh contains the following statements:

...
remove_module A
remove_module B
...

Let's also assume that A uses B. What if there's another contributor n who has the following modules: A, B and C with A using C and C using B. Wouldn't it be easier for said contributor to just create his own sof_remove_n.sh instead of having to find modules A and B in the initial file and inserting module C in between them? What if the file contains hundreds of lines?

3)

What would be the point of trying to remove/insert let's say hundreds of modules that you know for a fact can't be used on your platform?

Since the modification is not major why not just go for the more scalable option if possible?

marc-hb:
> Contributor 1 has modules A and C. C is being used by A.
> Contributor 2 has modules B and C. B is being used by C.

> Since C is a common module...

vendor-A depends on common-C which depends on vendor-B... how can a common module depend on a vendor module?

marc-hb:
> Wouldn't it be easier for said contributor to just create his own sof_remove_n.sh instead of having to find modules A and B in the initial file and inserting module C in between them? What if the file contains hundreds of lines?

That's more or less what exists already and it has required little effort. The "giant" file already used groups to make that easier, moving these groups to separate files does not seem like a ground-breaking improvement to me and it does complicate the scripting.

> Also, what would be the point of trying to remove/insert let's say hundreds of modules that you know can't be used on your platform?

Because it's simple and it works.

The only potential problem I see is this:
> Too many changes left and right and too much churn

So yes, conflicts are not fun. But they haven't happened in practice yet and based on past history I'm not sure they would have been that frequent. It's not like modules get created or renamed every day.

LaurentiuM1234:
> 



> ```
>     if [ -n "$VENDOR" ]; then
>         "$TOPDIR"/tools/kmod/sof_remove_"$VENDOR".sh || return $?
>     fi
> 
>     dlogi "VENDOR not specified. Running all removal subscripts."
> 
>     for remove_subscript in "$TOPDIR"/tools/kmod/sof_remove_*.sh; do
>         dlogi "Currently running: $remove_subscript"
>         "$remove_subscript" || return $?
>     done
> 
>    "$TOPDIR"/tools/kmod/sof_common_remove.sh
> ```
> 
> 
>      
>   
> 
> What if we had the following case:
> 
> 
> Contributor 1 has modules A and C. C is being used by A.
> Contributor 2 has modules B and C. B is being used by C.
> 
> 
> Since C is a common module then it would have to be placed in sof_common_remove.sh, right? That would mean that either A and B are removed before C or C is removed before A and B. First case wouldn't work on contributor 2's platform because B is used by C so C should be removed first. Second case wouldn't work on contributor 1's platform.

@marc-hb You're right, this is not possible, sorry.

I'll just add the i.MX modules to the existing sof_{remove/insert}.sh scripts.

dbaluta:
I agree we can add the imx modules to current scripts.

tools/kmod/sof_insert.sh
```diff
@@ -96,6 +98,20 @@ insert_module snd_sof_pci_intel_icl
 insert_module snd_sof_pci_intel_tgl
 insert_module snd_sof_pci_intel_mtl
 
+# core SOF driver
+insert_module snd_sof
+
```
dbaluta, line 103: I would move core SOF driver  before `top level ACPI/PCI SOF drivers.

LaurentiuM1234: Thanks ! Fixed.

marc-hb (approved):

marc-hb:
Sorry for the `Outdated` (and now "resolved") comments above, please ignore.

The 2 errors in https://sof-ci.01.org/softestpr/PR927/build79/devicetest/ are outdated.

dbaluta (approved):

PR #928, opened 2022-07-09
lib.sh: storage_check: add 'cht' (BSW Cyan) to slow devices

marc-hb:
See failures in daily run 13851.


marc-hb:
Unrelated suspend/resume failure in https://sof-ci.01.org/softestpr/PR928/build61/devicetest/, everything else green.

keqiaozhang (approved):

PR #929, opened 2022-07-12
Using a shell redirection instead of sof-logger's -o option

marc-hb:
As sof-logger must run as root, `sudo sof-logger -o ...` creates logs/
directories with a mix of user and root content. Use a plain >
redirection instead so all log files are owned by the user running the
script.


marc-hb:
Usual suspend/resume failures in https://sof-ci.01.org/softestpr/PR929/build64/devicetest/, everything else green.

paulstelian97 (approved):
That's what I normally do when manually testing so yea.

PR #930, opened 2022-07-12
[RFC]: Add support for i.MX power and platform dump

LaurentiuM1234:


tools/sof-dump-status.py
```diff
@@ -527,6 +557,13 @@ def dump_dapm(dapm, filter = "all"):
     # key is used. Here we output firmware path according to kernel's
     # match table.
     if ret_args.get('fwpath') is True:
+        sysinfo.loadFw()
+
+        if len(sysinfo.fw_lst) != 0:
+            fw_path = "/lib/firmware/imx/sof"
```
marc-hb, line 563: So `/sys/devices/soc0/soc_id` is IMX-only? Then it should be much clearer in comments and `imx` should be somewhere in the list name.

marc-hb (changes requested):

tools/sof-dump-status.py
```diff
@@ -82,6 +86,16 @@ def loadDMI(self):
             self.dmi[name] = line[idx:]
         pass
 
+    def loadFw(self):
+        self.fw_lst.clear()
```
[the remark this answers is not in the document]
LaurentiuM1234: This was also done for the sake of consistency. You're right, the list shouldn't be reset on a second call. I will fix this.

tools/sof-dump-status.py
```diff
@@ -32,6 +34,8 @@ def __init__(self):
         # some device use acpi-dev instead of pci-dev
         # https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-acpi-dev.c
         self._acpi_ids={"byt":"80860F28", "cht": "808622A8", "bdw":"INT3438"}
+        self._fw_map={"i.MX8QM":"imx8"}
+        self._dsp_map={"i.MX8QM":"556e8000.dsp"}
```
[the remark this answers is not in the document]
LaurentiuM1234: This was just bad naming from my part. This will be fixed.

tools/sof-dump-status.py
```diff
@@ -82,6 +86,16 @@ def loadDMI(self):
...
 
+    def loadFw(self):
+        self.fw_lst.clear()
+        exit_code, output = subprocess.getstatusoutput("cat /sys/devices/soc0/soc_id")
+        if exit_code != 0:
+            return
+        # Check if a mapping exists between SoC ID and firmware name
+        if output not in self._fw_map:
+            return
+        self.fw_lst.append({'name': self._fw_map[output], 'platform': output})
```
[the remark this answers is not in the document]
LaurentiuM1234: > This was just bad naming from my part. This will be fixed.

tools/sof-dump-status.py
```diff
@@ -82,6 +86,16 @@ def loadDMI(self):
             self.dmi[name] = line[idx:]
         pass
 
+    def loadFw(self):
```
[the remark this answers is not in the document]
LaurentiuM1234: I was trying to be consistent with the other `load` function but it just resulted in ambiguous naming. This will be fixed

tools/sof-dump-status.py
```diff
@@ -82,6 +86,16 @@ def loadDMI(self):
...
+        self.fw_lst.clear()
+        exit_code, output = subprocess.getstatusoutput("cat /sys/devices/soc0/soc_id")
```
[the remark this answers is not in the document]
dbaluta: @LaurentiuM1234 @marc-hb OK, so we need something like loadACPI or loadPCI for i.MX which logically should be loadDT.

Anyhow, this is different and we couldn't find an equivalent.

For now, we can go with a function that reads soc_id and it is named `loadDT`. If we use this name then we no longer need to add any comment beucase it is clear that this function is only for DT platforms (e.g i.MX). Similar with how ACPI is used for Intel.

tools/sof-dump-status.py
```diff
@@ -82,6 +86,16 @@ def loadDMI(self):
...
+        if exit_code != 0:
+            return
```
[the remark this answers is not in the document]
LaurentiuM1234: I can't really use stdout for debug/error messages because of the way the script is used but I could try using stderr (altough I don't think it's the best solution for debug messages). Another option would be to use some sort of log file.

tools/sof-dump-status.py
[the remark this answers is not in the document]
marc-hb: > The reason I used getstatusoutput was for the sake of consistency, but if that's not really a must then I will replace it with run.

Thanks! Consistency does matter but in this particular case I think consistency does not make much difference so it's better to stop using obsolete Python functions. This script is very far from perfect, feel free to fix issues like this elsewhere if you want.

paulstelian97 (changes requested):
Too many unrelated whitespace changes, which makes it hard to figure out what is actual changed code vs what is just a whitespace difference.

LaurentiuM1234:
> Too many whitespace changes in unrelated code. If you really do want to do this at least make a separate commit solely for that (don't combine coding style fixes with actual functionality changes in the same commit)

You're right @paulstelian97, I should have double checked before force pushing the changes. Now it should be better. Also, I've split the single commit into multiple ones. This should make the changes much more clearer.

tools/sof-dump-status.py
```diff
@@ -2,10 +2,20 @@
...
 
+
+class NoMapException(Exception):
+    """Raise if some element doesn't have an associated map."""
+
 class clsSYSCardInfo():
     def __init__(self):
+        # i.MX-specific maps
+        self.imx_soc_id_to_fw_name = {"i.MX8QM": "imx8"}
+        self.imx_soc_id_to_dsp_name = {"i.MX8QM": "556e8000.dsp"}
```
marc-hb, line 17: these are constant, make that clearer with capitals at the script's top level.

```suggestion
        IMX_SOC_ID_TO_DSP_NAME = {"i.MX8QM": "556e8000.dsp"}
```

I see you're trying to be consistent with pci_ids below but in this particular case please don't bother. This script has many design issues and these are just CONSTANTS so better do the Right Thing, make them look like constants and make your code more readable. Consistency does matter for complex and/or logical problems, not for defining constants...

tools/sof-dump-status.py
```diff
@@ -82,6 +92,31 @@ def loadDMI(self):
...
+            # load SoC family
+            _loadDT("family")
+            # load SoC revision
+            _loadDT("revision")
+            # load machine name
+            _loadDT("machine")
+        except IOError:
+            # if any of the above calls fails, then we're probably
+            # not on a DT-based platform so there should be no
+            # information in dt_info
```
marc-hb, line 117: This looks a bit weird. Is there no concept of required versus optional attributes?

tools/sof-dump-status.py
```diff
@@ -217,6 +252,23 @@ def loadPower(self):
...
 
+        # check to see if we're on a DT-based platform
+        self.loadDT()
+
+        if self.dt_info:
+            # check to see if we're on i.MX
+            if self.dt_info["soc_id"] in self.imx_soc_id_to_dsp_name:
+                dsp_name = self.imx_soc_id_to_dsp_name[self.dt_info["soc_id"]]
+            else:
+                raise NoMapException("Current DT-based platform doesn't have a soc_id to dsp_name map.")
```
marc-hb, line 263: This error handling is overkill. This is a test script, a stack trace is enough. In fact I would much prefer a stack trace, gets me right into the code and points me straight to where the problem is.

tools/sof-dump-status.py
```diff
@@ -452,7 +519,11 @@ def dump_dapm(dapm, filter = "all"):
         exit(0)
 
     if ret_args['power'] is True:
-        sysinfo.loadPower()
+        try:
+            sysinfo.loadPower()
+        except (IOError, NoMapException) as e:
+            print(e, file=sys.stderr)
+            exit(1)
```
marc-hb, line 526: Don't hide stack traces. This a test tool, stack traces are useful. Just let it fail. This is not C either, no need to implement error handling manually. Exceptions come for free. The less code the better.

tools/sof-dump-status.py
```diff
@@ -2,10 +2,20 @@
 
 import subprocess
 import os
+import sys
+
 from common import format_pipeline, export_pipeline
 
+
+class NoMapException(Exception):
+    """Raise if some element doesn't have an associated map."""
```
marc-hb, line 11: Overkill, see below (plus this looks like a pure duplicate of Python's built-in `KeyError` with no added value)

tools/sof-dump-status.py
```diff
@@ -439,6 +496,16 @@ def dump_dapm(dapm, filter = "all"):
...
     sysinfo = clsSYSCardInfo()
     if ret_args['platform'] is True:
+        # check to see if we're on a DT-based platform
+        sysinfo.loadDT()
+        if sysinfo.dt_info:
+            # check to see if we're on i.MX
+            if sysinfo.dt_info["soc_id"] in sysinfo.imx_soc_id_to_fw_name:
+                print(sysinfo.imx_soc_id_to_fw_name[sysinfo.dt_info["soc_id"]])
+            else:
+                print("Current DT-based platform doesn't have a soc_id to fw_name map.", file=sys.stderr)
```
marc-hb, line 506: See above.

tools/sof-dump-status.py
```diff
@@ -527,6 +598,11 @@ def dump_dapm(dapm, filter = "all"):
     # key is used. Here we output firmware path according to kernel's
     # match table.
     if ret_args.get('fwpath') is True:
+        if os.path.isdir("/lib/firmware/imx/sof"):
+            # we're currently on i.MX, just print the path
```
marc-hb, line 602: This looks like... a bug?
https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/imx

tools/sof-dump-status.py
```diff
@@ -294,6 +346,11 @@ def loadDAPM(self, filter = "all"):
                     self.dapm['name_lst'].append("'%s/%s'" %(path_name ,fname.name))
 
 if __name__ == "__main__":
+    def read_file(filepath):
```
marc-hb, line 349: Why not at the top-level?

marc-hb (changes requested):

tools/sof-dump-status.py
paulstelian97: DT doesn't specify the path to my knowledge, just the platform and the driver hardcodes the path. So for now this is probably correct. See for example the following excerpt from the Linux kernel code:

```
static struct sof_dev_desc sof_of_imx8qxp_desc = {
        .default_fw_path = "imx/sof",
        .default_tplg_path = "imx/sof-tplg",
        .default_fw_filename = "sof-imx8x.ri",
        .nocodec_tplg_filename = "sof-imx8-nocodec.tplg",
        .ops = &sof_imx8x_ops,
};

static struct sof_dev_desc sof_of_imx8qm_desc = {
        .default_fw_path = "imx/sof",
        .default_tplg_path = "imx/sof-tplg",
        .default_fw_filename = "sof-imx8.ri",
        .nocodec_tplg_filename = "sof-imx8-nocodec.tplg",
        .ops = &sof_imx8_ops,
};

static const struct of_device_id sof_of_imx8_ids[] = {
        { .compatible = "fsl,imx8qxp-dsp", .data = &sof_of_imx8qxp_desc},
        { .compatible = "fsl,imx8qm-dsp", .data = &sof_of_imx8qm_desc},
        { }
};
MODULE_DEVICE_TABLE(of, sof_of_imx8_ids);
```

tools/sof-dump-status.py
LaurentiuM1234: You're right. I will move them at the beginning of the script and use capitals.

tools/sof-dump-status.py
LaurentiuM1234: This isn't what I was trying to go for. `NoMapException` is meant to be used when you're trying to map some element to another using multiple maps.
Right now, I don't think that idea is very clear because imx_soc_id_to_dsp_name is the only map between a soc_id and a dsp_name but if more possible maps between soc_id and dsp_name were to be introduced I believe it would make more sense. Example:

```
            if self.dt_info["soc_id"] in self.imx_soc_id_to_dsp_name:
                dsp_name = self.imx_soc_id_to_dsp_name[self.dt_info["soc_id"]]
            elif self.dt_info["soc_id"] in self.platform1_soc_id_to_dsp_name:
                dsp_name = ...
            elif self.dt_info["soc_id"] in self.platform2_soc_id_to_dsp_name:
                dsp_name = ...
            elif self.dt_info["soc_id"] in self.platform3_soc_id_to_dsp_name:
                dsp_name = ...
            else:
                raise NoMapException("Current DT-based platform doesn't have a soc_id to dsp_name map.")
```

I believe that this:

```
            if self.dt_info["soc_id"] in self.imx_soc_id_to_dsp_name:
                dsp_name = self.imx_soc_id_to_dsp_name[self.dt_info["soc_id"]]
            else:
                raise NoMapException("Current DT-based platform doesn't have a soc_id to dsp_name map.")
```

could in fact be turned into this:

```
            dsp_name = self.imx_soc_id_to_dsp_name[self.dt_info["soc_id"]]
```
and if there's no i.MX mapping just let it raise `KeyError` but if more maps were to be introduced I believe the first example would be more suited.

tools/sof-dump-status.py
LaurentiuM1234: Because I believed it would be a good idea to have all the functions grouped together but if that's not needed then I see no reason why it shouldn't be placed at the top of the module.

tools/sof-dump-status.py
LaurentiuM1234: In the current implementation there isn't, but I think it would be a good idea to add said concepts. Also, I think I'm going to remove

```
# load SoC family
_loadDT("family")
# load SoC revision
 _loadDT("revision")
# load machine name
_loadDT("machine")
```

altogether since they aren't really used for anything at the moment. The only reason I left those in was to allow `dumpPlatform` to print information about the SoC in the future, but I think it would be a much better idea to just leave those out for now, add the concept of optional/required attributes and mark them as optional if they'll be required in the future.

tools/sof-dump-status.py
marc-hb: I think hiding functions away from the top-level is a bit unusual in Python (except for closures of course). For instance there is no `private` keyword https://docs.python-guide.org/writing/style/#we-are-all-responsible-users

This function is really super-generic, could be used anywhere.

Again I appreciate the consistency effort (I really do!) but having a giant `main()` function with all other functions inside seems like it was a pure mistake and waste of whitespace / indentation with no apparent benefit.

tools/sof-dump-status.py
marc-hb: I'm afraid we're talking across each other, sorry for the confusion.

I'm not questioning hardcoding this path. By nature this script duplicates and hardcodes a lot of stuff from the Linux kernel and that's OK, there isn't any easy way to automatically extract this information from the kernel sources and most of it is probably part of the "kernel ABI" or duplication hardware constants in the first place so no risk of diverging.

What I find brittle is that a simple `mkdir /lib/firmware/imx/sof` on a non-IMX system can break this script.

Above there is another `# check to see if we're on i.MX` decision made and it does not look at this path at all - so why two different ways? The other code looks at DT data instead, which seems MUCH more reliable.

tools/sof-dump-status.py
marc-hb: I'm still not convinced sorry.

```python
all_dsp_names = { **platform1_soc_id_to_dsp_name, **platform2_soc_id_to_dsp_name, ...
dsp_name = all_dsp_names[dt_info["soc_id"]]
```

2 lines, done! Less is more.

marc-hb (dismissed):

tools/sof-dump-status.py
LaurentiuM1234: Thanks ! This is so much better than what I was trying to go for.

LaurentiuM1234:
Thanks @marc-hb and @paulstelian97 for the comments ! I've made the following changes:

1) `read_file` is now placed at the top of the module. Since consistency was not a requirement there was really no point in keeping it grouped with the other functions.
2) `loadFw` was renamed to `loadDT` in order to avoid naming ambiguity and to suggest that this function is to be used on DT-based platforms.
3) `loadDT` now has the ability to distinguish between optional and required SoC-related information. Failing to read any of the files marked as required will result in dt_info being initialized with the empty dictionary.
4) Constants have been moved at the beginning of the file and contain only uppercase letters.
5) In order to avoid scenarios like this one:

```
            if self.dt_info["soc_id"] in self.imx_soc_id_to_dsp_name:
                dsp_name = self.imx_soc_id_to_dsp_name[self.dt_info["soc_id"]]
            elif self.dt_info["soc_id"] in self.platform1_soc_id_to_dsp_name:
                dsp_name = ...
            elif self.dt_info["soc_id"] in self.platform2_soc_id_to_dsp_name:
                dsp_name = ...
            elif self.dt_info["soc_id"] in self.platform3_soc_id_to_dsp_name:
                dsp_name = ...
            else:
                raise NoMapException("Current DT-based platform doesn't have a soc_id to dsp_name map.")
```

the dsp name, fw name and fw path are extracted like this:

```
dsp_name = SOC_ID_TO_DSP_NAME[self.dt_info["soc_id"]]
```

This makes the code much easier to read and scales very well. If there is no map between the soc_id and the required information (e.g: dsp name) then it will result in an exception which will simply break the script.

6) As a response to this:

> What I find brittle is that a simple mkdir /lib/firmware/imx/sof on a non-IMX system can break this script. Unlikely, sure but still brittle.

The fw path is now extracted using a map just like the fw and dsp names. This should offer more reliability because, in order to print the fw path, information about the SoC needs to be extracted and a map between the soc_id and fw path needs to exist.

A disadvantage to this implementation would be the fact that IMX_SOC_ID_TO_FW_PATH might end up containing multiple keys having the same value but it shouldn't be that bad if the number of platforms is limited. Also, this change allows the possibility of having different fw paths for the platforms owned by a contributor (if this can happen?). Also also, this should scale much better than the initial implementation.

7) `loadDT` doesn't print any debug message in case it fails to add data marked as required. The reason for this is because I believe messages regarding DT-based information loading wouldn't be relevant on platforms which are not DT-based.

8) `loadDT` now prints a debug message when loading optional SoC information fails. The reason I did this is because I believe those messages could be relevant on a DT-based platform. It's important to note that all the optional information should go after the required one because this way, if we're not on a DT-based platform, we can avoid the aformentoned debug messages.

tools/sof-dump-status.py
```diff
@@ -21,6 +24,7 @@ def read_file(filepath):
 
 class clsSYSCardInfo():
     def __init__(self):
+        # i.MX-specific maps
         self.dt_info = {}
```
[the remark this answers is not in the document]
LaurentiuM1234: Fixed. Thanks!

tools/sof-dump-status.py
```diff
@@ -2,10 +2,31 @@
 
 import subprocess
 import os
+import sys
+
 from common import format_pipeline, export_pipeline
 
+
+# i.MX specific maps
```
marc-hb, line 10: Are these maps DeviceTree-specific? If yes then the "Device Tree" keyword should be somewhere in this comment. Maybe even `_DT_` in the constant names?

tools/sof-dump-status.py
```diff
@@ -2,10 +2,31 @@
...
+IMX_SOC_ID_TO_FW_NAME = {"i.MX8QM": "imx8"}
+IMX_SOC_ID_TO_DSP_NAME = {"i.MX8QM": "556e8000.dsp"}
+IMX_SOC_ID_TO_FW_PATH = {"i.MX8QM": "/lib/firmware/imx/sof"}
+
+# maps containing the content of all platform-specific maps
+SOC_ID_TO_FW_NAME = {**IMX_SOC_ID_TO_FW_NAME}
+SOC_ID_TO_DSP_NAME = {**IMX_SOC_ID_TO_DSP_NAME}
+SOC_ID_TO_FW_PATH = {**IMX_SOC_ID_TO_FW_PATH}
```
marc-hb, line 18: Maybe a bit shorter:
```suggestion
FW_PATHS = {**IMX_FW_PATHS}
```

It's a matter of taste but in general I find it enough to have only the "to" field in maps names because it's then invoked like this:
```
FW_PATHS[self.dt_info["soc_id"]]
```
so `soc_id` is already there.

No big deal.

tools/sof-dump-status.py
```diff
@@ -82,6 +103,33 @@ def loadDMI(self):
...
+                    raise
+                else:
+                    print("[WARNING]: Unable to read content of {}.".format(filepath),
+                          file=sys.stderr)
+                    return
+
+        # read all required SoC-related information
+        try:
+            _loadDT("soc_id", is_required=True)
+        except IOError:
```
marc-hb, line 127: I don't understand the purpose of the `_loadDT()` indirection... what would be wrong with a "flatter", single `loadDT()` function that does everything in a single `try` block?

LaurentiuM1234: In the current implementation that would work, yes, but what if we need more SoC-related information? Wouldn't it be better to keep the internal `_loadDT`? Example:

```
# required SoC-related information
try:
    _loadDT("soc_id", is_required=True)
    _loadDT("family", is_required=True)
except IOError:
    dt_info = {}
    return

# optional SoC-related information
_loadDT("machine")
_loadDT("revision")
```

tools/sof-dump-status.py
LaurentiuM1234: They are. I will definitely add "Device Tree" to the comment so as to make it a bit clearer. I can also add `_DT_` to the names of all the constants if you think it will help others understand the code better but wouldn't it be a bit overkill, especially with the added comment?

tools/sof-dump-status.py
```diff
@@ -2,10 +2,31 @@
...
 import os
+import sys
+
 from common import format_pipeline, export_pipeline
 
+
+# i.MX specific maps
+IMX_SOC_ID_TO_FW_NAME = {"i.MX8QM": "imx8"}
+IMX_SOC_ID_TO_DSP_NAME = {"i.MX8QM": "556e8000.dsp"}
+IMX_SOC_ID_TO_FW_PATH = {"i.MX8QM": "/lib/firmware/imx/sof"}
```
aiChaoSONG, line 13: This is tedious, can you do this with a namedtuple?
```
FirmwareInfo = namedtuple('FirmwareInfo', ['fw_name', 'dsp_name', 'fw_path'])

imx_soc_id_to_fw_info = {'i.MX8QM': FrimwareInfo(fw_name='imx8', dsp_name='556e8000.dsp', fw_path='/lib/firmware/imx/sof')}
```

tools/sof-dump-status.py
```diff
@@ -2,10 +2,25 @@
 
 import subprocess
 import os
+from collections import namedtuple
 from common import format_pipeline, export_pipeline
 
+# DT-specific constants
+FIRMWARE_INFO = namedtuple("FIRMWARE_INFO", ["fw_name", "device_name", "fw_path"])
```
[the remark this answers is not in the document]
LaurentiuM1234: Thanks, fixed. Also changed comment to 'DT-specific data' since FirmwareInfo is not a constant.

aiChaoSONG (approved):

LaurentiuM1234:
Based on received feedback I've made the following changes:

1) `loadDT` no longer uses the internal `_loadDT`. Currently there was no point in keeping it since the only information we're using is the soc_id for now.
2) The implementation now uses a `namedtuple` instead of a bunch of dictionaries in order to extract information based on soc_id.
3) Added `DT` and a comment to the DT-specific data. This will help indicate that the data is DT-specific.

dbaluta (approved):

marc-hb (approved):

marc-hb:
There's been a major and unrelated regression (see sof 6020) https://sof-ci.01.org/softestpr/PR930/build96/devicetest/

Let's wait until CI is back on track and then merge.

marc-hb:
https://sof-ci.01.org/softestpr/PR930/build98/devicetest/ is still not perfect but MUCH better and the 3 failures are totally unrelated.

PR #931, opened 2022-07-13
check-pause-resume: handle volume MAX output in expect

fredoh9:
When volume is MAX, current expect misses the case. Why volume is MAX is
separate question. Momentary pop can be ingored but if volume is 100% DC
return error.
And added for ignoring dump-hw-params and last case is anything else,
it means unexpected outputs, will return error also.

This is example of output from aplay or arecord.
1. Playing
"##    +                                            | 11%"
2. When paused
"=== PAUSE ==="
3. When volume is MAX
"##################################################+| MAX"

related:
- https://github.com/thesofproject/linux/issues/3766

### DIFF
--- a/test-case/check-pause-release-suspend-resume.sh
+++ b/test-case/check-pause-release-suspend-resume.sh
@@ -121,6 +121,10 @@ rm -rf /tmp/sof-test.lock
 #   catch: Evaluate script and trap exceptional returns
 #   after ms: Ms must be an integer giving a time in milliseconds.
 #       The command sleeps for ms milliseconds and then returns.
+# FIXME: this test was broken by 46dadd0 ("Add mutual exclusion and journalctl logging")
+#       because it never was compatible with is_subtest()
+# FIXME: Need to handle more of aplay/arecord output. Need to apply similar fix with
+#       check-pause-resume.sh, multiple-pause-resume.sh
 expect <<AUDIO
 spawn $cmd -D $pcm -r $rate -c $channel -f $fmt -vv -i $dummy_file -q
 set i 1
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -96,18 +96,21 @@ do
     dlogi "Entering expect script with:
       $cmd $SOF_ALSA_OPTS $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q"
 
+    # NOTE: Purposely we don't handle 'Volume MAX' case. So that the MAX output can fall into
+    #       'anything else' case and error out.
+    # FIXME: share this expect script as a common function
     expect <<END
 spawn $cmd $SOF_ALSA_OPTS $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q
 set i 1
 expect {
-    "*#*+*\%" {
+    -re "#.*\%\r" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
         puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before pause"
         send " "
         after \$sleep_t
         exp_continue
     }
-    "*PAUSE*" {
+    "=== PAUSE ===" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
         puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before resume"
         send " "
@@ -116,6 +119,20 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    -re ".*\:.*\n"|"Hardware PCM card.*\n" {
+        puts "\rexpect ignore dump-hw-params"
+        after 10
+        exp_continue
+    }
+    default {
+        puts "\rError: timeout or eof detected, exit 1"
+        exit 1
+    }
+# this is conventional default, anything else hit here
+    "^?*\n" {
+        puts "\rError: unexpected output detected, exit 1. $expect_out(0,string)"
+        exit 1
+    }
 }
 exit 1
 END
@@ -123,6 +140,7 @@ END
     #flush the output
     echo
     if [ $ret -ne 0 ]; then
+        echo "Expect block return $ret"
         func_lib_lsof_error_dump "$snd"
         sof-process-kill.sh ||
             dlogw "Kill process catch error"
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -115,18 +115,22 @@ func_pause_resume_pipeline()
     #   after ms: Ms must be an integer giving a time in milliseconds.
     #       The command sleeps for ms milliseconds and then returns.
     dlogi "$pcm to command: $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i $file -q"
+
+    # NOTE: Purposely we don't handle 'Volume MAX' case. So that the MAX output can fall into
+    #       'anything else' case and error out.
+    # FIXME: share this expect script as a common function
     expect <<END &
 spawn $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i $file -q
 set i 1
 expect {
-    "*#*+*\%" {
+    -re "#.*\%\r" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
         puts "\r(\$i/$repeat_count) pcm'$pcm' cmd'$cmd' id'$idx': Wait for \$sleep_t ms before pause"
         send " "
         after \$sleep_t
         exp_continue
     }
-    "*PAUSE*" {
+    "=== PAUSE ===" {
         set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
         puts "\r(\$i/$repeat_count) pcm'$pcm' cmd'$cmd' id'$idx': Wait for \$sleep_t ms before resume"
         send " "
@@ -134,6 +138,20 @@ expect {
         incr i
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
+    }
+    -re ".*\:.*\n"|"Hardware PCM card.*\n" {
+        puts "\rexpect ignore dump-hw-params"
+        after 10
+        exp_continue
+    }
+    default {
+        puts "\rError: timeout or eof detected, exit 1"
+        exit 1
+    }
+# this is conventional default, anything else hit here
+    "^?*\n" {
+        puts "\rError: unexpected output detected, exit 1. $expect_out(0,string)"
+        exit 1
     }
 }
 exit 1


test-case/check-pause-resume.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+#set -e
```
fredoh9, line 3: accidently removed, will restore

test-case/check-pause-resume.sh
```diff
@@ -116,18 +116,27 @@ expect {
...
     if [ $ret -ne 0 ]; then
+        echo "Expect block return $ret"
         func_lib_lsof_error_dump "$snd"
         sof-process-kill.sh ||
             dlogw "Kill process catch error"
         exit $ret
     fi
     # check kernel log for each iteration to catch issues
     sof-kernel-log-check.sh "$KERNEL_CHECKPOINT" || die "Caught error in kernel log"
+    ps aux | grep arecord
```
fredoh9, line 141: No need

fredoh9:
multiple-pause-resume.sh need to be handled in same way. I will update the same shortly

marc-hb (changes requested):
> MAX case is same as playing case.

It it is the same then please don't copy/paste the code, it must be possible to match both and share the code.

Also, please add a fallback case that prints an error so no one wastes hours next time aplay prints something unexpected.

But more importantly: this test never failed like this. Why is MAX suddenly OK? This looks like a firmware bug?!


> multiple-pause-resume.sh need to be handled in same way. I will update the same shortly

Maybe it's time to move some new .expect file shared between tests?

fredoh9:
@marc-hb I will check how we can share the same code in expect. Need to search expect manual

fredoh9:
@marc-hb still looking for sharing the case in expect script. 

And also I'm trying to handle 'else' case but expect script didn't work like that, i thought it will match top to bottom but  "*" case hit all the time.

```bash
diff --git a/test-case/check-pause-resume.sh b/test-case/check-pause-resume.sh
index 170aff7..cdcd203 100755
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -116,6 +116,18 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    "*#*+*MAX" {
+        set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
+        puts "\r(\$i/$repeat_count) Wait for \$sleep_t ms before pause"
+        send " "
+        after \$sleep_t
+        exp_continue
+    }
+    "*" {
+        puts "\r Invalid output detected, Wait for 500ms and recheck"
+        set sleep_t 0.5
+       exp_continue
+    }
 }
 exit 1
 END

```

test-case/multiple-pause-resume.sh
```diff
@@ -135,6 +135,13 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    "*#*+*MAX" {
+        set sleep_t [expr int([expr rand() * $rnd_range]) + $rnd_min ]
```
greg-intel, line 139: Do our tests generally have a lot of randomness?  Doesn't that make failures less deterministic?

fredoh9: In general we don't like randomness in testing. In this pause/resume, we intentionally introduce randomness on ranger of interval to discover easily fw bug

fredoh9:
I missed `check-pause-release-suspend-resume.sh` has expect block as well.

marc-hb:
> And also I'm trying to handle 'else' case but expect script didn't work like that, i thought it will match top to bottom but "*" case hit all the time.

Strange. From https://www.tcl.tk/man/expect5.31/expect.1.html

> In the event that multiple patterns match, the one appearing first is used to select a body.

fredoh9:
This is the source code for VU meter output. Last line output will be "!clip"?
 
https://github.com/thesofproject/alsa-utils/blob/a566f8a0ed6d7ca5fd6ae2d224f3f28654a2f3be/aplay/aplay.c#L1760

```c
static void print_vu_meter_mono(int perc, int maxperc)
{
	const int bar_length = 50;
	char line[80];
	int val;

	for (val = 0; val <= perc * bar_length / 100 && val < bar_length; val++)
		line[val] = '#';
	for (; val <= maxperc * bar_length / 100 && val < bar_length; val++)
		line[val] = ' ';
	line[val] = '+';
	for (++val; val <= bar_length; val++)
		line[val] = ' ';
	if (maxperc > 99)
		sprintf(line + val, "| MAX");
	else
		sprintf(line + val, "| %02i%%", maxperc);
	fputs(line, stderr);
	if (perc > 100)
		fprintf(stderr, _(" !clip  "));
}
```

This is "PAUSE" output
https://github.com/thesofproject/alsa-utils/blob/a566f8a0ed6d7ca5fd6ae2d224f3f28654a2f3be/aplay/aplay.c#L1622

```c
static void check_stdin(void)
{
	unsigned char b;

	if (!interactive)
		return;
	if (fd != fileno(stdin)) {
		while (read(fileno(stdin), &b, 1) == 1) {
			if (b == ' ' || b == '\r') {
				while (read(fileno(stdin), &b, 1) == 1);
				fprintf(stderr, _("\r=== PAUSE ===                                                            "));
				fflush(stderr);
				do_pause();
				fprintf(stderr, "                                                                          \r");
				fflush(stderr);
			}
		}
	}
}
```

fredoh9:
> I missed check-pause-release-suspend-resume.sh has expect block as well.

check-pause-release-suspend-resume.sh is not being used by SOF CI yet. I can add FIXME note for it

fredoh9:
I tested locally and pushed my changes, I should whether check device test is going smooth.

thank you for your review.
@marc-hb @greg-intel @keqiaozhang @aiChaoSONG @miRoox @XiaoyunWu6666

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -169,7 +169,17 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    "*:*\n" {
+        puts "\rexpect ignore dump-hw-params"
+        after 10
+        exp_continue
+    }
+    "^?*\n" {
+        puts "\rError: unexpected output detected, exit 1"
```
marc-hb, line 178: Please print what was unexpected, should be something like `expect_out(0,string)` (untested)

test-case/multiple-pause-resume.sh
```diff
@@ -119,14 +119,14 @@ func_pause_resume_pipeline()
 spawn $cmd -D $dev -r $rate -c $channel -f $fmt -vv -i $file -q
 set i 1
 expect {
-    "*#*+*\%" {
+    -re "#.*\%|#.*MAX" {
```
marc-hb, line 122: Nope, MAX should print an error because it find regressions/bugs. It just found regression https://github.com/thesofproject/linux/issues/3750 that you documented well. This does not "fix" 3750, it hides the bug!

marc-hb (changes requested):

marc-hb:
Please submit the error handling commit in a different PR so we can fast track it.

plbossart:
@fredoh9 I am struggling to figure out what volume has to do with pause_push/pause_release. I went through the entire thread and I have zero understanding of the issue.

plbossart:
What is the relationship between pressing the space bar and volume control @fredoh9 ?

fredoh9:
> What is the relationship between pressing the space bar and volume control @fredoh9 ?

Right, nothing to do with. But expect script to look for aplay/arecord output to decide what to do. When first wrote the test script, we didn't expect MAX volume is one of possibility. That's why I said first problem is on expect script.

test-case/multiple-pause-resume.sh
fredoh9: I wanted to handle the MAX and get going. I agreed and updated. Starting pop case is ignored.

test-case/check-pause-release-suspend-resume.sh
fredoh9: Updated, I couldn't see meaningful output from my test but it seems useful when it hit the case

fredoh9:
From device test, multi-pause-resume.sh failure is fixed
https://sof-ci.01.org/softestpr/PR931/build82/devicetest

test-case/multiple-pause-resume.sh
```diff
@@ -135,6 +136,27 @@ expect {
...
+        after 500
+        exp_continue
+    }
+    -re "*:*\n"|"Hardware PCM card*" {
+        puts "\rexpect ignore dump-hw-params"
+        after 10
+        exp_continue
+    }
+    timeout {
+        puts "\rError: timeout detected, exit 1"
```
marc-hb, line 152:
```suggestion
     default {
        puts "\rError: expect time out or aplay is gone, exit 1"
```

marc-hb (dismissed):

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -121,6 +121,8 @@ rm -rf /tmp/sof-test.lock
 #   catch: Evaluate script and trap exceptional returns
 #   after ms: Ms must be an integer giving a time in milliseconds.
 #       The command sleeps for ms milliseconds and then returns.
+# FIXME: Need to handle more of aplay/arecord output. Need to apply similar fix with
+# check-pause-resume.sh, multiple-pause-resume.sh
```
marc-hb, line 125: Please also add FIXME: this test was broken by 46dadd0c563e ("Add mutual exclusion and journalctl logging") because it never was compatible with `is_subtest()`

test-case/multiple-pause-resume.sh
```diff
@@ -135,6 +136,27 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    -re "#.*MAX" {
+        if { \$maxcount > 10 } { puts "\rMAX Volume error!"; exit 1 }
+        puts "\rError: MAX volume detected (\$maxcount/10), Wait for 500ms"
+        incr maxcount
```
miRoox, line 142: Is it possible to reach MAX 10 times but just like "MAX -> xx% -> MAX -> ...". What I mean is: should `maxcount` be reset in other cases?

fredoh9: That's good point. Will do that.
But if pause/resume don't care about volume, I need to merge this case with normal volume case.

test-case/check-pause-release-suspend-resume.sh
fredoh9: Will add

fredoh9:
@marc-hb we need to combine "#.*\%"  AND "#.*MAX", Let's not worry any volume issue in pause-resume test.

And I tried to add "\n" but it didn't match well. Need to try more. I will update this PR later today.

test-case/multiple-pause-resume.sh
```diff
@@ -135,6 +136,27 @@ expect {
...
+        after 500
+        exp_continue
+    }
+    -re "*:*\n"|"Hardware PCM card*" {
```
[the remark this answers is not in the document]
greg-intel: Should there be a `.` before each `*`?  Or is it looking for literal `*`?

marc-hb:
> @marc-hb we need to combine "#.*%" AND "#.*MAX", Let's not worry any volume issue in pause-resume test.

It's not a volume issue, this test found by chance that the most basic `arecord` is broken. _You_ discovered that in #935 . No other test found that `arecord` is broken. This is precious coverage, it must not be lost.

test-case/check-pause-resume.sh
```diff
@@ -99,15 +99,21 @@ do
     expect <<END
 spawn $cmd $SOF_ALSA_OPTS $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q
 set i 1
+set maxcount 1
 expect {
-    "*#*+*\%" {
+    -re "#.*\%\r" {
+        puts "\rFred: normal volume, maxcount reset to 1\n"
```
fredoh9, line 105: Sorry, didn't delete debugging log, this and below
Hehe I deleted in multiple-pause-resume.sh, but not here.

test-case/multiple-pause-resume.sh
marc-hb: > And even if aplay or arecord is gone, timeout can handle the case.

I don't think so. Try passing a wrong --flag to aplay and you will see EOF but no timeout. I noticed this in a different sof-test and the result was PASS (now fixed)

fredoh9:
Device test result is all GREEN. https://sof-ci.01.org/softestpr/PR931/build91/devicetest/
Shellcheck errors are not real problem.

@marc-hb do you have any comments? So I can address comments all together.

marc-hb:
> Device test result is all GREEN. https://sof-ci.01.org/softestpr/PR931/build91/devicetest/

That's very worrying consider `arecord` is still producing garbage, isn't it?

fredoh9:
> That's very worrying consider `arecord` is still producing garbage, isn't it?

pause/resume test should not worry about pop or noise or volume max. If we need to check the capture output, we need to implement the feature, I filed an feature issue to internal issue tracker.

marc-hb:
> pause/resume test should not worry about pop or noise or volume max. 

No, this is not correct. While each test has a main focus, it should never have code that ignore/dismiss and hide other failures on purpose. Especially not very serious issues like this. NACK, the time of "green failures" is over.

Note I'm NOT saying tests should have additional logic and code to test other, unrelated features. I'm saying that no test should have **code that ignores unrelated failures on purpose**.

This is of course even more important  when we have absolutely nothing else that looks at captured audio; zero test coverage. But even if we had some test coverage elsewhere then we should not still not ignore a broken capture output here. Each test provides slightly different  because there's an infinite number of configurations, scenarios, uses cases and possibilities so we can never have too much test coverage.

test-case/check-pause-release-suspend-resume.sh
```diff
@@ -121,6 +121,10 @@ rm -rf /tmp/sof-test.lock
 #   catch: Evaluate script and trap exceptional returns
 #   after ms: Ms must be an integer giving a time in milliseconds.
 #       The command sleeps for ms milliseconds and then returns.
+# FIXME: this test was broken by 46dadd0 ("Add mutual exclusion and journalctl logging")
+#       because it never was compatible with is_subtest()
+# FIXME: Need to handle more of aplay/arecord output. Need to apply similar fix with
+#       check-pause-resume.sh, multiple-pause-resume.sh
```
plbossart, line 127: Why volume is MAX is separate question

No, that's why this fails in the first test failed in the first place.
If the ALSA settings are not correct, this PR is not useful.

Fix the ALSA settings first, then we can refine expect scripts.

plbossart (changes requested):
Completely lost on the directions. We know ALSA settings are not correct, why not fix them first?

test-case/check-pause-release-suspend-resume.sh
marc-hb: > Fix the ALSA settings first, then we can refine expect scripts.

As usual, keep the broken ALSA settings in your backpocket to "test the test" locally. A test that can't fail is useless (green failure).

fredoh9:
@marc-hb I dropped handling volume max from expect script.

plbossart:
> @marc-hb I dropped handling volume max from expect script.

erm, why? This was how we detected that ALSA settings were incorrect.

The goal of validation is not to make tests pass but highlight issues.

test-case/check-pause-resume.sh
```diff
@@ -116,13 +118,30 @@ expect {
         if { \$i > $repeat_count } { exit 0 }
         exp_continue
     }
+    -re ".*\:.*\n"|"Hardware PCM card.*\n" {
+        puts "\rexpect ignore dump-hw-params"
+        after 10
+        exp_continue
+    }
+# NOTE: we could use 'defatult' = timeoue + eof but eof is not the obvious problem.
+#       So handle timeout specifically
```
marc-hb, line 127: I don't understand. EOF means arecord died, how is that not a problem? Test with a wrong PCM and you should see.

marc-hb (commented):
> erm, why? This was how we detected that ALSA settings were incorrect.

> The goal of validation is not to make tests pass but highlight issues.

@plbossart at some point in time this PR matched MAX -> success. I object to that, now @fredoh9 removed MAX matching which means it will default to "unknown -> FAIL" which is I think what we both want.

test-case/check-pause-resume.sh
fredoh9: Glad I put my comments there. I thought eof didn't make sense for this test, but I think good point. Let me do some test for the case.
I have faced all the comments I could expect. haha :)

fredoh9:
> @plbossart at some point in time this PR matched "MAX -> success". I object to that, now @fredoh9 removed MAX matching which means MAX will default to "unknown -> FAIL" which is I think what we both want.

That's right. I will add more comments to clarify why purposely not to handle volume MAX case. My intention also to fail the test if we see unexpected output.

test-case/check-pause-resume.sh
fredoh9: I see typo, defatult -> default. WIll handle in next patch

marc-hb (approved):
Approving assuming this is well tested. Thanks @fredoh9 , I feel your pain!

miRoox (approved):
LGTM

marc-hb:
@plbossart OK?

plbossart:
> @plbossart OK?

We confirmed with Realtek that the audio settings are incorrect, so I am not sure why any changes are needed in sof-test. I would fix the audio settings first instead of making the tool work-around the bad settings.

marc-hb:
@fredoh9 can you confirm the test WILL STILL FAIL with this PR in the same system configuration as before?

marc-hb:
SOFCI TEST

marc-hb:
https://sof-ci.01.org/softestpr/PR931/build519/devicetest/index.html?model=MTLP_RVP_HDA&testcase=multiple-pause-resume-50 shows some "MAX" traces but now looking different. Saving the link for future investigation.

fredoh9:
The real problem is that the DMIC volume suddenly becomes 100% for no reason. I suspect something to do with electrical problem with DMIC transducer board, but don't know the root cause. Test failure rate varies between device/DMIC board.

The whole point of this PR is to match the expected output regardless of the DMIC output or volume level.

@marc-hb I'm looking at this PR now. Even with this PR, it silll fails when DMIC MAX volume happens. Looks the matching is broken. I will update the fix shortly.

fredoh9:
I scrolled up to read all the discussion, we concluded the volume MAX is failure case, not much thing to do with this PR.
Instead I will check the DMIC connections and replace with different DMIC board.

marc-hb:
> I scrolled up to read all the discussion, we concluded the volume MAX is failure case, not much thing to do with this PR.

This PR is all about that failure better.

marc-hb (changes requested):
I tested this script with `expect -d` and it works "by accident" either before or after the PR.
cc:
- https://github.com/thesofproject/linux/pull/5067#issuecomment-2177339566

marc-hb:
I realized the existing (and duplicated) expect script was hopelessly broken so I rewrote it and submitted it in:
- #1218

marc-hb:
- Alternative #1218 was merged so I think there are only some comments that are still valid in this PR.

- While testing #1218, I found that the "MAX" pop was still present on TGLU_SKU0A32 so I just re-opened this:
- https://github.com/thesofproject/linux/issues/3766

test-case/check-pause-resume.sh
```diff
@@ -96,18 +96,21 @@ do
     dlogi "Entering expect script with:
       $cmd $SOF_ALSA_OPTS $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q"
 
+    # NOTE: Purposely we don't handle 'Volume MAX' case. So that the MAX output can fall into
+    #       'anything else' case and error out.
+    # FIXME: share this expect script as a common function
```
marc-hb, line 101: @fredoh9 I re-submitted these comments in new #1222. I think all the rest has been superseded by merged #1218. OK to abandon this?

fredoh9:
Marc added better solution.

PR #934, opened 2022-07-14
[RFC]: Pipe journalctl command to grep instead of using the -g option

LaurentiuM1234:
The reason for this is to allow the tests based on journalctl -g "some string" to run on i.MX platforms. At this moment, running
journalctl -q -k -g "loading" results in the following error: Compiled without pattern matching support.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -604,7 +604,7 @@ disable_kernel_check_point()
 # shellcheck disable=SC2120
 sof_firmware_boot_complete()
 {
-    journalctl_cmd "$@" --grep 'sof.*firmware[[:blank:]]*boot[[:blank:]]*complete'
+    journalctl_cmd "$@" | grep -i 'sof.*firmware[[:blank:]]*boot[[:blank:]]*complete'
 }
 
 is_zephyr()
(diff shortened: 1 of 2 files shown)


marc-hb (changes requested):
`--grep` is case insensitive so to be equivalent this should be `grep -i`

Note piping is MUCH slower. Impact should hopefully be OK because we only search the current boot.

fredoh9:
I don't understand why current -g option is not working in i.MX platforms? Can you elaborate more?

marc-hb:
@fredoh9 some journalctl features can be removed at compile time to save space.

BTW @LaurentiuM1234 the commit message is incorrect; this problem is related to your particular Yocto configuration, it's not related to i.MXX

paulstelian97:
Um, we shouldn't have merge commits inside a pull request, at least in SOF firmware/test projects. Can a rebase be done so that the merge commit is unneeded?

LaurentiuM1234:
@paulstelian97 Yeah, sorry about that. Will start marking the PRs as drafts in case more mistakes like that happen again.
@marc-hb Thanks for pointing out the mistakes. Now the grep command is called with the -i command and the commit messages should be correct.

marc-hb (approved):

marc-hb:
I always start all my PRs as draft, it's a good habit.

PR #938, opened 2022-07-15
[SKIP CI] lib.sh: replace with axfer for playback and capture

fredoh9:
WIP: Initial attempt. Need to check how much change is required and
axfer is better in terms of SOF CI automation.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -558,13 +558,13 @@ aplay_opts()
 {
     dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
     # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    axfer transfer playback $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
 }
 arecord_opts()
 {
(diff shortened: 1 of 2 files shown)


fredoh9:
There are some more places to replace aplay/arecord command.

marc-hb:
https://sof-ci.01.org/softestpr/PR938/build88/devicetest/ shows compatibility issues.

If we ever switch I think we should switch more gradually, on a test by test basis.

Also, can `axfer` reproduce thesofproject/linux/issues/3766 like `arecord` does?

fredoh9:
> If we ever switch I think we should switch more gradually, on a test by test basis.

I agree

fredoh9:
put [SKIP CI] because check-signal-stop-start is not being used by PR testing.

Piggyback on check-signal-stop-start.sh to test signal pause/resume with Axfer.
The test results are PASS but there are some
`Fail to process frames: Interrupted system call`

Will look at that later.

https://github.com/thesofproject/alsa-utils/blob/a566f8a0ed6d7ca5fd6ae2d224f3f28654a2f3be/axfer/xfer-libasound.c#L784

[check-signal-pause-resume-test-axfer.txt](https://github.com/thesofproject/sof-test/files/9136152/check-signal-pause-resume-test-axfer.txt)
.

fredoh9:
will close for now, will re-visit later when available.

PR #943, opened 2022-07-26
Fix hijack.sh sudo() to use an array, support whitespace and behave

marc-hb:
Overriden commands should work exactly like real commands and support
whitespace and special characters exactly like the real command.

I usually disable the sudo() override locally when using `set -x`
to reduce trace noise considerably. The previous difference between the
sudo() override and the built-in sudo() cost me couple hours twice:

- WIP https://github.com/thesofproject/sof-test/pull/897#discussion_r928074591
- Merged commit 232d40ae3e0a73

No more!

Also fix the return value: 2 means SKIP.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -177,20 +177,25 @@ SUDO_LEVEL=""
 # overwrite the sudo command, sudo in the script can direct using sudo command
 sudo()
 {
+    local cmd=( "$@" )
+
     func_hijack_setup_sudo_level || true
     local cmd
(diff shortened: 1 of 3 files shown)


marc-hb:
`case-lib/hijack.sh` is shellcheck clean in https://github.com/thesofproject/sof-test/runs/7512927770?check_suite_focus=true, no new warning added in other files.

keqiaozhang (approved):
LGTM

marc-hb:
thanks @keqiaozhang

PR #946, opened 2022-07-29
lib.sh: fix recent check_error_in_file() . Exclude BSW

marc-hb:
2 commits. Main one:

Use -w to catch the words 'ERR' and 'ERROR' and avoid unrelated words
like 'ovERRuns' (unless of course they're prefixed with ERROR)

Fixes recent commit https://github.com/thesofproject/sof-test/commit/87aeb4d5eb79e5200581d444b0c19882d8b27bc8 ("check-sof-logger: catch ERROR in
firmware traces and fail")

Add -i: no reason to catch errors only in upper case.

Log the grep commands using set -x

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -363,7 +363,7 @@ check_error_in_file()
     local platf; platf=$(sof-dump-status.py -p)
 
     case "$platf" in
-        byt|bdw)
+        byt|bdw|bsw)
             # Maybe downgrading this to WARN would be enough, see #799
             #  src/trace/dma-trace.c:654  ERROR dtrace_add_event(): number of dropped logs = 8
             dlogw 'not looking for ERROR on BYT/BDW because of known DMA issues #4333 and others'
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -356,8 +356,6 @@ func_lib_start_log_collect()
     sudo "${loggerCmd[@]}" > "$logfile" &
 }
 
-# -B 2 shows the header line when the first etrace message is an ERROR
-# -A 1 shows whether the ERROR is last or not.
 check_error_in_file()
 {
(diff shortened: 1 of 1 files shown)


marc-hb:
A very strange
```
no such file no /sys/kernel/debug/sof/trace
```
... on one device in https://sof-ci.01.org/softestpr/PR946/build114/devicetest while `/sys/kernel/debug/sof/etrace` was there!?

Not caused by this PR.

marc-hb:
SOFCI TEST

marc-hb:
https://sof-ci.01.org/softestpr/PR946/build115/devicetest/?model=ADLP_RVP_NOCODEC_IPC4&testcase=check-sof-logger has the same missing `etrace`...

greg-intel:
Is the search case sensitive?

greg-intel (approved):

case-lib/lib.sh
```diff
@@ -375,7 +373,11 @@ check_error_in_file()
         dloge "file NOT FOUND: '$1'"
         return 1
     }
-    if grep -B 2 -A 1 -E 'ERRO?R?' "$1"; then
+    # -B 2 shows the header line when the first etrace message is an ERROR
+    # -A 1 shows whether the ERROR is last or not.
+    if (set -x
+        grep -B 2 -A 1 -i -w -e 'ERR' -e 'ERROR' "$1"
```
greg-intel, line 379: Is it weird that I find "-B2 -A1" more legible?

greg-intel (approved):

marc-hb:
SOFCI TEST

marc-hb:
Unrelated error is gone in https://sof-ci.01.org/softestpr/PR946/build119/devicetest, maybe thanks to https://github.com/thesofproject/sof/pull/5027

Issue #782, opened 2021-10-05
[BUG] filter out DPC: RP PIO error


plbossart:
On Fedora 34 with kernel 5.14.9:

````
===========================>>
[    0.744964] kernel: pci 0000:00:07.0: DPC: RP PIO log size 0 is invalid
[    0.747749] kernel: pci 0000:00:07.1: DPC: RP PIO log size 0 is invalid
<<===========================
2021-10-05 22:05:45 UTC [ERROR] Starting func_exit_handler(), exit status=1, FUNCNAME stack:
2021-10-05 22:05:45 UTC [ERROR]  main()  @  sof-test/test-case/verify-kernel-boot-log.sh
2021-10-05 22:05:45 UTC [INFO] Test Result: FAIL!
````

plbossart:
this was never added....

marc-hb:
Not an issue anymore, see #947

PR #947, opened 2022-08-01
sof-kernel-log-check: filter out errors unrelated to audio

plbossart:
These errors prevent sof-test from running with the default kernel on
Fedora36.

Additional issues were solved with a scattershot blacklist:
blacklist iwlwifi
blacklist bluetooth
blacklist btusb
blacklist btrtl
blacklist btbcm
blacklist btintel
blacklist btmtk

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -276,6 +276,10 @@ case "$platform" in
         # Bug Report: https://github.com/thesofproject/sof-test/issues/838
         # New TGLU_UP_HDA_ZEPHYR device reporting "TPM interrupt not working" errors.
         ignore_str="$ignore_str"'|kernel: tpm tpm0: \[Firmware Bug\]: TPM interrupt not working, polling instead'
+
+	# Errors reported on Dell SKU 0A32 early versions
+	ignore_str="$ignore_str"'|pci 0000:00:07\..: DPC: RP PIO log size 0 is invalid'
+	ignore_str="$ignore_str"'|int3472-discrete INT3472:00: INT3472 seems to have no dependents'
         ;;
     ehl)
         # i915 crtc logs can be ignored


tools/sof-kernel-log-check.sh
```diff
@@ -276,6 +276,10 @@ case "$platform" in
         # Bug Report: https://github.com/thesofproject/sof-test/issues/838
         # New TGLU_UP_HDA_ZEPHYR device reporting "TPM interrupt not working" errors.
         ignore_str="$ignore_str"'|kernel: tpm tpm0: \[Firmware Bug\]: TPM interrupt not working, polling instead'
+
+	# Errors reported on Dell SKU 0A32 early versions
+	ignore_str="$ignore_str"'|pci 0000:00:07\..: DPC: RP PIO log size 0 is invalid'
```
marc-hb, line 281: I wonder why you are seeing this only now, https://bugzilla.kernel.org/show_bug.cgi?id=209943 shows it's at least 2 years old.

marc-hb: Indeed you filed #782 a long time ago..

marc-hb (commented):
Can you provide some vague rationale why these are irrelevant? In other words, what is INT3472 for? Also, would it be possible to spell out the jargon fest that is DPC RP PIO?

In the past we've ignored USB/network/storage errors, then we lost months of work trying to understand problems like:
- https://github.com/thesofproject/linux/issues/3387

plbossart:
@marc-hb I have no idea what all these things are. I just tested with the 5.18 Fedora kernel and the first boot test failed because of these two things.

plbossart:
SOFCI TEST

greg-intel:
sof-ci/jenkins/pr-device-test check-suspend-resume-with-playback-5.sh timeout.
sof-ci/jenkins/pr-device-test/main check-suspend-resume-with-playback.sh timeout
sof-ci/jenkins/pr-device-test/stable-v2.2 verify-kernel-boot-log.sh Missing monitor issue.  

So, probably good enough.  I'll see about getting that missing monitor device taken care of.

PR #951, opened 2022-08-18
alsa_settings: CAVS_NOCODEC: reset values to 0dB/on with script

plbossart:
We should not encode specific mixer values that are tied to a
topology. The values should always be 0dB and all switches on, this
can be done with a script.

Credits to Marc Herbert for the sed regexp for the controls string.

### DIFF
--- a/alsa_settings/CAVS_NOCODEC.sh
+++ b/alsa_settings/CAVS_NOCODEC.sh
@@ -1,21 +1,11 @@
 set -e
 
-# SSP playback
-amixer -c sofnocodec cset name='PGA1.0 1 Master Playback Volume' 32
-amixer -c sofnocodec cset name='PGA3.0 3 Master Playback Volume' 32
-amixer -c sofnocodec cset name='PGA5.0 5 Master Playback Volume' 32
+# to set the volume as 0dB we have to use the scontrol interface
+amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" | while read -r mixer_name; do
+     amixer -Dhw:0 -- sset "$mixer_name" 0dB;
+done
 
-# SSP capture
-amixer -c sofnocodec cset name='PGA2.0 2 Master Capture Switch' on
-amixer -c sofnocodec cset name='PGA2.0 2 Master Capture Volume' 50
-
-amixer -c sofnocodec cset name='PGA4.0 4 Master Capture Switch' on
-amixer -c sofnocodec cset name='PGA4.0 4 Master Capture Volume' 50
-
-amixer -c sofnocodec cset name='PGA6.0 6 Master Capture Switch' on
-amixer -c sofnocodec cset name='PGA6.0 6 Master Capture Volume' 50
-
-# DMIC capture
-amixer -c sofnocodec cset name='Dmic0 Capture Switch' on
-amixer -c sofnocodec cset name='Dmic0 Capture Volume' 50
-amixer -c sofnocodec cset name='Dmic1 2nd Capture Volume' 50
+# to turn the switches on we have to use the control interface
+amixer -Dhw:0 controls | grep Switch | sed -e 's/.*numid=\([^,]*\),.*/\1/' | while read -r i; do
+    amixer -Dhw:0 cset numid=$i on;
+done
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -745,6 +745,13 @@ re_enable_ntp_sync()
 # param1: platform name
 set_alsa_settings()
 {
+    # If MODEL is defined, set proper gain for the platform
+    if [ -z "$1" ]; then
+	# treat as warning only
+	dlogw "NO MODEL is defined. Please define MODEL to run alsa_settings/MODEL.sh"
+	return 0
+    fi
+
     # ZEPHYR platform shares same tplg, remove '_ZEPHYR' from platform name
     local PNAME="${1%_ZEPHYR}"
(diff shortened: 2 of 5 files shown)


alsa_settings/CAVS_NOCODEC.sh
```diff
@@ -1,21 +1,11 @@
...
-amixer -c sofnocodec cset name='PGA6.0 6 Master Capture Switch' on
-amixer -c sofnocodec cset name='PGA6.0 6 Master Capture Volume' 50
-
-# DMIC capture
-amixer -c sofnocodec cset name='Dmic0 Capture Switch' on
-amixer -c sofnocodec cset name='Dmic0 Capture Volume' 50
-amixer -c sofnocodec cset name='Dmic1 2nd Capture Volume' 50
+# to set the volume as 0dB we have to use the scontrol interface
+amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" | while read -r i; do
+     amixer -Dhw:0 -- sset "$i" 0dB;
```
marc-hb, line 5: Can you please give `$i` a longer name? I don't know, maybe `mixer_name` or something. This code may be obvious to the expert `amixer` user but `$i` is very obscure for everyone else.

The other, second `while` loop does not have this problem thanks `numid=`.

plbossart: good suggestion, updated, thanks.

alsa_settings/CAVS_NOCODEC.sh
```diff
@@ -1,21 +1,11 @@
...
-amixer -c sofnocodec cset name='Dmic0 Capture Switch' on
-amixer -c sofnocodec cset name='Dmic0 Capture Volume' 50
-amixer -c sofnocodec cset name='Dmic1 2nd Capture Volume' 50
+# to set the volume as 0dB we have to use the scontrol interface
+amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" | while read -r mixer_name; do
+     amixer -Dhw:0 -- sset "$mixer_name" 0dB;
+done
+
+# to turn the switches on we have to use the control interface
+amixer -Dhw:0 controls | grep Switch | sed -e 's/.*numid=\([^,]*\),.*/\1/' | while read -r i; do
```
marc-hb, line 9: For shorter lines you can end lines with `|` (and some other characters like `&&` and then the shell automatically escapes the next newline.

marc-hb (commented):
codestyle looks good to me, someone else must approve the logical change.

aiChaoSONG:
@miRoox @keqiaozhang I think ansible is capable for this, so that we don't need alsa_settings in sof-test? please help to review.

miRoox:
I think we need it but at end of `volume-basic-test` rather than the alsa_settings of CAVS_NOCODEC.

plbossart:
> I think we need it but at end of `volume-basic-test` rather than the alsa_settings of CAVS_NOCODEC.

@miRoox we should not assume in which order the tests are run. First we should wipe all previous settings when deploying the kernel/firmware/sof-test on the DUT, and any tests that changes the setup should restore the values of course.

plbossart:
the shellcheck errors are not introduced by me...

marc-hb:
> First we should wipe all previous settings when deploying the kernel/firmware/sof-test on the DUT, and any tests that changes the setup should restore the values of course.

Ideally we always want a configuration reset BOTH before _and_ after testing in case something goes wrong. Examples:
- The test triggers a crash: no clean-up after.
- An interactive user reserves the system and does not really know how to use it: no clean-up before.
- An interactive user forgets to clean-up after he's done: no clean-up after.
- Etc.

BTW that's exactly why people use containers.

marc-hb:
> the shellcheck errors are not introduced by me...

This one looks pretty bad:
```
In test-case/volume-basic-test.sh line 70:
    for i in "${pgalist[@]}"
    ^-^ SC2165: This nested loop overrides the index variable of its parent.
```

plbossart:
well, *someone* will have to take over this PR. I don't have time or knowledge to fix prior issues.

This is now gating the changes for multiple pipelines on playback and capture in https://github.com/thesofproject/sof/pull/6131

@mengdonglin please check who can be assigned to work on this. thanks.

mengdonglin:
> well, _someone_ will have to take over this PR. I don't have time or knowledge to fix prior issues.
> 
> This is now gating the changes for multiple pipelines on playback and capture in [thesofproject/sof#6131](https://github.com/thesofproject/sof/pull/6131)
> 
> @mengdonglin please check who can be assigned to work on this. thanks.

Assign to myself @plbossart :)

plbossart:
The direction suggested with alsa-settings is actually completely wrong. This would only work for the cavs-nocodec case.

What needs to happen is that BEFORE and AFTER all tests impacting volume the PGA and switches are reset with a loop. This would have to be done on ALL platforms, not just nocodec.

In other words, the loops suggested in alsa_settings/CAVS_NOCODEC.sh need to be moved to a library and called from there.

mengdonglin:
@plbossart @marc-hb @keqiaozhang Can we merge this PR at first?  I'll fix the old shellcheck errors next week.  So we can avoid rebasing this PR.

mengdonglin:
@plbossart @marc-hb @keqiaozhang This PR will be replaced by #953

mengdonglin:
@plbossart I closed this PR as PR#953 has been merged. Please re-open if you think this PR is still needed. Thanks!

PR #953, opened 2022-08-29
Reset all PGA volume to 0dB

keqiaozhang:


marc-hb:
Can you clarify the relationship with
- #951?

keqiaozhang:
> Can you clarify the relationship with
> 
> * [alsa_settings: CAVS_NOCODEC: reset values to 0dB/on with script #951](https://github.com/thesofproject/sof-test/pull/951)?

I think https://github.com/thesofproject/sof-test/pull/951 is only reset the volume for nocodec devices. But we found the plain arecord issue(https://github.com/thesofproject/linux/issues/3804) on many other models recently:
http://sof-ci.sh.intel.com/#/result/planresultdetail/14968?model=ADLP_SKU0B00_SDCA_ZEPHYR&testcase=multiple-pause-resume-50
http://sof-ci.sh.intel.com/#/result/planresultdetail/14926?model=ADLP_BRYA_I2S&testcase=check-pause-resume-capture-100
...
This is because the PGA capture volume being turned to the maximum after the `volume_basic_test.sh`. So I added a function to reset the PGA volume to 0dB for all our models at the end of the volume test.

case-lib/lib.sh
```diff
@@ -768,4 +768,14 @@ set_alsa_settings()
     esac
 }
 
+reset_PGA_volume()
+{
+    # set all PGA* volume to 0dB
+    amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep PGA | \
```
marc-hb, line 774: You don't need the `\` backslash, ending a line with `|  ` is enough

marc-hb:
```suggestion
    amixer -Dhw:0 scontrols | sed -e "s/^.* '\(.*\)',.*/\1/" |grep PGA | 
```

test-case/volume-basic-test.sh
```diff
@@ -89,4 +89,7 @@ done
 #clean up background aplay
 pkill -9 aplay
 
+dlogi "Reset all PGA volume to 0dB"
+reset_PGA_volume || die "Failed to reset the PGA volume to 0dB."
```
marc-hb, line 93:
```suggestion
reset_PGA_volume || die "Failed to reset some PGA volume to 0dB."
```

test-case/volume-basic-test.sh
```diff
@@ -89,4 +89,7 @@ done
...
+
```
marc-hb, line 94: I know it's off-topic but please remove this spurious `exit 0` if you ever force-push a new version.

marc-hb (commented):
What do you think of this warning? I know it's off-topic but it looks like a serious bug; serious enough to require fixing before changing other, more minor problems.

```
In test-case/volume-basic-test.sh line 65:
for i in $(seq 1 $maxloop)
^-^ SC2167: This parent loop has its index variable overridden.


In test-case/volume-basic-test.sh line 70:
    for i in "${pgalist[@]}"
    ^-^ SC2165: This nested loop overrides the index variable of its parent.
```

test-case/volume-basic-test.sh
keqiaozhang: Fixed.

case-lib/lib.sh

keqiaozhang:
> serious enough to require fixing before changing other, more minor problems.

I will fix these warnings in another PR next.

fredoh9 (dismissed):
Thank you for finding the root cause why all PGAs are 100%, I totally forgot about volume test. Resetting to 0 db works.
And shellcheck fixes are also good catch.

Let volume test run first before alsabat test after this.

marc-hb:
> replace the index name to kctl for the nested loop
> it may overrides the index variable of its parent.

Thanks for the fix. I haven't read the code but it looks like a very serious fix? Maybe the title of the commit message should not be "fix shellcheck warning" like it's just pleasing shellcheck and nothing else. The reason why we use shellcheck is because sometimes it can find serious bugs like this one and it's worth all the other times when it reports non-serious issues.

keqiaozhang:
> Let volume test run first before alsabat test after this.

I will change the test case order.

keqiaozhang:
> Maybe the title of the commit message should not be "fix shellcheck warning" like it's just pleasing shellcheck and nothing else.

I have revised the commit message.

case-lib/lib.sh
```diff
@@ -768,4 +768,14 @@ set_alsa_settings()
     esac
 }
 
+reset_PGA_volume()
+{
+    # set all PGA* volume to 0dB
+    amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep PGA |
```
aiChaoSONG, line 774: for IPC4 the volume control are named as gain, I think you should also grep gain here.

keqiaozhang: Yes, you are right. But the volume-base-test.sh only support PGA control ATM. The volume test is skipped on all our IPC4 platforms. I can submit another PR for IPC4.

aiChaoSONG (approved):

aiChaoSONG:
merge for v2.3 test.

PR #956, opened 2022-09-09
Enable mtrace

keqiaozhang:


keqiaozhang:
Tested with different FW model:http://sof-ci.sh.intel.com/#/result/planresultdetail/15252. looks like everything is working fine.

case-lib/lib.sh
```diff
@@ -693,6 +693,19 @@ ipc4_used()
     return 0
 }
 
+is_ipc4_zephyr(){
+    local firmware_path firmware_name znum
+    fw_path=/sys/module/snd_sof_pci/parameters/fw_path
```
kv2019i, line 703: The above doesn't work if fw_path is not set. At least I had a local setup where I just use the standard path assumed by kernel for IPC4 FW. 

Maybe if fw_path is NULL, check for /lib/firmware/intel/avs/"$platf"/dsp_basefw.bin . This is what is in SOF kernel driver by default unless fw_path is passed as a module parameter. 

Above can be done in a seperate follow-up PR.

case-lib/lib.sh
```diff
@@ -728,16 +733,12 @@ logger_disabled()
...
 
-    ipc4_used && {
-        # TODO:
-        # Need to remove disabling sof-logger
-        # after sof-logger support for IPC4 has been provided in the future
-        dlogi 'Currenly sof-logger is not supported when running IPC4 mode'
-        dlogi 'SOF logs collection is globally disabled because DUT is running IPC4 mode'
+    if ipc4_used && ! is_ipc4_zephyr; then
+        # mtrace and sof-logger are not supported for closed source IPC4 Firmware
+        dlogi 'Currenly trace loggers are not supported for closed source IPC4 firmware.'
```
kv2019i, line 738: Let's just put "IPC4 FW logging only support with SOF Zephyr build"

kv2019i (approved):
Looks good! Couple of comments inline that can be fixed in a follow-up. Tested this locally and seems to work fine.

case-lib/lib.sh
```diff
@@ -293,6 +293,33 @@ find_ldc_file()
     printf '%s' "$ldcFile"
 }
 
+func_logger_collect()
```
marc-hb, line 314:
```suggestion
func_sof_logger_collect()
```

We have too many logging tools, important to label them well.

case-lib/lib.sh
```diff
@@ -693,6 +693,19 @@ ipc4_used()
...
+    local firmware_path firmware_name znum
+    fw_path=/sys/module/snd_sof_pci/parameters/fw_path
+
+    firmware_path=$(cat $fw_path)
+    firmware_name=dsp_basefw.bin
+
+    ipc4_used || return 1
+
+    znum=$(strings "/lib/firmware/$firmware_path/$firmware_name" | grep -c -i zephyr)
+    test "$znum" -gt 10
```
marc-hb, line 711: I recognize this code from the existing `is_zephyr()` function... which depends on the sof-logger .ldc file.

Please don't create a new, hidden and different `is_zephyr()` function here while keeping the old one. We need a single `is_zephyr()` function.

Either the old `is_zephyr()` function is still OK (because we still depend on sof-logger for DMA logs) and you call it here. Or it's not OK and you fix it; don't leave all the other `is_zephyr()` users broken.

marc-hb (changes requested):

case-lib/lib.sh
marc-hb: To find the firmware file you can try some combination of `./tools/sof-dump-status.py -P` and `./tools/sof-dump-status.py -p`. Again these are used in many other places, so if they don't work then them must be fixed.

@kv2019i , @plbossart, @ujfalusi  sof-test needs something like `/sys/module/snd_sof_pci/parameters/fw_path` but reliable that it can count on, can the kernel expose that? Same thing for the topology: up to now `sof-test/tools/sof-get-default-tplg.sh` has been grepping journalctl; that's a horrible and unreliable hack.


PS: actually the Linux kernel `request_firmware` API should expose all the firmware files it has  loaded from a filesystem, this sort of "manifest" seems like a bare minimum from a security perspective but that could take years.

case-lib/lib.sh
```diff
@@ -366,20 +366,25 @@ func_lib_start_log_collect()
...
-        >&2 dlogw "without sudo permission to run $SOFLOGGER command"
+        >&2 dlogw "without sudo permission to run logging command"
         return 3
     fi
 
     if [ "X$is_etrace" == "X0" ];then
-        log_file=$LOG_ROOT/slogger.txt
-        log_opt="-t"
-        func_logger_collect "$log_file" "$log_opt"
+        if is_ipc4_zephyr; then
```
marc-hb, line 374:
```suggestion
if is_zephyr && is_ipc4; then
```

case-lib/hijack.sh
```diff
@@ -100,6 +101,14 @@ function func_exit_handler()
 
     fi
 
+    if is_ipc4_zephyr; then
```
marc-hb, line 104:
```suggestion
if is_zephyr && is_ipc4; then
```

marc-hb (changes requested):
As with cavstool before, can you please fix `check-sof-logger.sh` first? This will reduce the impact when something goes wrong (as it happened with cavstool)

Note the earlier cavstool attempt  in sof-test #897 (to be abandoned?). It has relevant context about `func_lib_start_log_collect()` and others.

case-lib/lib.sh
keqiaozhang: Yes, I can do that, but we need to solve one problem, the full `fw_path` should be `/lib/firmware/intel/avs/"$platf"/"$key_type"/dsp_basefw.bin`. 

For example, the full fw path of TGLU_UP_HDA_IPC4ZPH is `/lib/firmware/intel/avs/tgl/community/dsp_basefw.bin`.  The key_type is defined in community boards :https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/sof-pci-dev.c#:~:text=/*%20all%20Up,%7D%3B.

I can use the DMI info to get the key_type, but better to let the kernel driver to expose the fw_path as Marc suggested below.

case-lib/lib.sh
keqiaozhang: > Please don't create a new, hidden and different `is_zephyr()` function here while keeping the old one. We need a single `is_zephyr()` function.

Yes, I also considered to reuse `is_zephyr()` function, but it depends on the sof-logger .ldc file. But IPC4 firmware has 
 no ldc file.  if we want to reuse this function, we need to switch to using firmware binary instead of ldc file.

keqiaozhang: > To find the firmware file you can try some combination of `./tools/sof-dump-status.py -P` and `./tools/sof-dump-status.py -p`

 `./tools/sof-dump-status.py -P` only works for IPC3 not IPC4. Part of path is fixed as `/lib/firmware/intel/sof/`,  it only check if the DUT is community board, if yes, the `fw_path` will be changed to `/lib/firmware/intel/sof/community`

keqiaozhang: > can the kernel expose that? Same thing for the topology: up to now `sof-test/tools/sof-get-default-tplg.sh` has been grepping journalctl; that's a horrible and unreliable hack.

I totally agree with this proposal, then we can get the reliable fw_path from the journalctl.

case-lib/lib.sh

case-lib/lib.sh

case-lib/lib.sh
kv2019i: I agree that would be nice @keqiaozhang  and @marc-hb , but let's not block this PR while waiting for a new kernel interface. The hardcoded paths we have in kernel are quite sticky in the end. We can't just change them in the kernel anymore as there are countless system deployed with FW files in the place former kernel versions expected. So unless user has overriden with "fw_path", we can fairly reasonably expect what the firmware path will be.

kv2019i (dismissed):

case-lib/lib.sh
greg-intel: "Currenly" --> "Currently"?

marc-hb:
Another important problem: there are only 3 platforms tested in https://sof-ci.01.org/softestpr/PR956/build135/devicetest/ and they're ALL part of the same audio hardware generation and they are all Zephyr. This is especially problematic for a PR like this one that has messy `is_zephyr` copy/paste/diverge.

Every sof-test PR needs this : https://sof-ci.01.org/linuxpr/PR3860/build714/devicetest/ (part of https://github.com/thesofproject/linux/pull/3860), we can't wait nightly tests to discover regressions.

keqiaozhang:
> Another important problem: there are only 3 platforms tested

I created a new test plan and covered all our test configs: xtos-ipc3, zephyr-ipc3. zephyr-ipc4 and cavs-ipc4:
http://sof-ci.sh.intel.com/#/result/planresultdetail/15402

The results look good. All failures are identified as known issues(sof-logger and IPC4 related issues). The new logging tool mtrace works as expected and sof-logger is not effected.

kv2019i (dismissed):
Seems to work great! And we really, really, really need this as currently we are running blind w.r.t. FW logs in IPC4 test runs.

marc-hb:
> I created a new test plan and covered all our test configs: xtos-ipc3, zephyr-ipc3. zephyr-ipc4 and cavs-ipc4:
http://sof-ci.sh.intel.com/#/result/planresultdetail/15402

This is good but it's a once-off. The (messy) logging code in sof-test is going to evolve more and again very soon and not just to fix the regressions introduced by this PR

> Every sof-test PR needs this : https://sof-ci.01.org/linuxpr/PR3860/build714/devicetest/ (part of https://github.com/thesofproject/linux/pull/3860), we can't wait nightly tests to discover regressions.

case-lib/lib.sh
marc-hb: So here's how I think we can  get this merged soon with a minimum number of reasonable and manageable short-term hacks.

The most important thing is to break down `is_ipc4_zephyr()` into multiple, smaller functions. `is_ipc4_zephyr()` combines too many different logic pieces and decisions in a single function right now. Some pieces are short-term and others not. They must be separated into different functions.

- The easiest thing is to extract is a new function `is_fw_file_zephyr($filename) -> return true/false`

- Then another small function `cavs_filename($platform)` that prints the "semi-harcoded" `/lib/firmware/intel/avs/$platform/dsp_basefw.bin` If you want, let `fw_path` take precedence when defined but that's probably optional.

- Then combine the two functions above in a third new function `avs_is_zephyr()` that gets the platform from `sof-dump-status.py -p`


- Finally, in the logging code do something like:
```shell

	if ipc4_used && avs_is_zephyr; then

```

keqiaozhang:
> can you please fix `check-sof-logger.sh` first?

1. sof-logger failure has nothing to do with mtrace, it's only for IPC3 test. they are old known issues and I don't think we can easily fix them all(https://github.com/thesofproject/sof/issues/6042).
2. `check-sof-logger.sh` is not suitable for IPC4 test, so we didn't add `check-sof-logger.sh` to the formal test plan of IPC4, maybe we can add a new case like `check-mtrace.sh` in future.

case-lib/lib.sh
keqiaozhang: I'm not sure I fully understand your point. 
IPC3 and IPC4 are using a completely different path:
IPC3: `/lib/firmware/intel/sof ` or `/lib/firmware/intel/sof/community/`
IPC4: `/lib/firmware/intel/sof-ipc4/$platform/community` or `/lib/firmware/intel/sof-ipc4/$platform/debug/`

So if we add a function like `is_fw_file_zephyr($filename) -> return true/false`, we first need to check `ipc4_used` or not. 
If we don't know the IPC type, then it's hard for us to know the specific `fw_path`, 
So the process should be:
1. check ipc type `ipc4_used`
2. get the  `fw_path`
3. check if the fw is zephyr

Almost the same steps as I did in function `is_ipc4_zephyr`.

kv2019i:
I do agree we need to add a separate test case for mtrace that just checks availability of mtrace. But @marc-hb is that a blocker for this PR? 

Btw, I filed a separate enhancement item for kernel to expose the chosen tplg+fw -> https://github.com/thesofproject/linux/issues/3867

kv2019i:
And @marc-hb and @keqiaozhang  , please note mtrace (SRAM logging) is not a replacement for sof-logger (DMA based logging). We will still have a new DMA-based logging system for IPC4, which is based on probes. For now, mtrace is just a new way to get lots out with no direct comparison in IPC3 SOF builds.

marc-hb:
> > can you please fix check-sof-logger.sh first?

> sof-logger failure has nothing to do with mtrace

Sorry for the confusion, I meant: "can you please add mtrace support to check-sof-logger first"?

> check-sof-logger.sh is not suitable for IPC4 test,

check-sof-logger.sh is generic enough to test any logs from anywhere.

> maybe we can add a new case like check-mtrace.sh in future.

Please no copy/paste/diverge.

marc-hb:
> I do agree we need to add a separate test case for mtrace that just checks availability of mtrace.

That's exactly what check-sof-logger.sh does for two types of SRAM logging already.

> But @marc-hb is that a blocker for this PR?

No but it would be safer. When we added cavstool we started with check-sof-logger which avoided causing a lot of intermittent failures (#897)

case-lib/lib.sh
marc-hb: > So if we add a function like is_fw_file_zephyr($filename) -> return true/false, we first need to check ipc4_used or not.
> ...
> Almost the same steps as I did in function is_ipc4_zephyr

Yes but `is_ipc4_zephyr` is monolithic and unmaintainable. Break it down in smaller, modular, re-usable and maintainable functions.

kv2019i:
@marc-hb wrote:
>> I do agree we need to add a separate test case for mtrace that just checks availability of mtrace.
> That's exactly what check-sof-logger.sh does for two earlier types of SRAM logging already.

I didn't realize that. So ack, then it's appropriate to also test mtrace.

>> But @marc-hb is that a blocker for this PR?
> No but it would be safer. When we added cavstool we started with check-sof-logger which avoided causing a lot of intermittent failures (#897)

Fair enough. The situation is a bit more urgent now as we have unusual number of P1 bugs open where issues are not happening on all hardware, and we have no visibility to FW logs in current CI results.

marc-hb:
To be clear I'm NOT blocking this PR on a `check-sof-logger` update. My main problems are with the monolithic and unmaintainable `is_ipc4_zephyr()` and the lack of sof-test PR testing in general which is especially relevant here considering the conditionals. https://sof-ci.01.org/softestpr/PR956/build137/devicetest/ is basically testing a single configuration!

case-lib/lib.sh
keqiaozhang: Sorry, I didn't figure out a better solution to break `is_ipc4_zephyr` down in smaller. 

I'm thinking to use firmware file to check zephyr in `is_zephyr`  instead of ldc file. 
We can refine this part after https://github.com/thesofproject/linux/issues/3867 is supported. 

Then combine these two functions `ipc4_used` +` is_zephyr`  to replace `is_ipc4_zephyr`.

any other suggestions?

marc-hb: > I'm thinking to use firmware file to check zephyr in is_zephyr instead of ldc file.

Yes, this is what I suggested above: `is_fw_file_zephyr($filename)`. We unfortunately need two different `is_zephyr_XXX()` functions for now. Better make that obvious rather than hide it buried deep in a big, monolithic `is_ipc4_zephyr()` that looks like it's calling the usual `is_zephyr()` while doing something different behind the scenes.

keqiaozhang:
@marc-hb  I have removed `is_ipc4_zephyr` function and added a new one `is_firmware_file_zephyr`. 
So we can use `is_ipc4 && is_firmware_file_zephyr` for IPC4 Zephyr firmware check.

The `is_firmware_file_zephyr` is not the final version, now we still need to check the ipc type first and use different ways to get the firmware name/path for IPC3 and IPC4. This part will be optimized after driver can expose the FW path to userspace via debugfs, then we can remove function `is_zephyr` and use `is_firmware_file_zephyr` instead.

PR testing is passed, will do more test next.

keqiaozhang:
SOFCI TEST

marc-hb (commented):
Thanks for the extra effort and also for the commit breakdown which makes this relatively complex refactoring so much easier to review. I need one night sleep to review the last commit (where all the real fun is), @kv2019i could you please also give it a close look? All the 5 previous commits just prepare the last one and they all look good to me.

keqiaozhang:
SOFCI TEST

case-lib/lib.sh
```diff
@@ -293,6 +293,24 @@ find_ldc_file()
...
+    local clogfile=$LOG_ROOT/mtrace.txt
+
+    if [ -z "$MTRACE" ]; then
+        MTRACE=$(command -v mtrace-reader.py) || {
+            dlogw 'No mtrace-reader.py found in PATH'
+            return 1
+        }
+    fi
+
+    local mtraceCmd="$MTRACE"
```
ujfalusi, line 307: In the last patch you use:
```bash
local mtraceBin; mtraceBin=mtrace-reader.py
```

Which one is the preferred one?

case-lib/lib.sh
```diff
@@ -685,8 +685,8 @@ ipc4_used()
 
     # If /sys/module/snd_sof_pci/parameters/ipc_type exists
     # If the value of file ipc_type is:
-    # 0: DUT runs IPC3 mode, ipc_used return 1(false)
-    # 1: DUT runs IPC4 mode, ipc4_used return 0(true)
+    # 0: DUT runs IPC3 mode, is_ipc4 return 1(false)
+    # 1: DUT runs IPC4 mode, is_ipc4 return 0(true)
```
ujfalusi, line 720: The comment does not make any sense
 0: DUT runs IPC3 mode, **is_ipc4 return 1(false)**
1: DUT runs IPC4 mode, **is_ipc4 return 0(true)**
if ipc_type is 1, this function returns 1 (is_ipc4 is true)

case-lib/lib.sh
```diff
@@ -672,6 +672,31 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+is_firmware_file_zephyr()
+{
+    local firmware_name firmware_path znum platform
+    # get the FW name and path
+    # TODO: optimize this part after driver can expose the FW path to userspace via debugfs
+    if is_ipc4; then
```
ujfalusi, line 686:
```bash
    if [ is_ipc4 ]; then
```

case-lib/lib.sh
```diff
@@ -366,18 +366,24 @@ func_lib_start_log_collect()
...
-        >&2 dlogw "without sudo permission to run $SOFLOGGER command"
+        >&2 dlogw "without sudo permission to run logging command"
         return 3
     fi
 
     if [ "X$is_etrace" == "X0" ];then
-        log_file=$LOG_ROOT/slogger.txt
-        log_opt="-t"
-        func_sof_logger_collect "$log_file" "$log_opt"
+        if is_ipc4 && is_firmware_file_zephyr; then
```
ujfalusi, line 374:
```bash
if [ is_ipc4 ] && [ is_firmware_file_zephyr ]; then
```

case-lib/lib.sh
```diff
@@ -366,18 +366,24 @@ func_lib_start_log_collect()
...
+        fi
     else
-        log_file=$LOG_ROOT/etrace.txt
-        log_opt=""
-        func_sof_logger_collect "$log_file" "$log_opt"
+        is_ipc4 || {
+            log_file=$LOG_ROOT/etrace.txt
+            log_opt=""
+            func_sof_logger_collect "$log_file" "$log_opt"
+        }
```
ujfalusi, line 386: I'm not following this syntact

kv2019i (approved):
Still ok to me. There are a few obviously wrong comments that are probably better to address immediately, but I did not see any showstoppers in functionality side.

case-lib/lib.sh
keqiaozhang: Because in  `func_mtrace_collect()`, the `local mtraceCmd="$MTRACE"` can be set when testing with IPC4-Zephyr platforms.

So in `hijack.sh`, if we test on non-IPC4 Zephyr platforms. the mtraceCmd will be empty, we cannot reuse `mtraceCmd` in `hijack.sh`. So I set mtrace separately in these 2 scripts.

case-lib/lib.sh
keqiaozhang: I just checked, for IPC3 mode, the ipc_type is `-1`(updated). 
I don't see any problems here. if `ipc_type=1`, `is_ipc4` function returns 0. which means `is_ipc4` is true.
We use `|| `here.
```shell
 [ "$ipc_type" -eq 1 ] || {
        return 1
    }
    return 0
```

case-lib/lib.sh
keqiaozhang: Sorry, is this really matter to you? We used it in this way in many places and there is no shellcheck warning.

case-lib/lib.sh
keqiaozhang: It means only non-ipc4 platforms have etrace support.

case-lib/lib.sh
ujfalusi: OK, I see. Thanks

case-lib/lib.sh
keqiaozhang: I just tried, if I do it like you suggested, the shellcheck will be failed.
```
In case-lib/lib.sh line 686:
    if [ is_ipc4 ]; then
         ^-----^ SC2078 (error): This expression is constant. Did you forget a $ somewhere?

```

ujfalusi: Sorry, this is fine, is_ipc4 is a function.

kv2019i (dismissed):

keqiaozhang:
Tested this PR with different models, mtrace works as expected on IPC4ZPH platforms(except for the check-sof-logger.sh) and sof-logger is not effected.
http://sof-ci.sh.intel.com/#/result/planresultdetail/15672.

2 on device PR tests also passed.

case-lib/hijack.sh
```diff
@@ -100,6 +101,14 @@ function func_exit_handler()
 
     fi
 
+    if is_ipc4 && is_firmware_file_zephyr; then
```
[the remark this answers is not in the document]
marc-hb: Duplicate of a resolved discussion below. Please resolve all comments like this one or none :-)

lgirdwood:
All - need some working mtrace sooner rather than later - we can polish it incrementally but we need to start getting the data out.

marc-hb (commented):
A couple small changes and we're good to go.

> All - need some working mtrace sooner rather than later - we can polish it incrementally but we need to start getting the data out.

This code is incredibly awkward, badly designed and brittle for what it does. @keqiaozhang is doing a delicate and amazing refactoring job but we have 1.9 persons total understanding it and this code is quite critical (as demonstrated by how much this PR is expected). One of the 1.9 persons is me and I have to spend literally 1 hour to make sense of it EVERY TIME I LOOK AT IT so I can't afford to make it worse than it already is.

Thanks everyone for your patience.

keqiaozhang:
I have updated this PR based on the last comments. Thanks you all for your review and patience. Tests are running.

case-lib/lib.sh
```diff
@@ -672,6 +672,31 @@ is_zephyr()
...
+    # TODO: optimize this part after driver can expose the FW path to userspace via debugfs
+    if is_ipc4; then
+        firmware_name=dsp_basefw.bin
+        fw_mod_para=/sys/module/snd_sof_pci/parameters/fw_path
+        if [ -s "$fw_mod_para" ]; then
+            firmware_path=$(cat $fw_mod_para)
+        else
+            # TODO: let the kernel driver expose the FW path
+            # and get the FW path by grepping journalctl.
+            return 1
```
marc-hb, line 696: We should `die` here because we don't know the answer. `return 1` means "not Zephyr"

case-lib/hijack.sh
```diff
@@ -24,7 +24,8 @@ function func_exit_handler()
     fi
 
     # when sof logger collect is open
-    if [ "X$SOF_LOG_COLLECT" == "X1" ]; then
+    if ! is_ipc4 && [ "X$SOF_LOG_COLLECT" == "X1" ]; then
+	local loggerBin wcLog; loggerBin=sof-logger
```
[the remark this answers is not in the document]
marc-hb: ping?

marc-hb (approved):
- You forgot to squash one fixup commit
- I don't see any mtrace output in internal test run 15728

kv2019i (approved):
Tested on local setup and seems to work fine.

keqiaozhang:
Sorry, I didn't get your point.




> * I don't see any mtrace output in internal test run 15728

Please check the cases under "TGLU_UP_HDA_IPC4ZPH" and "TGLU_RVP_NOCODEC_IPC4ZPH", I added 2 IPC4 platforms to check the mtrace. Others are XTOS-IPC3, Zephyr-IPC3, CAVS-IPC4.

case-lib/hijack.sh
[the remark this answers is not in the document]
keqiaozhang: > Please don't change or move this line. This refactoring is complex and risky enough, stick to the bare minimum required.

Hope I didn't misunderstand your point. PR is updated for 
```shell
local loggerBin wcLog; loggerBin=sof-logger
```

case-lib/lib.sh
keqiaozhang: Updated with die.

marc-hb (approved):

marc-hb:
> Please check the cases under ...

I see it now, sorry for the noise.

PR #957, opened 2022-09-13
tplgtool2: skip unknown vendor array

miRoox:
The topology might contain custom vendor array data, which is unrecorgnized by this "standard" topology parser. Previously, it didn't handle this case and will get stuck when parsing that part. Now it will skip the unknown vendor array and continue parsing the rest parts.


marc-hb:
Superseded by #958 I guess

PR #958, opened 2022-09-13
tplgtool2: parse vendor array for DAPM widgets only and raise error for unknown vendor types 

miRoox:
Previously, parsing `cavs-tgl-nocodec.tplg` got stuck because it treat the NHLT table in the manifest private data as a vendor array and didn't catch the unexpected vendor tuple types.

These 2 parts to fix it and improve soundness:

1. parse vendor array for DAPM widgets only
2. raise error for unknown vendor types

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -351,7 +351,7 @@ class TplgBinaryFormat:
             "size" / Int32ul, # size in bytes of the array, including all elements
             "type" / Enum(Int32ul, VendorTupleType),
             "num_elems" / Int32ul, # number of elements in array
-            "elems" / Array(this.num_elems, Switch(this.type, self._vendor_elem_cases))
+            "elems" / Array(this.num_elems, Switch(this.type, self._vendor_elem_cases, default=construct.Error)),
         )
         # expand the union of snd_soc_tplg_private for different sections:
         self._private_raw = Prefixed(
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -8,7 +8,7 @@ import re
 import sys
 import typing
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


aiChaoSONG (approved):

marc-hb:
Could this fix the missing topology in https://github.com/thesofproject/sof-test/pull/958 ?

cc: @plbossart

fredoh9:
It was stuck at a contruct function. glad found the fix.

```
...
File "/usr/lib/python3/dist-packages/construct/core.py", line 312, in _parsereport
    obj = self._parse(stream, context, path)
  File "/usr/lib/python3/dist-packages/construct/core.py", line 3917, in _parse
    keyfunc = evaluate(self.keyfunc, context)
  File "/usr/lib/python3/dist-packages/construct/core.py", line 198, in evaluate
    return param(context) if callable(param) else param
  File "/usr/lib/python3/dist-packages/construct/expr.py", line 184, in __call__
    def __call__(self, obj, *args):
```

fredoh9:
Looking at the failure, it released the stuck after 30min. I was wondering how it was bailed out from the stuck itself. It looks exit due to kernel Out of memory error. Then the script crashed and exit with 255.

`[ 1720.045898] kernel: Out of memory: Killed process 1457 (python3) total-vm:17467848kB, anon-rss:15470464kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:33780kB oom_score_adj:0`

marc-hb:
Out of memory looks like a really bad `construct` bug. Which machine does this tool currently run on? I'm curious about the `construct` version.

miRoox:
Tested on most of our in-use models: https://sof-ci.sh.intel.com/#/result/planresultdetail/15355?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=verify-kernel-boot-log, the errors on verify-kernel-boot-log is unrelated.

miRoox:
> Out of memory looks like a really bad `construct` bug. Which machine does this tool currently run on? I'm curious about the `construct` version.

It's not a `construct` bug, just like I implicitly write an infinitely recursive function in this case. But of course, it would be better if `construct` provides something like max iteration.

PR #959, opened 2022-09-14
check-playback: add check for FW performance

kv2019i:
**NOTE:** This PR is based on top of https://github.com/thesofproject/sof-test/pull/956

check-playback: add check for FW performance
    
Add a tool to analyze the performance traces that are emitted
by the FW low-latency scheduler, and use this tool in the check
playback test case.
    
The tool requires output of FW logs and currently only supports
the mtrace output available in SOF Zephyr IPC4 builds. This can
be extended later for other output types.
  
The sof-ll-timer-check.py can do some simple global check like
observing occurences of low-latency scheduler overruns and raise
an error in this case.
    
For most of the analysis, sof-ll-timer-check.py needs reference
data detailing what the expected performance level should be.
To implement this, a simple JSON database is added via
sof-ll-timer-check-db.json that is used to query the reference
data. As this data is manually maintained, it is expected
reference data is only used for a small set of key use-cases for
any given platform.

### DIFF
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -97,6 +97,15 @@ do
                     func_lib_lsof_error_dump "$snd"
                     die "aplay on PCM $dev failed at $i/$loop_cnt."
                 }
+
+		tplg_basename=$(basename $tplg)
+		platform=$(sof-dump-status.py -p)
+		is_ipc4_zephyr && {
+		    data_file=$LOG_ROOT/mtrace.txt
+		    test_reference_key="${platform}-${tplg_basename}-ipc4-zephyr-check-playback-${dev}"
+		    TOPDIR="$(dirname "${BASH_SOURCE[0]}")"/..
+		    $TOPDIR/tools/sof-ll-timer-check.py ${data_file} $test_reference_key $TOPDIR/tools/sof-ll-timer-check-db.json
+		}
             done
         done
     done
--- a/tools/sof-ll-timer-check.py
+++ b/tools/sof-ll-timer-check.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+'''Script to analyze performance data in SOF FW log output'''
+
+import sys
+import re
+import json
+from statistics import median
+
+# allowed error margin before error is raised for
+# lower observed performance (1.05 -> 5% difference
+# required to raise error)
+AVG_ERROR_MARGIN = 1.05
+
+max_vals = []
+avg_vals = []
+overruns = 0
+
+f = open(sys.argv[1])
+
+for line in f:
+    m = re.search('.*ll timer avg ([0-9]*), max ([0-9]*), overruns ([0-9]*)', line)
+    if m:
+        avg_vals.append(int(m.group(1)))
+        max_vals.append(int(m.group(2)))
+        overruns += int(m.group(3))
+
+median_avg_vals = median(avg_vals)
+print("Measurements:\t\t%d" % len(avg_vals))
+print("Median avg reported:\t%d" % median_avg_vals)
+print("Median max reported:\t%d" % median(max_vals))
+print("Highest max reported:\t%d" % max(max_vals))
+
+if overruns:
+    print("ERROR: %s overruns detected" % overruns, file=sys.stderr)
+    sys.exit(-1)
+
+if len(sys.argv) < 4:
+    print("No reference data for key '%s', unable to check performance against reference")
+    sys.exit(0)
+
+median_avg_ref = None
+dbfile = open(sys.argv[3])
+ref_key = sys.argv[2]
+ref_data_all = json.load(dbfile)
+
+for ref in ref_data_all:
+    if ref["test-key"] == ref_key:
+        median_avg_ref = ref["ll-timer-avg"]
+        break
+
+if not median_avg_ref:
+    print("No reference data for key '%s', unable to check performance against reference" % ref_key)
+    sys.exit(0)
+
+median_avg_ref_w_margin = median_avg_ref * AVG_ERROR_MARGIN
+if median_avg_vals > median_avg_ref_w_margin:
+    print("ERROR: ll-timer-avg median %d over threshold %d (%d without margin)" % (median_avg_vals, median_avg_ref_w_margin, median_avg_ref), file=sys.stderr)
+    sys.exit(-1)
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -7,6 +7,8 @@ trap 'func_exit_handler $?' EXIT
 function func_exit_handler()
 {
     local exit_status=${1:-0}
+    local loggerBin wcLog; loggerBin=sof-logger
+    local mtraceBin; mtraceBin=mtrace-reader.py
 
     # call trace
     if [ "$exit_status" -ne 0 ] ; then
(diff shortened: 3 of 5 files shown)


tools/sof-ll-timer-check.py
```diff
@@ -0,0 +1,59 @@
...
+    print("ERROR: %s overruns detected" % overruns, file=sys.stderr)
+    sys.exit(-1)
+
+if len(sys.argv) < 4:
+    print("No reference data for key '%s', unable to check performance against reference")
+    sys.exit(0)
+
+median_avg_ref = None
+dbfile = open(sys.argv[3])
+ref_key = sys.argv[2]
```
marc-hb, line 44: `argparse` requires a tiny bit more effort than `logging` but it's really worth it. Don't procrastinate. `git grep argparse` will find multiple examples in this repo https://docs.python.org/3/howto/argparse.html

tools/sof-ll-timer-check.py
```diff
@@ -0,0 +1,59 @@
...
+ref_data_all = json.load(dbfile)
```
marc-hb, line 45: Have you considered a `.init` file?
https://docs.python.org/3/library/configparser.html

It's less powerful and less flexible but more user-friendly and more programmer friendly.

Long story short: do you expect a _hierarchy_ of values?

tools/sof-ll-timer-check.py
```diff
@@ -0,0 +1,59 @@
...
+
+for ref in ref_data_all:
+    if ref["test-key"] == ref_key:
+        median_avg_ref = ref["ll-timer-avg"]
+        break
+
+if not median_avg_ref:
+    print("No reference data for key '%s', unable to check performance against reference" % ref_key)
```
marc-hb, line 53: You would probably need that code with a simpler .ini file, it would likely just fail with a meaning full stack trace and error message.

tools/sof-ll-timer-check.py
```diff
@@ -0,0 +1,59 @@
...
+
+import sys
+import re
+import json
+from statistics import median
+
+# allowed error margin before error is raised for
+# lower observed performance (1.05 -> 5% difference
+# required to raise error)
+AVG_ERROR_MARGIN = 1.05
```
marc-hb, line 13: This feels overkill / YAGNI... why not just bump the thresholds 5%? There are only two of them.

tools/sof-ll-timer-check.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
```
marc-hb, line 1: Most of what this script does is parsing a log file but there's neither "parsing" nor "log" in the name. This is a test code so I don't think this is an "implementation detail" that should be hidden.

test-case/check-playback.sh
```diff
@@ -97,6 +97,15 @@ do
...
                     die "aplay on PCM $dev failed at $i/$loop_cnt."
                 }
+
+		tplg_basename=$(basename $tplg)
+		platform=$(sof-dump-status.py -p)
+		is_ipc4_zephyr && {
+		    data_file=$LOG_ROOT/mtrace.txt
+		    test_reference_key="${platform}-${tplg_basename}-ipc4-zephyr-check-playback-${dev}"
+		    TOPDIR="$(dirname "${BASH_SOURCE[0]}")"/..
+		    $TOPDIR/tools/sof-ll-timer-check.py ${data_file} $test_reference_key $TOPDIR/tools/sof-ll-timer-check-db.json
```
marc-hb, line 107: Now it's just in check-playback.sh but I expect this to be found in multiple tests in the future. So please some new `check_firmware_load()` (pick the name) shell function / indirection that can be quickly deployed across many tests, quickly turned on/off test-suite wide, easily switched to a less crude method that does not parse the logs, etc.

marc-hb (changes requested):

marc-hb:
> Have you considered a .init file?
https://docs.python.org/3/library/configparser.html
> It's less powerful and less flexible but more user-friendly and more programmer friendly.
> Long story short: do you expect a hierarchy of values?

Answering myself: probably yes considering the huge range of hardware and software configurations, workloads etc.

Which brings the next question: should these values be hardcoded in sof-test commits? Most likely not because so far we've successfully kept sof-test and sof mostly independent, in concrete terms most sof-test versions can test most sof versions and configurations. So SOF performance thresholds should really not be hardcoded in sof-test, they should come from elsewhere. Not sure where unfortunately, in CI maybe?

Of course have small JSON _examples_ in sof-test is fine.

test-case/check-playback.sh
```diff
@@ -97,6 +97,15 @@ do
                     func_lib_lsof_error_dump "$snd"
                     die "aplay on PCM $dev failed at $i/$loop_cnt."
                 }
+
+		tplg_basename=$(basename $tplg)
+		platform=$(sof-dump-status.py -p)
+		is_ipc4_zephyr && {
+		    data_file=$LOG_ROOT/mtrace.txt
```
marc-hb, line 104: AFAIK this new test is merely scanning logs so it should be hardcoded to neither IPC4 nor Zephyr nor mtrace.

tools/sof-ll-timer-check.py
kv2019i: That should be considered, yeah. I frankly don't yet know whether even whole concept is pragmatic/feasible, so I just wanted to get some minimum-working-set and json was good for this.

tools/sof-ll-timer-check.py
kv2019i: @marc-hb I added this on purpose. Basicly anyone setting baseline values is going to run a test and copy values over to the database. Probably we add a option to the test script it self that if you pass "--reference-capture", script will run the test, measure performance, but instead of comparing against expected value, just write the measured values out in jsof/init format, so they can be incorporated in the database (after of course reviewed that values are a sane new baseline).

So to make above possible, it's better the margin is stored elsewhere.

kv2019i:
@marc-hb wrote:
> So SOF performance thresholds should really not be hardcoded in sof-test, they should come from elsewhere. Not sure where unfortunately, in CI maybe?

That's a very good point and answer is probably somewhere else (wherever the CI running sof-test keeps track of the device configurations). 

OTOH, considering this probably needs a PoC run in CI to get sense of usefulness of the approach, the initial database might just as well be in sof-test for simplicity. If proven useful and it becomes clear the database is going to be populated and maintained, the data needs to be split out. Oh well, probably this doesn't help much, so we could just as well put the database out from day 1.

test-case/check-playback.sh
kv2019i: @marc-hb  Well that's complicated. The log output structure is specific to Zephyr (it uses formatting of the Zephyr logging subsystem). IPC4 should not be a dependency, but we do need separate code to handle sof-logger output.

kv2019i: One big TODO that needs to be addresses is limiting the analysis to also cover logs that happened during the test run. Here full mtrace.txt is analyzed, but it is not guaranteed the DSP went to D3 between the test iterations. So some kind of capture of the timestamp of last line of log line before test starts, and pruning the log to only cover the new entries added during test execution, is needed.

greg-intel: Seems complicated...  
1) Can the analysis be ran last, after the log contributors have finished?
2) Can the analysis be ran as a "follow," so it gathers updates automatically?
3) Can the logs be backed up and purged after the initial run?  

It's possible 3 may not work, as it may still present the same challenges.

kv2019i:
Closing this, @btian1 will continue with a slightly modified approach.

PR #966, opened 2022-10-12
lib.sh: initialize with alsactl before setting proper alsa settings

fredoh9:
alsaclt init will try to initialize all sound devices to a default state. It doesn't guarantee to set default for all, but good to start with when proper alsa settings are desired.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -802,6 +802,10 @@ set_alsa_settings()
 {
     # ZEPHYR platform shares same tplg, remove '_ZEPHYR' from platform name
     local PNAME="${1%_ZEPHYR}"
+
+    # Initialize alsa settings first
+    alsactl init
+
     dlogi "Run alsa setting for $PNAME"
     case $PNAME in
         APL_UP2_NOCODEC | CML_RVP_NOCODEC | JSL_RVP_NOCODEC | TGLU_RVP_NOCODEC | ADLP_RVP_NOCODEC | TGLH_RVP_NOCODEC)
@@ -825,6 +829,9 @@ set_alsa_settings()
 
 reset_PGA_volume()
 {
+    # Initialize alsa settings first
+    alsactl init
+
     # set all PGA* volume to 0dB
     amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep PGA |
     while read -r mixer_name


case-lib/lib.sh
```diff
@@ -802,6 +802,10 @@ set_alsa_settings()
 {
     # ZEPHYR platform shares same tplg, remove '_ZEPHYR' from platform name
     local PNAME="${1%_ZEPHYR}"
+
+    # Initialize alsa settings first
+    alsactl init
+
```
kv2019i, line 808: @fredoh9  Note, this should also get the UCM defaults, but it is worth double-checking this does happen (alsactl also has "  -D,--ucm-defaults  execute also the UCM 'defaults' section" option but AFAIK this should not be needed)

Example default mixer settings for SOF HDA cards are here in alsa-ucm:
https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf

The "True.BootSequence" sections should be executed to put the card into a sane default state. 

FYI @greg-intel @marc-hb

greg-intel:
One of the device failures was related to `org.gnome.Shell@x11.service`

case-lib/lib.sh
```diff
@@ -802,6 +802,10 @@ set_alsa_settings()
 {
     # ZEPHYR platform shares same tplg, remove '_ZEPHYR' from platform name
     local PNAME="${1%_ZEPHYR}"
+
+    # Initialize alsa settings first
```
greg-intel, line 806: Maybe this can be an echo rather than a comment?  Likewise with the other comment below.  The only `set -x` I saw in this file were wrapped in sub-shells. Approving regardless.

greg-intel (approved):

case-lib/lib.sh
marc-hb: I agree with @greg-intel , this looks like a big change and it seems very "quiet": I tried this command on my system and absolutely nothing was printed. Also, `dlogi "Run alsa setting for $PNAME"` runs AFTER this, could be confusing.

So either a `dlogi` or maybe just this:
```diff
-  alsactl init
+  alsactl --debug init
```

marc-hb:
Merging this during the holiday season looks risky to me. Also, Fred is on vacation for a couple weeks.

case-lib/lib.sh
marc-hb: > So either a new dlogi or maybe just this...

@fredoh9 , ping?

fredoh9: I lost the momentum, need to review what was my intention at first

marc-hb:
SOFCI TEST

marc-hb:
LNL failure in https://sof-ci.01.org/softestpr/PR966/build659/devicetest/index.html is unrelated (https://github.com/thesofproject/linux/issues/5109)

The MTL timeout in https://sof-ci.01.org/softestpr/PR966/build657/devicetest/index.html is very unusual but can't possibly be related either.

Everything else is 100% green.

@fredoh9 , @plbossart , @ujfalusi , @kv2019i do we still want this `alsactl init` before every test? Single-line change, super easy to revert if needed.

Maybe this could help with some of the MAX volume issues? (e.g. https://github.com/thesofproject/linux/issues/5048, https://github.com/thesofproject/linux/issues/3766, etc.)

PR #972, opened 2022-10-18
[SKIP CI] Test-case: Test if acoustic capture of ambient is normal

singalsu:
Capture sound clip from microphone endpoint and automatically analyze if recording has normal characteristic. Detected issues are only PCM zero code or DC constant DC values, large offset, duplicated channels, saturated PCM code values, or abnormal FFT spectrum.

### DIFF
--- a/test-case/noise_capture_check.sh
+++ b/test-case/noise_capture_check.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+check_device_format () {
+    DEV=$1
+    FMT=$2
+    WAV=$(mktemp).wav
+    OCTAVE_SCRIPT="noise_capture_check('${WAV}')"
+    echo Testing "$DEV"
+    rm -f "$WAV"
+    arecord -D${DEV} $FMT -d 2 "$WAV"
+    octave --silent --no-gui --eval "$OCTAVE_SCRIPT"
+    status=$?
+    [ $status -eq 0 ] && echo Passed.
+    rm -f "$WAV"
+}
+
+check_device_format "hw:1,0" "-c 2 -r 48000 -f S16_LE"
+check_device_format "hw:1,6" "-c 2 -r 48000 -f S16_LE"
+check_device_format "hw:1,6" "-c 2 -r 48000 -f S32_LE"
+check_device_format "hw:1,7" "-c 2 -r 16000 -f S16_LE"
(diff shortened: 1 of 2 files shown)


singalsu:
@marc-hb Here's a draft version of ambient noise based mic capture test for try at some lab. Fan noise, speech, music etc. in background should pass while several kinds of problems should be detected.

test-case/noise_capture_check.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
marc-hb, line 1: The  shell script is useful for testing the octave script in isolation but for "real" tests it's probably not needed, existing `test-case/*capture*.sh` tests should probably be modified instead.

singalsu: Yes, that makes sense. Also would need a way to launch this only in devices with plugged in headset and/or analog or digital microphone available. Something like /etc/defaults/has_headset, etc?

test-case/noise_capture_check.sh
```diff
@@ -0,0 +1,23 @@
...
+check_device_format () {
+    DEV=$1
+    FMT=$2
+    WAV=$(mktemp).wav
+    OCTAVE_SCRIPT="noise_capture_check('${WAV}')"
+    echo Testing "$DEV"
+    rm -f "$WAV"
+    arecord -D${DEV} $FMT -d 2 "$WAV"
+    octave --silent --no-gui --eval "$OCTAVE_SCRIPT"
+    status=$?
```
marc-hb, line 14: This will always be zero because of `set -e`

test-case/noise_capture_check.sh
```diff
@@ -0,0 +1,23 @@
...
+    [ $status -eq 0 ] && echo Passed.
+    rm -f "$WAV"
+}
+
+check_device_format "hw:1,0" "-c 2 -r 48000 -f S16_LE"
+check_device_format "hw:1,6" "-c 2 -r 48000 -f S16_LE"
+check_device_format "hw:1,6" "-c 2 -r 48000 -f S32_LE"
+check_device_format "hw:1,7" "-c 2 -r 16000 -f S16_LE"
+check_device_format "hw:1,7" "-c 2 -r 16000 -f S32_LE"
```
marc-hb, line 23: Not compliant with
- #740

singalsu:
@singalsu reminder to myself to proceed with this after vacation

PR #973, opened 2022-10-20
lib.sh: simplify the check for firmware rtos

aiChaoSONG:
Previously we do a lot of test to get the firmware path, it is not necessary, because we can acquire it simply with journalctl.   Then we can test if the firmware loaded is using zephyr rtos by checking the occurrence of 'zephyr' in the firmware binary as before.

First (maybe the only) step to unlock sof-test for mtl

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -680,28 +680,23 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+# FIXME: the kernel driver should give us the FW path
+# https://github.com/thesofproject/linux/issues/3867
+get_firmware_path()
+{
+    journalctl_cmd -k |
+        awk '/sof.*request_firmware/ { sub(/^.*request_firmware/,""); last_loaded_file=$1 } END { print last_loaded_file }'
+}
+
 is_firmware_file_zephyr()
 {
-    local firmware_name firmware_path znum platform
-    # get the FW name and path
-    # TODO: optimize this part after driver can expose the FW path to userspace via debugfs
-    if is_ipc4; then
-        firmware_name=dsp_basefw.bin
-        fw_mod_para=/sys/module/snd_sof_pci/parameters/fw_path
-        if [ -s "$fw_mod_para" ]; then
-            firmware_path=$(cat $fw_mod_para)
-        else
-            # # FIXME: the kernel driver should give us the FW path
-            # https://github.com/thesofproject/linux/issues/3867
-            die "Failed to get the IPC4 FW path."
-        fi
-    else # for IPC3
-        platform=$(sof-dump-status.py -p)
-        firmware_name=sof-$platform.ri
-        firmware_path=$(sof-dump-status.py -P)
-    fi
+    local firmware_path znum
 
-    znum=$(strings "/lib/firmware/$firmware_path/$firmware_name" | grep -c -i zephyr)
+    firmware_path=$(get_firmware_path)
+    [ -n "$firmware_path" ] ||
+        die 'firmware path not found from journalctl, no firmware loaded or debug option disabled?'
+
+    znum=$(strings "/lib/firmware/$firmware_path" | grep -c -i zephyr)
     test "$znum" -gt 10
 }
 


ranj063 (approved):

marc-hb (changes requested):
My understanding of @plbossart's comment in https://github.com/thesofproject/linux/issues/3867#issuecomment-1285942941 is that we should always parse the debug logs, end of story. So this PR should not exist and #3867 should be closed?

Blocking this PR until we have clear, final agreement from everyone agrees on the final solution must be (even if we tolerate short term hacks in sof-test in the mean time).

case-lib/lib.sh
```diff
@@ -680,13 +680,26 @@ is_zephyr()
     test "$znum" -gt 10
 }
 
+get_ipc4_fw_name()
+{
+    local platf; platf=$(sof-dump-status.py -p)
+    case "$platf" in
+        tgl|adl)
```
plbossart, line 687: that will break sof-test on CML/SKL and not work for RPL either. 
don't we have a shortcut for cAVS platforms?

plbossart (dismissed):

case-lib/lib.sh
aiChaoSONG: @marc-hb @plbossart  it is not reliable to extract the firmware and topology used from kernel log, because it fails sometimes due to limited dmesg buffer.

Better fixes should be:
- CI records the what has been deployed and passes the information to sof-test.
- The kernel exposes what it has loaded(firmware/topology) to sysfs/proc.

comments?

marc-hb: > because it fails sometimes due to limited dmesg buffer.

I don't like parsing logs but `dmesg` is never a good excuse because there is simply never any reason to use `dmesg` in any script for the exact reason you just mentioned - and others. `journalctl -k` should always be used instead, it's also better for other reasons. Let's treat `dmesg` like a dirty word: it can be used in private but never mentioned in public :-)

Of course you already know that because you... did all the `journalctl` work!!
- https://github.com/thesofproject/sof-test/pull/639

marc-hb: > CI records what has been deployed and passes the information to sof-test.

If there is some bug or configuration issue in CI (and with IPC4 there has been VERY MANY!) then this will NOT match what the kernel actually loaded.

> The kernel exposes what it has loaded(firmware/topology) to sysfs/proc, https://github.com/thesofproject/linux/issues/3867

Either 3867 or parsing debug logs, looking for the last firmware that was loaded. We already (try to) do that in `verify-sof-firmware-load.sh`

aiChaoSONG: For tigerlake ipc4, sof-tgl.ri is renamed to dsp_basefw.bin, so the version major number can not be used to determine firmware name solely.

case-lib/lib.sh
```diff
@@ -682,26 +682,10 @@ is_zephyr()
...
-    else # for IPC3
-        platform=$(sof-dump-status.py -p)
-        firmware_name=sof-$platform.ri
-        firmware_path=$(sof-dump-status.py -P)
-    fi
+    local firmware_path znum
 
-    znum=$(strings "/lib/firmware/$firmware_path/$firmware_name" | grep -c -i zephyr)
+    firmware_path=$(journalctl_cmd | grep "request_firmware" | awk -F ' ' '{print $(NF-1)}')
+    znum=$(strings "/lib/firmware/$firmware_path" | grep -c -i zephyr)
```
plbossart, line 688: what makes you think 'zephyr' is used in the kernel @aiChaoSONG ?

aiChaoSONG: the function `is_firmware_file_zephyr()` is used to check if the firmware uses zephyr, if yes，mtrace-reader.py is used for firmware log. btw, I only simplify the function, don't change its behavior.

plbossart: the firmware path is not by default using the 'zephyr' prefix in any kernel descriptor.

Maybe that's what you use in your CI setup, but that would mean that you override the default locations.

That cannot be a solution.

marc-hb: @plbossart `strings` is searching the file content, not its path.

mengdonglin: When we previously align FW and topology layout, we agreed on CI DUTs:
- For SOF Zephyr+ IPC3, driver will request /lib/firmware/intel/sof-ipc3-zephyr/<key>/sof-<platform>.ri
- For SOF Zephyr + IPC4 on cavs platforms, driver will request /lib/firmware/intel/sof-ipc4/<platform name>/[key]/dsp_basefw.bin
- For SOF Zephyr + IPC4 on MTL, driver will request /lib/firmware/intel/sof-ipc4/<platform name>/[key]/sof-<platform name>.ri
- For IPC4 tester, driver will request for /lib/firmware/intel/avs/<platform name>/dsp_basefw.bin

Can we assume developers' personal test device also follow this agreement?  @plbossart @marc-hb @aiChaoSONG 
If yes, the firmware path contains `sof-ipc3-zephyr` or `sof-ipc4`, we know is Zephyr builds. If developers prefer to managing firmware in another way, we cannot totally depend on the firmware path.

miRoox:
We need this to enable mtrace on MTL

mengdonglin (dismissed):

case-lib/lib.sh
mengdonglin: > @plbossart `strings` is searching the file content, not its path.

The patch looks reasonable. And thanks for updating the commit message @aiChaoSONG

case-lib/lib.sh
```diff
@@ -682,26 +682,10 @@ is_zephyr()
...
-        fi
-    else # for IPC3
-        platform=$(sof-dump-status.py -p)
-        firmware_name=sof-$platform.ri
-        firmware_path=$(sof-dump-status.py -P)
-    fi
+    local firmware_path znum
 
-    znum=$(strings "/lib/firmware/$firmware_path/$firmware_name" | grep -c -i zephyr)
+    firmware_path=$(journalctl_cmd | grep "request_firmware" | awk -F ' ' '{print $(NF-1)}')
```
marc-hb, line 687: Not a problem for CI but we're not sure what a developer may have done with their system, so maybe take the last one:

```suggestion
    firmware_path=$(journalctl_cmd -k | grep "request_firmware" | tail -n 1 | awk -F ' ' '{print $(NF-1)}')
```

Extra `-k` to be on safe side (hopefully journalctl accepts multiple `-k`)

case-lib/lib.sh
kv2019i: This does make a dependency to dev_dbg print that is not on by default. Sof-test won't work without this.

kv2019i: @aiChaoSONG It would be good to fail with a clear error if the "request_firmware" line is not found (e.g. dynamic debug not enabled in kernel and/or the line from SOF loader.c is disabled). Now the test will silently conclude it is not Zephyr and do the wrong thing.

kv2019i (commented):
I think the patch looks good. The new dependency to
```
»       »       dev_dbg(sdev->dev, "request_firmware %s successful\n",
»       »       »       fw_filename);
```

... in SOF kernel driver should be more explicit though.

case-lib/lib.sh
```diff
@@ -682,26 +682,11 @@ is_zephyr()
...
-    else # for IPC3
-        platform=$(sof-dump-status.py -p)
-        firmware_name=sof-$platform.ri
-        firmware_path=$(sof-dump-status.py -P)
-    fi
+    local firmware_path znum
 
-    znum=$(strings "/lib/firmware/$firmware_path/$firmware_name" | grep -c -i zephyr)
+    firmware_path=$(journalctl_cmd | grep "request_firmware" | tail -n 1 |
+                    awk -F ' ' '{print $(NF-1)}')
```
libinyang, line 688: If you check the kernel source code, you will find there are some drivers which will dump "request_firmware" string. For example, ./drivers/net/wireless/intel/ipw2x00/ipw2200.c or ./drivers/net/wireless/realtek/rtl8xxxu/rtl8xxxu_core.c. This means there is potential conflict if we use the dmesg to get the firmware path. 

This should work on our CI machines currently. But customers outside intel may find it doesn't work. Or our later platform using different hw configuration will have issues to use dmesg.

case-lib/lib.sh
marc-hb: > Extra -k to be on safe side (hopefully journalctl accepts multiple -k)

`journalctl` does accept multiple `-k` so please add one here to be on the safe side. This function is in the "critical path" so we must make sure it never becomes slow no matter what `journalctl_cmd` becomes in the future.

case-lib/lib.sh
marc-hb: > It would be good to fail with a clear error if the "request_firmware" line is not found

Indeed.

If the user expects logging and logging fails then the test should fail. Corner cases that want to test without logging must use something like `SOF_LOGGING=none`


`[ -n "$firmware_path" ] || die 'no firmware loaded'`

marc-hb (changes requested):
Please create a new `get_firmware_path()` function that can be re-used elsewhere if needed (and easily changed to `debugfs` if/when  thesofproject/linux#3867 happens)

Do not drop the URL to https://github.com/thesofproject/linux/issues/3867, it has _not_ been closed as wontfix.

The commit subject is too vague / misleading. How about: `stop using PCI IDs to guess the firmware filename`, always grep kernel logs instead.


`awk` has a built-in `grep` feature, use it. In fact `awk` can implement `tail` too, see example in `tools/sof-get-default-tplg.sh` where we already do something very similar.


This should do `grep`, `tail`, fix the conflict spotted by @libinyang and everything else with a single `awk` command but please test it again
```shell
journalctl -k |
  awk '/sof.*request_firmware/ { sub(/^.*request_firmware/,""); last_loaded_file=$1 } END { print last_loaded_file }'
```

marc-hb:
> The commit subject is too vague / misleading. How about: stop using PCI IDs to guess the firmware filename, always grep kernel logs instead.

This patch stops using PCI IDs while PR #976 adds a new PCI ID at the same time. What is going on?

BTW have you looked for other places where `sof-dump-status.py -p / -P` are used? Don't they need the same fix?

mengdonglin:
> > The commit subject is too vague / misleading. How about: stop using PCI IDs to guess the firmware filename, always grep kernel logs instead.> 
> This patch stops using PCI IDs while PR #976 adds a new PCI ID at the same time. What is going on?

@marc-hb I think sof-dump-status.py is for dumping platform info to debug FW bugs. So PR #976 seems not relevant to this. We don't really care what SoC is used. 

> 
> BTW have you looked for other places where `sof-dump-status.py -p / -P` are used? Don't they need the same fix?

kv2019i (dismissed):
Thanks @aiChaoSONG , looks good to me!

mengdonglin (dismissed):

marc-hb (dismissed):
Can someone check the internal test results for BOTH stable-2.2 and main before merging this?

> @marc-hb I think sof-dump-status.py is for dumping platform info to debug FW bugs. So PR https://github.com/thesofproject/sof-test/pull/976 seems not relevant to this. We don't really care what SoC is used.

`sof-dump-status.py` is still used in many sof-test places for many things including finding the firmware filename, either directly or indirectly.

marc-hb (commented):
You dropped this
```
# FIXME: the kernel driver should give us the FW path
# https://github.com/thesofproject/linux/issues/3867
```

keqiaozhang (approved):

case-lib/lib.sh
```diff
@@ -680,28 +680,23 @@ is_zephyr()
...
-    else # for IPC3
-        platform=$(sof-dump-status.py -p)
-        firmware_name=sof-$platform.ri
-        firmware_path=$(sof-dump-status.py -P)
-    fi
+    local firmware_path znum
+
+    firmware_path=$(get_firmware_path)
+    [ -n "$firmware_path" ] ||
+        die 'firmware path not found from journalctl, no firmware loaded or debug option disabled?'
```
miRoox, line 697: die here will lose "Test Result: FAIL!" and make our CI treat it as TIMEOUT.

marc-hb: Thanks @miRoox! The `func_exit_handler()` should print "Test Result: something", did you find why it does not?

marc-hb: To reproduce simply change the `grep` value to make it fail.

miRoox: > The `func_exit_handler()` should print "Test Result: something", did you find why it does not?

As you said in #977, `func_exit_handler` will not be called again when exit inside it.

marc-hb: > As you said in https://github.com/thesofproject/sof-test/pull/977, func_exit_handler will not be called again when exit inside it.

Answering in #979 , let's move the discussion there.

Issue #969, opened 2022-10-14
[BUG] tplgtool2 doesn't complain about non-existing file


juimonen:
**Describe the bug**
You can do ./tplgtool2.py this_file_is_not_there.tplg without the tool complaining

**To Reproduce**
See above

**Expected behavior**
Error print with file not found

miRoox:
Fixed by #974

PR #974, opened 2022-10-25
tplgtool2.py: support filesystem globbing and multiple arguments

marc-hb:
Warning: this is a backwards-incompatible change! The previous behavior is now available ONLY when using `--tplgroot`.

Up to now the user interface made the unusual choice to implement globbing internally instead of relying on the operating system. Maybe for compatibility with ancient .BAT scripts? So before this commit, internal globbing had to be quoted like this to escape the shell globbing, otherwise it would reject too many arguments ("unrecognized extra argument")
```
   ./tplgtool2.py 'sof-*.tplg'             # old, now INVALID usage
```
This commit adds support for multiple arguments and such quoting is not necessary anymore. HOWEVER, for backwards-compatibility the old behavior is preserved when using `--tplgroot`:
```
   ./tplgtool2.py --tplgroot . 'sof-*.tplg' # UNCHANGED behavior
```
See new --help message for more details.

Also add a number of `assert` to fix usability issue #969 and error handling when relying on the older behavior.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -1160,11 +1160,13 @@ if __name__ == "__main__":
             description='A Topology Reader totally Written in Python.')
 
         parser.add_argument('--version', action='version', version='%(prog)s 1.0')
-        parser.add_argument('-t', '--tplgroot', type=str, help="load tplg file from tplg_root folder")
+        parser.add_argument('-t', '--tplgroot', type=str, help="""Load topology files matching \
+a patterns argument from the tplg_root folder.
+Enables internal globbing mode: the single positional argument is not a file but a comma-separated list of patterns.""")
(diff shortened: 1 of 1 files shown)


marc-hb:
I checked all the graphs in both https://sof-ci.01.org/softestpr/PR974/build64/devicetest and https://sof-ci.01.org/softestpr/PR974/build63/devicetest and they're all there.

The suspend/resume failure and the NTP failure are obviously irrelevant to this PR.

miRoox (approved):
LGTM,

seems related to #969.

juimonen (approved):
yes lgtm, fixes my issue #969 thanks!

marc-hb:
Let's wait for the 01.org migration (and any other system-wide issue) to be resolved and then merge this.

aiChaoSONG (approved):

PR #976, opened 2022-10-28
sof-dump-status.py: add ADL-N PCI DID

kv2019i:
Add PCI DID for ADL-N.

### DIFF
--- a/tools/sof-dump-status.py
+++ b/tools/sof-dump-status.py
@@ -37,7 +37,8 @@ class clsSYSCardInfo():
             # pci-icl.c
             "0x34c8":"icl", "0x38c8":"jsl", "0x4dc8":"jsl",
             # pci-tgl.c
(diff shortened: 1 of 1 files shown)


marc-hb (commented):
ADL-N found in https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/pci-tgl.c as expected.

@kv2019i, @plbossart  does this mean there will be new `sof-adl-n.ri` binaries in `/lib/firmware/sof` different from both `sof-adl.ri` and `sof-adl-s.ri` and everything else? Different even for community or just a key difference?

As a coincidence here's a good, recent picture of symbolic links in /lib/firmware:

https://github.com/thesofproject/sof-bin/commit/df24de4984e5d5a3b731375f91ab92c7a8159730

aiChaoSONG (approved):

aiChaoSONG:
@marc-hb Check Pierre's patch in the kernel, adl-n uses a different signing key, that means it will have its own production signed firmware, but for community signed, I think we can make a symbol link to adl firmware.

marc-hb:
While CI is still having problems this is zero risk, merging.

PR #985, opened 2022-11-22
Tools: Add more robustness to check_volume_levels.m measure

singalsu:


tools/check_volume_levels.m
```diff
@@ -187,15 +187,28 @@ function plot_levels(meas, tc, lm)
...
+			% this adds more robustness to test for controls
+			% apply delay.
+			dg = diff(gains(idx,j));
+			if max(abs(dg)) > dg_tol
+				dg_rev = dg(end:-1:1);
+				idx_add = length(dg) - find(abs(dg_rev) > 0.1, 1, 'first') + 1;
+			else
+				idx_add = 0;
+			end
+			avg_gain = mean(gains(idx + idx_add, j));
```
singalsu, line 220: The idx_add may push the end of indices outside length of gains(), need to limit gain averaging to available samples amount.

singalsu:
@aiChaoSONG @marc-hb I still get these "[WARNING] SOF_TEST_TOP_PID=1531 already defined, multiple lib.sh inclusions?".  I see only one line `source "$TESTDIR/case-lib/lib.sh"` in the script. How can I fix it?

tools/check_volume_levels.m
singalsu: Fixed now

tools/check_volume_levels.m
```diff
@@ -187,14 +187,36 @@ function plot_levels(meas, tc, lm)
...
+			dg = diff(gains(idx,j));
+			if max(abs(dg)) > dg_tol
+				n_idx = length(idx);
+				dg_rev = dg(end:-1:1);
+				idx_add = length(dg) - find(abs(dg_rev) > 0.1, 1, 'first') + 1;
+			        idx = idx + idx_add;
+				if idx(end) > size(gains, 1)
+					idx = idx(1):size(gains,1);
+				end
+				if idx(1) > size(gains, 1) || length(idx) < 0.5 * n_idx;
```
ShriramShastry, line 222: Additional semicolon is unnecessary in IF statement idx(1) > size(gains, 1) || length(idx) < 0.5 * n_idx`;`

ShriramShastry (commented):
Looks good !!

ShriramShastry (dismissed):
I am fine with the changes. Thank you!!

btian1 (commented):
title:

settling --> settting

aiChaoSONG:
> @aiChaoSONG @marc-hb I still get these "[WARNING] SOF_TEST_TOP_PID=1531 already defined, multiple lib.sh inclusions?". I see only one line `source "$TESTDIR/case-lib/lib.sh"` in the script. How can I fix it?

sof-test will write PID to a lock file in case you run multiple test case at the same time. maybe you interrupt the test, and the lock file not deleted?  try manually delete the lock file /tmp/sof-test-cardX.lock, X is the sof card ID.

singalsu:
> not clear with below statement:
> 
> This patch delays the check window if the gain settling seems to be still happening in the beginning of level check window. The settling is detected with differential of gain.
> 
> Could you use short sentence to describe?

By window I mean an acceptance criteria over time and gain in decibels. E.g. x milliseconds to y milliseconds, measured gain within a +/- b decibels. Actually I'm not evaluating every point in window but compute average gain from that measure window and compare to required. Any fail will fail the entire test.

By computing the differential of measured gain, I see from the values if the gain is constant or changing. The added code is shifting the measurement window to later in time if the gain is not constant in in the beginning.

Did this explain better - should I add this to commit message?

marc-hb:
> > @aiChaoSONG @marc-hb I still get these "[WARNING] SOF_TEST_TOP_PID=1531 already defined, multiple lib.sh inclusions?". I see only one line source "$TESTDIR/case-lib/lib.sh" in the script. How can I fix it?

> sof-test will write PID to a lock file in case you run multiple test case at the same time. maybe you interrupt the test, and the lock file not deleted? try manually delete the lock file /tmp/sof-test-cardX.lock, X is the sof card ID.

No, that would be a different error message. `SOF_TEST_TOP_PID=1531 already defined` really means that  the variable `SOF_TEST_TOP_PID` is already defined which is a serious sof-test coding problem, just look at `git grep -C 5 SOF_TEST_TOP_PID` and see for yourself.

I don't see how this could be related to matlab changes though. @singalsu is that independent of this PR? Can you reproduce with the current main branch and if yes please share how exactly in a new, different sof-test issue.

andrula-song (dismissed):

singalsu:
> I don't see how this could be related to matlab changes though. @singalsu is that independent of this PR? Can you reproduce with the current main branch and if yes please share how exactly in a new, different sof-test issue.

The error is not printed if I remove from test end line `sof-kernel-log-check.sh "$KERNEL_CHECKPOINT"`. It's added in commit 8af0a798. Maybe something that it needs is missing.

marc-hb:
@singalsu if you have a deterministic reproduction without this PR please file a new sof-test bug with complete reproduction steps.

singalsu:
@btian1 I improved the commit text of first commit, no other changes, is it OK now?

All, can you please re-review. The previous accepts were dropped from force push.

singalsu:
The PR update adds 4th commit that avoids need for parallel SOF repo checkout.

tools/check_volume_levels.m
```diff
@@ -282,3 +279,26 @@ function plot_levels(meas, tc, lm)
...
+	end
+
+	x = zeros(n, 1);
+	x(:,1) = amp(1)*sin(2*pi*f(1)*t);
+	for i=2:length(f)
+		x(:,1) = x(:,1) + amp(i)*sin(2*pi*f(i)*t+ph(i))';
+	end
+end
+
+function dbfs = level_dbfs(x)
```
marc-hb, line 300: Can you add a one-line comment with a pointer to the copies in sof? "A little copying is better than a little dependency." but it's even better when people are aware :-)

btian1 (dismissed):

tools/check_volume_levels.m
singalsu: Yep, I added the the notes.

singalsu:
@btian1 Can you please check again. Your approve got cleared. There's only comment text add difference to previous version in the last commit.

btian1 (dismissed):

marc-hb:
@ShriramShastry can you please review again the last changes?

Other matlab reviewers would be appreciated.

tools/check_volume_levels.m
```diff
@@ -183,18 +188,41 @@ function plot_levels(meas, tc, lm)
...
+			% this adds more robustness to test for controls
+			% apply delay.
+			dg = diff(gains(idx,j));
+			if max(abs(dg)) > dg_tol
+				n_idx = length(idx);
+				dg_rev = dg(end:-1:1);
+				idx_add = length(dg) - find(abs(dg_rev) > 0.1, 1, 'first') + 1;
+			        idx = idx + idx_add;
+				if idx(end) > size(gains, 1)
+					idx = idx(1):size(gains,1);
```
ShriramShastry, line 217: could be a space between `size(gains,1)` i.e. `size(gains, 1)`

tools/check_volume_levels.m
```diff
@@ -183,18 +188,41 @@ function plot_levels(meas, tc, lm)
...
+				end
+				if idx(1) > size(gains, 1) || length(idx) < 0.5 * n_idx
+					fprintf(1, 'Channel %d controls impact is delayed too much ', j);
+					fprintf(1, 'from %4.1f - %4.1fs\n', ts, te);
+					pass = 0;
```
ShriramShastry, line 222: Variable `pass` provides any additional benefit of resilience?
I believe the design is strong and do not recommend changing the tracking pass status. This is only useful for debugging reasons.

ShriramShastry (approved):
The implementation appears to be sound. Thank you !!

marc-hb:
@singalsu said he will address @ShriramShastry suggestions; waiting before merge.

tools/check_volume_levels.m
singalsu: I kept this as-is. The fail is done if the control has drifted too much in time from where it should have happened.

tools/check_volume_levels.m
singalsu: This is fixed now

singalsu:
@marc-hb I just checked that this passed with SOF v2.2 on UP2. I had to make a small change to the shell wrapper since the loopback I2S device in sof-apl-nocodec.tplg is no more hw:0,0 but hw:0,2: 

```
diff --git a/test-case/check-volume-levels.sh b/test-case/check-volume-levels.sh
index 0695320..d819835 100755
--- a/test-case/check-volume-levels.sh
+++ b/test-case/check-volume-levels.sh
@@ -40,7 +40,7 @@ setup_kernel_check_point
 tplg=${OPT_VAL['t']}
 
 TPLGREADER="$TESTDIR"/tools/sof-tplgreader.py
-PCM_ID=0
+PCM_ID=2
 CAP_FORMAT="S16_LE"
 CAP_RATE="48000"
 VOL_P30DB=80

```

I'd like to make fixes incrementally and address the finding of loopback I2S a separate patch. Its not impacting the Matlab code of this PR.

Also need to make the test to work with a test subset if there is no more mute switch, since those have been removed from IPC4 topologies.

tools/check_volume_levels.m
```diff
@@ -251,3 +279,30 @@ function plot_levels(meas, tc, lm)
...
+	n = round(fs*tlength);
+	t = (0:n-1)/fs;
+	nf = length(f);
+	if nf > 1
+		ph = rand(nf, 1)*2*pi;
+	else
+		ph = 0;
+	end
+
+	x = zeros(n, 1);
```
ShriramShastry, line 295: This is n rows and 1 column, right ?

tools/check_volume_levels.m
```diff
@@ -251,3 +279,30 @@ function plot_levels(meas, tc, lm)
...
+	x(:,1) = amp(1)*sin(2*pi*f(1)*t);
+	for i=2:length(f)
+		x(:,1) = x(:,1) + amp(i)*sin(2*pi*f(i)*t+ph(i))';
```
ShriramShastry, line 298: I think x(:,1) is no longer needed , Just x is sufficient

ShriramShastry (approved):
A minor comment !! If it is fine with you.

tools/check_volume_levels.m
```diff
@@ -251,3 +279,30 @@ function plot_levels(meas, tc, lm)
...
+	nf = length(f);
+	if nf > 1
+		ph = rand(nf, 1)*2*pi;
+	else
+		ph = 0;
+	end
+
+	x = zeros(n, 1);
+	x(:,1) = amp(1)*sin(2*pi*f(1)*t);
+	for i=2:length(f)
```
ShriramShastry, line 297: Why is the loop initialized from 2 rather than 1?

singalsu: It's because the first frequency f(1) sine wave is mixed before the loop. Any additional frequencies f(2) to f(end) are mixed in the loop. But yes, it can be simplified to e.g. 

```
	x = zeros(n, 1);
	for i=1:length(f)
		x(:,1) = x(:,1) + amp(i)*sin(2*pi*f(i)*t+ph(i))';
	end

```

tools/check_volume_levels.m
singalsu: Yep,

singalsu:
@ShriramShastry I updated the tone generator function as you proposed. Can you please review again?

I just checked on UP2 sof-apl-nocodec.tplg that the updated code works and test passes so the change was OK.

Next PR for volume checker should be to add support for IPC4 topologies without switch control, and solve the discovery of looped back I2S PCMs, or get the PCM IDs externally.

tools/check_volume_levels.m
singalsu: It's now simplified to just `x` . I was apparently worried about vector orientations when doing this and did not clean up afterwards. The mixing of sine preserved  the vector and did not make it a 2D matrix (a pitfall with Matlab language with column and row vectors).

tools/check_volume_levels.m
```diff
@@ -183,18 +188,41 @@ function plot_levels(meas, tc, lm)
...
 			idx = find(meas.t(idx0) > ts);
+
+			% Delay if settled gain is later in the window,
+			% this adds more robustness to test for controls
+			% apply delay.
+			dg = diff(gains(idx,j));
+			if max(abs(dg)) > dg_tol
+				n_idx = length(idx);
+				dg_rev = dg(end:-1:1);
+				idx_add = length(dg) - find(abs(dg_rev) > 0.1, 1, 'first') + 1;
```
ShriramShastry, line 214: Can 0.1 be superseded by dg tol?

tools/check_volume_levels.m
```diff
@@ -183,18 +188,41 @@ function plot_levels(meas, tc, lm)
...
 end
 
 function pass = check_levels(meas, tc, lm, verbose)
 	pass = 1;
+	dg_tol = 0.1;
 	gains =  meas.levels - lm.sine_dbfs;
 	sv = size(tc.volumes);
 	for j = 1:sv(2)
 		for i = 1:sv(1)
+			% Initial location to test
```
ShriramShastry, line 201: could you please add a space between `ts = tc.vctimes(i)+tc.meas(1); `as `ts = tc.vctimes(i) + tc.meas(1);`

tools/check_volume_levels.m
```diff
@@ -251,3 +279,29 @@ function plot_levels(meas, tc, lm)
 		y(:,j) = filter(b, a, x(:,j));
 	end
 end
+
+% This function is copy of
+% sof/tools/test/audio/test_utils/multitone.m
+function x = multitone( fs, f, amp, tlength )
+	n = round(fs*tlength);
+	t = (0:n-1)/fs;
```
ShriramShastry, line 287: can you please add space for the operation t = (0:n-1)/fs; like `t = (0:n-1) / fs;`

tools/check_volume_levels.m
```diff
@@ -251,3 +279,29 @@ function plot_levels(meas, tc, lm)
...
+	nf = length(f);
+	if nf > 1
+		ph = rand(nf, 1)*2*pi;
```
ShriramShastry, line 290: I think we can do a space for rand(nf, 1)*2*pi; like `rand(nf, 1) * 2 * pi;`

tools/check_volume_levels.m
```diff
@@ -251,3 +279,29 @@ function plot_levels(meas, tc, lm)
...
+	else
+		ph = 0;
+	end
+
+	x = zeros(n, 1);
+	for i=1:length(f)
+		x = x + amp(i)*sin(2*pi*f(i)*t+ph(i))';
```
ShriramShastry, line 297: can you please a space for the equation x = x + amp(i) * sin(2 * pi *f(i) * t + ph(i))';

ShriramShastry (commented):
A few sanity-related comments have been added.

Kindly include space where it is needed.  The implementation looks to have improved significantly.

Thank you

marc-hb:
> Kindly include space where it is needed.

There seems to be a couple matlab beautifier available, @ShriramShastry can you please recommend one? If no formatter is good then at least some codestyle document reference. Consistency is good but it needs to be well defined, we cannot have codestyle discussions repeats in every pull request.

Also: please do NOT reformat every matlab file and destroy git history. We need a reference for new and modified code, not for old, unchanged code. Too late.

tools/check_volume_levels.m
singalsu: Yep, it can be the same

singalsu:
@ShriramShastry The last commit contains the style changes for non-impacted code. Sine generator style changes are in the commit that adds it.

marc-hb:
> There is no Matlab style guide 

https://www.google.com/search?q=matlab+beautify
https://www.google.com/search?q=matlab+formatter

marc-hb:
@ShriramShastry all good now?
If there are any whitespace / style issues they can addressed later. This PR has been under review for a few months now...

marc-hb:
I learned that @ShriramShastry has been out of office. I looked at the changes since he last approved: https://github.com/thesofproject/sof-test/compare/d79bf2af7aeae751ff728c2585d37677f0663424..5eec1168b4ad26f19a744d69a00ea0fbe620eacf

I found only whitespace changes + this `for` loop change that he requested earlier https://github.com/thesofproject/sof-test/compare/d79bf2af7aeae751ff728c2585d37677f0663424..1218c95eaa5de33a300073367d823089c9f0b1e5

Merging.

marc-hb (approved):

marc-hb:
> > @singalsu if you have a deterministic reproduction without this PR please file a new sof-test bug with complete reproduction steps.

> I will but it requires merge of this PR. The script does not proceed to last line where sof-kernel-log-check.sh is called if there is a test fail. Or in other words the Matlab part of test always fails without this improvement. Somehow things (controls) happen in DUT slower than before and failure rate is 100%. Also we don't have tests for control latencies trend, as an idea for future.

@singalsu can you still reproduce?

marc-hb:
`SOF_TEST_TOP_PID=17709 already defined` just spotted in totally unrelated:

- https://github.com/thesofproject/sof-test/pull/1025#issuecomment-1512197485

PR #990, opened 2022-12-07
check-alsabat.sh: use 821Hz as default test frequency

kv2019i:
Lower the default test frequency from 997Hz to 821Hz. Most SOF topologies use 1000Hz tick rate, and test frequency close to this scheduling frequency can mask certain type of errors (like SOF bug

Link: https://github.com/thesofproject/sof/issues/6709
Suggested-by: Seppo Ingalsuo <seppo.ingalsuo@linux.intel.com>

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -34,7 +34,7 @@ OPT_NAME['c']='pcm_c'      	OPT_DESC['c']='pcm for capture. Example: hw:1,0'
 OPT_HAS_ARG['c']=1             OPT_VAL['c']=''
 
 OPT_NAME['f']='frequency'       OPT_DESC['f']='target frequency'
-OPT_HAS_ARG['f']=1             OPT_VAL['f']=997
+OPT_HAS_ARG['f']=1             OPT_VAL['f']=821
 
 OPT_NAME['n']='frames'          OPT_DESC['n']='test frames'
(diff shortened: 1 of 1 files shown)


kv2019i:
Tested with Intel custom test plan 18644 showing failure (not detected by 997Hz test).

lgirdwood (approved):

marc-hb:
@singalsu, @kv2019i , @lgirdwood should we run this test multiple times with a different frequency each time? If yes please provide a list.

singalsu:
> @singalsu, @kv2019i , @lgirdwood should we run this test multiple times with a different frequency each time? If yes please provide a list.

Yes, 599, 821, 997 Hz. But need to add also -k parameter, see #991.

997 Hz because it's a de-facto test frequency in industry. 599 Hz is another prime that I came up, could be some other prime as well.

keqiaozhang:
> Would someone have time to make a patch and include the -k stuff too?

@singalsu , I have added 3 alsabat cases to cover 599, 821 and 997 in CI web, we can pass the frequency to check-alsabat.sh with -F option, no need to make a patch.

Test reports:
IPC4:https://sof-ci.sh.intel.com/#/result/planresultdetail/19411
MTL:https://sof-ci.sh.intel.com/#/result/planresultdetail/19409

marc-hb:
@keqiaozhang #991 is not about passing the frequency, it adds a new sigma parameter.

PR #991, opened 2022-12-08
test-case: check-alsabat.sh: Lower the threshold for detecting tones

singalsu:
The default parameter sigma k is 3.0 when omitted from command line. This patch drops the threshold to half with 1.5 to better detect distortion in playback. Alsabat judges spectral components above the threshold as sine frequencies.

This change should increase the probability fail a test to avoid green failures to happen.

Signed-off-by: Seppo Ingalsuo <seppo.ingalsuo@linux.intel.com>

PR #992, opened 2022-12-09
alsa_settings: add alsa settings for IPC4 platforms

keqiaozhang:

### DIFF
--- a/alsa_settings/CAVS_NOCODEC_IPC4.sh
+++ b/alsa_settings/CAVS_NOCODEC_IPC4.sh
@@ -1,0 +1,15 @@
+set -e
+
+# SSP playback
+amixer -c sofnocodec cset name='gain.1.1 Playback Volume 1' 45
+amixer -c sofnocodec cset name='gain.15.1 Deepbuffer Volume' 45
+amixer -c sofnocodec cset name='gain.2.1 Main Playback Volume 2' 45
+amixer -c sofnocodec cset name='gain.3.1 Playback Volume 3' 45
+amixer -c sofnocodec cset name='gain.4.1 Main Playback Volume 4' 45
+amixer -c sofnocodec cset name='gain.5.1 Playback Volume 5' 45
+amixer -c sofnocodec cset name='gain.6.1 Main Playback Volume 6' 45
+
+# SSP capture
+amixer -c sofnocodec cset name='gain.17.1 Main Capture Volume 2' 45
+amixer -c sofnocodec cset name='gain.7.1 Main Capture Volume 1' 45
+amixer -c sofnocodec cset name='gain.8.1 Host Capture Volume' 45
--- a/alsa_settings/ADLP_RVP_SDW_IPC4ZPH.sh
+++ b/alsa_settings/ADLP_RVP_SDW_IPC4ZPH.sh
@@ -1,0 +1,8 @@
+set -e
+
+# enable playback
+amixer -c sofsoundwire cset name='Headphone Switch' on
+amixer -c sofsoundwire cset name='gain.0.1 1 Playback Volume 0' 45
+amixer -c sofsoundwire cset name='gain.15.1 Deepbuffer Volume' 45
+amixer -c sofsoundwire cset name='gain.1.1 2 Main Playback Volume' 45
+amixer -c sofsoundwire cset name='rt711 DAC Surr Playback Volume' 80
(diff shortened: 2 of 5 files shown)


miRoox (dismissed):
Why don't we ignore the difference between IPC3/IPC4, like trimming the model name to exclude the 4th part?

keqiaozhang:
> Why don't we ignore the difference between IPC3/IPC4, like trimming the model name to exclude the 4th part?

Because the tplgs are different, so the volume control names are different between IPC3 and IPC4. 
IPC3 volume control names are: `PGA7.0 7 Master ....`
IPC4 volume control names are: `gain. ..`

plbossart:
@ranj063 I think most of the settings will have to be changed with your topology cleanups, no?

keqiaozhang:
@ranj063 , please help to confirm if we need to change the amixer setting after your topology cleanups?

ranj063:
> @ranj063 , please help to confirm if we need to change the amixer setting after your topology cleanups?

@keqiaozhang @plbossart , Im going to punt this to @jsarha. He's working on cleaning up the mixer names.

keqiaozhang:
I have verified this patch with [PR991](https://github.com/thesofproject/sof-test/pull/991), the results are as expected:https://sof-ci.sh.intel.com/#/result/planresultdetail/19443

image

keqiaozhang:
@aiChaoSONG please help to review this PR, since we have enabled alsabat test on all our IPC4 platforms, so we need this PR to set the amixer settings before the test. This PR is well tested.

alsa_settings/CAVS_NOCODEC_IPC4.sh
```diff
@@ -0,0 +1,15 @@
+set -e
+
+# SSP playback
+amixer -c sofnocodec cset name='gain.1.1 Playback Volume 1' 45
+amixer -c sofnocodec cset name='gain.15.1 Deepbuffer Volume' 45
+amixer -c sofnocodec cset name='gain.2.1 Main Playback Volume 2' 45
+amixer -c sofnocodec cset name='gain.3.1 Playback Volume 3' 45
+amixer -c sofnocodec cset name='gain.4.1 Main Playback Volume 4' 45
+amixer -c sofnocodec cset name='gain.5.1 Playback Volume 5' 45
+amixer -c sofnocodec cset name='gain.6.1 Main Playback Volume 6' 45
```
plbossart, line 10: Can you remind me why we should care about the volume settings in the firmware? This should be 0dB by default, always. the way this is implemented will break for every topology change or rename of controls.

plbossart (changes requested):

alsa_settings/CAVS_NOCODEC_IPC4.sh
keqiaozhang: Yes, the default volume should be 0dB, but we cannot ensure that these volumes will not be modified before testing alsabat.

aiChaoSONG (approved):

alsa_settings/CAVS_NOCODEC_IPC4.sh
marc-hb: @keqiaozhang you mean because other tests may change them? Why?

I've seen a large number of discussions about ALSA settings and UCM for a few years now and I still don't know how the "ideal state" should look like. Exceptions and quirks are sometimes necessary but I still can't tell what is a quirk versus what is a Good Thing. Is this already documented somewhere or would a new https://github.com/thesofproject/sof-test/discussions help?

plbossart: > Yes, the default volume should be 0dB, but we cannot ensure that these volumes will not be modified before testing alsabat.

then use a script to find all gain/PGA for the firmware using regexps and reset the value separately from codec stuff.

keqiaozhang: > then use a script to find all gain/PGA for the firmware using regexps and reset the value separately from codec stuff

Can we merge this PR first? I will try your suggestion next. It's not only related to volume, but also related to switch.

marc-hb: > then use a script to find all gain/PGA for the firmware using regexps and reset the value separately from codec stuff.

@keqiaozhang is this what you're trying to do in #1010?

keqiaozhang: > @keqiaozhang is this what you're trying to do in #1010?

Yes, I'm working on it.

PR #996, opened 2023-01-04
hijack: hide the kernel initialization log on MTL

keqiaozhang:
It is not allowed to publish MTL full log externally. This PR will delete the irrelevant kernel initalization log after the test.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -7,6 +7,7 @@ trap 'func_exit_handler $?' EXIT
 function func_exit_handler()
 {
     local exit_status=${1:-0}
+    local keyword="Started Journal Service"
 
     # call trace
     if [ "$exit_status" -ne 0 ] ; then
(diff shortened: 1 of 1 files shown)


marc-hb (changes requested):
Booting is concurrent and racy so this is unreliable by design.

When a test fails it extracts some information from kernel logs which is another loophole.

We cannot have 100% certainty that other log files won't leak anything.

Also, let's not pollute the sof-test git history with temporary, Intel-only details. It must be possible to run any sof-test version against any sof version at any point in time.

Last but not least, anyone could submit a revert of this commit (or worse) and expose logs. So this is not secure at all!

Let's keep things simple and safe and keep ALL MTL logs private until it's not required anymore. Share only the red/green status in the mean time. In the very rare and unlikely case someone not from Intel suspects they broke MTL, they can ask for more information in the pull request.

marc-hb:
PS: Don't abuse `&&` that much. `&&` is NOT equivalent to if/then/else!  Here some differences below. Copy the following lines to your shell prompt and press "enter". They behave exactly the same in bash, dash, zsh and any other standard compliant shell.
```
if true; then false; else echo REAL_ELSE;     fi    # does not print anything
   true  &&   false   ||  echo NOT_LIKE_ELSE        # prints NOT_LIKE_ELSE !

( set -e ;   false && true;   echo HELLO )    # Prints HELLO
( set -e ; ( false && true) ; echo HELLO )    # Does NOT print HELLO!
```

Use '&&' only if you understand these examples really well. If you are not sure then stick to if/then/else. You can't go wrong with if/then/else.

PR #1002, opened 2023-02-09
verify-kernel-boot-log.sh: don't SKIP when i915 is missing

marc-hb:
`grep something missing_file` returns 2. When the file is missing this is misinterpreted as SKIP as seen in daily test run 20752. This happens because the grep is the last command in the subshell.

The `grep /sys/kernel/debug/dri/0/i915_display_info` and the including subshell are not part of the test; ignore their exit code and let the rest of the test code decide what the exit value of the test is.

Fixes commit aadd7707cb261

### DIFF
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -66,6 +66,7 @@ wait_is_system_running()
         DISPLAY=:0 xrandr --listmonitors
         DISPLAY=:1024 xrandr --listmonitors
         sudo grep -i connected /sys/kernel/debug/dri/0/i915_display_info
+        true
     )
     die "Some services are not running correctly"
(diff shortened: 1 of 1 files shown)


marc-hb:
One suspend/resume TIMEOUT in https://sof-ci.01.org/softestpr/PR1002/build142/devicetest/index.html, everything else green

fredoh9 (approved):
wow, good catch/fix!

greg-intel (approved):

marc-hb:
Thanks for the reviews; merging before daily test runs

marc-hb:
More context:
- https://github.com/thesofproject/sof-test/discussions/964
- https://github.com/thesofproject/sof-test/discussions/998

PR #1009, opened 2023-03-10
check_alsabat.sh: add format support

keqiaozhang:
add a -f option to support audio format.

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -33,8 +33,8 @@ OPT_HAS_ARG['r']=1             OPT_VAL['r']=48000
 OPT_NAME['c']='pcm_c'      	OPT_DESC['c']='pcm for capture. Example: hw:1,0'
 OPT_HAS_ARG['c']=1             OPT_VAL['c']=''
 
-OPT_NAME['f']='frequency'       OPT_DESC['f']='target frequency'
-OPT_HAS_ARG['f']=1             OPT_VAL['f']=821
+OPT_NAME['F']='frequency'       OPT_DESC['F']='target frequency'
+OPT_HAS_ARG['F']=1             OPT_VAL['F']=821
 
 OPT_NAME['k']='sigmak'		OPT_DESC['k']='sigma k value'
 OPT_HAS_ARG['k']=1             OPT_VAL['k']=1.5
@@ -52,7 +52,7 @@ pcm_p=${OPT_VAL['p']}
 pcm_c=${OPT_VAL['c']}
 rate=${OPT_VAL['r']}
 channel_c=${OPT_VAL['C']}
-frequency=${OPT_VAL['f']}
+frequency=${OPT_VAL['F']}
 sigmak=${OPT_VAL['k']}
 frames=${OPT_VAL['n']}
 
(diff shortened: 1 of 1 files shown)


dbaluta (dismissed):

test-case/check-alsabat.sh
```diff
@@ -25,22 +25,25 @@ OPT_NAME['p']='pcm_p'     	OPT_DESC['p']='pcm for playback. Example: hw:0,0'
 OPT_HAS_ARG['p']=1          	OPT_VAL['p']=''
 
 OPT_NAME['C']='channel_c'       OPT_DESC['C']='channel number for capture.'
-OPT_HAS_ARG['C']=1             OPT_VAL['C']='1'
+OPT_HAS_ARG['C']=1              OPT_VAL['C']='1'
 
 OPT_NAME['r']='rate'            OPT_DESC['r']='sample rate'
-OPT_HAS_ARG['r']=1             OPT_VAL['r']=48000
+OPT_HAS_ARG['r']=1              OPT_VAL['r']=48000
```
marc-hb, line 31: This is pure git noise, don't.
https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

marc-hb (changes requested):
This PR is hiding a backwards-incompatible change. I understand you did that to align with `alsabat` which is good but it will break existing users. This should at the very least be explained in a separate commit message, clearly visible in the git log.

- commit 1: rename -f[requency] option to -F for alsabat consistency
- commit 2: add new -f[format] option

With the NEW code, what happens when someone does `check-alsabat.sh -f 821`? Do they get a meaningful error message from alsabat? Add that error message in the first commit message.

test-case/check-alsabat.sh
keqiaozhang: Updated.

keqiaozhang:
Please help to review. If no objections, I will merge this PR myself, because I also need to update the parameters(`-f => -F`) in web side for alsabat related cases.

image

test-case/check-alsabat.sh
```diff
@@ -33,8 +33,11 @@ OPT_HAS_ARG['r']=1             OPT_VAL['r']=48000
 OPT_NAME['c']='pcm_c'      	OPT_DESC['c']='pcm for capture. Example: hw:1,0'
 OPT_HAS_ARG['c']=1             OPT_VAL['c']=''
 
-OPT_NAME['f']='frequency'       OPT_DESC['f']='target frequency'
-OPT_HAS_ARG['f']=1             OPT_VAL['f']=821
+OPT_NAME['f']='format'       OPT_DESC['f']='target format'
+OPT_HAS_ARG['f']=1             OPT_VAL['f']="S16_LE"
+
+OPT_NAME['F']='frequency'       OPT_DESC['F']='target frequency'
+OPT_HAS_ARG['F']=1             OPT_VAL['F']=821
```
miRoox, line 40: Will the change of `-f` break the backward compatibility? For example, daily MTL test runs for different frequencies with `-f`.

keqiaozhang: Yes, so we need to update the alsabat test config in web accordingly when merging this PR.

marc-hb (approved):

marc-hb:
Very small conflict with #991 that I just merged, need to rebase sorry.

keqiaozhang:
@marc-hb , rebased. Please give a check again. if no problems, I will merge it today.

marc-hb (approved):
Please make sure the test results are OK-ish.

PR #1011, opened 2023-03-14
Test-case: Relax sigma-k parameter in check-alsabat.sh

singalsu:
The value 1.5 has caused a false failure possibly due to noise in the test device output. Value 2.1 gives a small relaxation and passes the previously failed capture.

Fixes 92fe8a1c7cf465dc3820f538f09739754c65e08e

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -37,7 +37,7 @@ OPT_NAME['f']='frequency'       OPT_DESC['f']='target frequency'
 OPT_HAS_ARG['f']=1             OPT_VAL['f']=821
 
 OPT_NAME['k']='sigmak'		OPT_DESC['k']='sigma k value'
-OPT_HAS_ARG['k']=1             OPT_VAL['k']=1.5
+OPT_HAS_ARG['k']=1             OPT_VAL['k']=2.1
 
 OPT_NAME['n']='frames'          OPT_DESC['n']='test frames'
 OPT_HAS_ARG['n']=1             OPT_VAL['n']=240000


marc-hb:
Still one alsabat failure in https://sof-ci.01.org/softestpr/PR1011/build189/devicetest/index.html?model=CML_HEL_RT5682&testcase=check-alsabat-headset-playback


https://sof-ci.01.org/softestpr/PR1011/build190/devicetest/index.html is all green.

singalsu:
@marc-hb The quality of that sine wave is really bad, it should fail. I don't seem to be able to attach a screenshot to show, but it's bumpy looking due to excessive noise. I don't think they are FW bug caused glitches.

The measurement should be fixed (used some USB mono headset as sound card?) or exclude the alsabat test from that device. Or use just for this device "check-alsabat.sh -k 3.5" to relax a lot the test (maybe beyond being usable at all).

keqiaozhang:
> used some USB mono headset as sound card?
Issue fixed after replacing the audio splitter cable with a new one.

keqiaozhang (approved):
LGTM

marc-hb:
https://sof-ci.01.org/softestpr/PR1011/build189/devicetest/index.html?model=CML_HEL_RT5682&testcase=check-alsabat-headset-playback still failing. The rest is suspend/resume or PASS

singalsu:
> https://sof-ci.01.org/softestpr/PR1011/build189/devicetest/index.html?model=CML_HEL_RT5682&testcase=check-alsabat-headset-playback still failing. Same device `sh-cml-hel-rt5682-05`
> 
> The rest is suspend/resume or PASS

There is no glitch but audio  quality with this device and USB sound card (the latter is to blame) is so bad that it needs to fail.
 
Screenshot from 2023-03-16 10-55-08

You can't tell from waveform like this if it's a glitch or noise. The used headset adapter "sound card" does not fit a professional system.

Screenshot from 2023-03-16 10-57-44

singalsu:
Still not merged, the fail with https://sof-ci.01.org/sofpr/PR7291/build4734/devicetest/index.html would pass with this proposed value 2.1.

PR #1014, opened 2023-03-23
tplgtool2 add more information to the nodes in the picture

jsarha:
It is sometimes hard to draw connection between the nodes in the produced picture and user visible entities, like ALSA mixers and PCMs. The old behavior is behind '-w' or '--without_nodeinfo' flags.
    
The copier nodes are added with, copier type: AIF_IN, AIF_OUT, DAI_IN or DAI_OUT, and with copier widgets stream name.

I would have wanted to add the PCM names there in the AIF nodes, but unfortunately that info was not in the widgets of the TplgGraph, so I would have to bring it in from the GroupedTplg and somehow find the associated widgets for the PCMs.  I will look into that, but I think this already a good improvement. We could of course also just take care that the streams connected to PCMs would have names that are easily associated with the PCM names.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -818,6 +818,7 @@ class TplgGraph:
     def __init__(self, grouped_tplg: GroupedTplg):
         "Build graph from grouped topology data."
         self.show_core = 'auto'
+        self.without_nodeinfo = False
         self._tplg = grouped_tplg
         self._nodes_dict = TplgGraph._build_nodes_dict(grouped_tplg.widget_list)
         self._nodes_names_in_graph = TplgGraph._build_nodes_names_in_graph(grouped_tplg.graph_list, self._nodes_dict)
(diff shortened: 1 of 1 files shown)


marc-hb:
Can you reproduce the pylint "syntax error" in https://github.com/thesofproject/sof-test/actions/runs/4504287467/jobs/7928575215?pr=1014 ?

tools/tplgtool2.py
```diff
@@ -1214,8 +1229,8 @@ def get_tplg_paths(tplgroot: typing.Union[str, None], patterns: str) -> "list[Pa
         if not rootpath.is_dir():
             raise NotADirectoryError(rootpath)
         return [tplgpath
-            for pattern in arg_split(patterns) 
-            for tplgpath in do_glob(rootpath, pattern) 
+            for pattern in arg_split(patterns)
+            for tplgpath in do_glob(rootpath, pattern)
```
marc-hb, line 1233: Unless this fixes a warning please don't

jsarha:
Here is how our sof-tgl-nocodec topology2 picture looks after the change:
sof-tgl-nocodec

jsarha:
> Can you reproduce the pylint "syntax error" in https://github.com/thesofproject/sof-test/actions/runs/4504287467/jobs/7928575215?pr=1014 ?

The ubuntu 22 default pylint does not give such an error.

marc-hb:
> The ubuntu 22 default pylint does not give such an error.

Let's not waste time with a very old pylint version, upgrade submitted please review:
- #1015

paulstelian97 (dismissed):

tools/tplgtool2.py
```diff
@@ -840,13 +841,26 @@ def node_name_in_graph(self, node: Container) -> str:
         """
         return self._node_name_in_graph_from_name(node["widget"]["name"])
 
+    def node_sublabel(self, widget: Container) -> str:
+        match widget["widget"]["id"]:
```
miRoox, line 845: match-case is the new syntax in Python 3.10. GitHub doesn't highlight it.

tools/tplgtool2.py
```diff
@@ -840,13 +841,26 @@ def node_name_in_graph(self, node: Container) -> str:
...
+            case 'PGA':
+                sublabel = f'<BR ALIGN="CENTER"/><SUB>PGA: {widget["kcontrols"][0]["hdr"]["name"]}</SUB>'
```
miRoox, line 847: Is `widget["kcontrols"]` never empty?

miRoox:
SOFCI TEST

tools/tplgtool2.py
aiChaoSONG: it should be never empty for gain (ipc4) or PGA (ipc3) widget, because there should be (and only be) a mixer control for the widget, or the topology is just wrong. but alsatplg won't help to detect this I guess

miRoox (approved):

miRoox:
Maybe you can rebase the main branch (with https://github.com/thesofproject/sof-test/pull/1015) to resolve the pylint error.

aiChaoSONG (dismissed):

marc-hb:
> Maybe you can rebase the main branch (with https://github.com/thesofproject/sof-test/pull/1015) to resolve the pylint error.

You don't need to rebase, a empty `git commit --amend -C HEAD` + force-push is enough. Most CIs test the moving `pull/1014/merge`, NOT the static `pull/1014/head`.

Rebasing creates `git range-diff` noise https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#workflow-suggestions-that-help-reviewers

miRoox (approved):

PR #1016, opened 2023-03-27
Test-case: Add to check-alsabat.sh additional audio test

singalsu:
This patch adds run of check_wav_file_snr.m script for alsabat capture file to analyze with STFT spectra the audio quality. Dropped signal-to-noise ratio in a single FFT indicates presence of glitch. A low SNR result triggers run of additional glitch periodicity finder.


tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,355 @@
...
+
+if nargin < 1
+	error('Need to provide file name to analyze');
+end
+
+if exist('OCTAVE_VERSION', 'builtin')
+	pkg load signal;
+end
+
+%fp = '/home/singalsu/Downloads/';
```
singalsu, line 20: Commented code to be removed, please ignore.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,355 @@
...
+%wf = 'bat.wav.3Vbz0k'; % total mess
+%wf = 'bat.wav.zhJhMf'; % glitch in right channel
+%wf = 'bat.wav.Wodg37'; % low quality
+%wf = 'bat.wav.WR5of5'; % Low signal level but OK
+%wf = 'bat.wav.05IgGp'; % good
+%wf = 'bat.wav.3UiPoZ'; % good
+%fn = fullfile(fp, wf);
+
+param.t_ignore_from_start = 30e-3;
```
singalsu, line 29: Should be possible to enter the ignorefrom begin and end times from command line. And the test criteria should be printed to see them in log.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,355 @@
...
+	return
+end
+
+i1 = round(param.fs * param.t_ignore_from_start);
+i2 = length(x) - round(param.fs * param.t_ignore_from_end);
+[b, a] = butter(2, 0.95, 'high');
+y = abs(filter(b, a, x));
+z = y(i1:i2);
+thr = mean(z) + std(z);
+[~, locs]= findpeaks(z, 'MinPeakHeight', thr);
```
singalsu, line 134: this criteria with thr too easily claims some analog noise is periodic glitch. Should improve this to not unnecessarily report the finding. On the other hand finds quite well a periodic buffer issue.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,355 @@
...
+set(gca, 'ydir', 'normal');
+colorbar;
+grid on;
+title(tstr, 'interpreter', 'none');
+xlabel('Time (ms)');
+ylabel('Frequency (Hz)');
+
+end
+
+function fh = plot_levels(meas, t, idstr)
```
singalsu, line 320: Need to find a way to avoid these plots to appear as ASCII to log file. Just PNG file export to log dir is enough.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
+function pass = check_wav_file_snr(fn, test_tone_f, log_dir)
```
ShriramShastry, line 1: `pass` ,It appears to me that the method return value is never set.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+param.level_differential_tol = 1e-3;
+param.visible = 'off';
+param.do_plot = 1;
+param.do_print = 1;
+param.print_dir = log_dir;
+test_passed = true;
+
+% Read audio
+fprintf(1, 'Loading file %s\n', fn);
+[x, fs, frames, channels] = read_audio(fn);
```
ShriramShastry, line 35: The value given by the function call is not used by the code in `frames`.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+	idstr = sprintf('%s%s ch%d', basefn, extfn, ch);
+	plot_specgram(param, stft, f, t, idstr, fnstr);
+	plot_levels(param, meas, t, idstr, fnstr);
+
+	% Plot glitch waveform
+	if meas.num_glitches > 0
+		plot_glitch(param, x(:, ch), meas, idstr, fnstr);
+	end
+
+	all_meas(ch) = meas;
```
ShriramShastry, line 70: `all_meas` Please consider using zeros or ones for pre-allocation.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+y = abs(filter(b, a, x));
+z = y(i1:i2);
+thr = mean(z) + std(z);
+[~, locs]= findpeaks(z, 'MinPeakHeight', thr);
+dlocs = diff(locs);
+if isempty(dlocs)
+	return
+end
+
+[counts, centers] = hist(dlocs);
```
ShriramShastry, line 116: It is advised to change "`hist`" to "`histogram`."

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+	snr_db(i) = signal_db(i) - noise_db(i);
+end
+
+meas.noise_db = noise_db - param.win_gain;
+meas.signal_db = signal_db - param.win_gain;
+meas.snr_db = signal_db - noise_db;
+
+end
+
+function meas = check_tone_levels(stft, param, meas, t)
```
ShriramShastry, line 177: Input arguments `stft` and `t` might be unused or you are plans ?

ShriramShastry (commented):
Thank you !!
Initial review notes were added. I'll revisit the method and the detection procedures.

singalsu:
Example results from previous run
https://sof-ci.01.org/softestpr/PR1016/build232/devicetest/index.html
https://sof-ci.01.org/softestpr/PR1016/build234/devicetest/index.html

singalsu:
This version creates more compact graphics:

Screenshot from 2023-03-28 19-29-05

test-case/check-alsabat.sh
```diff
@@ -95,6 +96,22 @@ function __upload_wav_file
     done
 }
 
+function check_wav_file_snr
+{
+    for file in /tmp/bat.wav.*
+    do
+	if test -s "$file"; then
+	    dlogi "Checking wav file $file"
+	    cd "$TESTDIR"/tools || {
```
marc-hb, line 105: This seems overkill, `set -e` should be enough here. Why would this fail?

ShriramShastry (commented):
I've added a few more suggestions and shared my thoughts on the "`findpeaks()`" method; please take a look. I'll keep assessing the code. 
Much obliged

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+if meas.success
+	return
+end
+
+i1 = round(param.fs * param.t_ignore_from_start);
+i2 = length(x) - round(param.fs * param.t_ignore_from_end);
+[b, a] = butter(2, 0.95, 'high');
+y = abs(filter(b, a, x));
+z = y(i1:i2);
+thr = mean(z) + std(z);
```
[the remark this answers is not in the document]
singalsu: The mean of non-negative numbers is non-negative, there's abs() above. Variance is mean of squared differences so, non-negative. Standard deviation is square root of variance, so also non-negative. I don't think it can be negative. How?

ShriramShastry (commented):
A few more ideas on the `check_glitch_periodicity` function have been added. 
Thanks

singalsu:
I changed the SNR, signal, noise plot appearance to this

bat wav 05IgGp_level

The glitch find/plot can now also find glitches during gain ramp with added time domain glitch detect detect. It's not as good as FFT for tiny glitches but it can find coarse ones like the error in https://github.com/thesofproject/sof/issues/7298 .

test-case/check-alsabat.sh
singalsu: Seems that I did it to avoid the shellcheck warning, but now it's removed.

tools/check_wav_file_snr.m
singalsu: I don't know the size, so it's hard. Declaring similar structure with zero is not mandatory. This will be only interpreted code and since the amount of data in the array of  structs is low there's no performance issue. Since it's a function, and not a script there's no risk of data from previous call to interfere. So I'd leave it as is.

tools/check_wav_file_snr.m
singalsu: There's no histogram in Octave. Function hist seems to work with both.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,371 @@
...
+% All zeros or DC
+if abs(min(x) - max(x)) < eps
+	success = 0;
+else
+	success = 1;
+end
+
+end
+
+function meas = get_tone_levels(stft, param, f, t, signal_idx)
```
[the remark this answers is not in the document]
singalsu: Yep, I've cleaned it up. I moved them to params.

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,403 @@
...
+
+function meas = check_glitch_periodicity(x, param, meas, ch)
+
+if meas.success
+	return
+end
+
+i1 = round(param.fs * param.t_ignore_from_start);
+i2 = length(x) - round(param.fs * param.t_ignore_from_end);
+[b, a] = butter(2, 0.95, 'high');
```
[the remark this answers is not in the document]
singalsu: Since the alsabat test signal is always a low frequency sine wave, the high-pass filter does a good job in attenuating the test signal, only glitches appear at high level after that. It could be a notch too for the test frequency, but the low order high-pass converges faster. I didn't much experiment with 0.95 vs other cut-off.

The periodicity reporter still gets quite confused from noise, this part is not working well yet. The text should not be printed if the result is not reliable. Maybe I could disable this for for now until there's a better way.

test-case/check-alsabat.sh
```diff
@@ -95,6 +96,20 @@ function __upload_wav_file
...
 }
 
+function check_wav_file_snr
+{
+    for file in /tmp/bat.wav.*
+    do
+	if test -s "$file"; then
+	    dlogi "Checking wav file $file"
+	    cd "$TESTDIR"/tools
+	    octave --silent --no-gui --eval "check_wav_file_snr('$file', $frequency, '$LOG_ROOT/')" || {
```
singalsu, line 106: There should be ; in the end of octave command to avoid see "passed = 1" in test log

tools/check_wav_file_snr.m
```diff
@@ -0,0 +1,439 @@
...
+	end
+	fprintf(1, '\n');
+end
+
+end
+
+function success = signal_found(x)
+
+% All zeros or DC
+if abs(min(x) - max(x)) < eps
```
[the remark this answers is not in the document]
singalsu: I added this to fail if signal is all digital zeros. Eps should work in all cases to detect such. But I'll check that it works in both.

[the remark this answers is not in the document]
ShriramShastry: please refer the link, 
Matlab : [https://in.mathworks.com/help/matlab/ref/eps.html](url)
Octave:  [https://octave.sourceforge.io/octave/function/eps.html](url)
check works if `x` data limits are in bounds .

I'd like to know why `min(x)-max(x)` is taken into consideration rather than `max(x)-min.(x)`,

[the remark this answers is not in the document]
singalsu: On my computer both Matlab and Octave return the same eps

```
>> eps

ans =

   2.2204e-16
```
Do you have a suggestion how this line should be fixed to avoid the concern that it doesn't work? 10 * eps should be fine as well?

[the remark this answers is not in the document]
ShriramShastry: OK, I see the following information, and because the constant is small, I believe there is a low possibility of failure.
K>> sprintf('matlab =  %.32f, fixedOctave = %.32f',eps,2.2204e-16)

ans =

    'matlab =  0.00000000000000022204460492503131, fixedOctave = 0.00000000000000022204000000000000'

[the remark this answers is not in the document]
paulstelian97: The notes are an approximation. Octave should still return the value returned by matlab (which is a negative power of 2, being the ULP for double precision 1.0), it would be extremely weird if it didn't.

So there should be no concern about how it works or doesn't work. I don't know where the myth that the value is treated differently between Matlab and Octave comes from (shouldn't the value of epsilon come from the C standard library anyway for both? Epsilon is defined as the ULP for 1.0, that is the difference between 1.0 and the smallest double precision value greater than 1.0, which is 2^-52 indeed on any platform with the usual double precision floats)

2.2204e-16 is an approximation to the correct epsilon. You can use it if you want to, but there's no real benefit to that over the actual epsilon value.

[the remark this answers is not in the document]
ShriramShastry: I double-checked the `eps` findings in Matlab and Octave, and they are identical [`0.0000000000000002220446049250313080847263336181640625`]. Please disregard the comment. Thank you very much.

singalsu:
One more push, fixed a green failure with TGLU_RVP_NOCODEC_IPC4ZPH / sh-tglu-rvp-nocodec-ci-02 https://sof-ci.01.org/softestpr/PR1016/build251/devicetest/index.html, there was a glitch plotted but script didn't return error.

marc-hb (commented):
Could we make this new check optional at first? Then we could adjust the test plans to run check-alsabat twice: once in the old and lax mode, then a second time with this new and stricter check. This will give us a transition phase where the old and lax mode will pass and the new stricter mode will not yet in many configurations. The old mode will catch regressions while the new mode will show progressions.

If agreed then the easiest way to implement these two modes is probably to run the new octave check only when the user passes `-q SNR`.

An invalid 0dB SNR default value can detect when then user passes -q.

PR #1018, opened 2023-04-07
hijack: don't collect logger if SOF_LOGGING=none

fredoh9:
SOF_LOGGING=none means no logger log. This condition must be checked before the other conditions.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -101,8 +101,8 @@ function func_exit_handler()
 
     fi
 
-    if is_ipc4 && is_firmware_file_zephyr; then
-	local mtraceBin; mtraceBin=mtrace-reader.py
+    if [ "$SOF_LOGGING" != 'none' ] && is_ipc4 && is_firmware_file_zephyr; then
+        local mtraceBin; mtraceBin=mtrace-reader.py
         dlogi "pkill -TERM -f $mtraceBin"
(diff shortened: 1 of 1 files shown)


case-lib/hijack.sh
```diff
@@ -101,8 +101,8 @@ function func_exit_handler()
 
     fi
 
-    if is_ipc4 && is_firmware_file_zephyr; then
-	local mtraceBin; mtraceBin=mtrace-reader.py
+    if [ "$SOF_LOGGING" != 'none' ] && is_ipc4 && is_firmware_file_zephyr; then
+        local mtraceBin; mtraceBin=mtrace-reader.py
```
paulstelian97, line 105: Whitespace change: Good thing this isn't Python. Rest of the change isn't an issue so I can still approve.

paulstelian97 (approved):

marc-hb (approved):
Fix for unrelated shellcheck warning in https://github.com/thesofproject/sof-test/actions/runs/4635603962/jobs/8202801248?pr=1018 submitted in separate PR #1020

case-lib/hijack.sh
marc-hb: I hate unrelated whitespace fixes because they break git cherry-pick, merge, blame etc. but in this case the change is so close that git would conflict anyway.

PR #1019, opened 2023-04-07
[SKIP CI] run-all-test: make similar test cases with pr test

fredoh9:
run-all-test.h provides a good way to do local sanity tests. This change is to accommodate a known pr test plan.

+kernel-boot-log
+playback_all_formats
+capture_all_formats
+runtime-pm-status
+multiple-pipeline-all

-volume
-xrun-injection-playback
-xrun-injection-capture

I put 'SKIP CI', run-all-test.sh is not part of device test.

### DIFF
--- a/test-case/run-all-tests.sh
+++ b/test-case/run-all-tests.sh
@@ -21,6 +21,7 @@ very_short_tests()
 
 
 testlist="
+kernel-boot-log
 tplg-binary
 pcm_list
 sof-logger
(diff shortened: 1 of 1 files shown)


greg-intel:
Is it too late to switch the underscores to dashes in the file names, to match the formatting everywhere else?

test-case/run-all-tests.sh
```diff
@@ -169,6 +171,10 @@ interrupted_results()
 	exit 1
 }
 
+test_kernel-boot-log()
```
marc-hb, line 174: I agree with @greg-intel , please no mix in the same name.

```suggestion
test_kernel_boot_log()
```

greg-intel: I'm leaning towards renaming the few tests that have underscores to be dashes, as 95% of our tests are named using dashes.  I'm not sure how the underscore tests got through, I don't remember seeing them...

marc-hb: > I'm leaning towards renaming the few tests...

Consistency and simplicity are good and should absolutely be a goal for new code or code that is changing anyway for other reasons.

For unchanged code it's much more questionable. Is it worth the git churn and potential git conflicts with PRs in progress, branches, etc? It depends.

PR #1020, opened 2023-04-07
hijack.sh: properly disable shellcheck on fake logfile=$logfile 

marc-hb:
Fixes warning found in
https://github.com/thesofproject/sof-test/actions/runs/4635603962/jobs/8202801248?pr=1018 testing PR #1018

See inline comments. Fixes commit 0ccb4c80d2f8716 ("hijack.sh: more generic way to disable unknown $logfile shellcheck")

Different shellcheck versions may warn about this fake definition differently but life is too short to compare them, just disable both SC2154 and SC2269

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -52,7 +52,10 @@ function func_exit_handler()
         sleep 1
 
         # $logfile is defined in a different file (lib.sh)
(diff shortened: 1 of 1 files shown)


fredoh9 (approved):
thank you for the update

PR #1021, opened 2023-04-11
alsa_settings: adjust amixer setting for TGLU_RVP_SDW_IPC4ZPH

keqiaozhang:
Set the 'rt711 AMIC Volume' to 0 and turn up the volume of 'rt711 ADC 08 Capture Volume' a bit.


keqiaozhang:
@bardliao Could you please check with Relteak why does 'rt711 AMIC Volume' affect the waveform? Thanks!

bardliao:
> @bardliao Could you please check with Relteak why does 'rt711 AMIC Volume' affect the waveform? Thanks!

`rt711 AMIC Volume` is a boost gain. So, if we set it to a non-zero value, a gain will be added to the captured data.

fredoh9 (approved):
Thank you for the fix

Issue #1022, opened 2023-04-12
[BUG] multiple-pipeline-playback may start recording!!!!


plbossart:
**Describe the bug**
While using the sof-test scripts on a platform with limited capabilities, I came across a "feature" of the multiple-pipeline-playback/capture scripts.

Because of limited capabilities, I only allowed for 2 PCM playback devices by skipping all others with 
TPLG_IGNORE_LST['pcm']='HDA Digital,Port2'

This results in Port0 and Port1 being used.

However this also results in Port0 and Port1 being used for capture in parallel with playback. This seems to be a 'feature' where capture streams are added if there are fewer than 4 playback devices.

That's not supposed to happen, mixing playback and capture is a different test.

**To Reproduce**
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-lnl-nocodec.tplg /root/sof-test-main/test-case/multiple-pipeline-playback.sh -l 50
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-lnl-nocodec.tplg /root/sof-test-main/test-case/multiple-pipeline-capture.sh -l 50

**Expected behavior**
playback should only playback...

@mengdonglin @keqiaozhang @marc-hb @fredoh9 FYI, this problem needs to be fixed to allow the use of scripts even with limited capabilities.

mengdonglin:
@plbossart Probably we can look at this issue next week? This week @keqiaozhang is crazy busy with deployment updates for the new building service, this is a task we must finish with help of Yongan before Apr 30. We hope to make basic flow work this week and complete debugging/code review in the next 2 weeks.   Not sure if @fredoh9 has bandwidth this week.

keqiaozhang:
@plbossart and @plbossart 
This is intentional in our test script. If playback pipeline count is less than requested, capture pipelines will be started to ensure requested number of pipelines are running in parallel.

Please refer to:
https://github.com/thesofproject/sof-test/pull/570/files

keqiaozhang:
The default value of pipeline count is 4, so if you only want to test 2 playback pipelines in parallel, you can specify it with `-f p -c=2`.
```
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-lnl-nocodec.tplg /root/sof-test-main/test-case/multiple-pipeline-playback.sh -l 50 -f p -c 2
```

plbossart:
[...]

PR #1024, opened 2023-04-13
multiple-pipeline: simplify testing options with playback/capture/all 

fredoh9:
-f option was introduced to decide whether to fill the pipeline with
playback or capture first. But it causes more confusion. Now still keep
-f option, but to simplify the pipeline mode option.

playback mode : only plyaback pipeline
capture mode : only capture pipelines
all mode : run both playback and capture pipelines

Simple usage example:

* playback pipelines only
multiple-pipeline-playback.sh # playback with default values
multiple-pipeline-playback.sh -c 5 -l 15
or
multiple-pipeline.sh -f p -c 5 -l 15

* capture pipelines only
multiple-pipeline-capture.sh # capture with default values
multiple-pipeline-capture.sh -c 5 -l 15
or
multiple-pipeline.sh -f c -c 5 -l 15

* all pipelines, set pipeline count large enough (SOF CI need to change the command from '-f p' to '-f a')
multiple-pipeline.sh -f a -c 20 -l 50

fixes:
- https://github.com/thesofproject/sof-test/issues/1022

### DIFF
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -96,12 +96,15 @@ func_run_pipeline_with_type()
     func_pipeline_export "$tplg" "type:$direction $opt_filter"
     local -a idx_lst
     if [ ${OPT_VAL['r']} -eq 0 ]; then
-        idx_lst=("$(seq 0 $((PIPELINE_COUNT - 1)))")
+        # shellcheck disable=SC2207
+        idx_lst=($(seq 0 $((PIPELINE_COUNT - 1))))
     else
         # convert array to line, shuf to get random line, covert line to array
-        idx_lst=("$(seq 0 $((PIPELINE_COUNT - 1)) | sed 's/ /\n/g' | shuf | xargs)")
+        # shellcheck disable=SC2207
+        idx_lst=($(seq 0 $((PIPELINE_COUNT - 1)) | sed 's/ /\n/g' | shuf | xargs))
     fi
-    for idx in ${idx_lst[*]}
+
+    for idx in "${idx_lst[@]}"
     do
         channel=$(func_pipeline_parse_value "$idx" channel)
         rate=$(func_pipeline_parse_value "$idx" rate)
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -8,22 +8,16 @@
 ## Description:
 ##    Run multiple pipelines in parallel
 ##    Rule:
-##      1. playback first mode: playback pipelines will be started first, if playback
-##         pipeline count is less than requested pipeline count defined by -c option,
-##         capture pipelines will be started to ensure requested number of pipelines are
-##         running in parallel.
-##      2. capture first mode: same with playback mode, but capture pipelines will be
-##         started first.
-##      3. all pipelines mode: run all pipelines in parallel by
-##          a. use option '-f a' OR
-##          b. specify a very big requested pipeline count to -c option
+##      1. Playback mode: playback pipelines ONLY are used, test pipeline count
+##         (OPT_VAL['c']) is respected, but maxed out with playback pipelines.
+##      2. Capture mode: same with playback mode, but capture pipelines ONLY will
+##         be used.
+##      3. All pipelines mode: Run all playback and capture pipelines in parallel
+##         by using the option '-f a' and set the number of pipelines with -c.
+##         If you need to use all pipelines, set a high pipeline count such as -c 20.
 ## Case step:
 ##    1. acquire pipeline count that will be running in parallel
-##    2. start playback or capture pipelines
-##        a. Playback First Mode: start playback pipeline(s), if the requested pipeline count
-##        is not satisfied, start capture pipeline(s)
-##        b. Capture First Mode: start capture pipeline(s), if the requested pipeline count is
-##        not satisfied, start playback pipeline(s)
+##    2. start playback or capture pipelines or all pipelines
 ##    3. wait pipeline process(es) to be started and running
 ##    5. check process status & process count
 ##    6. running pipelines in parallel for a period of time defined by -w option
@@ -45,8 +39,8 @@ OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
 OPT_NAME['c']='count'    OPT_DESC['c']='test pipeline count'
 OPT_HAS_ARG['c']=1         OPT_VAL['c']=4
 
-OPT_NAME['f']='first'
-OPT_DESC['f']='Fill either playback (p) or capture (c) first or any (a) for all pipelines'
+OPT_NAME['f']='fill mode'
+OPT_DESC['f']='fill mode, either playback (p) or capture (c) or any (a) for all pipelines'
 OPT_HAS_ARG['f']=1         OPT_VAL['f']='p'
 
 OPT_NAME['w']='wait'     OPT_DESC['w']='duration of one (sub)test iteration'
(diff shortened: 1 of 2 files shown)


test-case/multiple-pipeline.sh
```diff
@@ -71,12 +65,18 @@ logger_disabled || func_lib_start_log_collect
...
-# in topology is less than the required pipeline count, all pipeline will be started.
+if [ "$f_arg" == "p" ]; then
+    func_pipeline_export "$tplg" "type:playback & ${MULTI_PIPELINE_FILTER}"
+elif [ "$f_arg" == "c" ]; then
+    func_pipeline_export "$tplg" "type:capture & ${MULTI_PIPELINE_FILTER}"
+elif [ "$f_arg" == "a" ]; then
+    func_pipeline_export "$tplg" "type:any & ${MULTI_PIPELINE_FILTER}"
+else
+    die "not supported option $f_arg"
+fi
```
marc-hb, line 76:
```shell
case "$f_arg" in
   'p') .... ;;

    *) die "unsupported option $f_arg"
```

fredoh9:
multiple-pipeline-all.sh from SOF CI is changed from '-f p' to '-f a'

marc-hb (commented):
Missing this:
```diff
- OPT_DESC['f']='Fill either playback (p) or capture (c) first or any (a) for all pipelines'
+ OPT_DESC['f']='Fill either playback (p) or capture (c) or any (a) for all pipelines'
```

Can you please change the following line and add at least one `if/then/else`?
```
[[ $PIPELINE_COUNT -gt ${OPT_VAL['c']} ]] && max_count=${OPT_VAL['c']} || max_count=$PIPELINE_COUNT
```
Short-circuit with a mix of `&&` and `||` is really unreadable.

Do our test plans need some corresponding name changes? Maybe not, maybe they were vague enough :-)

test-case/multiple-pipeline.sh
```diff
@@ -8,22 +8,16 @@
...
-##      3. all pipelines mode: run all pipelines in parallel by
-##          a. use option '-f a' OR
-##          b. specify a very big requested pipeline count to -c option
+##      1. Playback mode: playback pipelines ONLY are used, test pipeline count
+##         (OPT_VAL['c']) is respected, but maxed out with playback pipelines.
+##      2. Capture mode: same with playback mode, but capture pipelines ONLY will
+##         be used.
+##      3. All pipelines mode: Run all playback and capture pipelines in parallel
+##         by using the option '-f a' and set the number of pipelines with -c.
+##         If you need to use all pipelines, set a high pipeline count such as -c 20.
```
plbossart, line 17: what happens if the -c option is lower than the total of pipelines?

I also feel that this option could quickly run in trouble, with e.g. the nocodec topology we have multiple mixers and multiple copiers on capture, so we could end-up with 10+ streams in parallel. That's likely to fail in practice.

We may need to have two counts for capture and playback each, it's more pragmatic to me than trying to run 20 pipelines.

plbossart (commented):
Thanks for taking the lead on this @fredoh9, much appreciated. See a couple of comments below, I am not sure having a large number of pipelines is realistic/feasible.

test-case/multiple-pipeline.sh
fredoh9: If -c option has less than total pipeline. first fill with playback pipelines, rest of number will be filled with capture. so capture pipeline will be limited by the number.

For example, with -c 20, MTLP NOCODEC tplg has playback 4 pcm + capture 7 pcm = total 11 pcm. so mulitple-pipeline-all will start all 11 pcm at same time. If -c 8 is set, playback 4 pcm + capture 4 pcm will be used. If any pcm need to be skipped, there is a way to filter out by pcm name.

If we have two counts, there needs to be a little more condition checking between two counts. still trivial but I need more requirement for all(playback + capture) case.

fredoh9:
> Missing this:
> 
> ```diff
> - OPT_DESC['f']='Fill either playback (p) or capture (c) first or any (a) for all pipelines'
> + OPT_DESC['f']='Fill either playback (p) or capture (c) or any (a) for all pipelines'
> ```
> 
> Can you please change the following line and add at least one `if/then/else`?
> 
> ```
> [[ $PIPELINE_COUNT -gt ${OPT_VAL['c']} ]] && max_count=${OPT_VAL['c']} || max_count=$PIPELINE_COUNT
> ```

Addressed above comments from @marc-hb.
I rebased top of main unnecessarily. diff from the previous commit has a bit more.

paulstelian97 (dismissed):

test-case/multiple-pipeline.sh
```diff
@@ -70,13 +64,22 @@ logger_disabled || func_lib_start_log_collect
...
+    *) die "not supported option $f_arg";;
+esac
+func_pipeline_export "$tplg" "$type_filter & ${MULTI_PIPELINE_FILTER}"
+
+# respect number of pipelines requested but max will be real pipeline count
+if [[ $PIPELINE_COUNT -gt ${OPT_VAL['c']} ]]; then
+    max_count=${OPT_VAL['c']}
+else
+    max_count=$PIPELINE_COUNT
+fi
```
marc-hb, line 82: maybe this:
```shell
min() { printf '%d' $(( "$1" < "$2"  ? "$1" : "$2" )); }

max_count=$(min "${OPT_VAL['c']}" $PIPELINE_COUNT)
```

marc-hb (commented):
codestyle looks good, someone else must approve the logical changes.

For the record, missing blobs in
https://sof-ci.01.org/softestpr/PR1024/build295/devicetest/index.html
https://sof-ci.01.org/softestpr/PR1024/build296/devicetest/index.html
https://sof-ci.01.org/softestpr/PR1024/build297/devicetest/index.html

EDIT: fixed by Greg, it was some ssh key issue.

marc-hb:
If you want you can fix the single shellcheck warning in https://github.com/thesofproject/sof-test/actions/runs/4746518584/jobs/8430264948?pr=1024 with this small change, preferably in a separate commit:

```diff
--- a/test-case/multiple-pipeline.sh
+++ b/test-case/multiple-pipeline.sh
@@ -102,7 +102,7 @@ func_run_pipeline_with_type()
         # convert array to line, shuf to get random line, covert line to array
         idx_lst=("$(seq 0 $((PIPELINE_COUNT - 1)) | sed 's/ /\n/g' | shuf | xargs)")
     fi
-    for idx in ${idx_lst[*]}
+    for idx in "${idx_lst[@]}"
     do
         channel=$(func_pipeline_parse_value "$idx" channel)
         rate=$(func_pipeline_parse_value "$idx" rate)

```

fredoh9:
@marc-hb when I put double quote, I have expansion/substitution error. lost now

```
In test-case/multiple-pipeline.sh line 105:
    for idx in ${idx_lst[@]}
               ^-----------^ SC2068 (error): Double quote array expansions to avoid re-splitting elements.
```

Error message,
```
/home/ubuntu/sof-test/case-lib/pipeline.sh: line 81: [[: 0
                                                                                          1
                                                                                            2
                                                                                             3: syntax error in expression (error token is "1
    2
     3")
 /home/ubuntu/sof-test/case-lib/pipeline.sh: line 86: ${PIPELINE_0
                                                                         1
                                                                          2
                                                                           3[channel]}: bad substitution
```

marc-hb (commented):
Everything looks great to me, @plbossart can you please take a look and confirm it addresses the bug you filed as you wanted it?

plbossart (approved):

marc-hb (approved):

PR #1028, opened 2023-05-03
tplgtool2.py: Use ifelse to avoid dependency on Python 3.10

yongzhi1:


ranj063 (dismissed):
Thanks you so much for this, missing commit message!

fredoh9:
@yongzhi1 can you elaborate more why old code had dependency problem? Do you see some errors with Python 3.10?

marc-hb (approved):
If this is all it takes to work with Python 3.8 then great!


Next time @yongzhi1 please provide a commit message with such information to save reviewers' time:

From https://docs.python.org/3/reference/compound_stmts.html#the-match-statement

> New in version 3.10.

tools/tplgtool2.py
```diff
@@ -842,13 +842,12 @@ def node_name_in_graph(self, node: Container) -> str:
...
-        match widget["widget"]["id"]:
-            case 'PGA':
-                sublabel = f'<BR ALIGN="CENTER"/><SUB>PGA: {widget["kcontrols"][0]["hdr"]["name"]}</SUB>'
-            case 'AIF_IN' | 'AIF_OUT' | 'DAI_IN' | 'DAI_OUT':
-                sublabel = f'<BR ALIGN="CENTER"/><SUB>{widget["widget"]["id"]}: {widget["widget"]["sname"]}</SUB>'
-            case _:
-                sublabel = ""
+        if widget["widget"]["id"] == 'PGA':
+             sublabel = f'<BR ALIGN="CENTER"/><SUB>PGA: {widget["kcontrols"][0]["hdr"]["name"]}</SUB>'
+        elif widget["widget"]["id"] in ('AIF_IN' , 'AIF_OUT' , 'DAI_IN' , 'DAI_OUT'):
```
marc-hb, line 847: A tuple is a bit surprising, I think a list would be more... "pythonic"
```suggestion
        elif widget["widget"]["id"] in ['AIF_IN' , 'AIF_OUT' , 'DAI_IN' , 'DAI_OUT']:
```

marc-hb (approved):

marc-hb (changes requested):
Change is good but please fix the pylint warnings: there are only 4 of them!

Use a separate commit if unrelated.

marc-hb:
For one of the pylint warnings do this:
```diff
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -841,6 +841,7 @@ class TplgGraph:
         """
         return self._node_name_in_graph_from_name(node["widget"]["name"])
 
+    # pylint: disable=no-self-use
     def node_sublabel(self, widget: Container) -> str:
         match widget["widget"]["id"]:
             case 'PGA':
```
For the others just fix the indentation.

tools/tplgtool2.py
yongzhi1: > A tuple is a bit surprising, I think a list would be more... "pythonic"

Ack, will fix in next update.

yongzhi1:
V2 update:
1. Add commit description.
2. fix pylint reported warnings.

Thanks everyone, I learned more from your reviews :)
('no-self-use' was removed https://pylint.readthedocs.io/en/latest/whatsnew/2/2.14/summary.html#removed-checker)

marc-hb:
> 'no-self-use' was removed

OK, I see `# pylint: disable=no-self-use` warns about "useless disable" with a newer pylint, which is pretty stupid from pylint because now we have to choose between a clean run either before or after the removal but can't have both. Bummer.

Please add the line to we can have a clean pylint with Ubuntu 22 which is the latest LTS used by 95% of Ubuntu users (https://ubuntu.com/about/release-cycle) and the latest container provided by Github (https://github.com/actions/runner-images)

PS: funny how you use an old Python version but a recent pylint :-)

yongzhi1:
V3 update:

Added  _# pylint: disable=no-self-use_ to suppress warning:
```tools/tplgtool2.py:844:4: R0201: Method could be a function (no-self-use)```

yongzhi1:
> I don't see it. Wrong branch pushed?
Sorry one more time :)

marc-hb (approved):

marc-hb:
Something's wrong, we don't have sof-ci/jenkins hardware tests! Looking into it.

marc-hb:
> On the internal server I can see that test run 24899 has tested latest commit https://github.com/thesofproject/sof-test/commit/e67fda92f100bcf3d4f35729ccb2ba8a9725b0fd (17:35 UTC) just fine

Actually, test run 24899 was supposed to test the _previous_ commit 9d932e4 (17:21 UTC) but you force-pushed only 14 minutes later which confused CI. I documented this in internal issue 413.

Now the "real" results triggered by e67fda92f100b (17:35 UTC) have arrived, they're published and they look good.

Sorry for the confusion. A good reminder to avoid force-pushes too close to each other.

fredoh9 (approved):
thank you for updating commit text.

Issue #979, opened 2022-11-09
[BUG] "Test Result: FAIL!" is missing when firmware path is not found from journalctl


miRoox:
**Describe the bug**
When firmware path is not found from journalctl on an IPC4 test, it will immediately exit without report "Test Result: FAIL!". So our CI will treat it as TIMEOUT.

**To Reproduce**
It's hard to reproduce, but we can mock the case like #978.

**Expected behavior**
Report "Test Result: FAIL!" at the end of the log.

**Detail Info**
1) Branch name and commit hash of the 3 repositories: sof (firmware/topology), linux (kernel driver) and sof-test (test case)
    * Kernel: https://github.com/thesofproject/linux/commit/551c11c310ac
    * SOF: https://github.com/thesofproject/sof/commit/a0d9a4e4acf7
    * SOF-TEST: 2e1573c
2) Test report ID (if you find it from test report)
    * ID: 17476
3) Test DUT Model (or a brief discribtion about the device)
    * MODEL: TGLU_RVP_NOCODEC_IPC4ZPH

**Screenshots or console output**

First occurrence:

```
2022-11-08 22:01:29 UTC N ===== Testing: (Round: 50/50) (PCM: DMIC [hw:0,10]) (Loop: 1/1) =====
2022-11-08 22:01:29 UTC N no file prefix, use /dev/null as dummy capture output
2022-11-08 22:01:29 UTC N arecord   -Dhw:0,10 -r 48000 -c 4 -f S32_LE -d 1 /dev/null -v -q
Hardware PCM card 0 'sof-nocodec' device 10 subdevice 0
Its setup is:
  stream       : CAPTURE
  access       : RW_INTERLEAVED
  format       : S32_LE
  subformat    : STD
  channels     : 4
  rate         : 48000
  exact rate   : 48000 (48000/1)
  msbits       : 32
  buffer_size  : 4096
  period_size  : 1024
  period_time  : 21333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 1024
  period_event : 0
  start_threshold  : 1
  stop_threshold   : 4096
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
  appl_ptr     : 0
  hw_ptr       : 0
declare -- cmd="journalctl_cmd --since=@1667944672"
[...]

Issue #291, opened 2020-07-20
Add test for no-i915 configurations


kv2019i:
SOF driver now supports configurations where the HDMI/DP audio via integrated graphics (i915) is disabled. No separate driver or topology is needed to make this work. However, currently the SOF CI is not testing this configuration and we've had multiple regresssions and bugs that only occur in this configuration. Some examples:

- https://github.com/thesofproject/linux/pull/2287
- https://github.com/thesofproject/linux/pull/2270

mengdonglin:
@plbossart  @xiulipan and @RanderWang aligned on how to support this configuration in SOF CI. It seems we cannot support this in sof-test, but in CI test framework.

plbossart:
@mengdonglin can you elaborate? I don't get at all the explanations.

xiulipan:
@plbossart I think @mengdonglin would like us to add some fail case in CI. Platform with non-i915 support will be one of this kind of test

marc-hb:
@kv2019i is this still needed?

fredoh9:
I came here to add this PR, but it is already mentioned.

https://github.com/thesofproject/sof-test/pull/1033

PR #1033, opened 2023-05-04
lib: wait for fw loaded before starting any test

fredoh9:
This is going to check whether the sof-test can be started after boot/reboot.

If FW LOADING is not a prerequisite for the test, this can be skipped, by defining NO_POLL_FW_LOADING=true

Another benefit with this change, if FW is not loaded it will be failed instead of TIMEOUT.

### DIFF
--- a/case-lib/config.sh
+++ b/case-lib/config.sh
@@ -90,3 +90,13 @@ FALLBACK_TO_PROC=${FALLBACK_TO_PROC:-false}
 
 # Skip the storage checks if they aren't required
 ENABLE_STORAGE_CHECKS=${ENABLE_STORAGE_CHECKS:-0}
+
+# start_test in lib.sh will check whether FW is loaded before every test case.
+# If FW LOADING is not a prerequisite for the test, or if you just want to skip
+# the check, then define NO_POLL_FW_LOADING to true.
+# NO_POLL_FW_LOADING=true
+
+# Set MAX Polling time to check FW Loading. If FW is already loaded, it will
+# return immediately. Default value is set to 60 seconds, because i915 driver
+# timeout is 60 seconds.
+MAX_WAIT_FW_LOADING=${MAX_WAIT_FW_LOADING:-70}
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -74,6 +74,18 @@ start_test()
         return 0
     }
 
+    # Check for whether SOF fW is loaded or not before starting any test
+    # without --since=@"$KERNEL_CHECKPOINT", it will be scan for current boot
+    # fw loaded -> sof module removed case can't be detected.
+    # Mainly it is going to check whether the sof-test can be started after boot/reboot
+    if [ -z "$NO_POLL_FW_LOADING" ]; then
+        if poll_wait_for 1 "$MAX_WAIT_FW_LOADING" sof_firmware_boot_complete; then
+            dlogi "Good to start the test, FW is loaded!"
+        else
+            die "FW is not loaded for $MAX_WAIT_FW_LOADING"
+        fi
+    fi
+
     export SOF_TEST_TOP_PID="$$"
     local prefix; prefix="ktime=$(ktime) sof-test PID=${SOF_TEST_TOP_PID}"
     local ftemp; ftemp=$(mktemp --tmpdir sof-test-XXXXX)


fredoh9:
typo in title, staring -> starting.
will fix when next push is required

case-lib/config.sh
```diff
@@ -90,3 +90,13 @@ FALLBACK_TO_PROC=${FALLBACK_TO_PROC:-false}
...
+
+# start_test in lib.sh will check whether FW is loaded before every test case.
+# If FW LOADING is not a prerequisite for the test, or if you just want to skip
+# the check, then define NO_POLL_FW_LOADING to true.
+# NO_POLL_FW_LOADING=true
+
+# Set MAX Polling time to check FW Loading. If FW is already loaded, it will
+# return immediately. Default value is set to 60 seconds, because i915 driver
+# timeout is 60 seconds.
+MAX_WAIT_FW_LOADING=${MAX_WAIT_FW_LOADING:-60}
```
marc-hb, line 102: If i915 driver is 60 seconds, shouldn't this be 65 seconds?

Also: can you please add at least one reference for more context? Maybe some of these:
- https://github.com/thesofproject/sof-test/issues/291
- https://github.com/thesofproject/linux/issues/3042
- https://github.com/thesofproject/sof-test/issues/979
- other?

marc-hb (commented):
@plbossart should we add a counter and print a warning when waiting more than 10s?

Note we already discussed this in test-specific #862. There we set a 10s maximum 1 year ago and it has never spuriously failed.

case-lib/config.sh
fredoh9: > If i915 driver is 60 seconds, shouldn't this be 65 seconds?

Good point, I think about this too. But there are some other delays already applied top of this. so it should have plenty of room

plbossart, line 102: what is the relationship between DSP firmware load and i915? is this that we first request the i915 modules before downloading the firmware?

and if yes, should we use the same routine as for the kmod remove/modprobe test where we have an iteration that stops on BOOT_COMPLETE.

plbossart (commented):
couple of comments below @fredoh9, nice work indeed.

case-lib/config.sh
marc-hb: > But there are some other delays already applied top of this

Once we merge this then we should get rid of those.

marc-hb: > and if yes, should we use the same routine as for the kmod remove/modprobe test where we have an iteration that stops on BOOT_COMPLETE.

That's exactly the suggestion I made in internal bug 387 and what @fredoh9 is doing here: he's re-using the functions I wrote in #862

marc-hb (commented):
> @plbossart should we add a counter and print a warning when waiting more than 10s?

I think you missed this question

marc-hb:
> Another benefit with this change, if FW is not loaded it will be failed instead of TIMEOUT.

Because it's the "last chance" TIMEOUT in our CI it's very long, how long Fred?.

Moreover we power-cycle the device after such a CI TIMEOUT. When combined with another unrelated issue #979, the whole test run ends up in a very time-consuming tail spin on that device.

case-lib/config.sh
marc-hb: Also: this is only a maximum. The test will start as soon as the firmware has booted.

fredoh9:
> Because it's the "last chance" TIMEOUT in our CI it's very long, how long Fred?.

TIMEOUT value is dependent on each test case. check-suspend-resume-50.sh has 6000 second. TIMEOUT has become an expensive way to waste resources.

case-lib/config.sh
fredoh9: okay, I will update default value to 70. if you plan to remove all extra delays, I need some extra margin

fredoh9:
> > @plbossart should we add a counter and print a warning when waiting more than 10s?
> 
> I think you missed this question

@plbossart @marc-hb I can print out warning every 10 sec. instead of zero output for MAX wait time

marc-hb (approved):

keqiaozhang (approved):
Nice solution.

paulstelian97 (commented):
This feels good but incomplete. Before merging this I would like to see a good analysis -- do some tests actually need that variable right now or can it be done later?

marc-hb:
@paulstelian97 this will help a lot in scenarios where the GPU driver is missing or failing and audio it timing out and where other issues make the situation much worse. I just unresolved the comment where I asked to provide some links for more context; you could start with #979.

dbaluta (approved):

paulstelian97 (approved):

fredoh9:
> That would be great. I think we could merge this and then add these warnings later?

yes, let's merge this and see. If  we need anything improvement, I will address them together

fredoh9:
I will monitor upcoming the test results

marc-hb:
I think @lyakh found an issue with this new feature: kernel logs seem to be missing when the firmware does not boot. Example: https://sof-ci.01.org/sofpr/PR7584/build7585/devicetest/index.html

In jenkins logs generic_test/13429/
```
10:58:40  cmd: scp -r ubuntu@sh-adlp-rvp-sdw-01.sh.intel.com:~/sof-test/test-case/../logs/verify-firmware-presence/last/* results/2023-05-12-03:55:59/sh-adlp-rvp-sdw-01/verify-firmware-presence
10:58:40  return code: 1
10:58:40  cmd output: scp: /home/ubuntu/sof-test/test-case/../logs/verify-firmware-presence/last/*: No such file or directory
```

This is very likely because the new code runs "too early".

Issue #729, opened 2021-07-03
fix all shellcheck warnings


marc-hb:
First things first

greg-intel:
Current status..
![image](https://github.com/thesofproject/sof-test/assets/93791892/83d105fd-62d9-4f8d-887b-17403e6c0af8)

Not many issues left with files that are used consistently.

greg-intel:
Another Update

![image](https://github.com/thesofproject/sof-test/assets/93791892/a9ddf57e-b017-4da8-98e9-f5fa1eb69665)

greg-intel:
![image](https://github.com/thesofproject/sof-test/assets/93791892/5bbb1673-86bf-4d9b-b5a7-2087dc01c9d7)

There are currently seven open PRs, once they're all merged I believe this ticket can be closed, @marc-hb ?

greg-intel:
Think I'm caught up on all the PRs[.](https://intel-my.sharepoint.com/:x:/r/personal/greg_galloway_intel_com/_layouts/15/Doc.aspx?sourcedoc=%7B271FBDB1-6140-45F9-B602-FAC8BC89177D%7D&file=Shellcheck%20status%20of%20SOF%20Test%20Suite.xlsx&action=default&mobileredirect=true)

marc-hb:
Could you please update and re-share the spreadsheet to reflect the new status?

greg-intel:
![image](https://github.com/thesofproject/sof-test/assets/93791892/28503c06-55e3-4101-909e-119d5e4ca73c)

@marc-hb see any remaining candidates in the ones I left behind?

greg-intel:
Also of note:
https://github.com/thesofproject/sof-test/pull/1019

marc-hb:
> Could you please update and re-share the spreadsheet to reflect the new status?

I meant can you make the fixed tests green and with a "0" to see what's left at a glance? Scanning "19 FIXED" lines one by one does not give any overview (and I'm not even sure what it means)

greg-intel:
I guess the previous image was already up-to-date.  Here, I've color coded the status to make it easier to see which files were left out, all files that were fixed or were "0" are green for column C.
![image](https://github.com/thesofproject/sof-test/assets/93791892/4c9a1dd0-4745-4f51-ad7f-a15a6e383edb)

marc-hb:
Thanks, now this gives an overview. "the new status" meant not interested in the number of issues when you started to work on this.

[...]

PR #1041, opened 2023-05-25
check-capture.sh: fix quoting and "ret=$?" warnings

greg-intel:
Per issue https://github.com/thesofproject/sof-test/issues/729, "Fix all shellcheck warnings".

Prior to this commit there were 13 issues in this file, now there are none.

check-capture.sh was run before and after the fixes, no change in output.

Signed off by Greg Galloway [greg.galloway@intel.com](mailto:greg.galloway@intel.com)

### DIFF
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -65,21 +65,21 @@ func_pipeline_export "$tplg" "type:capture & ${OPT_VAL['S']}"
 
 for round in $(seq 1 $round_cnt)
 do
-    for idx in $(seq 0 $(expr $PIPELINE_COUNT - 1))
+    for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
     do
-        channel=$(func_pipeline_parse_value $idx channel)
-        rate=$(func_pipeline_parse_value $idx rate)
-        fmt=$(func_pipeline_parse_value $idx fmt)
-        dev=$(func_pipeline_parse_value $idx dev)
-        pcm=$(func_pipeline_parse_value $idx pcm)
-        type=$(func_pipeline_parse_value $idx type)
-        snd=$(func_pipeline_parse_value $idx snd)
+        channel=$(func_pipeline_parse_value "$idx" channel)
+        rate=$(func_pipeline_parse_value "$idx" rate)
+        fmt=$(func_pipeline_parse_value "$idx" fmt)
+        dev=$(func_pipeline_parse_value "$idx" dev)
+        pcm=$(func_pipeline_parse_value "$idx" pcm)
+        type=$(func_pipeline_parse_value "$idx" type)
+        snd=$(func_pipeline_parse_value "$idx" snd)
 
         if [ ${OPT_VAL['F']} = '1' ]; then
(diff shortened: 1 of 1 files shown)


test-case/check-capture.sh
```diff
@@ -94,9 +94,9 @@ do
                     dlogi "using $file as capture output"
                 fi
 
-                arecord_opts -D$dev -r $rate -c $channel -f $fmt_elem -d $duration $file -v -q
-                if [[ $? -ne 0 ]]; then
-                    func_lib_lsof_error_dump $snd
+                if ! arecord_opts -D"$dev" -r "$rate" -c "$channel" -f "$fmt_elem" -d $duration "$file" -v -q;
+                then
+                    func_lib_lsof_error_dump "$snd"
```
marc-hb, line 99: Nit: I prefer `arecord_opts ... || { fail ...` because it reads more like plain english and avoids a negation. Also, `!` can't be copied to the command line (try it!)

test-case/check-capture.sh
```diff
@@ -65,21 +65,21 @@ func_pipeline_export "$tplg" "type:capture & ${OPT_VAL['S']}"
...
+        type=$(func_pipeline_parse_value "$idx" type)
+        snd=$(func_pipeline_parse_value "$idx" snd)
 
         if [ ${OPT_VAL['F']} = '1' ]; then
-            fmt=$(func_pipeline_parse_value $idx fmts)
+            fmt=$(func_pipeline_parse_value "$idx" fmts)
         fi
 
-        for fmt_elem in $(echo $fmt)
+        for fmt_elem in $fmt
```
marc-hb, line 82: The previous code was really weird, I don't understand what it was trying to achieve so I'm a bit concerned that the fix seems "too simple". @aiChaoSONG you added this code in commit aedc851c46d0, can you please review this PR?

marc-hb: @greg-intel can you please test the -F option before and after this PR and compare the logs, try to spot any difference?

marc-hb (commented):
Mostly OK, just one small concern below and the lack of a commit title (`check-capture.sh: <blank>`). I suggest `check-capture.sh: fix quoting and "ret=$?" warnings`)

test-case/check-capture.sh
greg-intel: I didn't spot a difference, but I think maybe the device only had one format.  I *did* (during testing) check behavior of this line when there were two items in the list, and it iterated over each individually.  (I added " testing" to fmt, it ran the known value first and the second generated an error because it didn't exist.)

test-case/check-capture.sh
aiChaoSONG: if this pr only fix shellcheck warnings, maybe we leave this to next PR.

test-case/check-capture.sh
aiChaoSONG: I could not recall why it is `$(echo $fmt)`, let's do a test see if `$fmt` is ok, if yes, it will be good then.

marc-hb (approved):

PR #1042, opened 2023-05-25
check-kmod-load-unload-after-playback.sh: fix source reference and quote issues.

greg-intel:
Fixes per ticket https://github.com/thesofproject/sof-test/issues/729

Six issues to zero.

Fixed the source reference to match that of other repos.  Fixed a single quote issue (a single quote block contained a variable).

Signed off by Greg Galloway [greg.galloway@intel.com](mailto:greg.galloway@intel.com)

### DIFF
--- a/test-case/check-kmod-load-unload-after-playback.sh
+++ b/test-case/check-kmod-load-unload-after-playback.sh
@@ -86,7 +86,7 @@ do
             dlogi "platform doesn't support runtime pm, skip waiting" && break
         [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
         sleep 1
-        if [ "$i" -eq 15 ]; then
+        if [ "$i" -ge 15 ]; then
             die "dsp is not suspended after 15s, end test"
         fi
     done
--- a/test-case/check-kmod-load-unload-after-playback.sh
+++ b/test-case/check-kmod-load-unload-after-playback.sh
@@ -32,10 +32,11 @@ set -e
 ##    check kernel log and find no errors
 ##
 
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 case_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
 
-OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_NAME['t']='tplg'     OPT_DESC['t']="tplg file, default value is env TPLG: $TPLG"
 OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
 
 OPT_NAME['l']='loop'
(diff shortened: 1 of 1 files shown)


test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -85,7 +86,7 @@ do
             dlogi "platform doesn't support runtime pm, skip waiting" && break
         [[ $(sof-dump-status.py --dsp_status 0) == "suspended" ]] && break
         sleep 1
-        if [ $i -eq 15 ]; then
+        if [ "$i" -eq 15 ]; then
```
marc-hb, line 89: Can you please add one more commit with this?

```suggestion
        if [ "$i" -ge 15 ]; then
```

Much safer in case someone fails to update both duplicates at the same time.

test-case/check-kmod-load-unload-after-playback.sh
```diff
@@ -32,10 +32,11 @@ set -e
 ##    check kernel log and find no errors
 ##
 
-source $(dirname ${BASH_SOURCE[0]})/../case-lib/lib.sh
+# shellcheck source=case-lib/lib.sh
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 case_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)
 
-OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $TPLG'
+OPT_NAME['t']='tplg'     OPT_DESC['t']="tplg file, default value is env TPLG: $TPLG"
```
marc-hb, line 39: That's a useful bug fix!

marc-hb (approved):

test-case/check-kmod-load-unload-after-playback.sh
greg-intel: Done.

marc-hb (approved):
> Changed -eq 15 to be -ge 15 per Marc.  

This sounds a bit like "just because X asked" which is definitely not enough of a reason to do a code change!

Now if you want to credit someone else for an incredibly good idea then by all means do it but that's not really how this commit message looks right now.

Also, this particular change is very far from ground-breaking.

Never mind, does not matter. Don't fix it.

test-case/check-kmod-load-unload-after-playback.sh
ranj063: Why no sign-off in either commit?

greg-intel:
> > Changed -eq 15 to be -ge 15 per Marc.
> 
> This sounds a bit like "just because X asked" which is definitely not enough of a reason to do a code change!
> 
> Now if you want to credit someone else for an incredibly good idea then by all means do it but that's not really how this commit message looks right now.
> 
> Also, this particular change is very far from ground-breaking.
> 
> Never mind, does not matter. Don't fix it.

Sorry, that wasn't my intent.  I was calling it out because you're much more familiar with the test suite and what's being tested to know what situations were covered.  I had originally tried to be "too smart" and over-explain, but then I realized that might get me in trouble (and that I was probably wrong on some counts) so I simplified my message.  I felt this message lent more credibility to the change.

test-case/check-kmod-load-unload-after-playback.sh
greg-intel: Oversight on my part, sorry.

marc-hb:
The rare test  failures happened in other tests not modified by this PR. Merging.

PR #1051, opened 2023-06-06
check-ipc-flood.sh: Shellcheck fixes

greg-intel:
Fixed lib.sh source reference to avoid shellcheck complaints.

Re-write conditionals to satisfy shellcheck requirements.

Signed-off-by: Greg Galloway <greg.galloway@intel.com>


greg-intel:
The errors did not occur in the changed code.  Actually, the changed code is only tested as part of the daily test, it's set to NA for PR testing.

test-case/check-ipc-flood.sh
```diff
@@ -33,12 +34,12 @@ lpc_loop_cnt=${OPT_VAL['c']}
 ipc_flood_dfs=${OPT_VAL['f']}
 loop_cnt=${OPT_VAL['l']}
 
-[[ ! "$(sof-kernel-dump.sh|grep 'sof-audio'|grep 'Firmware debug build')" ]] && dlogw "${BASH_SOURCE[0]} need debug version firmware" && exit 2
+! sof-kernel-dump.sh | grep sof-audio | grep -q "Firmware debug" && dlogw "${BASH_SOURCE[0]} need debug version firmware" && exit 2
```
marc-hb, line 37: Please take this opportunity to remove the negation, split the too long line and remove the hardcoded exit code

```shell
sof-kernel-dump.sh | grep sof-audio | grep -q "Firmware debug" ||
     skip "need debug version firmware"
```

test-case/check-ipc-flood.sh
```diff
@@ -33,12 +34,12 @@ lpc_loop_cnt=${OPT_VAL['c']}
...
 
 func_lib_check_sudo
 
 dlogi "Check sof debug fs environment"
-[[ "$(sudo file $ipc_flood_dfs|grep 'No such file')" ]] && dlogw "${BASH_SOURCE[0]} need $ipc_flood_dfs to run the test case" && exit 2
+sudo file $ipc_flood_dfs | grep 'No such file' && dlogw "${BASH_SOURCE[0]} need $ipc_flood_dfs to run the test case" && exit 2
```
marc-hb, line 42: Similar comments as above

```shell
if sudo file $ipc_flood_dfs | grep 'No such file'; then
  skip "${BASH_SOURCE[0]} need $ipc_flood_dfs to run the test case"
fi
```

marc-hb:
Don't spill the commit title over the commit body like this:

```
check-ipc-flood.sh: The previous source reference for lib.sh triggers…

… shellcheck errors.
```

The commit title must stand alone.

greg-intel:
> Don't spill the commit title over the commit body like this:
> 
> ```
> check-ipc-flood.sh: The previous source reference for lib.sh triggers…
> 
> … shellcheck errors.
> ```
> 
> The commit title must stand alone.

In the future I'll try to keep the ~titles~ first commit lines to 70 characters, everything looked fine in the git logs, so github is just aggressive with the first line of a commit.

test-case/check-ipc-flood.sh
greg-intel: I was intentionally trying to do the minimum to resolve the shellcheck issues.  It also wouldn't have occurred to me to use skip at all, what does it do? I don't see it in lib.sh (just skip_test) but I don't think that presents as an error.

marc-hb: > I was intentionally trying to do the minimum to resolve the shellcheck issues.

That's fair but you did change the code, I mean you removed `[[ ... ]]` so you already crossed the "code change" line, it's not just adding quotes to please shellcheck's paranoia about special characters.

Note I'm not requesting any changes outside the lines that you are already changing and re-testing. That's why I'm requesting a full clean-up of these lines (and only of these lines): because you can leverage the context and knowledge and test environment that you already have about these lines.

> It also wouldn't have occurred to me to use skip at all, what does it do? I don't see it in lib.sh (just skip_test) but I don't think that presents as an error.

Sorry I meant `skip_test`. It should behave the same without hardcoding the magic value `2`, please test and make sure.

marc-hb (changes requested):
>  so github is just aggressive with the first line of a commit.

My bad, it looks like Github is making it harder and harder to look at the actual commit message. In  this case it displays the commit subject as truncated when it's not, the same way as the _actually_ truncated PR title...

marc-hb (changes requested):
Final code looks code but as mentioned in a different PR don't provide one tentative fix in one commit and then another, different fix for the same line in a latter commit. Get it right the first time in a single commit.

greg-intel:
> Final code looks code but as mentioned in a different PR don't provide one tentative fix in one commit and then another, different fix for the same line in a latter commit. Get it right the first time in a single commit.

Fixed.

marc-hb (approved):

greg-intel:
Check-ipc-flood tests were either passing or NA.

marc-hb:
> Check-ipc-flood tests were either passing or NA.

Always copy the URLs because they will be lost on the next force push.

PR #1052, opened 2023-06-07
alsa_settings: set capture gain to 0dB for MTLP_RVP_SDW

fredoh9:
USB sound card is upgraded to stereo recording. After re-evaluating the output, the capture volume was too high. Set to 0dB.

23 is 0 dB
43 is 15 db
1 step = 0.75dB

### DIFF
--- a/alsa_settings/MTLP_RVP_SDW.sh
+++ b/alsa_settings/MTLP_RVP_SDW.sh
@@ -7,4 +7,4 @@ amixer -c sofsoundwire cset name='rt711 FU05 Playback Volume' 60
 # enable headset capture
 amixer -c sofsoundwire cset name='Headset Mic Switch' on
 amixer -c sofsoundwire cset name='rt711 FU0F Capture Switch' on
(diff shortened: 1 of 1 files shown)


plbossart (approved):

fredoh9:
@marc-hb without this alsabat capture test will be failed. This is low risk PR, can we merge?

marc-hb:
Lots of MTL failures in https://sof-ci.01.org/softestpr/PR1051/build409/devicetest/index.html too.

fredoh9:
the errors are not related, but looking at the problem

fredoh9:
hmm... i'm able to reproduce the problem with just daily builds and sof-test main. confirmed nothing to do with this pr.
Checking the merge/build history.

keqiaozhang:
@fredoh9 , I have submitted a similar PR to fix the volume issue on MTL-SDW after replacing the USB sound card. https://github.com/thesofproject/sof-test/pull/1053 was merged.

marc-hb:
#1053 has a different value and also changed the Playback Volume.

PR #1056, opened 2023-06-13
lib.sh: mtrace doesn't require ldc file on IPC4_Zephyr platforms

keqiaozhang:
We use mtrace on IPC4 Zephyr platforms and it doesn't require ldc file. So add an exception for IPC4 Zephyr platforms before the ldc checking.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -740,11 +740,14 @@ logger_disabled()
 {
     local ldcFile
     # Some firmware/OS configurations do not support logging.
-    ldcFile=$(find_ldc_file) || {
-        dlogi '.ldc dictionary file not found, SOF logs collection disabled'
-        return 0 # 0 is 'true'
-    }
-
+    # mtrace doesn't require ldc file on IPC4-Zephyr platforms.
+    if ! is_ipc4; then
+        ldcFile=$(find_ldc_file) || {
+            dlogi '.ldc dictionary file not found, SOF logs collection disabled'
+            return 0 # 0 is 'true'
+        }
+    fi
+    
     # Disable logging when available...
     if [ ${OPT_VAL['s']} -eq 0 ]; then
         return 0


marc-hb (changes requested):
~This would also disable mtrace~. How did you test this?

marc-hb:
What problems does this solve? Just a warning maybe?

marc-hb (commented):
Sorry I got confused because I get confused every time I look at this really bad area of the code. To help me (and everyone else) get less confused next time please add this:

```
# Should _background_ log collection be disabled?
# This test should be called background_logger_disabled()
# Once-off log collection at the end of the test is not affected.
logger_disabled()
```


I still don't see why this patch is needed; what problem does it solve and in which configuration?

If you find the message confusing the please change it to `'.ldc dictionary file not found, background SOF logs collection disabled'`

In the future we will have complete, background logs collection with Zephyr too.

keqiaozhang:
> This would also disable mtrace. How did you test this?

Why would it also disable mtrace? I checked it on CAVS 2.5 IPC4 platforms, mtrace works well without ldc file and CI tests passed without side-effects.


>  What problems does this solve? Just a warning maybe?

sof-logger only work on IPC3 platforms and corresponding ldc files are also required. But on IPC4 platforms, we use mtrace which doesn't require the ldc file. So if ldc file is missing on IPC4 platforms, then sof log collection will be disabled during the test.
``` 
.ldc dictionary file not found, SOF logs collection disabled
```

So this PR will skip checking the ldc file on IPC4 platforms to make sure mtrace works properly.

keqiaozhang:
Moreover, if no ldc file, check-sof-logger.sh will also fail on IPC4 platforms, it has a hard dependency. Please correct me, but I think ldc file has nothing to do with IPC4 zephyr firmware.

marc-hb:
Just remove that bit entirely. I wanted this "feature" (see #811) but no one else requested it so I bet no one else is using it. Everyone is probably using SOF_LOGGING=none. Or rather: everyone is using mtrace because no one works on stable-v2.2.

So just remove that bit. We need the logging logic in sof-test to become simpler, not more complicated.

marc-hb:
> But on IPC4 platforms, we use mtrace which doesn't require the ldc file.

OK but:
1) FOR NOW the Zephyr build script still builds the .ldc files for all platforms for consistency with non-Intel Zephyr IPC3 and for simplicity
2) you should always deploy the ENTIRE build-sof-staging which includes FOR NOW the (useless on Intel) .ldc file.

In other words: if you're missing the useless .ldc file right now, you have a deployment bug that you should also look into. I'm OK with removing that code because we will stop building .ldc in the future but right now this PR is trying to hide a bug in your deployment code. Not good.

keqiaozhang:
I think we have 3 scenarios now.
1. no logger mode, we can continue to use "SOF_LOGGING=none"
2. IPC3-XTOS platforms, these platforms use sof-logger for logging.  e.g. stable-v2.2 is LTS, we need to support it.
3. IPC4 Zephyr platforms, these platforms use mtrace for logging, no need to deploy the .ldc file. 

So the right approach is to simplify our logic and only consider these 3 scenarios. 
New deployment will deploy the ENTIRE build-sof-staging to DUT. Moving on, we should stop checking the ldc file on IPC4 Zephyr platforms. 

I will try to simplify the sof-test logic for logging and move the .ldc dependency in check_sof-logger.sh.

marc-hb:
> I think we have 3 scenarios now.

You dropped non-Intel IPC3 Zephyr on the main branch.

> IPC4 Zephyr platforms, these platforms use mtrace for logging, no need to deploy the .ldc file.

Deployment MUST always deploy complete packages. Deployment must never mess with individual files. Deployment must know nothing about individual files.

> New deployment will deploy the ENTIRE build-sof-staging to DUT. 

Great.

> I will try to simplify the sof-test logic for logging and move the .ldc dependency in check_sof-logger.sh.

No, the logging test should not have a different logic. The logging test should only inspect the SAME logs as the other tests, it should just keep searching for different things but in the SAME log files. 

> So just remove that bit. We need the logging logic in sof-test to become simpler, not more complicated.

Does this solve your current problem? Whatever it is.

keqiaozhang:
I just want to get ride of ldc file on IPC4 platforms. 
Now sof-logger check highly dependent on ldc file. The test script will first check whether the ldc file exists, if ldc file does not exist, sof-logger check fails. But mtrace doesn't need it.  It doesn't make sense. 

I entirely agree that we must deploy complete packages. But it cannot solve all problems.
e.g. how to deal ADL-IPC4 platform? Rename the sof-tgl.ldc to sof-adl.ldc after package deployment?

So my idea is:
1. deploy complete packages, but do not check the ldc files on IPC4 platforms.
2. use `is_firmware_file_zephyr` instead of` is_zephyr` on IPC4 platforms, use firmware binary to determine if it is a zephyr firmware, not use ldc file. In this way, we don't need to change when we stop building .ldc on main branch in the future.
3. Remove the ldc file dependencies in `check-sof-logger.sh` for IPC4 platforms. Check if the LDC file exists only on the IPC3 platforms.

keqiaozhang:
I have submitted a PR https://github.com/thesofproject/sof-test/pull/1065 based on my above comment.

marc-hb:
superseded by newer #1065, close?

PR #1057, opened 2023-06-14
lib: start polling for fw boot complete only when sof card is not pre…

fredoh9:
…sent

SOF soundcard is already listed it is ready to start the test. If not, set the KERNEL_CHECKPOINT and start polling for FW boot complete. This is a bug fix for fceeee02a95eece3f033bd23cdff37fa91714640.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -77,17 +77,19 @@ start_test()
         return 0
     }
 
-    # Check for whether SOF fW is loaded or not before starting any test
-    # without --since=@"$KERNEL_CHECKPOINT", it will be scan for current boot
-    # fw loaded -> sof module removed case can't be detected.
-    # Mainly it is going to check whether the sof-test can be started after boot/reboot
+    # Check for whether SOF fW is loaded or not before starting any test.
+    # Only start the polling for firmware boot complete when SOF soundcard is not available
+    # setup_kernel_check_point has already -1 second to avoid TOCTOU race condition
+    is_sof_used || {
     if [ -z "$NO_POLL_FW_LOADING" ]; then
-        if poll_wait_for 1 "$MAX_WAIT_FW_LOADING" sof_firmware_boot_complete; then
+        setup_kernel_check_point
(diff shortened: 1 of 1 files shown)


fredoh9:
without --since=@"$KERNEL_CHECKPOINT" is the biggest the problem from fceeee02a95eece3f033bd23cdff37fa91714640

marc-hb:
Summary of an off-line review with @fredoh9 :
- There's a tiny TOCTOU race window between `is_sof_used()` and `setup_kernel_check_point`. But @fredoh9 found that there's already a -1 in `setup_kernel_check_point`, so we're good. Just need a comment.
- By fixing a bug, this PR makes it impossible to design any test where the modules are not already loaded. Fred will add an escape.

fredoh9:
There is no TOCTOU race condition problem, because setup_kernel_check_point() has -3 or -1 second

marc-hb:
> There is no TOCTOU race condition problem, because setup_kernel_check_point() has -3 or -1 second

That's a bit of hack :-) It's never going to happen but in computer science theory you could have things like:
- the `lib.sh` process being suspended by the OS for more than 1 second, or
- the firmware being UNloaded just after the -3 seconds and before the `is_sof_used()` `grep`.

Current code is good enough in practice but please add a comment, something like "TOCTOU race avoided thanks to -1 second inside `setup_kernel_check_point()`

fredoh9:
Note, NO_POLL_FW_LOADING variable is already available to skip this polling

marc-hb (approved):

fredoh9:
Looks good for device test results also. Merging now

marc-hb:
Here's an interesting side-effect.

https://github.com/thesofproject/sof/pull/7698  broke the MTL build. Tests were started anyway without firmware (tracked in internal issue 139). Each test took 70s to timeout in https://sof-ci.01.org/sofpr/PR7698/build9948/devicetest

The whole PR test plan uselessly hogged 3 MTL devices for more than 200 minutes.

PR #1064, opened 2023-06-19
tools: add the initial version of trace analyzer

aiChaoSONG:
This patch adds the initial version of SOF mtrace
analyzer. The initial version of the analyzer is
able to output the performance statistics per module, for example:
```
                     COMP_NAME  COMP_ID     AVG_OF_CPU_AVG   AVG_OF_CPU_PEAK  PEAK_OF_CPU_AVG  PEAK_OF_CPU_PEAK
                    mixout.2.1  1-0x30000   2.29             3.12             2.30             12.67          
                      gain.2.1  1-0x60001   2.58             3.38             2.69             9.43           
dai-copier.HDA.Analog.playback  1-0x40001   4.78             4.90             4.79             4.96           
        host-copier.0.playback  0-0x40000   4.31             4.47             4.31             4.91           
                      gain.1.1  0-0x60000   2.55             3.71             2.71             9.52           
                     mixin.1.1  0-0x20000   3.56             3.65             3.59             3.75
```
mtrace file used in above example: 
[playback_mtrace1.txt](https://github.com/thesofproject/sof-test/files/12066989/playback_mtrace1.txt)
dmesg: 
[playback_dmesg1.txt](https://github.com/thesofproject/sof-test/files/12066991/playback_dmesg1.txt)

run below command with above mtrace and dmesg:
`./sof-perf-analyzer.py playback_mtrace1.txt --dmesg=playback_dmesg1.txt`

If no dmesg is provided, component name will be missing:
```
                     COMP_NAME  COMP_ID     AVG_OF_CPU_AVG   AVG_OF_CPU_PEAK  PEAK_OF_CPU_AVG  PEAK_OF_CPU_PEAK
                          None  1-0x30000   2.29             3.12             2.30             12.67          
                          None  1-0x60001   2.58             3.38             2.69             9.43           
                          None  1-0x40001   4.78             4.90             4.79             4.96           
                          None  0-0x40000   4.31             4.47             4.31             4.91           
                          None  0-0x60000   2.55             3.71             2.71             9.52           
                          None  0-0x20000   3.56             3.65             3.59             3.75           

```

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -1,0 +1,280 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2023 Intel Corporation. All rights reserved.
+
+'''
+The sof_perf_analyzer helps to analyze SOF trace output generated during the running of
+aplay and arecord commands or a test case in sof-test repository. It takes human readable
+trace file (contents from plain mtrace or the decoded dictionary trace) as input, and
+output below analysis results:
+    - Average and maximum MCPS of cpu average and cpu peak per SOF audio component
+
+Example of performance logging:
+[    4.612041] <inf> component: comp_copy: comp:1 0x40001 perf comp_copy samples 48 period
+ 1000 cpu avg 459 peak 473
+
+There is no audio component name information in SOF trace, so auxiliary linux kernel log
+is used to extract component name.
+
+Hardcoded value explained:
+DSP_CLK: The DSP clock frequency in Hz.
+DSP_TIMER: The DSP timer clock frequency in Hz.
+'''
+
+import re
+import pathlib
+import argparse
+from typing import TextIO
+from typing import Optional
+from typing import Generator
+from dataclasses import dataclass
+
+@dataclass()
+class TraceItem:
+    '''The structural representation for a single, parsed line of trace'''
+    timestamp: float
+    level: str
+    context: str # the trace context registered with LOG_MODULE_REGISTER
+    func: str
+    # The user message filled to log functions, exclude timestamp, log level,
+    # log context and log function, which is prepended by zephyr logging system
+    msg: str
+
+@dataclass(frozen=True)
+class Component:
+    '''The identifier for a SOF audio component'''
+    ppln_id: str
+    comp_id: str
+
+@dataclass(frozen=True)
+class CompPerfSample:
+    '''The dataclass for SOF audio component performance info'''
+    timestamp: float
+    samples: int
+    period: int
+    cpu_avg: int
+    cpu_peak: int
+
+@dataclass(frozen=True)
+class CompPerfStats:
+    '''The dataclass for SOF audio component statistics'''
+    avg_cpu_avg: float   # average of cpu average
+    avg_cpu_peak: float  # average of cpu peak
+    peak_cpu_avg: float  # peak of cpu average
+    peak_cpu_peak: float # peak of cpu peak
+
+# Currently, we don't have test case to test low-power mode, in which
+# the DSP may not operate in the maximum clock frequency. Let't hardcode
+# this value to the maximum clock frequency for now.
+DSP_CLK = 400000000
+# The DSP timer clock frequency is fixed despite of DSP_CLK
+DSP_TIMER = 38400000
+UINT32_MAX = 4294967295
+AUDIO_PERIOD = 1000
+
+TICK_TO_MCPS = DSP_CLK / DSP_TIMER / AUDIO_PERIOD
+
+# Once DSP timer count to UINT32_MAX, it wrapped to zero, which causes
+# the timestamp wrap to zero, too. This ts_shift is used to correct
+# the timestamp value on wrap. UINT32_MAX / DSP_TIMER is added to it at
+# every wrap.
+#
+# pylint: disable=C0103
+ts_shift = 0
+# pylint: disable=C0103
+args = None
+
+# This map gives access to the list of component performance samples
+# collected for a given component
+perf_info: dict[Component, list[CompPerfSample]] = {}
+perf_stats: dict[Component, CompPerfStats] = {}
+component_name: dict[Component, str] = {}
+
+TraceItemGenerator = Generator[TraceItem, None, None]
+
+def collect_perf_info(trace_item: TraceItem):
+    '''Parse and collect performace trace information(msg field of TraceItem) to a
+    Component->list[CompPerfSample] mapping for further analysis.
+    '''
+    msg = trace_item.msg.split()
+    ppln_id = msg[0].split(':')[1]
+    perf_val = int(msg[5]), int(msg[7]), int(msg[10]), int(msg[12])
+    perf_sample = CompPerfSample(trace_item.timestamp, *perf_val)
+    comp_perf_info_list = perf_info.setdefault(Component(ppln_id, msg[1]), [])
+    comp_perf_info_list.append(perf_sample)
+
+def dispatch_trace_item(trace_item: TraceItem):
+    '''Dispatch trace item to cosponding trace collecting function. In a TraceItem,
+    we have log timestamp, log level, log context and log function, with them, dispatch
+    could be easily implemented.
+    '''
+    if trace_item.func == 'comp_copy':
+        collect_perf_info(trace_item)
+
+def skip_to_first_trace(trace_item_gen: TraceItemGenerator) -> Optional[TraceItem]:
+    '''The current sof-test test case may collect some traces belonging to previous
+    test case due to mtrace is configured in deferred mode. This function consumes
+    those traces from the generator, and return the first trace item of current test.
+    '''
+    try:
+        while item := next(trace_item_gen):
+            # On test running, the SOF firmware is reloaded to DSP, timer is reset to 0.
+            # The first trace must have a timestamp with integral part equals to 0.
+            if int(item.timestamp) == 0:
+                return item
+    except StopIteration:
+        return None
+    return None
+
+def make_trace_item(fileio: TextIO) -> TraceItemGenerator:
+    '''Filter and parse a line of trace in string form into TraceItem object, for example:
+    '[    2.566046] <inf> component: comp_copy: comp:0 0x40000 perf comp_copy samples
+    48 period 1000 cpu avg 413 peak 425' -> TraceItem(timestamp=2.566046, level='inf',
+    context='component', func='comp_cppy', msg='comp:0 0x40000 perf comp_copy samples
+    48 period 1000 cpu avg 413 peak 425')
+
+    In this function, we do the first stage filtering and parsing. The firmware trace file
+    contains some information that is not quite formated, for example:
+    '*** Booting Zephyr OS build v3.4.0-rc2 ***', those messages are and dropped. Other messages
+    with timestamp, log level, log context, log function are parsed to a TraceItem for timestamp
+    correction and second stage filtering and parsing.
+    '''
+    for line in fileio:
+        # Filter extra lines that are not formal traces, for example, the banner,
+        # which don't contain timestamp and trace level.
+        if match_obj := re.search(r'\[.+\] <(dbg|inf|wrn|err)>', line):
+            # Sometimes, A trace output may be incomplete and mixed with next
+            # trace output in a single line, for example: '[    0.071590] <inf>
+            # pipe[    0.071751] <inf> host_comp: host_get_copy_bytes_normal: comp:1
+            # 0x40003 no bytes to copy, available samples: 0, free_samples: 384'.
+            # Under this circumstance, we ignore the first trace and extract the
+            # second trace from the line. This is done by regarding the end position
+            # of the matched substring as sentinel, timestamp and trace level are
+            # both some specific offset from the sentinel.
+            span_end_pos = match_obj.span()[1]
+            trace_lvl = line[span_end_pos - 4: span_end_pos - 1]
+            timestamp = float(line[span_end_pos - 19: span_end_pos - 7].strip())
+
+            # The rest after removing timestamp and log level
+            rest = line[span_end_pos + 1:].split(': ')
+            ctx, func = rest[0:2]
+            msg = ': '.join(rest[2:]).strip()
+            yield TraceItem(timestamp, trace_lvl, ctx, func, msg)
+
+def process_trace_file():
+    '''The top-level caller for processing the trace file'''
+    with open(args.filename, 'r', encoding='utf8') as file:
+        trace_item_gen = make_trace_item(file)
+        trace_prev = skip_to_first_trace(trace_item_gen)
+        if trace_prev is None:
+            raise Exception('No valid trace in provided file')
+        for trace_curr in trace_item_gen:
+            # pylint: disable=W0603
+            global ts_shift
+            old_ts_shift = ts_shift
+            # On wrap happened, the timestamp of current trace should be much more smaller
+            # than the previous one. In practice, it is possible that the timestamp of
+            # current trace is slightly smaller than the previous one, this could be a
+            # bug in SOF. Add a 50s shift to make sure timestamp correction work properly.
+            if trace_curr.timestamp < trace_prev.timestamp - 50:
+                ts_shift = ts_shift + UINT32_MAX / DSP_TIMER
+            trace_prev.timestamp += old_ts_shift
+            dispatch_trace_item(trace_prev)
+            trace_prev = trace_curr
+        trace_prev.timestamp += ts_shift
+        dispatch_trace_item(trace_prev)
+
+def process_kmsg_file():
+    '''Process the dmesg to get the component ID to component name mapping,
+    component name is acquired from the line that contains 'Create widget',
+    component ID is acquired from the next line. Example:
+    [  334.818435] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
+    [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
+
+    In practice, sof-test only capture kernel message and firmware trace generated during a test
+    case run. Mostly in manual tests, if the kernel message file contains multiple firmware runs
+    with overlapping information, the last one wins.
+    '''
+    with open(args.kmsg, encoding='utf8') as f:
+        ppln_id = None
+        for line in f:
+            if match_obj := re.search(r'Create widget', line):
+                span_end_pos = match_obj.span()[1]
+                line_split = line[span_end_pos + 1:].split()
+                widget_name = line_split[0]
+                # In the linux kernel, IDA is used to allocated pipeline widget instance ID,
+                # this ID later is used for pipeline creation, thus becomes pipeline ID in the
+                # firmware. Note that ppln_id variable will be assigned properly at pipeline widget
+                # creation, because it is always the first one to be created before all other
+                # widgets in the same pipeline and pipelines are created sequentially.
+                if widget_name.startswith('pipeline'):
+                    ppln_id = line_split[2]
+                next_line = next(f)
+                widget_id = next_line.split('|')[0].split(':')[-1].strip()
+                # convert to the same ID format in mtrace
+                widget_id = '0x' + widget_id[:-6:-1]
+                component_name[Component(ppln_id, widget_id)] = widget_name
+
+def analyze_perf_info():
+    '''Calculate performance statistics from performance information'''
+    for comp, perf_info_list in perf_info.items():
+        len_perf_info = len(perf_info_list)
+        avg_cpu_avg = sum(e.cpu_avg for e in perf_info_list) / len_perf_info
+        avg_cpu_peak = sum(e.cpu_peak for e in perf_info_list) / len_perf_info
+        peak_cpu_avg = max(e.cpu_avg for e in perf_info_list)
+        peak_cpu_peak = max(e.cpu_peak for e in perf_info_list)
+        perf_stats[comp] = CompPerfStats(avg_cpu_avg, avg_cpu_peak, peak_cpu_avg, peak_cpu_peak)
+
+def print_perf_info():
+    '''Output SOF trace performance statistics'''
+    if len(perf_stats):
+        max_name_len = max(len(name) for name in component_name.values())
+        name_fmt = '{:>' + f'{max_name_len}' + '},'
+        title_fmt = name_fmt + ' {:>10}, {:>15}, {:>15}, {:>15}, {:>16}'
+        title = title_fmt.format('COMP_NAME', 'COMP_ID', 'AVG_OF_CPU_AVG', 'AVG_OF_CPU_PEAK',
+                             'PEAK_OF_CPU_AVG', 'PEAK_OF_CPU_PEAK')
+        print(title)
+
+        stats_fmt = name_fmt + ' {:>10}, {:>15,.2f}, {:>15,.2f}, {:>15,.2f}, {:>16,.2f}'
+        for comp, perf in perf_stats.items():
+            avg_cpu_avg_mcps = perf.avg_cpu_avg * TICK_TO_MCPS
+            avg_cpu_peak_mcps = perf.avg_cpu_peak * TICK_TO_MCPS
+            peak_cpu_avg_mcps = perf.peak_cpu_avg * TICK_TO_MCPS
+            peak_cpu_peak_mcps = perf.peak_cpu_peak * TICK_TO_MCPS
+
+            comp_name = component_name.get(comp, 'None')
+            comp_id = f'{comp.ppln_id}-{comp.comp_id}'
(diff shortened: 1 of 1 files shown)


btian1:
Possible to display component name instead of only display comp ID?
there is a table in dmesg log, with dmesg | grep module

Also there are task(each pipeline) and core performance data as well, please also add it and display together.
pipeline: [    9.584976] <inf> ll_schedule: do_task_run: perf_cycle task 0xa01106c0, 0x201b0U cpu avg 1518 peak 4098
core: [    9.583993] <inf> ll_schedule: zephyr_domain_thread_fn: ll timer avg 3428, max 6624, overruns 0

and for memory:
[    0.369206] <inf> mem_allocator: heap_alloc_aligned: heap allocated: 31368 free: 819032 max allocated: 31368

The last line memory log shows current max allocation.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+            self._perf_stats[comp] = CompPerfStats(cpu_avg, cpu_max)
+
+    def summarize(self):
+        '''Output SOF trace analysis summary'''
+        if len(self._perf_stats):
+            print('COMP\t\tCPU_AVG\t\tCPU_PEAK')
+            for comp, perf in self._perf_stats.items():
+                cpu_avg_mcps = perf.cpu_avg * TICK_TO_MCPS
+                cpu_peak_mcps = perf.cpu_peak * TICK_TO_MCPS
+                print(f'{comp.comp_id}\t\t{cpu_avg_mcps:.2f}\t\t{cpu_peak_mcps:.2f}')
```
marc-hb, line 62: https://docs.python.org/3/howto/logging.html

singalsu:
Could this kind of CSV output format be possible, for spreadsheet and database export. Also for easy sum all the average and max values. Just an example, no need for exactly like this:

```
                             Name,        Mtrace id, avg MCPS, max MCPS
            host-copier.0.playback,   comp:0 0x40000,     3.55,     6.85
                          gain.1.1,   comp:0 0x60000,     2.85,     8.73
                         mixin.1.1,   comp:0 0x20000,     3.02,     3.26
                        mixout.2.1,   comp:1 0x30000,     2.27,     2.83
                          gain.2.1,   comp:1 0x60001,     2.86,    16.29
                         eqiir.2.1,   comp:1 0xa0000,     2.08,     2.20
                         eqfir.2.1,   comp:1 0xb0000,     1.98,     9.84
                           drc.2.1,   comp:1 0xf0000,     3.22,     3.41
    dai-copier.HDA.Analog.playback,   comp:1 0x40001,     4.18,     4.46
                                  ,                 ,    26.00,    57.88
```

The association of mtrace ids with kernel output is quite tricky. The "comp:a"  number also matters in finding the right component name, not just the hex number.

aiChaoSONG:
@btian1 @singalsu  Your requests is reasonable, and I will add in further updates. 

@marc-hb  do we really need logging in such a script with only one or two prints?

Could you please review the code? Then Keqiao can enable performance build, and integrate this script into test case.  So that we could have a simple working sof-test that reports performance statistics.

Later I will add more features:
    - parse dmesg for component names
    - output csv or any format we need
    - draw a graph for the component MCPS vs time
    - ...
In the initial version, I don't want to make the PR too big, which make it hard to review.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+
+CPU_CLK = 400000000
+CPU_TIMER = 38400000
+AUDIO_PERIOD = 1000
+TICK_TO_MCPS = CPU_CLK / CPU_TIMER / AUDIO_PERIOD
+
+TraceItem = namedtuple('TraceItem', ['timestamp', 'level', 'context', 'func', 'msg'])
+CompId = namedtuple('CompId', ['ppl_id', 'comp_id'])
+CompPerfInfo = namedtuple('CompPerfInfo', ['timestamp', 'samples', 'period', 'cpu_avg', 'cpu_peak'])
+CompPerfStats = namedtuple('CompPerfStats', ['cpu_avg', 'cpu_peak'])
```
marc-hb, line 22: Take a look at https://docs.python.org/3/library/dataclasses.html, I started using them in `xtensa-build-zephyr.py` and I've been very happy with them. They may be better for some of these.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+    The TraceAnalyzer class for SOF trace analysis, provides statics for below items:
+        - SOF component performance
+    '''
+    def __init__(self):
+        self._perf_info = {}
+        self._perf_stats = {}
+
+    def _add_perf_trace(self, trace_item: TraceItem):
+        # Example of msg: ['comp:0', '0x20000', 'perf', 'comp_copy', 'samples', '48',
+        # period', '1000', 'cpu', 'avg', '322', 'peak', '333']
```
marc-hb, line 35: Can you give the _unparsed_ example instead? Otherwise it's impossible to review the parsing code.

Also, if the input format changes then the unparsed example can be compared to the failing input.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+TICK_TO_MCPS = CPU_CLK / CPU_TIMER / AUDIO_PERIOD
+
+TraceItem = namedtuple('TraceItem', ['timestamp', 'level', 'context', 'func', 'msg'])
+CompId = namedtuple('CompId', ['ppl_id', 'comp_id'])
+CompPerfInfo = namedtuple('CompPerfInfo', ['timestamp', 'samples', 'period', 'cpu_avg', 'cpu_peak'])
+CompPerfStats = namedtuple('CompPerfStats', ['cpu_avg', 'cpu_peak'])
+
+class TraceAnalyzer:
+    '''
+    The TraceAnalyzer class for SOF trace analysis, provides statics for below items:
```
marc-hb, line 26: ?
```suggestion
    The TraceAnalyzer class for SOF trace analysis, provides statistics for below items:
```

Looks like the start of a --help message.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+CPU_TIMER = 38400000
+AUDIO_PERIOD = 1000
+TICK_TO_MCPS = CPU_CLK / CPU_TIMER / AUDIO_PERIOD
+
+TraceItem = namedtuple('TraceItem', ['timestamp', 'level', 'context', 'func', 'msg'])
+CompId = namedtuple('CompId', ['ppl_id', 'comp_id'])
+CompPerfInfo = namedtuple('CompPerfInfo', ['timestamp', 'samples', 'period', 'cpu_avg', 'cpu_peak'])
+CompPerfStats = namedtuple('CompPerfStats', ['cpu_avg', 'cpu_peak'])
+
+class TraceAnalyzer:
```
marc-hb, line 24: Is this always going to be a singleton? If yes then I don't understand why it's a `class` at all. A couple globals and a few functions would be simpler.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2023 Intel Corporation. All rights reserved.
+
+'''
+A tool to analyze SOF trace file.
```
marc-hb, line 7: Please this a little bit longer

Input
- Trace from which tool? (we have too many). If mtrace then rename to `mtrace_perf_analyzer.py` or something more specific.
- What kind of data does it analyze?
- Over which period of time / what kind of input? Does it support streaming or just whole file? Maybe streaming in the future? (I saw a `yield`)

Output
- what kind of analysis does this do? Sliding window average? etc.
- There are some hardcoded CLK and other values: they deserve to be mentioned in the --help, at least briefly.

tools/trace_analyzer.py
```diff
@@ -0,0 +1,98 @@
...
+        if len(self._perf_stats):
+            print('COMP\t\tCPU_AVG\t\tCPU_PEAK')
+            for comp, perf in self._perf_stats.items():
+                cpu_avg_mcps = perf.cpu_avg * TICK_TO_MCPS
+                cpu_peak_mcps = perf.cpu_peak * TICK_TO_MCPS
+                print(f'{comp.comp_id}\t\t{cpu_avg_mcps:.2f}\t\t{cpu_peak_mcps:.2f}')
+
+    @staticmethod
+    def make_trace_item(file) -> TraceItem:
+        '''Convert a line of trace into TraceItem object'''
```
marc-hb, line 66: Please provide an example of an unparsed input line, see above.

tools/trace_analyzer.py
marc-hb: Sorry I didn't realize this was the main (and only) output.

marc-hb (commented):
> In the initial version, I don't want to make the PR too big, which make it hard to review.

That's very good, but...

> @marc-hb do we really need logging in such a script with only one or two prints?

Did you look a the example? Python's logging module is built-in and requires less than 5 lines of configuration at the top. It's much harder to add later when there are many `print` to replace, causing git conflicts.

Also, it should be possible ASAP to make the different between logging and actual output. For instance: logging to stdout and output to a file. No need to add this now but need to start with something reasonable flexible like python's built-in logging module.

This being said, I think you don't have any logging yet, just data output. So ignore `logging` for now.

> Later I will add more features:
> - output csv or any format we need

You need to think about the output format(s) right now, that's the most important design decision and the one that is very hard to change once the script has any user which is... tomorrow!

> Then Keqiao can enable performance build, and integrate this script into test case. 


http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2878263
http://www.catb.org/~esr/writings/taoup/html/textualitychapter.html
etc.

tools/trace_analyzer.py
aiChaoSONG: dataclass, namedtuple and typing.NamedTuple are all factory classes,in our use cases, we don't require the flexibility provided by dataclass, namedtuple and typed.NamedTuple both are good for our use. We could use it for future flexibility, but for now, I will add `frozen=true`

marc-hb: I didn't suggest dataclass because it's more flexible. I mentioned it because `namedtuple()` is older and IMHO less readable.

Some interesting stuff at:
https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases

>  We could use it for future flexibility,

So you will not switch to dataclass?

> but for now, I will add `frozen=true`

So you will switch to dataclass?

I'm confused.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+            stat = stats_fmt.format(comp_name, comp_id, avg_cpu_avg_mcps, avg_cpu_peak_mcps,
+                                    peak_cpu_avg_mcps, peak_cpu_peak_mcps)
+            print(stat)
+
+def parse_args():
+    '''Parse command line arguments'''
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                     description=__doc__)
+    parser.add_argument('filename')
+    parser.add_argument('--dmesg', type=pathlib.Path, required=False,
```
marc-hb, line 232: Don't use the name `dmesg`. `dmesg` is just one of the ways to access `kmsg` kernel logs and it's probably the worst way because it's a limited size ring buffer where logs expire. It also has no timestamp, not hostname and no log level. Also make sure this script is compatible with `kern.log` and `journalctl -k`. Being compatible with `dmesg` is fine but it should never look like it recommends `dmesg`.

```suggestion
    parser.add_argument('--kmsg', type=pathlib.Path, required=False,
```

marc-hb (changes requested):

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    try:
+        while item := next(trace_item_gen):
+            if int(item.timestamp) == 0:
+                return item
+    except StopIteration:
+        return None
+    return None
+
+def make_trace_item(fileio: TextIO) -> TraceItemGenerator:
+    '''Convert a line of trace in string form into TraceItem object, for example:
```
marc-hb, line 118: Nit:
```suggestion
    '''Parse a line of trace in string form into TraceItem object, for example:
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+        CompPerfInfo(trace_item.timestamp, *perf_val)
+        )
+
+def dispatch_trace_item(trace_item: TraceItem):
+    '''Dispatch trace item to cosponding analyzing function'''
+    if trace_item.func == 'comp_copy':
+        collect_perf_info(trace_item)
+
+def get_first_valid_trace(trace_item_gen: TraceItemGenerator) -> Optional[TraceItem]:
+    '''The current sof-test test case may collect some traces belongs to previous
```
marc-hb, line 105: Nit
```suggestion
    '''The current sof-test test case may collect some traces belonging to previous
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    perf_info.setdefault(Comp(ppl_id, msg[1]),[]).append(
+        CompPerfInfo(trace_item.timestamp, *perf_val)
+        )
+
+def dispatch_trace_item(trace_item: TraceItem):
+    '''Dispatch trace item to cosponding analyzing function'''
+    if trace_item.func == 'comp_copy':
+        collect_perf_info(trace_item)
+
+def get_first_valid_trace(trace_item_gen: TraceItemGenerator) -> Optional[TraceItem]:
```
marc-hb, line 104:
```suggestion
def skip_to_first_trace(trace_item_gen: TraceItemGenerator) -> Optional[TraceItem]:
```

The simple code in this low-level function does not seem to have any high level concept of what is "valid" or not. Keep its name as simple as what is code does - and more accurate.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    '''The current sof-test test case may collect some traces belongs to previous
+    test case due to mtrace is configured in deferred mode. This function consumes
+    those invalid traces from the generator, and return the first valid trace item.
+    '''
+    try:
+        while item := next(trace_item_gen):
+            if int(item.timestamp) == 0:
```
marc-hb, line 111: Maybe a very short comment explaining why there is always a 0 timestamp (as opposed to 0.01)

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+from typing import Dict
+from typing import List
+from typing import TextIO
+from typing import Optional
+from typing import Generator
+from dataclasses import dataclass
+
+@dataclass()
+class TraceItem:
+    '''The structural representation for a single line of trace'''
```
marc-hb, line 30: Nit
```suggestion
    '''The structural representation for a single, parsed line of trace'''
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    timestamp: float
+    level: str
+    context: str # the trace context registered with LOG_MODULE_REGISTER
+    func: str
+    msg: str
+
+@dataclass(frozen=True)
+class Comp:
```
marc-hb, line 38: Nit:
```suggestion
class Component:
```
Really not much longer and so much more readable.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    '''The identifier for a SOF component'''
+    ppl_id: str
```
marc-hb, line 40: I would just go for `pipeline_id` but OK, this is used a lot. Please include at least 1 more letter = all the consonants.
```suggestion
    ppln_id: str
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+        dispatch_trace_item(trace_prev)
+
+def process_dmesg_file():
+    '''Process the dmesg to get the component ID to component name mapping,
+    component name is acquired from the line that contains 'Create widget',
+    component ID is acquired from the next line. Example:
+    [  334.818435] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
+    [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]'''
```
marc-hb, line 175: What happens when the input given to this function is made of multiple FW runs with overlapping and possibly conflicting information? Fails or last one wins? This needs to be at least documented.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+        if trace_prev is None:
+            raise Exception('No valid trace in provided file')
+        for trace_curr in trace_item_gen:
+            # pylint: disable=W0603
+            global ts_shift
+            old_ts_shift = ts_shift
+            # On wrap happened, the timestamp of current trace should be much more smaller
+            # than the previous one. In practice, it is possible that the timestamp of
+            # current trace is slightly smaller than the previous one, this could be a
+            # bug in SOF. Add a 50s shift to make sure timestamp correction work properly.
```
marc-hb, line 159: How can we trust performance information provided by the firmware if timestamps are wrong?

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    msg: str
+
+@dataclass(frozen=True)
+class Comp:
+    '''The identifier for a SOF component'''
+    ppl_id: str
+    comp_id: str
+
+@dataclass(frozen=True)
+class CompPerfInfo:
```
marc-hb, line 44:
```suggestion
class CompPerfSample:
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+ts_shift = 0
+# pylint: disable=C0103
+args = None
+
+# The dictionary for performance information
+perf_info: Dict[Comp, List[CompPerfInfo]] = {}
+# The dictionary for performace statistics
+perf_stats: Dict[Comp, CompPerfStats] = {}
+# Component ID to name map acquired from dmesg
+comp_name_map: Dict[Comp, str] = {}
```
marc-hb, line 86: Most of the text in the comments merely duplicates the names.

```python
# This map gives access to the list of samples collected for a given component
perf_info: Dict[Comp, List[CompPerfInfo]] = {}
```

A `_map` suffix is almost never useful because it's almost always obvious from the definition and usage.

```python
components_name: Dict[...

...

components_name[my_component] ...
```

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+
+TraceItemGenerator = Generator[TraceItem, None, None]
+
+def collect_perf_info(trace_item: TraceItem):
+    '''Collect trace information in mtrace to a Comp->List[CompPerfInfo] mapping'''
```
marc-hb, line 91: This is a second level of parsing. It's not immediately obvious that your parsing is split into two different steps, that should be easy to find in the comments but I had to read and understand the code (I hope!) to see that.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+    msg = trace_item.msg.split()
+    ppl_id = msg[0].split(':')[1]
+    perf_val = int(msg[5]), int(msg[7]), int(msg[10]), int(msg[12])
+    perf_info.setdefault(Comp(ppl_id, msg[1]),[]).append(
```
marc-hb, line 95: It's confusing to find a .setdefault() and an .append() in function where most of the code is parsing. They're logically very different tasks so please keep them separate.
First I read the name "collect_perf_info" so I do expect some .append(). Then I see parsing code so I think the function is misnamed. Then I finally see some `.append()` at the very end.  It's hard to follow, each function should perform only one logical task.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+from dataclasses import dataclass
+
+@dataclass()
+class TraceItem:
+    '''The structural representation for a single line of trace'''
+    timestamp: float
+    level: str
+    context: str # the trace context registered with LOG_MODULE_REGISTER
+    func: str
+    msg: str
```
marc-hb, line 35: I found this `msg` field to be the most confusing part of the design. Maybe it's OK but it needs at least more comments.

This structure represents a parsed one line of logs / messages. So why is a field still called "msg"? Is this the entire line kept for recording purposes?

The purpose of this script is to extract performance numbers. Yet there is no performance number in a parsed line of logs?

See other comment about your 2-steps parsing below.

tools/sof-perf-analyzer.py
```diff
@@ -0,0 +1,253 @@
...
+                return item
+    except StopIteration:
+        return None
+    return None
+
+def make_trace_item(fileio: TextIO) -> TraceItemGenerator:
+    '''Convert a line of trace in string form into TraceItem object, for example:
+    '[0.349880] <inf> pipe: pipeline_connect: comp:0 0x40000 connect buffer 0 as sink' ->
+    TraceItem(timestamp=0.349880, level='inf', context='pipe', func='pipeline_connect',
+              msg='comp:0 0x40000 connect buffer 0 as sink')
```
marc-hb, line 121: So the entire purpose of this script is to parse performance information from the logs but there's no obvious performance information in the log example...?

Also, please have another (or the same) example like this at the very top.

marc-hb (changes requested):
I think the general structure looks OK but the code could be much easier to read, review and maintain with a fair number of small changes.

tools/sof-perf-analyzer.py
aiChaoSONG: The timestamp is not wrong, it just wrapped if a test is running too long (> 111.85 seconds). That why we do timestamp correction, add 111.85 seconds at every wrap.

keqiaozhang (approved):
LGTM

tools/sof_perf_analyzer.py
```diff
@@ -0,0 +1,278 @@
...
+        avg_cpu_peak = sum(e.cpu_peak for e in perf_info_list) / len_perf_info
+        peak_cpu_avg = max(e.cpu_avg for e in perf_info_list)
+        peak_cpu_peak = max(e.cpu_peak for e in perf_info_list)
+        perf_stats[comp] = CompPerfStats(avg_cpu_avg, avg_cpu_peak, peak_cpu_avg, peak_cpu_peak)
+
+def print_perf_info():
+    '''Output SOF trace performance statistics'''
+    if len(perf_stats):
+        title_fmt = '{:>30}  {:10}  {:<15}  {:<15}  {:<15}  {:<15}'
+        title = title_fmt.format('COMP_NAME', 'COMP_ID', 'AVG_OF_CPU_AVG', 'AVG_OF_CPU_PEAK',
```
singalsu, line 235: If a separator character is added the output format would be CSV that eases import to spreadsheet. In US or C locale the separator is comma character. Could this be changed?

aiChaoSONG: sure, let's have this.

btian1:
What's the output looks like for now?
Also, please note, there is another pr to adding peak mcps counter: #8008

aiChaoSONG:
Updates:  
- alignment will auto adjust according to the longest component name;
- add comma between fields to make it easy import to spreedsheet

```
                     COMP_NAME,    COMP_ID,  AVG_OF_CPU_AVG, AVG_OF_CPU_PEAK, PEAK_OF_CPU_AVG, PEAK_OF_CPU_PEAK
                    mixout.2.1,  1-0x30000,            2.29,            3.12,            2.30,            12.67
                      gain.2.1,  1-0x60001,            2.58,            3.38,            2.69,             9.43
dai-copier.HDA.Analog.playback,  1-0x40001,            4.78,            4.90,            4.79,             4.96
        host-copier.0.playback,  0-0x40000,            4.31,            4.47,            4.31,             4.91
                      gain.1.1,  0-0x60000,            2.55,            3.71,            2.71,             9.52
                     mixin.1.1,  0-0x20000,            3.56,            3.65,            3.59,             3.75
```

andrula-song (approved):

btian1 (commented):
these are all component performance data, perf log further provide pipe and core perf data:
pipe(task):
[   26.851691] <inf> ll_schedule: do_task_run: perf_cycle task 0xbe0ba700, 0x20180U cpu avg 1324 peak 3418
[   26.856726] <inf> ll_schedule: do_task_run: perf_cycle task 0xbe0bb480, 0x20180U cpu avg 1261 peak 3382

core:

[   27.875740] <inf> ll_schedule: zephyr_domain_thread_fn: ll timer avg 2800, max 5376, overruns 0

core also have overrun flags, you can also report this out as fatal error when it set to 1.

aiChaoSONG:
> these are all component performance data, perf log further provide pipe and core perf data: pipe(task): [ 26.851691] ll_schedule: do_task_run: perf_cycle task 0xbe0ba700, 0x20180U cpu avg 1324 peak 3418 [ 26.856726] ll_schedule: do_task_run: perf_cycle task 0xbe0bb480, 0x20180U cpu avg 1261 peak 3382
> 
> core:
> 
> [ 27.875740] ll_schedule: zephyr_domain_thread_fn: ll timer avg 2800, max 5376, overruns 0
> 
> core also have overrun flags, you can also report this out as fatal error when it set to 1.

For component performance measurement, this information is not that important. If you also want it, you can add it later to this script.

aiChaoSONG:
Merge for now to unlock further work.

PR #1065, opened 2023-06-20
Refine logger

keqiaozhang:


case-lib/lib.sh
```diff
@@ -741,29 +741,29 @@ is_ipc4()
...
+            }
+        else
+            ldcFile=$(find_ldc_file) || {
+                dlogi '.ldc dictionary file not found, SOF logs collection disabled'
+                return 0 # 0 is 'true'
+            }
+        fi
     fi
 
     return 1
```
greg-intel, line 769: An error condition should have a message stating what the error is before returning.

kv2019i (approved):
Thanks @keqiaozhang ! This will pave way to enable SysT logging later on -> https://github.com/thesofproject/sof/issues/7703

case-lib/lib.sh
```diff
@@ -741,29 +741,29 @@ is_ipc4()
...
     if [ "$SOF_LOGGING" == 'none' ]; then
         dlogi 'SOF logs collection globally disabled by SOF_LOGGING=none'
         return 0
-    fi
-
-    if is_ipc4 && ! is_firmware_file_zephyr; then
-        dlogi 'IPC4 FW logging only support with SOF Zephyr build'
-        dlogi 'SOF logs collection is globally disabled.'
-        return 0
+    else
```
lyakh, line 754: no need for this `else`

case-lib/lib.sh
```diff
@@ -741,29 +741,29 @@ is_ipc4()
...
-    ldcFile=$(find_ldc_file) || {
-        dlogi '.ldc dictionary file not found, SOF logs collection disabled'
-        return 0 # 0 is 'true'
-    }
 
     # Disable logging when available...
-    if [ ${OPT_VAL['s']} -eq 0 ]; then
-        return 0
-    fi
+    [ ${OPT_VAL['s']} -eq 0 ] && return 0
```
marc-hb, line 746: This is not needed and it looks incompatible with `set -e`. It's actually compatible with `set -e` but that's confusing.

It's also less readable for some people and distracting git churn in a too complex and too messy area of the code.

marc-hb (changes requested):
Thanks! the logic looks good, please fix the minor code style issues.

case-lib/lib.sh
marc-hb: +1. Less indentation and less nesting: easier to read.

https://medium.com/swlh/return-early-pattern-3d18a41bba8

marc-hb:
Please re-add check-sof-logger in some PR testing before merging this? Depending on recent daily test results.

We used to have frequent failures but we're probably not testing these configurations. In any case it should run at least for MTL.

marc-hb:
https://sof-ci.01.org/softestpr/PR1065/build474/devicetest/index.html and https://sof-ci.01.org/softestpr/PR1065/build474/devicetest/index.html are all green (but 10 days old now)

https://sof-ci.01.org/softestpr/PR1065/build475/devicetest/index.html is mostly `Not Tested`.

There's also been other changes merged. Re-running.

marc-hb:
SOFCI TEST

Issue #1059, opened 2023-06-15
[FEATURE] configurable FW Loading polling time per platform


fredoh9:
MAX_WAIT_FW_LOADING variable has default value of 70 seconds, The default value was decided high to cover MTL i915 probe timeout. In normal production device, FW loading time is less than a second. It doesn't mean MAX_WAIT_FW_LOADING to waste 70 seconds for every test in every platform, it will escape the polling as soon as FW loading is done.

One concern is raised, even 10 second timeout also pretty luxury but it may hide fw loading delay regression problem. If we can configure 
* MAX_WAIT_FW_LOADING=70s for MTL and newer platforms which might have same i915 driver problem,
* MAX_WAIT_FW_LOADING=10s for other platforms

And with this configurable feature we can set back 10 seconds for MTL when i915 drivers are upsteamed, in near future.

Related PRs,
- https://github.com/thesofproject/sof-test/pull/1057
- https://github.com/thesofproject/sof-test/pull/1044

marc-hb:
Here's an interesting datapoint.

https://github.com/thesofproject/sof/pull/7698  broke the MTL build. Tests were started anyway without firmware (tracked in internal issue 139). Each test took 70s to timeout in https://sof-ci.01.org/sofpr/PR7698/build9948/devicetest

The whole PR test plan uselessly hogged 3 MTL devices for more than 200 minutes.

PR #1067, opened 2023-06-24
lib.sh: make MAX_WAIT_FW_LOADING platform dependent

marc-hb:
We don't want to wait for 70s in every test on every platform when the firmware does not load. When multiplied by the number of tests in a plan this can cause a considerable waste of hardware resources.

i915 timeout aside, 10s has proved over a long period to be more than enough.

Fixes: #1059


marc-hb:
Single, isolated suspend/resume failure in https://sof-ci.01.org/softestpr/PR1067/build493/devicetest/index.html?model=JSL_RVP_NOCODEC&testcase=check-suspend-resume-with-playback-5

Everything else green.

fredoh9 (approved):
Looks good, thank you for the update

PR #1070, opened 2023-06-27
check-volume-levels.sh: Shellcheck fixes.

greg-intel:
Three instances of...
SC2295 (info): Expansions inside ${..} need to be quoted separately, otherwise they match as patterns.

Added quotes to $search in this line:
`CAP_SWITCH=${tmp#*$search}` -->  `CAP_SWITCH=${tmp#*"$search"}`

Tested/verified on jf-adlp-rvp-nocodec-8

### DIFF
--- a/test-case/check-volume-levels.sh
+++ b/test-case/check-volume-levels.sh
@@ -294,13 +294,13 @@ do_preparations () {
     tmp=$(amixer controls | grep -e "$CAP_PGA.*Volume" || true )
     test -n "$tmp" || die "No control with name Volume found in $CAP_PGA"
     search="name="
-    CAP_VOLUME=${tmp#*$search}
+    CAP_VOLUME=${tmp#*"$search"}
     export CAP_VOLUME
     dlogi "Capture volume control name is $CAP_VOLUME"
 
     tmp=$(amixer controls | grep -e "$CAP_PGA.*Switch" || true )
     test -n "$tmp" || die "No control with name Switch found in $CAP_PGA"
-    CAP_SWITCH=${tmp#*$search}
+    CAP_SWITCH=${tmp#*"$search"}
     export CAP_SWITCH
     dlogi "Capture switch control name is $CAP_SWITCH"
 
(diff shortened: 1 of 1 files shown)


marc-hb:
> Originally skipped CI, due to the test itself not being part of ci testing.

Please don't leave temporary stuff like this in commit messages.

So how come this is still not running jenkins? This still very confusing. Would it be possible to simplify everything like this:
- remove all the skipping from the first review
- close this duplicate ?

Or maybe the other way round.

greg-intel:
> > Originally skipped CI, due to the test itself not being part of ci testing.
> 
> Please don't leave temporary stuff like this in commit messages.
> 
> So how come this is still not running jenkins? This still very confusing. Would it be possible to simplify everything like this:
> 
> * remove all the skipping from the first review
> * close this duplicate ?
> 
> Or maybe the other way round.

Doesn't matter if it doesn't run on Jenkins.  The file that was changed isn't part of the PR testing.  
As for why, maybe it was still seeing the description rather than the commit, which I've edited to remove the skip part.

marc-hb:
> which I've edited to remove the skip part.

Thanks but this is still in the commit message of the current 95bc7c44a20db09d0a and it does not belong there:  "Originally skipped CI, due to the test itself not being part of ci testing."


> Doesn't matter if it doesn't run on Jenkins.

The reason it matters is because I don't want to have to even think about "SKIP whatever" in the context of this particular PR; we already spent way too much time on that. Neither in the commit message, nor in the PR description nor in the test results. Just the most simple possible PR please.

greg-intel:
> Thanks but this is still in the commit message of the current [95bc7c4](https://github.com/thesofproject/sof-test/commit/95bc7c44a20db09d0a4438b1a1f2525bfecaca23) and it does not belong there: "Originally skipped CI, due to the test itself not being part of ci testing."

That's context, explaining how this test differs from others.

greg-intel:
SOFCI TEST

marc-hb:
> Doesn't matter if it doesn't run on Jenkins. 

Wait, the other reason it matters is "why?". Why is Jenkins still skipping this after you removed all "skip" keywords? Is this expected? Seems wrong. This is off-topic (hence a simpler PR to avoid these questions) but it should really worry the Jenkins owner.

> SOFCI TEST

OK, now we may never know.

> That's context, explaining how this test differs from others.

Fair enough, something like "not currently tested in CI" is fine. I think it's the "Originally" that I don't get and I don't like.

marc-hb (approved):

PR #1071, opened 2023-06-29
lib.sh,hijack.sh: restore ability to be sourced interactively

marc-hb:
Run `start_test()` and `func_exit_handler()` only in "normal" usage, not
when sourced interactively (for interactive debug) and not as a subtest.

With this commit it's not possible to do this without any side-effect:

```
( source case-lib/lib.sh )
```

More to the point, it's now possible to run this and start interactive
testing without any unwanted actions:

```
source case-lib/lib.sh
get_firmware_path
```

- #1170 
- #1033

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -2,11 +2,19 @@
 
 SUDO_CMD=$(command -v sudo)
 
-trap 'func_exit_handler $?' EXIT
+# Register cleanup callback only when we're sourced by lib.sh itself
+# sourced by just other file level. Not when sourced interactively and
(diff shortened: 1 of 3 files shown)


marc-hb:
https://sof-ci.01.org/softestpr/PR1071/build516/devicetest/index.html is failing because of (complex) thesofproject/sof#7875

marc-hb:
Need to test this with SubTests more.

marc-hb:
Too complex and not flexible enough: we want some scripts not to call start_test() at all. See #1170 and friends.

PR #1068, opened 2023-06-27
volume-basic-test.sh: don't skip "playback volume" controls

marc-hb:
Looking for "PGA" is not enough.

Other minor fixes.

cc:
- #931

PR #1074, opened 2023-06-29
volume-basic-test.sh: catch amixer failures and log all controls

marc-hb:
This (1) provides some logging (2) avoids `skip_test` if amixer fails


marc-hb:
Just 1 suspend/resume failure in https://sof-ci.01.org/softestpr/PR1074/build530/devicetest/index.html

marc-hb:
This is the more "cosmetic" subset of #1068 which requires more discussion - while this one can hopefully be expedited.

keqiaozhang (approved):
A very small change and can catch the corner case when amixer fails. LGTM.

PR #1078, opened 2023-07-05
sof-kernel-log-check.sh: add a filter for Dell TGLH

keqiaozhang:
This patch will ignore CPU the stepping warning after boot.


plbossart (approved):

marc-hb:
PS: the commit title does really not describe the change...

PR #1081, opened 2023-07-08
verify-tplg-binary.sh: create .png in $LOG_ROOT instead of /tmp

marc-hb:
3 commits. Main one:

------
verify-tplg-binary.sh: create .png in $LOG_ROOT instead of /tmp

It's better next to the "other" logs. Also means there's a single, usual
place to clean up.

We'd rather not create it at all but it's not optional.

### DIFF
--- a/test-case/verify-tplg-binary.sh
+++ b/test-case/verify-tplg-binary.sh
@@ -44,7 +44,7 @@ main()
     # This one can find more problems, see sof-test#1054
     dlogi "Checking topology file with tplgtool2.py: $tplg_path"
     ( set -x
-      tplgtool2.py -D /tmp/ "$tplg_path" ) || {
+      tplgtool2.py -D "${LOG_ROOT}" "$tplg_path" ) || {
         ret=$?
         die "tplgtool2.py returned $ret"
     }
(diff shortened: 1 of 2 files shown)


test-case/verify-tplg-binary.sh
```diff
@@ -44,7 +44,7 @@ main()
     # This one can find more problems, see sof-test#1054
     dlogi "Checking topology file with tplgtool2.py: $tplg_path"
     ( set -x
-      tplgtool2.py -D /tmp/ "$tplg_path" ) || {
```
RanderWang: do we need to update CI test ?

RanderWang (commented):
LGTM.  One concern: do we need to update CI test ?

test-case/verify-tplg-binary.sh
marc-hb: Thanks @RanderWang , can you clarify which test?

This line was added only last week so I can't imagine something already depending on it.

Maybe you're referring to the graph that is appearing in test results? That is elsewhere, much much older and unrelated.

marc-hb:
https://sof-ci.01.org/softestpr/PR1081/build586/devicetest/index.html and 
https://sof-ci.01.org/softestpr/PR1081/build585/devicetest/index.html
 => unrelated suspend/resume failures

https://sof-ci.01.org/softestpr/PR1081/build584/devicetest/index.html
=> known multi pause resume

test-case/verify-tplg-binary.sh
RanderWang: OK, I get your idea. I first considered some CI test may use this path

RanderWang (approved):
LGTM

PR #1082, opened 2023-07-10
tplgtool2: dump core ID for any widget not scheduled on core 0

yongzhi1:
Dump core ID if some widgets are running on core 1 or 2 and the rest defaults to core 0.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -738,8 +738,8 @@ class GroupedTplg:
 
     @cached_property
     def coreids(self):
-        "All available core IDs."
-        cores = set()
+        "Core 0 is the default value."
+        cores = {0}
         for widget in self.widget_list:
             core = self.get_core_id(widget)
             if core is not None:


marc-hb:
Thanks @yongzhi1 ! Could you share two before/after screenshots for comparison? For just one sample, representative topology?

yongzhi1:
Hi, @marc-hb, just an example https://github.com/thesofproject/sof/commit/41340b08d509d471fd36387d6930fcc84b316a06

Without the patch, core ID is not displayed even though the tplg contains both core 0 & 2:

sof-mtl-nocodec

With the patch (same as -c=always):
sof-mtl-nocodec-v2

tools/tplgtool2.py
```diff
@@ -739,7 +739,7 @@ def print_pcm_info(self):
     @cached_property
     def coreids(self):
         "All available core IDs."
-        cores = set()
+        cores = set([0])
```
marc-hb, line 742: I found the code quite cryptic. I think these two changes would have really helped:
- A very short comment
- Not using a [ ] list indirection which looks like it's a set of lists.
```suggestion
        # Core 0 is the default value
        cores = {0}
```

marc-hb (changes requested):
OK I think _maybe_ I understand the fix now. Because this is a bug fix right? If I'm correct then the message should say that it is a bug fix and then explain the bug first and then explain the fix. For now the sentence "Dump core ID if some widgets are running on core 1 or 2 and the rest defaults to core 0." sounds like a feature.

So I think the bug is: if only core 2 is specified (and 0 is the default), then `cores = {2}` which means `len(cores) == 1` and the code wrongly believes that the topology is single core and does not display anything. Is that the bug getting fixed, did I understand this correctly?

So if I understood this correctly, then my next question (after adding the missing commit message) is: is it possible to create a topology where _everything_ is on core 2?

yongzhi1:
> So I think the bug is: if only core 2 is specified (and 0 is the default), then `cores = {2}` which means `len(cores) == 1` and the code wrongly believes that the topology is single core and does not display anything. Is that the bug getting fixed, did I understand this correctly?

Yes, your understanding is correct.

> 
> So if I understood this correctly, then my next question (after adding the missing commit message) is: is it possible to create a topology where _everything_ is on core 2?

I am not sure the 3 cores are identical, and any of them can serve as the primarily core. From mere topology perspective, it's doable to create a tplg based on secondary core only.

tools/tplgtool2.py
yongzhi1: Ack

yongzhi1:
https://github.com/thesofproject/sof-test/pull/1083 is more thorough treatment, so close this one. Thanks @marc-hb !!

PR #1083, opened 2023-07-11
tplgtool2.py: fix --show_core handling of the lack of coreID

marc-hb:
Fixes `--show_core` display option issues initially reported by Yong Zhi in #1082

Components that don't have any core ID specified and using a default core were mishandled in several ways:

- Default values did not count in the `is_multicore()` function! So, when using a single CORE_ID=2 the count was wrong and `is_multicore()` wrongly returned false. See #1082

- Default values were never shown even when using `--show_core always`

- It was impossible to show the difference between the default coreID (no value) and the explicit value zero.

Fix all these issues by giving the default value a real status in the code.

A side-effect of this commit is that the `--show_core auto` option gets rid of the "multicore" idea and will now display coreIDs when there is a mix of "default" and "zero". This is a good thing. tplgtool2.py is about analyzing and visualizing what is in .tplg files exactly. It shouldn't be a tool that tries (and fails) to be smart and to guess whether a topology is multicore. It shouldn't try to hide the difference between no coreID and coreID zero.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -742,13 +742,15 @@ class GroupedTplg:
         cores = set()
         for widget in self.widget_list:
             core = self.get_core_id(widget)
-            if core is not None:
+            if core is None:
+                cores.add(-1) # placeholder for the lack of core ID
+            else:
                 cores.add(core)
         return cores
 
     @cached_property
-    def is_multicore(self):
-        "Check if this topology is a multi-core topology."
+    def has_core_differences(self):
+        "Check whether this topology uses more than one coreID"
         return len(self.coreids) > 1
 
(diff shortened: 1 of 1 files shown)


fredoh9 (approved):
Looks good to me!

marc-hb:
Only 2 known issues in https://sof-ci.01.org/softestpr/PR1083/build601/devicetest/index.html
and https://sof-ci.01.org/softestpr/PR1083/build599/devicetest/index.html

MTL https://sof-ci.01.org/softestpr/PR1083/build600/devicetest/index.html is all green..

Amazingly, not a single suspend/resume failure.

yongzhi1 (approved):

PR #1086, opened 2023-07-18
check-sof-logger.sh: cat correct error file on mtrace failure

marc-hb:
Fixes commit https://github.com/thesofproject/sof-test/commit/14313031c43ad8e0e58d91b519295bf86359688c ("check-sof-logger: add support for mtrace
log with IPC4")

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #1088, opened 2023-07-18
check-sof-logger: Switch to use is_firmware_file_zephyr

ujfalusi:
The is_zephyr check looks at the number of 'zephyr' strings in the LDC file while the is_firmware_file_zephyr is doing the same on the actual firmware file.
It has been observed that the is_zephyr function can get the detection wrong and reports false for a zephyr firmware.

Note also that virtually only the check-sof-logger is using the is_zephyr call, all other test cases use the is_firmware_file_zephyr call.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -76,7 +76,7 @@ run_loggers()
     # the same time, so $data_file will hopefully not be long.
     local collect_secs=2
 
-    if is_zephyr; then
+    if is_firmware_file_zephyr; then
         # Collect logs from Zephyr logging backends
 
         if is_ipc4 ; then
@@ -136,7 +136,7 @@ run_loggers()
         }
     fi
 
-    if is_zephyr; then
+    if is_firmware_file_zephyr; then
         # Zephyr logging backends
 
         if is_ipc4 ; then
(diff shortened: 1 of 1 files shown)


kv2019i (approved):

marc-hb (approved):
My only minor concern is: now this test requires the firmware to have been loaded at least once on this linux boot so the firmware file name can be extracted from the kernel logs. There's a good chance this is already required for some other reason, so be it.

Approved assuming the tests are passing. Note PR testing does not test MTL yet.

While in the check-sof-logger.sh context please review one line long but important fix #1086 in the same test.

ujfalusi:
@marc-hb, it is kind of hard to test if the firmware logging is working without a loaded firmware, no?

marc-hb:
> @marc-hb, it is kind of hard to test if the firmware logging is working without a loaded firmware, no?

The test unloads and reloads the firmware, using the firmware boot banner as a sanity check. So _in theory_ this test can work while being the first to load the firmware. But now this won't work anymore because `is_firmware_file_zephyr` is called first.

Minor new requirement, no big deal.

ujfalusi:
> > @marc-hb, it is kind of hard to test if the firmware logging is working without a loaded firmware, no?
> 
> This test unloads and reloads the firmware, using the firmware boot banner as a sanity check. So _in theory_ this test can work while being the first to load the firmware. But now this won't work anymore because `is_firmware_file_zephyr` is called first.

Hrm, I'm not sure about that as the test will first reloads the drivers, thus boots the firmware then it will starts the logger to fetch the logs:
https://github.com/thesofproject/sof-test/blob/main/test-case/check-sof-logger.sh#L257-L259

Or am I looking at it in a wrong way?

ujfalusi:
All check-sof-logger cases are passing

marc-hb:
> as the test will first reloads the drivers, thus boots the firmware then it will starts the logger to fetch the logs:

Then we're good.

I guess I had `logger_disabled()->is_firmware_file_zephyr()` in mind. This one is called very early but it's fortunately not used by check-sof-logger.sh right now (I think someone asked for that a long time ago)

These dependencies are very brittle, seems more complicated than booting an OS :-)

PR #1089, opened 2023-07-27
Test-case: test-speaker.sh: Use pink noise for non-48 kHz rates

singalsu:
The speaker test fails with -t wav internally for other than "-r 48000" because it is not re-sampling the wav "Front center" etc. clips. Using pink noise for non-48 kHz works around this limitation and allows to test with all sample rates. Also the test are not depending on spoken channels names. Channels order can be subjectively checked with the 48 kHz test.

### DIFF
--- a/test-case/test-speaker.sh
+++ b/test-case/test-speaker.sh
@@ -43,8 +43,15 @@ do
     dev=$(func_pipeline_parse_value "$idx" dev)
     snd=$(func_pipeline_parse_value "$idx" snd)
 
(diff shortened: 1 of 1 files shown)


singalsu:
None of the CI fails are related, since this is a bug fix for CI I added P1 priority. Please review!

juimonen (approved):
fine by me

btian1 (approved):

fredoh9 (approved):
thank you for the update/fix

PR #1092, opened 2023-08-29
test-case: add more error handling for check-performance

aiChaoSONG:
The error won't be reported if process exit unexpectedly, add error handling code for this.

### DIFF
--- a/test-case/check-performance.sh
+++ b/test-case/check-performance.sh
@@ -47,6 +47,9 @@ setup_kernel_check_point
 func_lib_check_sudo
 func_pipeline_export "$tplg" "type:any"
 
+aplay_num=0
+arecord_num=0
+
 for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
 do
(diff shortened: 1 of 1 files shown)


keqiaozhang (approved):
LGTM

test-case/check-performance.sh
```diff
@@ -62,11 +65,24 @@ do
...
             arecord_opts -D "$dev" -c "$channel" -r "$rate" -f "$fmt" -d "$duration" /dev/null -q &
+            arecord_num=$((arecord_num+1))
         fi
 done
 
+sleep 1 # waiting stable streaming of aplay/arecord
+dlogi "Number of aplay/arecord process started: $aplay_num, $arecord_num"
+
+real_aplay_num=$(ps --no-headers -C aplay | wc -l)
+real_arecord_num=$(ps --no-headers -C arecord | wc -l)
```
marc-hb, line 79: You should use `wait $PIDs` instead of `ps`. It's more reliable and it passes error codes.

PR #1093, opened 2023-08-31
tools: sof_perf_analyzer: fix component ID

aiChaoSONG:
Previously, component ID in mtrace and kernel
is different, which makes it hard to find the
component from kernel message with a component
ID from mtrace. After it is fixed in SOF side,
now module_id take the first 16 bits, instance_id
takes next 8 bits. We should do a fix accordingly.

Link: https://github.com/thesofproject/sof/issues/8051


fredoh9 (approved):
looks good, thank you for the fix

PR #1097, opened 2023-09-05
add support to output sof performance data to csv

aiChaoSONG:
Two commits:
- support to output csv for performance data in sof_perf_analyzer
- enable the feature in sof-test case

An example of CSV output: [sof_perf.csv](https://github.com/thesofproject/sof-test/files/12519387/sof_perf.csv)


mengdonglin (approved):

keqiaozhang (approved):
LGTM

PR #1098, opened 2023-09-06
lib.sh: use txt suffix for csv file

aiChaoSONG:
Our sof-ci web cannot recognize csv file,
so on opening the test result, the csv file
will be downloaded. Change the file suffix
to txt, so that we can view it in the web.


marc-hb (approved):

aiChaoSONG:
running example, see internal report 31304.

PR #1099, opened 2023-09-06
tplgtool2.py: show widget CPC in topology graph

aiChaoSONG:
This patch adds support to show widget CPC value
int topology graph.

sof-hda-generic-4ch


marc-hb (approved):
Could you:
- Share a sample screenshot in the PR?
- Split in 2 commits? First the no-op refactor, second commit adds the new feature.

mengdonglin (approved):

PR #1100, opened 2023-09-06
sof_perf_analyzer: add skip-to-first-trace option

aiChaoSONG:
In CI test, some traces from previous test case will apprear in the mtrace of current test case, this is a know issue.

Previously, we always skip to the first valid trace of this test case, this could raise trouble for developer use of this script.

This patch adds skip-to-first-trace option, for CI test, this option should be set (True), and for developer use, this option should be left as it is (False).

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -901,9 +901,9 @@ perf_analyze()
     dlogi "Checking SOF component performance"
     if [ -e "$LOG_ROOT/mtrace.txt" ]; then
         if [ -e "$LOG_ROOT/dmesg.txt" ]; then
-            perf_cmd="sof_perf_analyzer.py --kmsg=$LOG_ROOT/dmesg.txt --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt "
+            perf_cmd="sof_perf_analyzer.py --skip-to-first-trace --kmsg=$LOG_ROOT/dmesg.txt --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt"
         else
-            perf_cmd="sof_perf_analyzer.py --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt"
+            perf_cmd="sof_perf_analyzer.py --skip-to-first-trace --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt"
         fi
         dlogc "$perf_cmd"
         eval "$perf_cmd" || {
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -26,7 +26,6 @@ import re
 import pathlib
 import argparse
 from typing import TextIO
-from typing import Optional
 from typing import Generator
 from dataclasses import dataclass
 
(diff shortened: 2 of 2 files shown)


singalsu (approved):
This patch fixes the problem that I had. Thank you very much!

tools/sof_perf_analyzer.py
```diff
@@ -295,6 +296,10 @@ def parse_args():
                         help='Kernel message file captured with journalctl or other log utility')
     parser.add_argument('--out2csv', type=pathlib.Path, required=False,
                     help='Output SOF performance statistics to csv file')
+    parser.add_argument('-s', '--skip-to-first-trace', action="store_true",  default=False,
+                        help='''In CI test, some traces from previous test case will apprear in
```
marc-hb, line 300:
```suggestion
                        help='''In CI test, some traces from previous test case will appear in
```

tools/sof_perf_analyzer.py
```diff
@@ -295,6 +296,10 @@ def parse_args():
...
+the mtrace of current test case, this flag is used to denote if we
+want to skip those traces, for CI test, set the flag to True''')
```
marc-hb, line 302: Please don't "hide" the actual logic behind abstract names. This is test code, we need as few abstractions as possible.
```suggestion
want to skip until the first line with a timestamp between 0 and 1s. For CI test, set the flag to True''')
```

tools/sof_perf_analyzer.py
```diff
@@ -169,8 +165,13 @@ def process_trace_file():
...
         trace_item_gen = make_trace_item(file)
-        trace_prev = skip_to_first_trace(trace_item_gen)
-        if trace_prev is None:
+        trace_prev = None
+        try:
+            if args.skip_to_first_trace:
+                trace_prev = skip_to_first_trace(trace_item_gen)
+            else:
+                trace_prev = next(trace_item_gen)
+        except:
```
marc-hb, line 174: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/broad-exception-caught.html

Why do you need to catch anything here? Just let the original exception propagate to the top. This is test code, we want as few abstractions and layers as possible when things go wrong.

marc-hb (changes requested):
The pylint warning may go away if you drop the too broad except?

tools/sof_perf_analyzer.py
aiChaoSONG: The raised exception has no useful information, I want to add some useful information here, I will use below code, which is better, add message to the exception, and raise it again.
```
        try:
            if args.skip_to_first_trace:
                trace_prev = skip_to_first_trace(trace_item_gen)
            else:
                trace_prev = next(trace_item_gen)
        except StopIteration as si:
            si.args = 'No valid trace in provided file',
            raise 
```

marc-hb (dismissed):
Nit: could you add a blank line before 190?

Note: still a Github Draft

case-lib/lib.sh
```diff
@@ -901,9 +901,9 @@ perf_analyze()
     dlogi "Checking SOF component performance"
     if [ -e "$LOG_ROOT/mtrace.txt" ]; then
         if [ -e "$LOG_ROOT/dmesg.txt" ]; then
-            perf_cmd="sof_perf_analyzer.py --kmsg=$LOG_ROOT/dmesg.txt --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt "
+            perf_cmd="sof_perf_analyzer.py -s --kmsg=$LOG_ROOT/dmesg.txt --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt "
         else
-            perf_cmd="sof_perf_analyzer.py --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt"
+            perf_cmd="sof_perf_analyzer.py -s --out2csv $LOG_ROOT/sof_perf.txt $LOG_ROOT/mtrace.txt"
```
marc-hb, line 906: Please use the long option so people don't have to look it up. Short options save repetitive typing but it's not a problem here.

marc-hb (approved):

aiChaoSONG:
Thanks Marc, the test is good, see internal result 31419, merge

PR #1102, opened 2023-09-07
Add support for MIPI Sys-T logging

kv2019i:
Add initial support for handling MIPI Sys-T Cat logging in SOF firmware:

- identify FW that emits logging in Sys-T Cat dictionary format
- use systprint tools to decode logs  ( https://github.com/MIPI-Alliance/public-mipi-sys-t )
- no flag day needed, this PR is compatible with other logging types (so can be merged without dependencies)

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -339,6 +339,27 @@ find_ldc_file()
     printf '%s' "$ldcFile"
 }
 
+# Prints the syst collateal file found on stdout, logs on stderr.
+find_syst_dict_file()
+{
+    local dict_file
+
+    if [ -n "$SOFSYST" ]; then
+        dict_file="$SOFSYST"
+        >&2 dlogi "SOFSYST=${SOFSYST} overriding default locations"
+    else
+        # reuse LDC logic to set subdir
+        local subdir; subdir=$(get_ldc_subdir)
+        dict_file=/lib/firmware/"${subdir}"/mipi_syst_collateral.xml
+    fi
+
+    [[ -e "$dict_file" ]] || {
+        >&2 dlogi "MIPI Sys-T Collateral file $dict_file not found"
+        return 1
+    }
+    printf '%s' "$dict_file"
+}
+
 func_mtrace_collect()
 {
     local clogfile="$1"
@@ -355,6 +376,44 @@ func_mtrace_collect()
     # Cleaned up by func_exit_handler() in hijack.sh
     # shellcheck disable=SC2024
     sudo "${mtraceCmd[@]}" >& "$clogfile" &
+}
+
+func_lib_log_post_process()
+{
+    # SyS-T log output a Zephyr feature, no need postprocess
+    # if non-Zephyr firmware file under test
+    is_firmware_file_zephyr || return 0
+
+    local logfile="$LOG_ROOT"/mtrace.txt
+    local outfile="$LOG_ROOT"/mtrace-decoded.csv
+
+    # postprocess is not required if the test
+    # configuration does not emit SysT-Cat log encoding
+    grep -q "SYS-T RAW DATA:" "$logfile" || return 0
+
+    # systprint tool is available via
+    # - https://github.com/zephyrproject-rtos/mipi-sys-t
+    # - https://github.com/MIPI-Alliance/public-mipi-sys-t
+    if [ -z "$SYSTPRINT" ]; then
+        SYSTPRINT=$(command -v systprint) || {
+            dlogw "No systprint found in PATH, unable to post-process file $logfile"
+            # TODO: make this fatal
+            return 0
+        }
+    fi
+
+    local dict_file
+    dict_file=$(find_syst_dict_file) || {
+        dlogw "SyS-T dict file not found, unable to post-process file $logfile"
+        # TODO: make this fatal
+        return 0
+    }
+
+    $SYSTPRINT -c "$dict_file" "$logfile" >"$outfile" || {
+        dlogw "Error running systprint, unable to post-process file $logfile"
+        # TODO: make this fatal
+        return 0
+    }
 }
 
 func_sof_logger_collect()
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -277,14 +277,22 @@ main()
     # then for the 'FW ABI' banner
     for f in "${stdout_files[@]}"; do
         local tracef="$LOG_ROOT/logger.$f.txt"
+        local syst_prefix="SYS-T RAW DATA:"
         test -e "$tracef" || die "$tracef" not found
 
         local tool_banner boot_banner
         if is_ipc4 && is_firmware_file_zephyr && [ "$f" = 'etrace' ]; then
             # mtrace
-            # No specific tool banner, just check some logs are visible.
-            tool_banner=' .*<.*>'
-            boot_banner='FW ABI.*tag.*zephyr'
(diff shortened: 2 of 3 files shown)


test-case/check-playback.sh
```diff
@@ -102,4 +102,6 @@ do
     done
 done
 
+logger_disabled || func_lib_log_post_process
```
aiChaoSONG, line 105: I think this is too early to do log post process. at this time, mtrace-reader is still running. better to add this to exit handler, see case-lib/hijack.sh.

kv2019i (commented):
Forgot to post my own comments.

test-case/check-playback.sh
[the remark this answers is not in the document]
kv2019i: Thanks @aiChaoSONG , this should be handled in exit handler. And now done in latest revision.

case-lib/lib.sh
```diff
@@ -357,6 +384,44 @@ func_mtrace_collect()
     sudo "${mtraceCmd[@]}" >& "$clogfile" &
 }
 
+func_lib_log_post_process()
+{
+    # nothing to do unless dictionary backends of Zephyr
+    # are used
+    if ! is_ipc4 && ! is_firmware_file_zephyr; then
+        return 0
```
marc-hb, line 392: You can avoid all the negations and the `if/then/else` using the common "assert || return" style

```shell
 is_ipc4 || is_firmware_file_zephyr || return 0
```

BTW this condition looks weird, could use a comment.

case-lib/lib.sh
```diff
@@ -357,6 +384,44 @@ func_mtrace_collect()
...
+            dlogw 'No systprint found in PATH'
+            return 1
+        }
+    fi
+
+    local logfile="$LOG_ROOT"/mtrace.txt
+    local outfile="$LOG_ROOT"/mtrace-decoded.txt
+
+    grep -q "SYS-T RAW DATA:" $logfile || {
+        return 0
```
marc-hb, line 409: Is this normal? At the very least a `dlogw` if not.

If normal then you can drop the `{}`

case-lib/lib.sh
```diff
@@ -357,6 +384,44 @@ func_mtrace_collect()
...
+    local dict_file
+    dict_file=$(find_syst_dict_file) || {
+        dloge 'Sys-T dict file not found, unable to postprocess!'
+        return 1
+    }
+
+    echo $SYSTPRINT -p -c $dict_file $logfile
+    $SYSTPRINT -p -c $dict_file $logfile >$outfile || {
+        dlogw 'Error running sysprint'
+        return 1
```
marc-hb, line 420: Either use `dloge` or return 0

case-lib/lib.sh
```diff
@@ -357,6 +384,44 @@ func_mtrace_collect()
...
+    }
+
+    local dict_file
+    dict_file=$(find_syst_dict_file) || {
+        dloge 'Sys-T dict file not found, unable to postprocess!'
+        return 1
+    }
+
+    echo $SYSTPRINT -p -c $dict_file $logfile
+    $SYSTPRINT -p -c $dict_file $logfile >$outfile || {
```
marc-hb: This type of copy/paste/diverge can easily be avoided with `set -x`

```shell
(set -x
 $SYSTPRINT -p -c $dict_file $logfile >$outfile ) ||
  dlog ...
```

case-lib/lib.sh
```diff
@@ -339,6 +339,33 @@ find_ldc_file()
...
+find_syst_dict_file()
+{
+    local dict_file
+    # if user doesn't specify file path of sof-*.ldc, fall back to
+    # /etc/sof/sof-PLATFORM.ldc, which is the default path used by CI.
+    # and then on the standard location.
+    if [ -n "$SOFSYST" ]; then
+        dict_file="$SOFSYST"
+        >&2 dlogi "SOFSYST=${SOFSYST} overriding default locations"
+    else
```
marc-hb, line 352: `printf` and early `return` here avoids the `else` and the extra indentation and is easier to follow. Don't even need the `dict_file` variable yet.

case-lib/lib.sh
```diff
@@ -339,6 +339,33 @@ find_ldc_file()
...
+        local platf; platf=$(sof-dump-status.py -p) || {
+            >&2 dloge "Failed to query platform with sof-dump-status.py"
```
marc-hb, line 354: You don't seem to be using `platf` in this function... is this obsolete code maybe?

test-case/check-sof-logger.sh
```diff
@@ -277,14 +277,22 @@ main()
...
         if is_ipc4 && is_firmware_file_zephyr && [ "$f" = 'etrace' ]; then
             # mtrace
-            # No specific tool banner, just check some logs are visible.
-            tool_banner=' .*<.*>'
-            boot_banner='FW ABI.*tag.*zephyr'
+            if head -n 5 "$tracef" | grep -q "$syst_prefix" ; then
+                # MIPI Sys-T Catalog format, just search for Sys-T prefix
+                # that is available without decoding tools and collateral
+                tool_banner=$syst_prefix
+                banner_banner=$syst_prefix
```
marc-hb, line 290: Typo? Spotted by shellcheck. Shellcheck saves lives, use it.

marc-hb (dismissed):

case-lib/lib.sh
kv2019i: Fixed.

test-case/check-sof-logger.sh
marc-hb: ping?

case-lib/lib.sh
```diff
@@ -339,6 +339,29 @@ find_ldc_file()
     printf '%s' "$ldcFile"
 }
 
+# Prints the syst collateal file found on stdout, logs on stderr.
+find_syst_dict_file()
+{
+    local dict_file
+    # if user doesn't specify file path of sof-*.ldc, fall back to
+    # /etc/sof/sof-PLATFORM.ldc, which is the default path used by CI.
```
marc-hb, line 347: You copied this comment from the `find_ldc_file()` function that does this fallback but it does not seem to happen here. Which is OK because `/etc/sof` is bad, everything should be in a single place `/lib/firmware/intel/sof`. So just the comment is old and wrong?

case-lib/lib.sh
```diff
@@ -357,6 +380,41 @@ func_mtrace_collect()
...
+        SYSTPRINT=$(command -v systprint) || {
+            dlogw "No systprint found in PATH, unable to post-process file $logfile"
+            return 0
+        }
+    fi
+
+    local dict_file
+    dict_file=$(find_syst_dict_file) || {
+        dlogw "SyS-T dict file not found, unable to post-process file $logfile"
+        return 0
```
marc-hb, line 409:
```suggestion
        # TODO: make this fatal
        return 0
```

kv2019i:
Marked as ready to review. This should now be ready for review for merge. The changes should have no impact to test configurations that do not use SyS-T logging, but provides capability to test SyS-T logging enabled images with sof-test (and start doing deployment of SyS-T configurations to CI).

marc-hb (approved):

marc-hb:
Warning: jenkins has not run yet. Internal issue 480.

The summary screen is deceptive:

Screenshot 2023-09-19 at 10 08 15

dabekjakub (approved):

kv2019i:
@marc-hb and @thesofproject/test-maintainers , could this be merged?

PR #1111, opened 2023-10-09
check-sof-logger.sh: more robust fix to runtime-PM race

kv2019i:
It turns out on some test platforms, the launch of logger processes can take significant amount of time, so the solution to create some DSP activity just before launching the logger processes, is not reliable enough.

Implement alternative solution where the DSP activity is triggered after loggers are started.

Fix tested with a modified mtrace-reader.py that injects arbitrary delays at startup.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -102,17 +102,12 @@ run_loggers()
 
     # This test is not really supposed to run while the DSP is busy at
     # the same time, so $data_file will hopefully not be long.
-    local collect_secs=2
+    local collect_secs=3
 
     # Wait until SOF ALSA devices are registered. This is last step
     # of the SOF driver probe, so when the ALSA card is registered,
     # the driver probe is complete.
     wait_for_sof_alsa_card
-
-    # Avoid racing with runtime-pm by opening the PCM nodes just before logger
-    # is started. This is still not bullet-proof, but at least we are
-    # pinging the DSP as close as possible to start of log gathering.
-    ping_pcm_devices
 
     if is_firmware_file_zephyr; then
         # Collect logs from Zephyr logging backends
@@ -160,6 +155,17 @@ run_loggers()
              "$loggerBin" -t -f 3 -l "$ldcFile" \
              > "$data_file" 2> "$error_file" & dmaPID=$!
     fi
+
+    # Delay between logger start and triggering DSP activity
+    local logger_settle_time_secs=1
+
+    sleep "$logger_settle_time_secs"
+    # Avoid racing with runtime-pm by opening the PCM nodes when
+    # the loggers are running.
+    ping_pcm_devices
+
+    dlogi "Last kernel activity:"
+    journalctl_cmd -g ' sof' | tail -n 3
 
     sleep "$collect_secs"
 


test-case/check-sof-logger.sh
```diff
@@ -161,6 +159,14 @@ run_loggers()
              > "$data_file" 2> "$error_file" & dmaPID=$!
     fi
 
+    sleep "$logger_settle_time_secs"
+    # Avoid racing with runtime-pm by opening the PCM nodes when
+    # the loggers are running.
+    ping_pcm_devices
+
+    dlogi "Collecting logs now, last kernel activity:"
```
marc-hb, line 167:
```suggestion
    dlogi "Last kernel activity:"
```

I find the "Collecting logs NOW,..." message misleading for two reasons:
- This code is trying to work around the fact that we don't know _when exactly_ log collection starts.
- Seeing `journalctl` called immediately after "collecting now" gives the wrong impression that they're related to each other (I think they're not)

test-case/check-sof-logger.sh
```diff
@@ -102,18 +102,16 @@ run_loggers()
 
     # This test is not really supposed to run while the DSP is busy at
     # the same time, so $data_file will hopefully not be long.
-    local collect_secs=2
+    local collect_secs=3
+
+    # Delay between logger start and triggering DSP activity
+    local logger_settle_time_secs=1
```
marc-hb, line 108: I appreciate the long and descriptive name but I don't think this needs to be defined 50 lines before the only place where it's used. Can you keep it all in one place? This is not C89 :-)

test-case/check-sof-logger.sh
```diff
@@ -161,6 +159,14 @@ run_loggers()
              > "$data_file" 2> "$error_file" & dmaPID=$!
     fi
 
+    sleep "$logger_settle_time_secs"
+    # Avoid racing with runtime-pm by opening the PCM nodes when
+    # the loggers are running.
+    ping_pcm_devices
+
+    dlogi "Collecting logs now, last kernel activity:"
+    journalctl_cmd |grep "kernel: sof" |tail -1 || true
```
marc-hb, line 168:
```suggestion
    journalctl_cmd -g ' sof' | tail -n 3
```
There is already a `-k` in `journalctl_cmd()`.

`tail` always succeeds (unless some true disaster happened, and then we want to stop)

marc-hb (changes requested):
Mostly OK.

marc-hb:
https://sof-ci.01.org/softestpr/PR1111/build786/devicetest/index.html is all green so maybe you finally nailed it?

test-case/check-sof-logger.sh
kv2019i: Ack, will fix.

test-case/check-sof-logger.sh
kv2019i: Updated this as well. I had "kernel: sof" on purpose to not matching random " sof" strings in kernel output, but after a few tests, this seems good enough.

kv2019i:
@marc-hb @fredoh9 Please re-review. I did a custom test run 33338 that covers all daily test configurations and all tests passed with this PR now, so I think this finally covers all cases.

marc-hb (approved):
LGTM

PR #1113, opened 2023-10-11
sof_perf_analyzer: fix exception when kernel message not provided

aiChaoSONG:
If kernel message file is not provided, the component_name dict will be empty, in calculating the maximum length of component names, we calls the max function with an empty sequence, which raises below error:
    "ValueError: max() arg is an empty sequence"

This patch checks if component_name dict is empty, and on empty, provides a default value.

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -247,7 +247,10 @@ def format_perf_info() -> list[str] | None:
     '''Format SOF trace performance statistics'''
     if len(perf_stats):
         lines: list[str] = []
-        max_name_len = max(len(name) for name in component_name.values())
+        if args.kmsg:
+            max_name_len = max(len(name) for name in component_name.values())
+        else:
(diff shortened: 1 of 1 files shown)


tools/sof_perf_analyzer.py
```diff
@@ -247,7 +247,10 @@ def format_perf_info() -> list[str] | None:
     '''Format SOF trace performance statistics'''
     if len(perf_stats):
         lines: list[str] = []
-        max_name_len = max(len(name) for name in component_name.values())
+        if component_name:
```
marc-hb, line 250:
```suggestion
        if args.kmsg:
```
Either the user provided kmsg and then it must be correct. Or they didn't and we don't have component names.

We don't want to silently ignore bad kmsg arguments.

This also saves an explanation why `component_name` can be empty.

marc-hb (changes requested):

btian1 (approved):

marc-hb:
SOFCI TEST

PR #1114, opened 2023-10-13
audio: perf: add module cpc calculation in performance script

btian1:
Add module CPC calculation support in performance scripts.

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -27,6 +27,9 @@ from dataclasses import dataclass
 
 import pandas as pd
 
+# currently, keep CPC = avg_cpu_peak * CPC_MARGIN
+CPC_MARGIN = 1.15
+
 @dataclass()
 class TraceItem:
     '''The structural representation for a single, parsed line of trace'''
@@ -234,7 +237,10 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-
+    comp_peak_avg_cycles = perf_info.groupby('COMP_ID')['CPU_PEAK'].mean()
+    perf_stats['MODULE_CPC'] = comp_peak_avg_cycles * CPC_MARGIN
+    # change data type from float to int
+    perf_stats['MODULE_CPC'] = perf_stats['MODULE_CPC'].astype(int)
 
 def print_perf_info():
     '''Output SOF performance info'''


tools/sof_perf_analyzer.py
```diff
@@ -102,6 +100,8 @@ def collect_perf_info(trace_item: TraceItem):
     ppln_id = msg[0].split(':')[1]
     comp_id = int(msg[1], 16)
     perf_val = int(msg[5]), int(msg[7]), int(msg[10]), int(msg[12])
+    global AUDIO_PERIOD
+    AUDIO_PERIOD = int(msg[7])
```
marc-hb, line 104: All other `msg[x]` fields go to either `CompPerfSample` or to `Component`. Why should `msg[7]` be special and require a `global` (and pylint warning)

marc-hb (dismissed):
> this is a follow up with: #8245

You mean https://github.com/thesofproject/sof/pull/8245, please fix the link.

Please explain what happens when:
- people use the old script on new code after 8245
- people use the new script on old code after 8245

Do they get an error message, silently corrupted data, other?

Think about git bisect for instance (but not just)

tools/sof_perf_analyzer.py
```diff
@@ -256,6 +256,8 @@ def format_perf_info() -> list[str] | None:
                                  'PEAK(MAX)/AVG(AVG)')
         lines.append(title + '\n')
 
+        TICK_TO_MCPS = AUDIO_PERIOD / 1000000
+
```
andrula-song, line 260: Does this calculation really right?

marc-hb:
I don't understand: the first version 0b7cbb43f62812e looks totally different from the current version e76f9c1f91d284828. Both the code changes and commit messages look totally different. What happened?

btian1:
> I don't understand: the first version [0b7cbb4](https://github.com/thesofproject/sof-test/commit/0b7cbb43f62812e791800bcb424e2652c1c5bfec) looks totally different from the current version [e76f9c1](https://github.com/thesofproject/sof-test/commit/e76f9c1f91d284828ecc4c0b3bd375d5fcd1a9f1). Both the code changes and commit messages look totally different. What happened?

#1115 merged， it include patch 1 actually, so only left patch 2 with pandas format adding.

tools/sof_perf_analyzer.py
```diff
@@ -228,13 +231,15 @@ def analyze_perf_info():
         perf_info.groupby('COMP_ID')['CPU_AVG_MCPS'].max(),
         perf_info.groupby('COMP_ID')['CPU_PEAK_MCPS'].min(),
         perf_info.groupby('COMP_ID')['CPU_PEAK_MCPS'].mean(),
-        perf_info.groupby('COMP_ID')['CPU_PEAK_MCPS'].max()
+        perf_info.groupby('COMP_ID')['CPU_PEAK_MCPS'].max(),
+        perf_info.groupby('COMP_ID')['CPU_PEAK'].mean() * CPC_MARGIN,
```
aiChaoSONG, line 235: You don't really want to add it here, you can assign a new column just like 'PEAK(MAX)/AVG(AVG)'

aiChaoSONG (changes requested):

tools/sof_perf_analyzer.py
```diff
@@ -234,7 +237,9 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-
+    perf_stats['MODULE_CPC'] = perf_stats['CPU_PEAK(AVG)'] * perf_info['PERIOD'][1] * CPC_MARGIN
+    # change data type from float to int
+    perf_stats['MODULE_CPC'] = perf_stats['MODULE_CPC'].astype(int)
```
andrula-song, line 242: is it suitable to use int here?

tools/sof_perf_analyzer.py
btian1: @aiChaoSONG ， changed, please check again

btian1:
image

andrula-song:
LGTM, but in this way we must guarantee the period must be the same for each component.

andrula-song (dismissed):

tools/sof_perf_analyzer.py
```diff
@@ -234,7 +237,9 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-
+    perf_stats['MODULE_CPC'] = perf_stats['CPU_PEAK(AVG)'] * perf_info['PERIOD'][1] * CPC_MARGIN
```
aiChaoSONG, line 240: This doesn't look good, `perf_info['PERIOD'][1]` is a single value for one component, it is used by all component in this code. 

the assignment to `perf_stats['MODULE_CPC']` could be done in a single line of code.  The code here will degrade the performance.

`astype` will do `floor` to the data, this means 1.9 becomes 1, make sure this is the expected behavior.

aiChaoSONG (changes requested):

tools/sof_perf_analyzer.py
btian1: > This doesn't look good, `perf_info['PERIOD'][1]` is a single value for one component, it is used by all component in this code.
any good ideas for better implementation?
> 
> the assignment to `perf_stats['MODULE_CPC']` could be done in a single line of code. The code here will degrade the performance.
could show some sample code? I don't know how to implement.
> 
> `astype` will do `floor` to the data, this means 1.9 becomes 1, make sure this is the expected behavior.
yes, however, this is cycle level truncation, cycle level impact is very minor.

aiChaoSONG (approved):

andrula-song (approved):

PR #1121, opened 2023-10-30
lib.sh: generate html instead csv for performance test

aiChaoSONG:
HTML format is more friendly to CI than csv format. It is easy to read in CI.


aiChaoSONG:
SOFCI TEST

keqiaozhang (approved):
Tested and LGTM.

aiChaoSONG:
See internal report 33858, the result looks good. it is html talbe now, we could add style sheet to it later, see https://pandas.pydata.org/docs/user_guide/style.html
Screenshot from 2023-10-30 16-06-18

Issue #1108, opened 2023-10-02
[BUG] check-performance.sh failed with ValueError: invalid literal for int()


fredoh9:
**Describe the bug**

check-performance.sh failed with ValueError: invalid literal for int()

```
2023-10-02 14:41:20 UTC [REMOTE_INFO] Checking SOF component performance
2023-10-02 14:41:20 UTC [REMOTE_COMMAND] sof_perf_analyzer.py --skip-to-first-trace --kmsg=/home/ubuntu/sof-test/logs/check-performance/2023-10-02-14:40:46-4847/dmesg.txt --out2csv /home/ubuntu/sof-test/logs/check-performance/2023-10-02-14:40:46-4847/sof_perf.txt /home/ubuntu/sof-test/logs/check-performance/2023-10-02-14:40:46-4847/mtrace.txt
Traceback (most recent call last):
  File "/home/ubuntu/sof-test/tools/sof_perf_analyzer.py", line 323, in 
    main()
  File "/home/ubuntu/sof-test/tools/sof_perf_analyzer.py", line 316, in main
    process_kmsg_file()
  File "/home/ubuntu/sof-test/tools/sof_perf_analyzer.py", line 222, in process_kmsg_file
    widget_id = int('0x' + widget_id[-6:], 16)
ValueError: invalid literal for int() with base 16: '0x cmd 1'
2023-10-02 14:41:20 UTC [REMOTE_ERROR] SOF component performace analysis tool exit with error
2023-10-02 14:41:20 UTC [REMOTE_INFO] Test Result: FAIL!
```

Most recent failure link is
planresultdetail/32602?model=MTLP_SDW_AIOC-ace1_0-ipc4&testcase=check-performance

**To Reproduce**
Test with MTLP_SDW_AIOC
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-mtl-rt711-l0-rt1316-l23-rt714-l1.tplg MODEL=MTLP_SDW_AIOC-ace1_0-ipc4 SOF_TEST_INTERVAL=5 ~/sof-test/test-case/check-performance.sh

marc-hb:
From: @aiChaoSONG 
Date: Monday, 25 September 2023 at 19:03
Subject: RE: Daily report summary -WW39.1

The module name parsing from the kernel message is based on pattern like this:
```
[...]

PR #1122, opened 2023-10-31
tools: sof_perf_analyzer: fix widget ID parsing error

btian1:
In existing code, we parse component name from the line
that contains 'Create widget', and component ID from the
next line, which normally is the ipc tx message:

Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]

However, due to the concurrent nature of linux kernel, the
assumption that `Create widget` message and `ipc tx` message
are always adjacent is wrong. An exception will be raised
if the next line is not the 'ipc tx' message.

The first step to fix the issue is to make a kernel PR to print
all the information we need in a single 'Create widget' line,
thus to avoid the requirement for multiple lines of kernel message, 
see https://github.com/thesofproject/linux/pull/4709.

And next，this patch will parse the new 'Create widget' line to
get component name and ID.

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -175,12 +175,14 @@ def process_trace_file():
 
 def process_kmsg_file():
     '''Process the dmesg to get the component ID to component name mapping,
-    component name is acquired from the line that contains 'Create widget',
-    component ID is acquired from the next line. Example:
-    [  334.818435] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
-    Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
-    [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
-    ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
+    they are acquired from the line that contains 'Create widget':
+        [   59.622645] snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+        Create widget host-copier.0.capture (pipe 3) - ID 4, instance 3, core 0"
+
+    By design in the kernel, pipeline ID is the instance ID of pipeline
+    widget, so it is acquired from the line that contains 'Create pipeline':
+        [   59.622134] snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+        Create pipeline pipeline.3 (pipe 3) - instance 3, core 0
 
     In practice, sof-test only capture kernel message and firmware trace generated during a test
     case run. Mostly in manual tests, if the kernel message file contains multiple firmware runs
@@ -190,7 +192,7 @@ def process_kmsg_file():
     with open(args.kmsg, encoding='utf8') as f:
         ppln_id = None
         for line in f:
-            if match_obj := re.search(r'Create widget', line):
+            if match_obj := re.search(r"Create (widget|pipeline)", line):
                 span_end_pos = match_obj.span()[1]
                 line_split = line[span_end_pos + 1:].split()
                 widget_name = line_split[0]
@@ -200,11 +202,15 @@ def process_kmsg_file():
                 # creation, because it is always the first one to be created before all other
                 # widgets in the same pipeline and pipelines are created sequentially.
                 if widget_name.startswith('pipeline'):
-                    ppln_id = line_split[2]
-                next_line = next(f)
-                widget_id = next_line.split('|')[0].split(':')[-1].strip()
-                # convert to the same ID format in mtrace
-                widget_id = int('0x' + widget_id[-6:], 16)
+                    # remove ending comma(,) with [:-1]
+                    ppln_id = int(line_split[5][:-1])
+                    continue
+                # remove ending comma(,) with [:-1]
+                module_instance_id = int(line_split[7][:-1])
+                # remove ending comma(,) with [:-1]
+                widget_id = int(line_split[5][:-1])
+                # final module id are composed with high16(module instance id) + low16(module id)
+                widget_id |= module_instance_id << 16
                 comp_name[str(Component(ppln_id, widget_id))] = widget_name
 
     col_data = pd.DataFrame(


tools/sof_perf_analyzer.py
```diff
@@ -204,8 +205,29 @@ def process_kmsg_file():
...
+                    print("no valid widget ID found, try with bind information")
+                    not_found[widget_name] = ppln_id
+                    continue
                 comp_name[str(Component(ppln_id, widget_id))] = widget_name
+            elif match_obj := re.search(r' bind ', line):
+                if not_found:
+                    # parse current line to get widget name
+                    span_end_pos = match_obj.span()[1] - 1
+                    line_split = line[span_end_pos + 1:].split()
+                    widget_name = line_split[0].split(':')[0]
```
aiChaoSONG, line 220: we got two widget_name (s) and two widget_id (s) from the bind message, you need to handle both, what if the not_found widget is the second widget in bind message?

tools/sof_perf_analyzer.py
```diff
@@ -204,8 +205,29 @@ def process_kmsg_file():
...
+                    # parse next line get widget ID
+                    next_line = next(f)
+                    widget_id = next_line.split('|')[0].split(':')[-1].strip()
+                    widget_id = int('0x' + widget_id[-6:], 16)
+                    # get ppln_id from dictionary
+                    if not_found[widget_name]:
+                        ppln_id = not_found[widget_name]
+                        comp_name[str(Component(ppln_id, widget_id))] = widget_name
+                        # delete not_fund
```
aiChaoSONG, line 229: comment is redundant, `del ...` is self-explanatory.

tools/sof_perf_analyzer.py
```diff
@@ -204,8 +205,29 @@ def process_kmsg_file():
...
+                    not_found[widget_name] = ppln_id
+                    continue
                 comp_name[str(Component(ppln_id, widget_id))] = widget_name
+            elif match_obj := re.search(r' bind ', line):
+                if not_found:
+                    # parse current line to get widget name
+                    span_end_pos = match_obj.span()[1] - 1
+                    line_split = line[span_end_pos + 1:].split()
+                    widget_name = line_split[0].split(':')[0]
+                    # parse next line get widget ID
```
aiChaoSONG, line 221: parse next line to get widget ID.

tools/sof_perf_analyzer.py
```diff
@@ -204,8 +205,29 @@ def process_kmsg_file():
...
+                    next_line = next(f)
+                    widget_id = next_line.split('|')[0].split(':')[-1].strip()
+                    widget_id = int('0x' + widget_id[-6:], 16)
+                    # get ppln_id from dictionary
+                    if not_found[widget_name]:
+                        ppln_id = not_found[widget_name]
```
aiChaoSONG, line 227: `if ppln_id := not_found[widget_name]`, this removes the repetition of `not_found[widget_name]`

aiChaoSONG (changes requested):

tools/sof_perf_analyzer.py
```diff
@@ -204,8 +205,29 @@ def process_kmsg_file():
...
                 # convert to the same ID format in mtrace
-                widget_id = int('0x' + widget_id[-6:], 16)
+                try:
+                    widget_id = int('0x' + widget_id[-6:], 16)
+                except ValueError:
+                    print("no valid widget ID found, try with bind information")
+                    not_found[widget_name] = ppln_id
+                    continue
                 comp_name[str(Component(ppln_id, widget_id))] = widget_name
+            elif match_obj := re.search(r' bind ', line):
```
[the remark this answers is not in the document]
btian1: it is not search 'bind', it is ' bind ', with current dmesg logs, all the search is valid
this is simple and works, how about keep as for now? since this works, it can keep implementation
simple.
image

tools/sof_perf_analyzer.py
btian1: the code are making sure we are getting the first widget name and first widget ID.
as you see, there is [0] for widget_name, we always use first to match the case.

tools/sof_perf_analyzer.py
btian1: at this moment, ppln_id is still valid? 
if not_found[widget_name]
this condition means: with current parsed widget_name, if meet this condition, means we need setup dictionary for this widget, any logic wrong?

tools/sof_perf_analyzer.py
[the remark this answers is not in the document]
marc-hb: > with current dmesg logs, all the search is valid

This is called "programming by coincidence"

tools/sof_perf_analyzer.py
```diff
@@ -182,11 +182,19 @@ def process_kmsg_file():
     [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
     ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
 
+    However, for some cases, the above lines does not contain widget ID information, in this
```
marc-hb, line 185: Which cases? And what exactly is missing in these cases and why? An entire line and if yes which one? A part of a line and if yes which part? You must be much more specific, otherwise it's impossible to review the code without knowing what it is trying to parse.

tools/sof_perf_analyzer.py
```diff
@@ -204,8 +212,31 @@ def process_kmsg_file():
                 next_line = next(f)
                 widget_id = next_line.split('|')[0].split(':')[-1].strip()
                 # convert to the same ID format in mtrace
-                widget_id = int('0x' + widget_id[-6:], 16)
+                try:
+                    widget_id = int('0x' + widget_id[-6:], 16)
+                except ValueError:
+                    print(f'no valid widget ID found for {widget_name}, try with bind message')
+                    not_found[widget_name] = ppln_id
+                    continue
```
marc-hb, line 220: Don't mix `next()` and `continue` in the same `for` loop, this is incredibly hard to follow.

marc-hb (changes requested):

marc-hb:
Long before we search the info in a different place, could we have a "quicker fix" where the entire script does not crash but replaces the ID with some "Unknown" category or something?

tools/sof_perf_analyzer.py
marc-hb: BTW there `next_line_if_there_is_one = next(f, None)` if needed. It should probably be used here like this:

```python
next_line = next(f, None)
if not next_line:
   # use some "NotFound" placeholder
   # so the code does not crash later
   # other cleanup if needed.
   raise StopIteration

```

tools/sof_perf_analyzer.py
btian1: https://github.com/thesofproject/sof-test/issues/1108

tools/sof_perf_analyzer.py
btian1: 1. continue is correct, still need check the following lines in for loop.
2. ok

tools/sof_perf_analyzer.py
marc-hb: Thanks, I had heard about 1108 but besides for extraordinary complex situations anyone with basic Python knowledge should be able to understand the code without opening bugs which are not even linked from the source. This code is not an extraordinary complex situation, it's just parsing some logs so it must be easy to follow. Thanks for adding an example!

```suggestion
However, for some cases, the line after "Creating" does not contain widget ID information, in this
```

Also explain why the line after "Creating" does not always contain widget ID information, let's not "program by coincidence".

tools/sof_perf_analyzer.py
marc-hb: > 1. continue is correct, still need check the following lines in for loop.

I didn't say it was incorrect. I said it was unreadable.

Now I understand the `bind` fallback is actually in a totally different part of the logs. In that case use two different `for` loops. There's no need to look for both at the same time, it makes the code incredibly hard to follow.

marc-hb (changes requested):

tools/sof_perf_analyzer.py
```diff
@@ -274,8 +271,8 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-    comp_peak_avg_cycles = perf_info.groupby('COMP_ID')['CPU_PEAK'].mean()
-    perf_stats['MODULE_CPC'] = comp_peak_avg_cycles * CPC_MARGIN
+    comp_peak_max_cycles = perf_info.groupby('COMP_ID')['CPU_PEAK'].max()
+    perf_stats['MODULE_CPC'] = comp_peak_max_cycles
```
aiChaoSONG, line 275: `perf_stats['MODULE_CPC'] = perf_info.groupby('COMP_ID')['CPU_PEAK'].max()` single line is good.

Let's make this commit a standalone PR, because
- it has nothing to do with widget id issue
- it can be merged quickly

aiChaoSONG:
@btian1 For the widget id issue, I am afraid we have to refine code and comments according Marc's review comments. Let's sync tomorrow.

tools/sof_perf_analyzer.py
aiChaoSONG: @btian1 Move `comp_name[str(Component(ppln_id, widget_id))] = widget_name` to the try block, and remove `continue`

aiChaoSONG (dismissed):

tools/sof_perf_analyzer.py
```diff
@@ -237,8 +271,7 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-    comp_peak_avg_cycles = perf_info.groupby('COMP_ID')['CPU_PEAK'].mean()
-    perf_stats['MODULE_CPC'] = comp_peak_avg_cycles * CPC_MARGIN
+    perf_stats['MODULE_CPC'] = perf_info.groupby('COMP_ID')['CPU_PEAK'].max()
```
marc-hb, line 274: This second commit does not seem related at all to the first one. Is it?

tools/sof_perf_analyzer.py
```diff
@@ -182,11 +179,19 @@ def process_kmsg_file():
     [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
     ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
 
+    However, for some cases, the line after "Creating" does not contain widget ID information,
```
marc-hb, line 182: In which cases? Is that not a kernel bug that this PR would then hide?

cc: @plbossart , @ranj063

tools/sof_perf_analyzer.py
```diff
@@ -201,11 +206,40 @@ def process_kmsg_file():
...
+                        # reach EOF, exit.
+                        break
+                    widget_ids = next_line.split(': ')[-2].split('|')
+                    widget_ids = int('0x' + widget_ids[0][-6:], 16), int(widget_ids[1], 16)
+
+                    widgets = dict(zip(widget_names, widget_ids))
+
+                    # search and set missed comp_name
+                    for widget_name, widget_id in widgets.items():
+                        if ppln_id := not_found.get(widget_name):
```
marc-hb, line 240: This `not_found` makes the code significantly more complicated for no obvious benefit, please get rid of it. It should be as simple as:

1) Use the `for line` loop to build some `id + name` dictionary. Don't touch `comp_name` yet. When you find a `bind` line, simply give `Creating` lines precedence in the temporary dictionary.

2) AFTER the `for line` loop, use the temporary and complete dictionary to create `comp_name`.

marc-hb (changes requested):

tools/sof_perf_analyzer.py
aiChaoSONG: @marc-hb Component class have pipeline_id and comp_id, two parts here. 
from `create widget` message we could parse widget(component) name and pipeline id, and from next line we parse comp_id. on next line parsing error, we only have pipeline id and widget name.

from bind message, we parse widget name, and from its next line, we parse comp_id. bind message is complementary information if we failed to parse some 'create message'. 

in both `create widget` and bind message parsing, two lines are involved. we cannot assume it is atomic. because the concurrent nature of the linux kernel.

We are use two types of information to get more Component class here.  there is still probability a component in the topology is missing in our comp_name dict, but the probability is low as we parse more and more information ('create widget', 'bind message', 'unbind message'), but in practice, with 'create widget' and bind message,  the error rate can be think as low as zero.

marc-hb: Thanks @aiChaoSONG for the explanation. The length of your explanation demonstrates that the data being parsed is complex for various reasons. It's not possible to review this parsing code for correctness without understanding the data it's trying to parse. So examples are very useful but they are not enough. I'm afraid this is one of these rare cases where it's very hard to make the code "self-explanatory" and where a lot more comments than usual are required in the source code. For sure a lot of what you just wrote should make its way to the source code.

People's assignments come and go and we cannot have only 1.5 people in the team who can understand this code (or any code for that matter)

"“Programs must be written for people to read, and only incidentally for machines to execute.” - this is especially true in Python code which is very slow to execute (but usually much faster to read).

Also, forcing oneself to explain the code (with comments) very often results in finding simplifications; writing comments is a good example of https://en.wikipedia.org/wiki/Rubber_duck_debugging

tools/sof_perf_analyzer.py
plbossart: @btian1 what exactly is the problem statement? You are suggesting a solution without giving any context or alternate suggestion. That's not very helpful, I have no idea what this patch does.

plbossart (dismissed):
no idea what this is about

plbossart:
"When there is no widget ID in kernel logs, try to find the widget ID in bind related logs."

Has anyone tried to change the kernel logs to make the problem go away?

marc-hb (changes requested):
> in both create widget and bind message parsing, two lines are involved. we cannot assume it is atomic. because the concurrent nature of the linux kernel.
> [...]
> there is still probability a component in the topology is missing in our comp_name dict, but the probability is low as we parse more and more information

I just had a quick chat with @plbossart and he told me that there has been zero effort to simplify the kernel logs... is this real?

tools/sof_perf_analyzer.py
```diff
@@ -274,8 +271,7 @@ def analyze_perf_info():
     perf_stats.columns = ['CPU_AVG(MIN)', 'CPU_AVG(AVG)', 'CPU_AVG(MAX)',
                           'CPU_PEAK(MIN)', 'CPU_PEAK(AVG)', 'CPU_PEAK(MAX)']
     perf_stats['PEAK(MAX)/AVG(AVG)'] = perf_stats['CPU_PEAK(MAX)'] / perf_stats['CPU_AVG(AVG)']
-    comp_peak_avg_cycles = perf_info.groupby('COMP_ID')['CPU_PEAK'].mean()
-    perf_stats['MODULE_CPC'] = comp_peak_avg_cycles * CPC_MARGIN
+    perf_stats['MODULE_CPC'] = perf_info.groupby('COMP_ID')['CPU_PEAK'].max()
```
plbossart, line 274: it's also arguable that the CPC value needs to be e.g. 1.2 the max measured value to account for unknowns such as cache and IPC.

 CPC = measured_peak_cpu * CPC_MARGIN

tools/sof_perf_analyzer.py
btian1: @plbossart , please check: https://github.com/thesofproject/sof-test/issues/1108
there is a possibility of log missed for kernel log output, in this case, this patch try to parse bind related info to get widget ID, widget name and so on.

@marc-hb , I checked kernel code, there should have no bug in kernel, just kernel log missing with a possibility:
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/ipc4.c#L222
@plbossart , please double confirm.

tools/sof_perf_analyzer.py
btian1: @aiChaoSONG shall we go with itertools.pairwise？

tools/sof_perf_analyzer.py
btian1: ok, peak is already much larger than avg(2-5 times), unless all module peak happen in the same time, then the budget should be ok.

tools/sof_perf_analyzer.py
btian1: yes, if you want to have a separate PR, I will do it.

tools/sof_perf_analyzer.py
plbossart: I will double-confirm that I don't understand the problem statement and why you are not trying to fix the issue (if it's real) at the source by updating the kernel logs.

btian1: Thanks, @plbossart , let me use logs to describe:
expected logs:
[61230.667725] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget mixin.20.1 instance 0 - pipe 20 - core 0
[61230.667728] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x40000002|0xa: MOD_INIT_INSTANCE [data size: 40]

we can get name(mixin.20.1) and its module ID: 2(0x40000002).

unexpected logs(may not have errror in kernel):

[  262.012482] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget alh-copier.SDW0-Capture.0 instance 8 - pipe 11 - core 0
[  262.012486] kernel: snd_sof_intel_hda_common:hda_dsp_stream_trigger: sof-audio-pci-intel-mtl 0000:00:1f.3: FW Poll Status: reg[0xa0]=0x2024001e successful

[  261.979993] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget pipeline.31 instance 4 - pipe 31 - core 0
[  261.980114] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x33000000|0x1: GLB_SET_PIPELINE_STATE

[61230.948775] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget gain.15.1 instance 4 - pipe 15 - core 0
[61230.948777] kernel: snd_sof:sof_pcm_trigger: sof-audio-pci-intel-mtl 0000:00:1f.3: pcm: trigger stream 0 dir 0 cmd 1

so we get those info from bind related logs:
[61230.670335] kernel: snd_sof:sof_ipc4_route_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: bind host-copier.2.playback:0 -> gain.20.1:0
[61230.670338] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x45000004|0x6: MOD_BIND

get name(host-copier.2.playback), ID 4(0x45000004).

this is this patch's purpose.

plbossart: I am unable to follow what's missing or not right in the logs above.

It's rather disappointing that after multiple asks I still get this sort of raw dump of information without clarity on the problem statement. This is starting to be a waste of everyone's time if I am honest.

marc-hb (dismissed):
> It's rather disappointing that after multiple asks I still get this sort of raw dump of information without clarity on the problem statement. This is starting to be a waste of everyone's time if I am honest. - @plbossart 


_About two weeks ago, 4 days after this PR was first submitted_, I asked this:
https://github.com/thesofproject/sof-test/pull/1122#discussion_r1382264896

> Which cases? And what exactly is missing in these cases and why? An entire line and if yes which one? A part of a line and if yes which part? You must be much more specific, otherwise it's impossible to review the code without knowing what it is trying to parse.

That question was marked as "resolved" without an answer.

At the same time I asked basically the same question in https://github.com/thesofproject/sof-test/issues/1108. No answer either.


It's _pointless_ to submit a fix to a problem that has not been analyzed, understood and explained. It's called "programming by coincidence".


1. **Why are the kernel logs incomplete, how and when?** 
1.  **Why can't the kernel logs be fixed instead?** Or at least made somewhat better.

Don't answer in this PR here, answer in the bug 1108

tools/sof_perf_analyzer.py
bardliao: @plbossart I chatted with @btian1. The purpose is to connect the widget name and the IPC primary data. Like `mixin.20.1` <-> `0x40000002`.
The existing code get the information from dmesg log when it looks for `Create widget` and parse the log in 2 lines. It works in most cases, but as you know, the kernel is multi thread and there is no guarantee that the `ipc tx :` log will be right after the `Create widget` log. And the existing code will return error in that case.
The commit will use another key work `bind` to get the IPC primary data. @btian1 @aiChaoSONG Please correct me if I was wrong.

It should be easy to "solve" the issue if we can add msg->primary to 
```
dev_dbg(sdev->dev, "Create widget %s instance %d - pipe %d - core %d\n",
                swidget->widget->name, swidget->instance_id, swidget->pipeline_id, swidget->core);
```
Note that I don't mean that we need to fix it from kernel. But trying to explain the issue they are trying to fix.

marc-hb: > It should be easy to "solve" the issue if we can add msg->primary to

Sounds great to me! @plbossart ?

plbossart: The IPC is serialized so the order will not vary. There might be other things in between IPCs but they will be by construction based on different/irrelevant information.

So yes if we can change the logs there's no issue. This isn't an ABI, we can tune the logs to the needs of profiling tools.

ranj063: Sorry for jumping in a bit late. This is not an error in the kernel ogs nor an unexpected sequence of logs. When widgets in a pipeline are set up, the first widget in the pipeline sets up the pipeline before the widget itself gets set up. So please do not assume that the log ```dev_dbg(sdev->dev, "Create widget %s instance %d - pipe %d - core %d\n",...```. If you do , please fix it instead of assuming incorrect sequences.

plbossart: not able to follow your answer @ranj063, there's something missing in the sentence "do not assume that the log"

ujfalusi: The swidget is a Linux SOF concept and it represents all sorts of 'Sof WIDGETs'.
The pipeline is also a swidget and that is not  a module, it has no 'module ID'

ujfalusi: My proposal is that the kernel should print two version for swidget creation.
For 'modules': `Create widget for module: %s, module id: %u, instance id: %u, pipeline: %u, core: %u`
For 'pipelines': `Create widget for pipeline: %s, id: %u, instance id: %u, core: %u`

The pipeline_id is a Linux driver concept, it is good to have it out, but not sure what would be the best way.

tools/sof_perf_analyzer.py
```diff
@@ -175,12 +175,14 @@ def process_trace_file():
...
-    [  334.818442] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3:
-    ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
+    component name and ID are acquired from the line that contains 'Create widget',
+    and pipeline ID is acuquired from the line that contains 'Create pipeline'
+    Below are example logs for parsing:
+    [   59.622134] snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    Create pipeline pipeline.3 (pipe 3) - instance 3, core 0
+
+    [   59.622645] snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3:
+    Create widget host-copier.0.capture (pipe 3) - ID 4, instance 3, core 0
```
aiChaoSONG, line 186:
```
'''Process the dmesg to get component ID to component name mapping,
they are acquired from the line that contains 'Create widget':
    EXAMPLE_OF_CREATE_WIDGWT_LINE
By design in the kernel, pipeline ID is the instance ID of pipeline
widget, so it is acquired from the line that contains 'Create pipeline':
    EXAMPLE_OF_CREATE_WIDGET_LINE
'''
```

tools/sof_perf_analyzer.py
```diff
@@ -190,21 +192,24 @@ def process_kmsg_file():
...
-                # firmware. Note that ppln_id variable will be assigned properly at pipeline widget
-                # creation, because it is always the first one to be created before all other
-                # widgets in the same pipeline and pipelines are created sequentially.
+                # with sof kernel PR: https://github.com/thesofproject/linux/pull/4709
+                # pipeline and module creation are using separate print in kernle log.
+                # So, for pipeline case, pipeline instance ID will be extracted from pipeline
+                # logs. For module creation, module instance ID and module ID will be
+                # extracted accordingly.
+                # ppln_id, widget_id and widget instance ID will be combined together to
+                # column COMP_ID as the unique ID for each module inside pipelines.
```
aiChaoSONG, line 205: keep my original comments, it explains why instance id of pipeline widget is the widget pipeline id for the widgets of the pineline

tools/sof_perf_analyzer.py
```diff
@@ -190,21 +192,24 @@ def process_kmsg_file():
...
-                    ppln_id = line_split[2]
-                next_line = next(f)
-                widget_id = next_line.split('|')[0].split(':')[-1].strip()
-                # convert to the same ID format in mtrace
-                widget_id = int('0x' + widget_id[-6:], 16)
+                    ppln_id = int(line_split[5][:-1])
+                    continue
+
+                module_instance_id = int(line_split[7][:-1])
+                widget_id = int(line_split[5][:-1])
```
aiChaoSONG, line 211: same, explain `[:-1]` with a simple comment.

tools/sof_perf_analyzer.py
```diff
@@ -190,21 +192,24 @@ def process_kmsg_file():
...
+                # extracted accordingly.
+                # ppln_id, widget_id and widget instance ID will be combined together to
+                # column COMP_ID as the unique ID for each module inside pipelines.
                 if widget_name.startswith('pipeline'):
-                    ppln_id = line_split[2]
-                next_line = next(f)
-                widget_id = next_line.split('|')[0].split(':')[-1].strip()
-                # convert to the same ID format in mtrace
-                widget_id = int('0x' + widget_id[-6:], 16)
+                    ppln_id = int(line_split[5][:-1])
```
aiChaoSONG, line 207: explain `[:-1]`, no one know why if you don't explain.
`remove ending comma(,) with [:-1]`

aiChaoSONG (changes requested):
Commit title and body:
```
tools: sof_perf_analyzer: fix widget ID parsing error

In existing code, we parse component name from the line
that contains 'Create widget', and component ID from the
next line, which normally is the ipc tx message:

Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]

However, due to the concurrent nature of linux kernel, the
assumption that `Create widget` message and `ipc tx` message
are always adjacent is wrong. An exception will be raised
if the next line is not the 'ipc tx' message.

The first step to fix the issue is to make a kernel PR to print
all the information we need in a single 'Create widget' line,
thus to avoid the requirement for multiple lines of kernel message, 
see https://github.com/thesofproject/linux/pull/4709.

And next，this patch will parse the new 'Create widget' line to
get component name and ID.

Fixes: https://github.com/thesofproject/sof-test/issues/1108
```

Always describe the issue (what), root cause of the issue (why), and how it is fixed(how).

aiChaoSONG:
see internal test report 35019， test result is good.

aiChaoSONG (approved):

PR #1124, opened 2023-10-31
sof-kernel-log-check.sh: ignore unrelated ACPI error on LNL platform

keqiaozhang:


plbossart (dismissed):

tools/sof-kernel-log-check.sh
```diff
@@ -261,6 +261,9 @@ ignore_str="$ignore_str"'|asix .-.+\..:.\.. en.+: Failed to .+'
 # https://github.com/intel-innersource/drivers.audio.ci.sof-framework/issues/246
 ignore_str="$ignore_str"'|mei_me 0000:00:16\..\: .+'
 
+# ignore the ACPI error on LNL.
+ignore_str="$ignore_str""|kernel: ACPI: \\\\: Can't tag data node"
```
marc-hb, line 265: Do we have any idea what this error means? We do use ACPI. If yes please add a comment. Is it safe to ignore across all products? If not there is a `lnl` platform section below.

keqiaozhang: The errors come from https://github.com/thesofproject/linux/blob/bf67942088295fc31042c8d8d71a212c53a9134c/drivers/acpi/property.c#L366C1-L386C2.  
It means that data nodes failed to tie to ACPI handles. I don't know much about this, but from my perspective it's harmless to audio.

aiChaoSONG (approved):

marc-hb (approved):

keqiaozhang:
> @keqiaozhang any idea why jenkins did not pick that up?

It seems that Jenkins failed to post the status to GitHub, I filed a bug https://github.com/intel-innersource/drivers.audio.ci.sof-framework/issues/525 to track it.  I spent hours on it, but no clues so far.

PR #1125, opened 2023-11-03
verify-kernel-boot-log: add new detect_log_flood()

marc-hb:
We need to catch issues like
https://github.com/thesofproject/sof-test/discussions/998 before they bring the entire system down


marc-hb:
The results are not published because of internal issue 525 but the tests ran. I checked the results internally (34134, 34135 and 34136) and they're all good.

fredoh9 (approved):
this is good idea!

PR #1126, opened 2023-11-03
multiple-pause-resume.sh: Check for aplay/arecord pid when waiting

ujfalusi:
It has been observed that the expect process has been terminated (pidof returns with no PID) but the child aplay is still running. This only happens with ChainDMA enabled PCMs where the time to stop the stream takes more than 100ms (from trigger stop to PCM close). With the right timing on such PCM the pidof expect will break the wait loop and we progress to the next iteration which - again with right timing and setup - might fail if the aplay is still in a process of closing and the next iteration includes the same PCM.

Add the aplay and arecord commands to the pidof command to make sure that the iteration has been completed before starting a new one. We will still catch timeouts and in that case we  should kill the stuck processes.

### DIFF
--- a/test-case/multiple-pause-resume.sh
+++ b/test-case/multiple-pause-resume.sh
@@ -157,22 +157,28 @@ do
             func_pause_resume_pipeline "$idx"
             pid_lst=("${pid_lst[@]}" $!)
         done
-        # wait for expect script finished
-        dlogi "wait for expect process finished"
+        # wait for aplay/arecord finished
+        dlogi "wait for expect/aplay/arecord process finished"
         iwait=$max_wait_time
         while [ $iwait -gt 0 ]
         do
             iwait=$((iwait - 1))
             sleep 1s
-            [[ ! "$(pidof expect)" ]] && break
+            [[ ! "$(pidof expect aplay arecord)" ]] && break
         done
-        # fix aplay/arecord last output
+        # Catch timout after the wait loop
         echo
-        if [ "$(pidof expect)" ]; then
+        if [ "$(pidof expect aplay arecord)" ]; then
             dloge "Still have expect process not finished after wait for $max_wait_time"
-            # list aplay/arecord processes
+            # list expect/aplay/arecord processes
+            pgrep -a -f expect || true
             pgrep -a -f aplay || true
             pgrep -a -f arecord || true
+
+            # kill aplay/arecord
+            sof-process-kill.sh ||
+                dlogw "Kill process catch error"
+
             exit 1
         fi
         # now check for all expect quit status


ujfalusi:
400 iteration on local device:
```
TPLG=/lib/firmware/intel/sof-ipc4-tplg/sof-hda-generic.tplg MODEL=TGLU_UP_HDA_IPC4ZPH SOF_TEST_INTERVAL=5 ~/work/sof-test/test-case/multiple-pause-resume.sh -l 400 -r 5
....
declare -- cmd="kernel_logger_cmd --since=@1699008086"
2023-11-03 10:41:44 UTC [INFO] Starting func_exit_handler(0)
2023-11-03 10:42:58 UTC [INFO] pkill -TERM -f mtrace-reader.py
2023-11-03 10:42:58 UTC [INFO] nlines=1361430 /home/sof/work/sof-test/logs/multiple-pause-resume/2023-11-03-10:47:48-6916/mtrace.txt
2023-11-03 10:44:12 UTC [INFO] ktime=9574 sof-test PID=22522: ending
2023-11-03 10:44:12 UTC [INFO] Test Result: PASS!

```

marc-hb (commented):
Except for cleanups, we should stop using pkill and pgrep and use pid_list instead.

keqiaozhang:
SOFCI TEST

marc-hb:
Still no jenkins test results published :-(

Screenshot 2023-11-09 at 12 54 26

marc-hb:
A fix for https://github.com/thesofproject/sof/issues/8445 has been merged. @ujfalusi is this still needed? Naive question sorry.

ujfalusi:
> A fix for [thesofproject/sof#8445](https://github.com/thesofproject/sof/issues/8445) has been merged. @ujfalusi is this still needed? Naive question sorry.

While now ChainDMA will stop faster (and pause) and we might not see the issue this patch is fixing, I'm fine to close it, but we should note that the `pidof expect` is not an indication that aplay/arecord has been actually terminated.

marc-hb:
@ujfalusi I completely rewrote the Expect script in commit 006946cceaac. Can you re-test and see if that makes any difference for this PR?

marc-hb:
SOFCI TEST

marc-hb:
SOFCI TEST

marc-hb (approved):

marc-hb:
This looks simple and useful: merged.

@ujfalusi can you please take a look at failure https://sof-ci.01.org/softestpr/PR1126/build682/devicetest/index.html?model=LNLM_RVP_NOCODEC&testcase=multiple-pause-resume-50 and the previous one above? Does it give the extra information that you were expecting?

ujfalusi:
Both shows that after a set of timeout there is an expect and aplay running.
Looks to me that the internal expect timeout is not triggered and the main script kills expect/aplay, then the expect fort he aplay process hits EOL (aplay is killed).

I'm not sure what to make of the whole issue, but for sure we had a stall aplay process.

What I'm not sure with the rewritten expect script is the impact of the `sof-process-kill.sh` call... It kills the stall processes and that forces the expect to exit as well, but not sure what would happen if it is not done.

The history of this change was that in certain cases the test reported an error on subsequent iteration, depending on few factors and that was hard to piece back together.

PR #1128, opened 2023-11-16
Introduce itertools.pairwise

aiChaoSONG:
In the sof_perf_analyzer, there are some conditions that we need to process two items for a generator, for example, in timestamp correction, we need two compare two timestamps from two trace item, and in kernel message parsing, we need two lines of log to extract component name, component ID and pipeline ID. itertools.pairwise could provide great help under those condition, and help to simplify the code more or less in code lines or readability.

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -21,6 +21,7 @@ is used to extract component name.
 import re
 import pathlib
 import argparse
+import itertools
 from typing import TextIO
 from typing import Generator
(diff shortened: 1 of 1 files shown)


keqiaozhang:
SOFCI TEST

aiChaoSONG:
For kernel message analysis, we won't analyze a pair of messages anymore by making change to kernel log, see https://github.com/thesofproject/linux/pull/4709, itertools.pairwise is not that required, close.

PR #1133, opened 2023-11-30
kmod: sof_remove: add dependencies between platforms

plbossart:
With https://github.com/thesofproject/linux/pull/4703, TGL and ICL depend on CNL, and LNL on MTL. The order in which the modules are removed needs to respect those dependencies.

Link: https://sof-ci.01.org/linuxpr/PR4703/build611/devicetest/index.html

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -76,16 +76,21 @@ remove_module snd_usb_audio
 #-------------------------------------------
 # Top level devices
 # ACPI is after PCI due to TNG dependencies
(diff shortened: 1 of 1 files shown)


marc-hb (approved):
Can you confirm the order did not matter in the past?

In other words, no backwards compatibility issue?

plbossart:
> Can you confirm the order did not matter in the past?
> 
> In other words, no backwards compatibility issue?

yes, the dependency is added by a set Kconfig select additions which did not exist before.

PR #1136, opened 2023-12-01
tools: sof_perf_analyzer: change module CPC calculation

btian1:
Previously, module CPC is calculated by multiplying the mean value of cpu_peak with a margin scalar, however, abnormal peak cycles are finally proved that it is caused by IRQ, with further team discussion, an agreement was reached to use module average multiplied with a margin as new way for CPC calculation.

### DIFF
--- a/tools/sof_perf_analyzer.py
+++ b/tools/sof_perf_analyzer.py
@@ -27,8 +27,14 @@ from dataclasses import dataclass
 
 import pandas as pd
 
-# currently, keep CPC = max(module peak) * CPC_MARGIN
-CPC_MARGIN = 1.1
+# CPC_MARGIN is set to 1.5, because there is some inactive code for some module
+# due to unmet condition. For example:
+# volume: ramp operation only run on volume change, but we don't do volume
+# change in our test.
+
+# So, we set a relative high margin for avg cycles.
+# CPC = AVG(module) * CPC_MARGIN
+CPC_MARGIN = 1.5
 
 @dataclass()
 class TraceItem:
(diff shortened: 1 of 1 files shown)


marc-hb:
Is this basically hiding very short audio glitches?

tools/sof_perf_analyzer.py
```diff
@@ -27,8 +27,13 @@
...
 import pandas as pd
 
-# currently, keep CPC = max(module peak) * CPC_MARGIN
-CPC_MARGIN = 1.1
+# CPC_MARGIN set to 1.5 because some modules have not running code, for example
+# volume: volume have two kind of ramp type, linear and log, it only be runned when
+# volume change happens.
+# copier: there is attenuation inside, normally, it will not be runned.
+# So we set a relative high margin for avg cycles.
+# CPC = AVG(module) * CPC_MARGIN
```
aiChaoSONG, line 35: passive voice for run is run, not runned.
I don't think copier example is a good one, I don't think attenuation will impact the `copier_process` ops

```
CPC_MARGIN is set to 1.5, because there is some inactive code for some module
due to unmet condition. For example:
volume: ramp operation only run on volume change.

So, we set a relative high margin for avg cycles.
CPC = AVG(module) * CPC_MARGIN
```

aiChaoSONG (commented):
code looks good, let's revise commit message a little bit:
```
In daily performance test, we have a high and abnormal peak
cycle for some module, recently, we find this abnormal peak
cycle is caused by IRQ.

Previously, module CPC is calculated by multiplying the mean
value of cpu_peak with a margin scalar. With further team
discussion based on the new finding, an agreement is reached
to use the product of module average cycle and a margin as
module CPC.
```

tools/sof_perf_analyzer.py
btian1: for copier, there are three main part, copier frame work, copier copy, and attenuation, it is a suitable case for set margin to 1.5.

Anyway, I don't care which case is suitable or not, it is in my mind, I changed based on your suggestion.

marc-hb:
> @marc-hb ， I am not able to open this link, could you stick to this PR itself?

I'm not sure about the copyright sorry. I never had any problem loading that page.

> does above link related with this PR?

Yes, it explains that averages are generally useless. Even more useless in "real-time" situation like audio .

aiChaoSONG:
> With further team discussion based on the new finding, an agreement is reached
> to use the product of module average cycle and a margin as module CPC.

Can we have at lease one team member that agrees with this method review this pull request?

aiChaoSONG (approved):

mengdonglin:
@lgirdwood has agreed to use average cycles with a fix ratio for CPC to upstream in FW manifest. So that we can use CPC in FW manifest to monitor potential performance regression in daily performance test in the future. 

So the method in this PR 'CPC = AVG(module) * CPC_MARGIN' with CPC_MARGIN =1.5 looks good to me. @aiChaoSONG  @marc-hb

aiChaoSONG:
Failure not related

PR #1141, opened 2023-12-21
test-case: add socwatch parsing toool to generate graph

fredoh9:
It can generate a graph from text output. matplotlib is new dependency.

### DIFF
--- a/tools/sof-socwatch-parser.py
+++ b/tools/sof-socwatch-parser.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+# Fred: Add new dependency
+import matplotlib.pyplot as plt
+
+# Define lists to store C-State names and percentages
+c_states = []
+percentages = []
+
+# Check if the input file parameter is provided
+if len(sys.argv) != 2:
+    print("Usage: python script_name.py input_file.txt")
+    sys.exit(1)
+
+input_file = sys.argv[1]
+
+# Read the input file and extract relevant data
+with open(input_file, "r") as file:
+    lines = file.readlines()
+
+# Iterate through the lines and extract C-State data
+for line in lines:
+    # When line starts with PC
+    if line.strip().startswith("PC"):
+        columns = line.strip().split(',')
+        c_state = columns[0].strip()
+        percentage = float(columns[1].strip())
+        c_states.append(c_state)
+        percentages.append(percentage)
+
+# Plotting the data
+plt.figure(figsize=(10, 6))
+plt.bar(c_states, percentages, color='blue')
+plt.xlabel('C-State')
+plt.ylabel('Residency (%)')
+plt.title('Package C-State Residency')
+plt.ylim(0, 100)  # Set y-axis limit to percentage values
+plt.xticks(rotation=45)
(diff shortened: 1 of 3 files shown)


tools/sof-socwatch-parser.py
```diff
@@ -0,0 +1,52 @@
...
+import matplotlib.pyplot as plt
+
+# Define lists to store C-State names and percentages
+c_states = []
+percentages = []
+
+# Check if the input file parameter is provided
+if len(sys.argv) != 2:
+    print("Usage: python script_name.py input_file.txt")
+    sys.exit(1)
```
marc-hb, line 15: `argparse` is really easy to use and there are plenty examples.

tools/sof-socwatch-parser.py
```diff
@@ -0,0 +1,52 @@
...
+
+input_file = sys.argv[1]
+
+# Read the input file and extract relevant data
+with open(input_file, "r") as file:
+    lines = file.readlines()
+
+# Iterate through the lines and extract C-State data
+for line in lines:
```
marc-hb, line 24: For now it does not look like you  need to read the whole file at once.

tools/sof-socwatch-parser.py
```diff
@@ -0,0 +1,52 @@
...
+# Get the directory and base filename of the input file
+input_dir = os.path.dirname(input_file)
+input_filename = os.path.basename(input_file)
+
+# output_file has same path and same base filename but .png extension
+output_file = os.path.join(input_dir, os.path.splitext(input_filename)[0] + '.png')
+plt.savefig(output_file)
+
+# Display the plot
+#plt.show()
```
marc-hb, line 52: This can be a new argparse option later. No need to add it now but good reason to use argparse now.

tools/sof-socwatch-parser.py
marc-hb, line 1: The main feature of this script is not parsing, parsing is only a means to an end. The end is `matplotlib`, so the script should be named something like `sof-socwatch-plot.py`

marc-hb: Please use the usual `main()` function:
https://docs.python.org/3/library/__main__.html
https://stackoverflow.com/questions/419163/what-does-if-name-main-do

- https://github.com/thesofproject/sof-test/issues/740

marc-hb (changes requested):
Are you sure SoCWatch has no better option? Surely we're not the first ones trying to re-use SoCWatch output. Either some already available python library or at least a more "parser-friendly" output.

EDIT: 1 minor pylint warning:
https://github.com/thesofproject/sof-test/actions/runs/7290667116/job/19867965010?pr=1141

marc-hb:
> Either some already available python library or at least a more "parser-friendly" output.

There is apparently some `-r` option.

I've also found mentions of a new "library interface". It wasn't ready at the time but this was last year, so maybe it's usable now.

marc-hb:
Demoting to draft as @fredoh9 is on vacation.

marc-hb:
SOFCI TEST

marc-hb:
@fredoh9 close, re-open if you find time for this?

BTW there's a git conflict

PR #1142, opened 2023-12-28
Codeowners: add code owner to performance script

btian1:
Add btian1 and andrula-song as script: sof_perf_analyzer.py codeowner.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,3 +9,4 @@
 # will be requested to review.
 
 tools/kmod/*            @thesofproject/test-maintainers @plbossart
+tools/sof_perf_analyzer.py @thesofproject/test-maintainers @btian1 @andrula-song


andrula-song (approved):

marc-hb:
GitHub reports this does not work because you don't have "write" (=maintainer) access.

The easiest way to become maintainer is to be a member of "test-maintainers". Then you will be notified of every PR, which is not many.

Per-file maintainership is really micro-management, I don't think this repo has enough traffic to require such micro-management.

btian1:
> GitHub reports this does not work because you don't have "write" (=maintainer) access.
> 
> The easiest way to become maintainer is to be a member of "test-maintainers". Then you will be notified of every PR anyway (which is not many) and this CODEOWNERS change becomes pointless.
> 
> Per-file maintainership is really micro-management, I don't think this repo has enough traffic to require such micro-management.

@marc-hb ， are we qualified add to test-maintainers?

btian1:
closing, as we are accepted by test-maintainer application.

PR #1144, opened 2024-01-03
[SKIP SOF-TEST] Simple dependency-free ALSA test rig for PCM capture analysis.

andyross:
Just drop this script on a test device to run it.  No tools to build, no dependencies to install.  Confirmed to run on Python 3.8+ with nothing more than the core libraries and a working libasound.so.2 visible to the runtime linker.

When run without arguments, the tool will record from the capture device for the specified duration, then emit the resulting samples back out the playback device without processing (except potentially to convert the sample format from s32_le to s16_le if needed, and to discard any channels beyond those supported by the playback device).

Passing --chirp-test enables a playback-to-capture latency detector: the tool will emit a short ~6 kHz wave packet via ALSA's mmap interface (which allows measuring and correcting for the buffer latency from the userspace process) and simultaneously loop on short reads from the capture device looking for the moment it arrives.

Passing --echo-test enables a capture-while-playback test.  The script will play a specified .wav file ("noise.wav" by default) for the specified duration, while simultaneously capturing, and report the "power" (in essentially arbitrary units, but it's linear with actual signal energy assuming the sample space is itself linear) of the captured data to stdout at the end of the test.

### DIFF
--- a/tools/capture-test.py
+++ b/tools/capture-test.py
@@ -1,0 +1,397 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright 2024 Google LLC
+# Author: Andy Ross <andyross@google.com>
+import os
+import re
+import sys
+import time
+import struct
+import random
+import argparse
+import ctypes as C
+
+HELP_TEXT="""
+Simple dependency-free ALSA test rig for PCM capture analysis.
+
+Just drop this script on a test device to run it.  No tools to build,
+no dependencies to install.  Confirmed to run on Python 3.8+ with
+nothing more than the core libraries and a working libasound.so.2
+visible to the runtime linker.
+
+When run without arguments, the tool will record from the capture
+device for the specified duration, then emit the resulting samples
+back out the playback device without processing (except potentially to
+convert the sample format from s32_le to s16_le if needed, and to
+discard any channels beyond those supported by the playback device).
+
+Passing --chirp-test enables a playback-to-capture latency detector:
+the tool will emit a short ~6 kHz wave packet via ALSA's mmap
+interface (which allows measuring and correcting for the buffer
+latency from the userspace process) and simultaneously loop on short
+reads from the capture device looking for the moment it arrives.
+
+Passing --echo-test enables a capture-while-playback test.  The script
+will play a specified .wav file ("noise.wav" by default) for the
+specified duration, while simultaneously capturing, and report the
+"power" (in essentially arbitrary units, but it's linear with actual
+signal energy assuming the sample space is itself linear) of the
+captured data to stdout at the end of the test.
+
+The tool supports a "--disable-rtnr" for the specific case of RTNR
+noise reduction, which has been observed to suppress the chirp test
+occasionally.  It does not otherwise change the runtime configuration
+of the ALSA device and expects it to be configured by the user for the
+specific case under test.
+"""
+
+def parse_opts():
+    global opts
+    ap = argparse.ArgumentParser(description=HELP_TEXT,
+                                 formatter_class=argparse.RawDescriptionHelpFormatter)
+    ap.add_argument("--disable-rtnr", action="store_true", help="Disable RTNR noise reduction")
+    ap.add_argument("-c", "--card", type=int, default=0, help="ALSA card index")
+    ap.add_argument("--pcm", type=int, default=16, help="Output ALSA PCM index")
+    ap.add_argument("--cap", type=int, default=18, help="Capture ALSA PCM index")
+    ap.add_argument("--rate", type=int, default=48000, help="Sample rate")
+    ap.add_argument("--chan", type=int, default=2, help="Output channel count")
+    ap.add_argument("--capchan", type=int,
+                      help="Capture channel count (if different from output)")
+    ap.add_argument("--capbits", type=int, default=16, help="Capture sample bits (16 or 32)")
+    ap.add_argument("--noise", default="noise.wav",
+                      help="WAV file containing 'noise' for capture")
+    ap.add_argument("--duration", type=int, default=3, help="Capture duration (seconds)")
+    ap.add_argument("--chirpcyc", type=int, default=120, help="Repetitions of chirp waveform")
+    ap.add_argument("--chirp-test", action="store_true",
+                      help="Test latency with synthesized audio")
+    ap.add_argument("--echo-test", action="store_true", help="Test simultaneous capture/playback")
+
+    opts = ap.parse_args()
+    if not opts.capchan:
+        opts.capchan = opts.chan
+    opts.base_test = not (opts.chirp_test or opts.echo_test)
+
+class ALSA:
+    """
+    Tiny ctypes stub.  Wraps the alsa API such that errno returns (at
+    least ones that look like an errno) become OSErrors and don't need
+    to be checked.  Includes a generalized alloc() that wraps all the
+    _sizeof() predicates and allocates from the (safe/collected) python
+    heap.  Provides a simple spot for putting (manually-derived)
+    constants.  The ALSA C API is mostly-structless and quite simple, so
+    this tends to work well without a lot of ctypes use except for an
+    occasional constructed integer or byref() pointer.
+    """
+    PCM_STREAM_PLAYBACK = 0
+    PCM_STREAM_CAPTURE = 1
+    PCM_FORMAT_S16_LE = 2
+    PCM_FORMAT_S32_LE = 10
+    PCM_ACCESS_MMAP_INTERLEAVED = 0
+    PCM_ACCESS_RW_INTERLEAVED = 3
+    def __init__(self):
+        self.lib = C.cdll.LoadLibrary("libasound.so.2")
+    def __getattr__(self, name):
+        fn = getattr(self.lib, name)
+        if name.endswith("_name"): # These return strings!
+            fn.restype = C.c_char_p
+            return lambda *args: fn(*args).decode("utf-8")
+        return lambda *args: ALSA.err_wrap(fn(*args))
+    @staticmethod
+    def err_wrap(ret):
+        if -200 < ret < 0:
+            raise OSError(os.strerror(-ret))
+        return ret
+    def alloc(self, typ):
+        return (C.c_byte * getattr(self.lib, f"snd_{typ}_sizeof")())()
+    class pcm_channel_area_t(C.Structure):
+        _fields_ = [("addr", C.c_ulong), ("first", C.c_int), ("step", C.c_int)]
+
+def pcm_init_stream(pcm, rate, chans, fmt, access):
+    hwp = alsa.alloc("pcm_hw_params")
+    alsa.snd_pcm_hw_params_any(pcm, hwp)
+    alsa.snd_pcm_hw_params_set_format(pcm, hwp, fmt)
+    alsa.snd_pcm_hw_params_set_channels(pcm, hwp, chans)
+    alsa.snd_pcm_hw_params_set_rate(pcm, hwp, rate, alsa.PCM_STREAM_PLAYBACK)
+    alsa.snd_pcm_hw_params_set_access(pcm, hwp, access)
+    alsa.snd_pcm_hw_params(pcm, hwp)
+
+def ctl_disable_rtnr():
+    """
+    Noise reduction likes to squash our chirp on capture.  Walk the list
+    of controls, looking for an RTNR enable control, if one exists, and
+    set it to false.  Unbelievably cumbersome API to do this: call
+    elem_list once on an empty struct to get the element count, then
+    allocate, then call it again.  Then for each element we can check
+    the name directly, but need to allocate an "id" struct to query an
+    abstract identifier, that we use with a separately-allocated "value"
+    (on which we set the dyncmically typed data) to send the command to
+    the kernel.
+    """
+    dev = f"hw:{opts.card}".encode("ascii")
+    ctl = C.c_ulong()
+    alsa.snd_ctl_open(C.byref(ctl), dev, 0)
+    elist = alsa.alloc("ctl_elem_list")
+    alsa.snd_ctl_elem_list(ctl, elist)
+    nelem = alsa.snd_ctl_elem_list_get_count(elist)
+    alsa.snd_ctl_elem_list_alloc_space(elist, nelem)
+    alsa.snd_ctl_elem_list(ctl, elist)
+    for i in range(nelem):
+        name = alsa.snd_ctl_elem_list_get_name(elist, i)
+        if re.match(r'RTNR.*\s+rtnr_enable.*', name):
+            print(f"Disabling control: {name}")
+            eid = alsa.alloc("ctl_elem_id")
+            val = alsa.alloc("ctl_elem_value")
+            alsa.snd_ctl_elem_list_get_id(elist, i, C.byref(eid))
+            alsa.snd_ctl_elem_value_set_id(val, eid)
+            alsa.snd_ctl_elem_value_set_boolean(val, 0, False)
+            alsa.snd_ctl_elem_write(ctl, val)
+    alsa.snd_ctl_close(ctl)
+
+def pcm_play_buf(data):
+    data = bytearray(data)
+    addr = C.addressof((C.c_byte * 1).from_buffer(data))
+    off = 0
+    n = int(len(data) / (2 * opts.chan))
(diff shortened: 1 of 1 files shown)


andyross:
From https://github.com/thesofproject/sof/pull/8681 , see there for previous discussion

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+    PCM_STREAM_PLAYBACK = 0
+    PCM_STREAM_CAPTURE = 1
+    PCM_FORMAT_S16_LE = 2
+    PCM_FORMAT_S32_LE = 10
+    PCM_ACCESS_MMAP_INTERLEAVED = 0
+    PCM_ACCESS_RW_INTERLEAVED = 3
+    def __init__(self):
+        self.lib = c.cdll.LoadLibrary("libasound.so.2")
+    def __getattr__(self, name):
+        if name.startswith("snd_"):
```
marc-hb, line 86: As found by pylint, what happens when name does NOT start with `snd_`? Add an `assert(False)` at the end maybe?

marc-hb (dismissed):
A good few, minor codestyle issues left otherwise LGTM but we need a reviewer who knows ALSA well.

Unlike other repos, we run `pylint` in this one and most of these few warnings seem valid and easy to fix:
https://github.com/thesofproject/sof-test/actions/runs/7393258262/job/20112906366?pr=1144

Please use `pylint: disable=W1234` for the ones you can't or don't want to fix.

kv2019i (approved):
Thanks @andyross ! While there is obvious overlap with alsabat and rest of alsa-utils, the "drop-in" capability will for sure have value in many cases.

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+        exit(0)
+
+    pcm_play_buf(buf)
+
+    os.waitpid(pid, 0)
+    msg = os.read(rfd, 9999).decode("ascii")
+    print("Capture energy:", msg)
+
+# Simplest test: Just capture opts.duration seconds worth of data,
+# convert to playback format, and play it.
```
plbossart, line 349: what conversions are you referring to?
- sample rate?
- channels?
- word size/valid bits?

it's not usual to record from 4 mics but only have 2 speakers.

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+    n = int(len(data) / (2 * opts.chan))
+    n = min(n, opts.rate * opts.duration)
+
+    pcm = c.c_long(0)
+    dev = f"hw:{opts.card},{opts.pcm}".encode("ascii")
+    alsa.snd_pcm_open(c.byref(pcm), dev, alsa.PCM_STREAM_PLAYBACK, 0)
+    pcm_init_stream(pcm, opts.rate, opts.chan, alsa.PCM_FORMAT_S16_LE,
+                    alsa.PCM_ACCESS_RW_INTERLEAVED)
+    while n > 0:
+        f = alsa.snd_pcm_writei(pcm, c.c_ulong(addr + off), n)
```
plbossart, line 154: can I ask why writei() is used here, and elsewhere the mmap_begin/commit is used? why not use *one* method for all uses?

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+    alsa.snd_pcm_reset(pcm)
+    while True:
+        alsa.snd_pcm_avail_update(pcm)
+        alsa.snd_pcm_mmap_begin(pcm, c.byref(area), c.byref(offset), c.byref(frames))
+        committed = alsa.snd_pcm_mmap_commit(pcm, offset, frames)
+        ring_frames += committed
+        if committed == 0:
+            break
+
+    silence = bytes(2 * opts.chan * ring_frames)
```
plbossart, line 185: is there any reason to restrict the playback usage to S16_LE?

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+# here for performance reasons, just one heap allocation and copy.
+def pcm_do_capture(duration):
+    pcm = c.c_long(0)
+    fmt = alsa.PCM_FORMAT_S32_LE if opts.capbits == 32 else alsa.PCM_FORMAT_S16_LE
+    capsz = 4 if opts.capbits == 32 else 2
+    dev = f"hw:{opts.card},{opts.cap}".encode("ascii")
+    alsa.snd_pcm_open(c.byref(pcm), dev, alsa.PCM_STREAM_CAPTURE, 0)
+    pcm_init_stream(pcm, opts.rate, opts.capchan, fmt, alsa.PCM_ACCESS_RW_INTERLEAVED)
+    frames_remaining = duration * opts.rate
+    buf_frames = int(opts.rate / 1000) # 1ms blocks
```
plbossart, line 219: doesn't work for 44.1kHz

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+        name = alsa.snd_ctl_elem_list_get_name(elist, i)
+        if re.match(r'RTNR.*\s+rtnr_enable.*', name):
+            print(f"Disabling control: {name}")
+            eid = alsa.alloc("ctl_elem_id")
+            val = alsa.alloc("ctl_elem_value")
+            alsa.snd_ctl_elem_list_get_id(elist, i, c.byref(eid))
+            alsa.snd_ctl_elem_value_set_id(val, eid)
+            alsa.snd_ctl_elem_value_set_boolean(val, 0, False)
+            alsa.snd_ctl_elem_write(ctl, val)
+    alsa.snd_ctl_close(ctl)
```
plbossart, line 139: not following how all other controls are initialized? The commit message states 'dependency-free ALSA' but the control parts will use whatever UCM or alsactl restored.

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+    buflist = []
+    buf_frames = c.c_ulong(buf_frames)
+    while frames_remaining > 0:
+        f = alsa.snd_pcm_readi(pcm, addr, buf_frames)
+        t = time.perf_counter()
+        frames_remaining -= f
+        buflist.append((t, bytes(buf[0:f * fsz])))
+    return buflist
+
+# A programmatically-detectable chirp/pop signal for testing latency.
```
plbossart, line 232: not following which latency you are measuring, does this not add playback+capture continuous latencies together?

tools/capture-test.py
```diff
@@ -0,0 +1,373 @@
...
+        if energy > (0.1/8):
+            os.write(wpipe, f"{t}".encode("ascii"))
+            return
+
+def echo_child(wpipe):
+    energy = 0
+    for rec in pcm_do_capture(opts.duration):
+        energy += cap_to_playback(rec[1])[1]
+
+    # Normalize energy to "half-swing per second" here, just to make
```
plbossart, line 300: not following the definition of 'half-swing', did you mean -3dBFS?

tools/capture-test.py
andyross: MTL indeed has 4 channel s32_le data on the capture pipeline, which is a hardship when dealing with AEC integration as you need to haul the data to your host and use audacity or whatever to check it.

tools/capture-test.py
andyross: writei() is simpler for "just play the buffer", but I need mmap to correct for already-buffered-data on capture, note that the capture path checks the (pre-filled) buffer state synchronously with the chirp insertion and timestamp, so we can know with higher precision when the chirp reaches the front of the DMA stream.

tools/capture-test.py
andyross: Not understanding, rate/1000 gives you the number of samples per millisecond.  It's true that 44100 doesn't divide evenly, but the code doesn't care.  This is just a conveniently small buffer size to do the analysis with later.

tools/capture-test.py
andyross: That's exactly right.  You can set it up offline or reload the driver.  Honestly that's a feature, not a bug.  I want to exercise this code in different circumstances.  The RTNR stuff is just a workaround that was important enough to include directly.  Longer term, sure, it would be nice to have it grow more capability.

tools/capture-test.py
andyross: Samples are in the range [-1,1], so one full swing is 2.0.  A single unit in the representation is half that.  But yes, 10^(-3/10) ~= 0.5, so sure.

tools/capture-test.py
andyross: Actually what I'm **really** measuring is a whitebox test inside AEC, which is doing essentially the same thing this python code is doing to watch the chirp go by.  This is here as a general smoke test from userspace.  But to be specific: it's measuring the time between the arrival of the chirp signal at the head of the userspace mmap buffer to the return from snd_pcm_readi() that contains it.

tools/capture-test.py
andyross: It's all any of the devices I'm using have the capability to emit, basically.  Agreed that a more general tool would be nice, but there's cost to code size too.

tools/capture-test.py
plbossart: all I am saying is that there should be a clear point on when the code no longer works.
We've have countless issues with such rounding that work fine in configuration A and break in configuration B.

tools/capture-test.py
andyross: I just removed it.  That was chaff, there's little value to checking for "valid" API use in an untyped environment.  If you want to pull out some random symbol out of the shared library and try to call it, there's no reason for the script to try to stop you.

tools/capture-test.py
plbossart: it's fine to keep S16_LE, it's just not clear if this was intentional due to lack of time or by design due to XYZ restrictions.

tools/capture-test.py
plbossart: Not following your explanation, sorry.

tools/capture-test.py
marc-hb: > Honestly that's a feature, not a bug. I want to exercise this code in different circumstances.

Then this is an IMHO important part of the user interface. Please add one sentence about that in the HELP_TEXT.

tools/capture-test.py
andyross: 1. Fill the mmap playback buffer[1] with silence
2. Start playing
3. Spin until there is just enough space in the buffer to accept the chirp, write it
4. Check buffer stats and take timestamp.

Now we know exactly how many samples are "ahead" of the chirp but still in host memory, so we can subtract them from the measured time it takes the chirp to appear on the capture stream.  You can't do this with snd_pcm_writei() as the buffering is internal to the userspace API.

[1] Which is obviously the actual DMA buffer, though the API doesn't promise this.

tools/capture-test.py
andyross: You'd be OK with an assert that the number of frames is >zero?  I could add a "chunk_size" tunable I guess too, I just don't see much value there.

Again, the specific size of these buffers isn't important.  All the analysis happens on lists of them and doesn't care about how many samples each buffer contains.  They're kept small to (1) provide reasonably fine-grained timestamp behavior, since we can only timestamp between API calls, and (2) avoid doing a big allocation and buffer copy synchronously with the capture, since this is python and I don't want to deal with performance headaches.

tools/capture-test.py
plbossart: not all DMAs provide a precise position, it's a recurring problem with audio - and it's only made worse with the bursty nature of the fimware-controlled DMAs.

You probably have at least one ms of error each way for the Intel HDaudio hw. no idea for others.

andyross: No doubt.  But we do what we can.  The point was just to not have an unmeasurable 0-85ms of latency in the userspace process, or to have to guess at "how long after the first writei() call does the DMA stream start running", etc...  These numbers aren't the whole story, but they're at least reliable.

FWIW: I'm getting a pretty consistent 30-31ms round trip on MTL with AEC enabled, which seems not totally unreasonable (10ms buffer for AEC alone, plus 0-1ms getting into and out of the DP thread due to scheduling cadence mismatch) but also that it probably has room for improvement.    mt8195 is a disaster in comparison: 100ms on the device I'm using, which AFAICS isn't in the firmware at all but in some hardware component (you can also hear it squash short chirp bursts, which is why the default length is so large).

tools/capture-test.py
```diff
@@ -0,0 +1,395 @@
...
+def main():
+    parse_opts()
+    if opts.disable_rtnr:
+        ctl_disable_rtnr()
+    if opts.base_test:
+        base_test()
+    if opts.chirp_test:
+        chirp_test()
+    if opts.echo_test:
+        echo_test()
```
marc-hb, line 390: Kill four birds with one stone:
- get rid of the pylint warning
- make echo_test() more generic
- make `opts` less global
- do not add one level of indentation to the entire echo_test() code

```suggestion
   with open(opts.noise, "rb").read()[HEADER_LEN:] as buf:
        echo_test(buf)
```

marc-hb (dismissed):
Found a few more nits + easy fixes for 2 pylint warnings but Python codestyle looks great.

Need other people (@plbossart , @aiChaoSONG , @singalsu , @ranj063, others?) to review the audio logic and ALSA API usage.

tools/capture-test.py
andyross: That looks clever, but interestingly this doesn't work.  It turns out that the "with" syntax isn't integrated with the internal reference counter at all[1], it's done at the class level by forcing the value assigned to implement `__enter__/__exit__` methods.  A file object has that, but a bytes object doesn't (because it's a heap block handled by the interpreter refcount!).   So the result is "`AttributeError: __enter__`"

Really that "with" warning is a false positive.  What it's detecting is the call to open().  But the object gets used and discarded in just the one line of code ("with" requires enter/exit, but the actual file descriptor gets closed out of the refcounter), so there's nothing to clean up.  To get this right, pylint would need to do lifetime analysis of the result of open() and not just detect that it's called, and it's not that smart.

[1] Which, sigh, is very much on brand for python.  All about top-level API definitions and never about clean integration with a core interpreter engine.

tools/capture-test.py
```diff
@@ -0,0 +1,395 @@
...
+class ALSA:
+    """
+    Tiny ctypes stub.  Wraps the alsa API such that errno returns (at
+    least ones that look like an errno) become OSErrors and don't need
+    to be checked.  Includes a generalized alloc() that wraps all the
+    _sizeof() predicates and allocates from the (safe/collected) python
+    heap.  Provides a simple spot for putting (manually-derived)
+    constants.  The ALSA C API is mostly-structless and quite simple, so
+    this tends to work well without a lot of ctypes use except for an
+    occasional constructed integer or byref() pointer.
```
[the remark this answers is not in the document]
andyross: You have to write ctypes code directly, e.g. the one struct definition in the class here, or the fact that the return value of functions ending with "name" is forced to be a c_char_p (the default is a return value of "int", which breaks because it chops off the top bits of the pointer).   Ctypes isn't super clean as a FFI, but it's not so bad.  But 90% of ALSA code in this script looks just like ALSA code in C, with the advantage that the errno returns become exceptions.

tools/capture-test.py
marc-hb: I guess the following could have worked (the previous suggestion was hasty copy/paste) but I understand it is a false positive in the first place.

```python
with open(opts.noise, "rb") as reader:
    echo_test(reader)
```

marc-hb (approved):
Same as before: code style looks good, someone else must approve the audio and ALSA API parts

tools/capture-test.py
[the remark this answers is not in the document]
marc-hb:
```suggestion
    occasional constructed integer or byref() pointer, e.g. pcm_channel_area_t below.
```
I read too fast and didn't realize this script has already sample solutions for this.

tools/capture-test.py
kv2019i: I think this is ok with suitable disclaimers. E.g. the 30-31ms number sounds in the right ballpark and matches data got (without AEC) on with ALSA latency tester https://www.alsa-project.org/wiki/Test_latency.c and jack-delay https://wiki.archlinux.org/title/Professional_audio#Latency_verification . Both tools have larger more dependencies, but can be used on any Linux setup to verify the numbers.

kv2019i (approved):
I'm good with this. In majority of cases, we'll be using aplay/arecord et al, so I don't think this has a be a perfect copy. But I do think this can be a very valueble tool on constrained environments and I can see value in having this available in sof-test.

marc-hb:
@plbossart any objection?

PR #1146, opened 2024-01-04
test-case: test-speaker.sh: use -F to pass sample format

kv2019i:
Unlike other alsa-utils, -f is not used to pass sample format, but instead -F needs to be used. speaker-test also doesn't complain if a sample format is passed to -f, which makes this not so easy to catch (especially as S16_LE works with most PCMs).

### DIFF
--- a/test-case/test-speaker.sh
+++ b/test-case/test-speaker.sh
@@ -50,8 +50,10 @@ do
 	sound_type="pink"
     fi
 
-    dlogc "speaker-test -D $dev -r $rate -c $channel -f $fmt -l $tcnt -t $sound_type -P 8"
-    speaker-test -D "$dev" -r "$rate" -c "$channel" -f "$fmt" -l "$tcnt" -t "$sound_type" -P 8 2>&1 |tee "$LOG_ROOT"/result_"$idx".txt
+    dlogc "speaker-test -D $dev -r $rate -c $channel -F $fmt -l $tcnt -t $sound_type -P 8"
+    # note -F has different semantics than in other alsa-utils, and speaker-test(1)
+    # man page does not cover this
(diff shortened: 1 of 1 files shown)


marc-hb (commented):
Please add a comment/warning that `man speaker-test` is 11 years out of date BUT `speaker-test -h` is up to date.

https://github.com/alsa-project/alsa-utils/tree/master/speaker-test

Screenshot 2024-01-04 at 13 08 14

Similar to https://github.com/thesofproject/sof-test/pull/1148

> speaker-test also doesn't complain if a sample format is passed to -f,

Welcome to C: the language of ignored errors and memory corruption.

kv2019i:
@marc-hb wrote:
> Please add a comment/warning that `man speaker-test` is 11 years out of date BUT `speaker-test -h` is up to date.

Isn't that a bit redundant? Maybe the man page will get updated next year and then we have a stale comment in sof-test. Not sure if this is worth a comment. Lesson learn, please check the source code for alsa-utils

marc-hb:
> Maybe the man page will get updated next year and then we have a stale comment in sof-test.

Then it will be out of date the next year.

> Not sure if this is worth a comment. Lesson learn, please check the source code for alsa-utils

Checking the source code was my first impulse. Then I realized `-h` is maintained and up to date. So yes, I think using an _inconsistent_ and half-documented option that's been misused for years it's worth a one-line comment.

kv2019i:
@marc-hb wrote:
>> Not sure if this is worth a comment. Lesson learn, please check the source code for alsa-utils
> 
> Checking the source code was my first impulse. Then I realized `-h` is maintained and up to date. So yes, I think using an _inconsistent_ and half-documented option that's been misused for years it's worth a one-line comment.

Fair enough, PR updated.

marc-hb (approved):
Very small fix, thanks for the comment.

PR #1148, opened 2024-01-04
check-alsabat.sh: explain that we run two alsabat instances

marc-hb:
This threw me off.

Also add warning that https://manpages.org/alsabat is incomplete and never tells that it is possible to use two separate alsabat instances, nor how to invoke these (answer: omit -P or -C)

### DIFF
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -2,15 +2,24 @@
 
 ##
 ## Case Name: check alsabat
+##
 ## Preconditions:
 ##    This test case requires physical loopback between playback and capture.
 ##    playback <=====>  capture
 ##    nocodec : no need to use hw loopback cable, It support DSP loopback by quirk
+##
 ## Description:
-##    run alsabat test on the specified pipelines
+##    Run two alsabat instances concurrently, one on each specified PCM: playback
+##    and capture.
+##
+##    Warning: as of January 2024, "man alsabat" is incomplete and
+##    documents only the "single instance" mode where a single alsabat
+##    process performs both playback and capture.
+##
 ## Case step:
 ##    1. Specify the pcm IDs for playback and catpure
 ##    3. run alsabat test
+##
 ## Expect result:
 ##    The return value of alsabat is 0
 ##


marc-hb:
Only suspend/resume failures.

test-case/check-alsabat.sh
```diff
@@ -2,15 +2,24 @@
...
 ##    nocodec : no need to use hw loopback cable, It support DSP loopback by quirk
+##
 ## Description:
-##    run alsabat test on the specified pipelines
+##    Run two alsabat instances concurrently, one on each specified PCM: playback
+##    and capture.
+##
+##    Warning: as of January 2024, "man alsabat" is incomplete and
+##    documents only the "single instance" mode where a single alsabat
+##    process performs both playback and capture.
```
kv2019i, line 17: Like with the "speaker-test" case, not sure about this comment. With the energy spent adding notes here, why not send a patch to upstream alsabat man page to address this?

kv2019i (approved):
The note on two alsabat instance is good, given the tool can be used in different ways.

test-case/check-alsabat.sh
marc-hb: > why not send a patch to upstream alsabat man page to address this?

Because:
- I have countless commits in this repo but never sent an ALSA patch in my life and it's not in any of my current objectives to learn how to do that.
- I have no idea who this "dual instance" mode actually works. It's a bit more complicated than just a missing `-f` format option.
- Assuming such a patch would be merged quickly, it would still take years to trickle back to the latest Ubuntu LTS, other Linux distros etc. sof-test changes are deployed to all test devices in less than 24 hours.
- In #1146 I found that the man page for speaker-test is 11 years old, so as a whole these man pages look like a lost cause not worth the effort.
- Even if fix this documentation gap, I will miss other gaps. So the generic part of this comment would still be useful. Yes I'm saying we can hardly have too many warnings not to trust ALSA man pages.

marc-hb:
The GLK and CML "ALL FAIL" suspend/resume https://sof-ci.01.org/softestpr/PR1148/build159/devicetest/index.html should have been fixed by Keqiao recently; re-running.

Only one suspend/resume failure in https://sof-ci.01.org/softestpr/PR1148/build158/devicetest/index.html

marc-hb:
SOFCI TEST

marc-hb:
Only 1 suspend/resume failure in https://sof-ci.01.org/softestpr/PR1148/build166/devicetest/index.html, everything else green.

PR #1152, opened 2024-01-16
[SKIP SOF-TEST] [CI TEST] [DON'T REVIEW] filter out all PCM on Port0 for multi-core test

keqiaozhang:


mengdonglin (approved):

mengdonglin:
I understand 2nd commit is to simulate current use scenario. Hope we can add a parameter not to overlook PCMs on port 0 later on.

keqiaozhang:
> I understand 2nd commit is to simulate current use scenario. Hope we can add a parameter not to overlook PCMs on port 0 later on.

My local code base was not clean. Just updated it.

Issue #1151, opened 2024-01-16
[BUG] sof-test output is messy when run test case locally


aiChaoSONG:
**Describe the bug**
When run test case locally, the output of the test case is messy. Log doesn't start from the beginning of a line. This makes the log hard to follow.
![Screenshot from 2024-01-16 16-31-02](https://github.com/thesofproject/sof-test/assets/52959957/0716df4b-6818-4cb6-94e6-9bb948a2f356)

**To Reproduce**
Just run a test case, say check-playback.sh on any device, and check the output.

EDIT, duplicate of:
- #945

aiChaoSONG:
`sudo -E "${mtraceCmd[@]}" >& "$clogfile" & `  this will mitigate the issue a little bit.

marc-hb:
Let's leave open as low priority because it's just a workaround and we have no idea why it works.

marc-hb:
I've just observed the same effect in a totally different (build) context. I think it happens when multiple processes write to the same terminal concurrently and one of them uses control codes (e.g. color). I suspect control codes are not "atomic" which means another problem can insert characters in the middle and break them.

Maybe the workaround #1154 disables control codes by hiding the terminal.

This would also explains why CI results are not affected: they don't use a terminal.

marc-hb:
I  got tired of this and I tested further. This is a problem only when using `sudo + &`
```
/bin/sudo    sleep 1 &   # garbage
/bin/sudo -b sleep 1     # clean
/bin/sudo    sleep 1     # clean
/bin/sudo -C sleep 1     # clean
```

```diff
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -463,7 +463,7 @@ func_sof_logger_collect()
     dlogi "Starting ${loggerCmd[*]} > $logfile &"
     # Cleaned up by func_exit_handler() in hijack.sh
     # shellcheck disable=SC2024
-    sudo "${loggerCmd[@]}" > "$logfile" &
+    /bin/sudo sleep 1 &
 }
```

PR #1154, opened 2024-01-17
lib.sh: Workaround sof-test output alignment issue

aiChaoSONG:
If sof-test test case is run manually and locally, the output of sof-test is messy, the lines are not aligned left at the beginning of each line, some
lines start in the middle of a line, which makes
the output log hard to follow.

This patch helps to workaround the above issue.

Link: https://github.com/thesofproject/sof-test/issues/1151

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -373,8 +373,10 @@ func_mtrace_collect()
     local mtraceCmd=("$MTRACE")
     dlogi "Starting ${mtraceCmd[*]} >& $clogfile &"
     # Cleaned up by func_exit_handler() in hijack.sh
+    # The use of 'bash -c' is a workaround for below issue:
+    # https://github.com/thesofproject/sof-test/issues/1151
     # shellcheck disable=SC2024
-    sudo "${mtraceCmd[@]}" >& "$clogfile" &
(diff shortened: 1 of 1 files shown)


case-lib/lib.sh
```diff
@@ -374,7 +374,7 @@ func_mtrace_collect()
     dlogi "Starting ${mtraceCmd[*]} >& $clogfile &"
     # Cleaned up by func_exit_handler() in hijack.sh
     # shellcheck disable=SC2024
-    sudo "${mtraceCmd[@]}" >& "$clogfile" &
+    sudo bash -c "${mtraceCmd[@]} >& $clogfile &"
```
marc-hb, line 377: The shellcheck warning is valid here: using `@` does not make sense anymore if you include as part of a bigger string. Use `*` as done in the `dlogi` above.

https://mywiki.wooledge.org/Quotes#Expand_argument_lists

marc-hb (changes requested):
No objection to the workaround if it works but please add a comment and link to the bug, otherwise there's no reason to do this.

In fact there's a reason NOT to do this (it's not whitespace compatible) but your workaround is more important.

marc-hb (approved):

PR #1155, opened 2024-01-19
test-case: test-speaker.sh: fix test for PCMs that only support 32bit

kv2019i:
The speaker-test "wav" test method supports only 48000Hz and 16bit files. If a topology has a PCM that does not support S16_LE, a different speaker-test method must be used.


marc-hb (approved):

marc-hb (changes requested):
The warning in https://github.com/thesofproject/sof-test/actions/runs/7585249087/job/20660669613?pr=1155 is unfortunately valid, relevant and very easy to fix. Sorry!

marc-hb (approved):

PR #1157, opened 2024-01-31
test-socwatch.sh: no need to load socwatch module manually

keqiaozhang:
We only use socwatch tool to read the power data for power measurement in this case, we can directly use the latest released version, so there is no need to pre-compile the driver.

### DIFF
--- a/test-case/test-socwatch.sh
+++ b/test-case/test-socwatch.sh
@@ -38,6 +38,7 @@ OPT_NAME['u']='unload-audio'  OPT_DESC['u']='unload audio modules for the test'
 OPT_HAS_ARG['u']=0            OPT_VAL['u']=0
 
 : "${SOCWATCH_PATH:=$HOME/socwatch}"
+SOCWATCH_VERSION=$("$SOCWATCH_PATH"/socwatch --version |grep Version)
 
 # reference cmd: sudo ./socwatch -t 20 -s 5 -f cpu-cstate -f pkg-pwr -o fredtest5
 #SOCWATCH_CMD="./socwatch"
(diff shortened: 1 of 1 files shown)


marc-hb (changes requested):
Please don't delete all the corresponding code. We are barely getting started in this area which means things will change in the future and we may need this again soon.

In fact it should be very easy to make the module loading and unload optional with just this:

```diff
- sudo "$SOCWATCH_PATH"/drivers/insmod-socwatch -q
+ sudo "$SOCWATCH_PATH"/drivers/insmod-socwatch -q || true
```

```diff
- check_socwatch_module_loaded || die "socwatch module not loaded"
+ check_socwatch_module_loaded || dlogi "socwatch module not loaded"
```

test-case/test-socwatch.sh
```diff
@@ -51,19 +52,20 @@ loop_count=${OPT_VAL['l']}
...
 {
-    lsmod | grep -q socwatch || die "socwatch is not loaded"
+    lsmod | grep -q socwatch || dlogi "socwatch is not loaded"
 }
 
 socwatch_test_once()
 {
     local i="$1"
     dlogi "===== Loop($i/$loop_count) ====="
+    dlogi "SoCWatch version: ${SOCWATCH_VERSION}"
```
marc-hb, line 62: I think you could have just run the command here but no big deal.

marc-hb (approved):
I don't find the commit message very accurate but... good enough.

We do need to load the driver manually; this PR just makes that optional.

keqiaozhang:
Test report: https://sof-ci.ostc.intel.com/#/result/planresultdetail/37633

marc-hb (approved):

marc-hb:
@fredoh9 can we merge this?

fredoh9 (approved):
Looks good.

PR #1158, opened 2024-02-07
[DNM] test-case/multiple-pipeline.sh: ignore aplay/arecord early exit

kv2019i:
Ignore early exit to allow runnning tests for longer without stopping in case https://github.com/thesofproject/sof/issues/8770 occurs.


marc-hb:
SOFCI TEST

Issue #1159, opened 2024-02-14
[BUG] bad regexp in sof-kernel-log-check.sh


plbossart:
**Describe the bug**
sof-test shows a set of "grep: warning: stray \ before :"

This is root caused to this regexp:

cd2eab8 (Keqiao Zhang          2022-09-22 14:46:43 +0800 262) ignore_str="$ignore_str"'|mei_me 0000:00:16\..\: .+'

the last \: is problematic.

**To Reproduce**

````
journalctl --priority=err | grep -vE "|mei_me 0000:00:16\..\: .+"
grep: warning: stray \ before :
````

**Expected behavior**
no warnings

**Detail Info**
    * SOF-TEST: 65951c4 ("kmod: fix hda-sdw-bpt dependency")

marc-hb:
I found this https://github.com/koalaman/shellcheck/issues/2573

> The interpretation of an ordinary character preceded by an unescaped ( '\' ) is undefined, except for: [...]"
> grep is giving a warning since 3.8

Ubuntu 22.04 has grep 3.7

Could you drop the antislash and send a PR?

plbossart:
````
root@fedora:~/sof-test# grep --version
grep (GNU grep) 3.11
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others; see
<https://git.savannah.gnu.org/cgit/grep.git/tree/AUTHORS>.

grep -P uses PCRE2 10.42 2022-12-11
````

I don't know if dropping the backslash is the right solution. Is ````:```` an 'ordinary' character?

marc-hb:
> Is : an 'ordinary' character?

Yes. Why? Because the "stray \" error message says so :-)

Also, because it's not in that list: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04

PR #1160, opened 2024-02-16
tools/sof-kernel-log-check.sh: remove stray `\`

plbossart:
Newer versions of grep complain about a regexp using a stray backslash in a regexp.

Closes: https://github.com/thesofproject/sof-test/issues/1159


marc-hb (approved):

marc-hb:
Failures known and unrelated

PR #1161, opened 2024-02-21
verify-kernel-boot-log: use dmidecode to print the BIOS version

marc-hb:
This will help with issues like
https://github.com/thesofproject/linux/issues/4823 and many others.


fredoh9:
Good idea to print BIOS version

test-case/verify-kernel-boot-log.sh
```diff
@@ -28,6 +28,9 @@ main()
 
     print_module_params
 
+    # 7 lines should be enough for all devices
+    sudo dmidecode | grep -A7 -i 'BIOS[[:blank:]]*Information'
```
fredoh9, line 32: how about
sudo dmidecode -s bios-version

marc-hb: The output of -s is just one string.

I think I found something better: `dmidecode -t bios`

fredoh9: One line isn't enough?

-t 0 is too verbose to me
```
$ sudo dmidecode --type 0
# dmidecode 3.3
Getting SMBIOS data from sysfs.
SMBIOS 3.6 present.
# SMBIOS implementations newer than version 3.5.0 are not
# fully supported by this version of dmidecode.

Handle 0x0015, DMI type 0, 26 bytes
BIOS Information
	Vendor: Intel Corporation
	Version: MTLPFWI1.R00.3424.D88.2311150537
	Release Date: 11/15/2023
	ROM Size: 14 MB
	Characteristics:
		PCI is supported
		BIOS is upgradeable
		BIOS shadowing is allowed
		Boot from CD is supported
		Selectable boot is supported
		EDD is supported
		Print screen service is supported (int 5h)
		8042 keyboard services are supported (int 9h)
		Serial services are supported (int 14h)
		Printer services are supported (int 17h)
		ACPI is supported
		USB legacy is supported
		LS-120 boot is supported
		ATAPI Zip drive boot is supported
		BIOS boot specification is supported
		Function key-initiated network boot is supported
		Targeted content distribution is supported
		UEFI is supported
	Firmware Revision: 1.71
```

marc-hb:
Known and recent CAVS regression https://github.com/thesofproject/sof/issues/8875 observed in https://sof-ci.01.org/softestpr/PR1161/build230/devicetest/index.html, everything else green.

https://sof-ci.01.org/softestpr/PR1161/build231/devicetest/index.html and https://sof-ci.01.org/softestpr/PR1161/build232/devicetest/index.html are all green.

All BIOS versions successfully printed everywhere.

test-case/verify-kernel-boot-log.sh
marc-hb: This is only ~30 lines and I don't think anyone ever looks at the logs of this test except when there is something wrong.

As you can see, dmidecode reports multiple different versions: one at the top and another one or two at the bottom.

For instance https://sof-ci.01.org/softestpr/PR1161/build232/devicetest/index.html?model=CML_RVP_SDW-ipc3&testcase=verify-kernel-boot-log has 2 versions reported at the bottom.
 
I agree we don't care about the stuff in the middle but I'd like to keep both versions and to keep the sof-test code very simple and robust.

We have other, unusable test logs with many thousands of lines, so big they make the browser slow: https://github.com/thesofproject/sof/issues/8761
So 30 lines is a very small problem in comparison.

marc-hb: > I agree we don't care about the stuff in the middle but...

We could drop all the stuff in the middle like this but it's brittle and IMHO not worth the risk.

```
sudo dmidecode -t 0 | grep -v -e $'\t'Characteristics -e  ^$'\t'$'\t' # bashism 

# dmidecode 3.3
Getting SMBIOS data from sysfs.
SMBIOS 3.2 present.

Handle 0x0012, DMI type 0, 26 bytes
BIOS Information
	Vendor: Intel Corporation
	Version: CMLSFWR1.R00.1245.D00.1906150738
	Release Date: 06/15/2019
	ROM Size: 11 MB
	BIOS Revision: 245.0
	Firmware Revision: 1.44



```

mengdonglin (approved):

PR #1164, opened 2024-03-29
kernel-boot-log.sh: add new show_daemons_session_display()

marc-hb:
We started showing part of i915_display_info a long time ago but only when `systemctl is-system-running` failed. Because of recent HDMI issues like https://github.com/thesofproject/linux/issues/4861 and others we want to show more and we want to show it always.


marc-hb:
The test results finally expose the very wide variety according to which our systems are configured :-(

kv2019i (approved):

Issue #1166, opened 2024-04-03
[BUG] sof-test scripts shall not assume settings based on kernel parameters


plbossart:
Filing this as a bug because I truly think there's a conceptual issue.

Overtime, sof-test has become dependent on kernel parameters, such as 
- ipc_type. This should ONLY be used for CAVS/IPC4 tests. In all 'normal' case the ipc_type is a given for each platform, e.g. anything MTL+ is ipc4
the following snippet is completely wrong

````bash
is_ipc4()
{           
    local ipc_type
    ipc_file=/sys/module/snd_sof_pci/parameters/ipc_type
            
    # If /sys/module/snd_sof_pci/parameters/ipc_type does not exist
    # the DUT is running IPC3 mode <<< OF COURSE NOT ON MTL+!!!
    ipc_type=$(cat $ipc_file) || {
        return 1
    }
            
    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
    # If the value of file ipc_type is:
    # -1: DUT runs IPC3 mode, is_ipc4 return 1(false)
    # 1: DUT runs IPC4 mode, is_ipc4 return 0(true)
    if [ "$ipc_type" -eq 1 ]; then
        return 0
    fi
    return 1
}
````

- fw_path. the default is set for the kernel. overrides should only be used for things like debug key.
- tplg_path

The premise of sof-test is that the test should run without any assumption on the environment, other than what is clearly described in env-check.sh

sof-test is used by CI, it shall not depend on CI settings.

marc-hb:
>   the DUT is running IPC3 mode <<< OF COURSE NOT ON MTL+!!!

I agree it's wrong. I guess it wasn't when this code was written.

I think we all agree 100% that sof-test should be independent of CI. I filed and fixed many issues along this line. "Conceptual" issue sounds exaggerated: I think this is just a short-term hack and bug that lingered because of a lack of maintenance.

But now the question is: how can userspace reliably tell which IPC version is currently active? To avoid sof-test hacks, the kernel must expose interfaces that allow it to be tested.

Similar problem and question in
- https://github.com/thesofproject/linux/issues/3867

plbossart:
well, if we are in the questions stage, why would sof-test need to know which IPC type is used?

Apart from the logging I don't see any reason... And the code is just awfully convoluted between is_firmware_zephyr and is_ipc4 - which means the same ...

plbossart:
just using the existence of this directory would be enough to know which logger to use:

[...]

PR #1167, opened 2024-04-03
case-lib: is_ipc4: use new debugfs file instead of kernel parameter

plbossart:
First step to remove use of kernel parameters

Link: https://github.com/thesofproject/linux/pull/4902
Link: https://github.com/thesofproject/sof-test/issues/1166

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -891,6 +891,10 @@ print_module_params()
     # all snd_hda* modules
     grep -H ^ /sys/module/snd_hda*/parameters/* || true
     echo "----------------------------------------"
+
+    echo "--------- Printing debugfs settings ----------"
+    grep -H ^ /sys/kernel/debug/sof/fw_profile/* || true
+    echo "----------------------------------------------"
 }
 
 # "$@" is optional, typically: --since=@"$epoch".
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -296,9 +296,16 @@ fake_kern_error()
 get_ldc_subdir()
 {
     local subdir='intel/sof' # default
-    if test -e /sys/module/snd_sof_pci/parameters/fw_path; then
-        local fw_path
-        fw_path=$(cat /sys/module/snd_sof_pci/parameters/fw_path)
+    local fw_path
+    local fw_path_info='/sys/kernel/debug/sof/fw_path'
+
+    # either $fw_path_info exists, OR we redefine $fw_path_info with
+    # the backwards-compatible alternative based on kernel parameter
+    test -e $fw_path_info ||
+	fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
+
+    if fw_path=$(cat $fw_path_info); then
+	# "cat" was succesful
         if [ "$fw_path" != '(null)' ]; then
             subdir=${fw_path%/} # strip any trailing slash
             subdir=${subdir%/community}
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -904,7 +904,7 @@ grep_firmware_info_in_logs()
     # dump the version info and ABI info
     # "head -n" makes this compatible with set -e.
     journalctl_cmd "$@" | grep "Firmware info" -A1 | head -n 12
-    # For dumping the firmware information when DUT runs IPC4 mode 
+    # For dumping the firmware information when DUT runs IPC4 mode
     journalctl_cmd "$@" | grep "firmware version" -A1 | head -n 12
     # dump the debug info
     journalctl_cmd "$@" | grep "Firmware debug build" -A3 | head -n 12
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -813,21 +813,32 @@ is_firmware_file_zephyr()
 is_ipc4()
 {
     local ipc_type
-    ipc_file=/sys/module/snd_sof_pci/parameters/ipc_type
 
-    # If /sys/module/snd_sof_pci/parameters/ipc_type does not exist
-    # the DUT is running IPC3 mode
-    ipc_type=$(cat $ipc_file) || {
-        return 1
-    }
-
-    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
-    # If the value of file ipc_type is:
-    # -1: DUT runs IPC3 mode, is_ipc4 return 1(false)
-    # 1: DUT runs IPC4 mode, is_ipc4 return 0(true)
-    if [ "$ipc_type" -eq 1 ]; then
-        return 0
+    if ipc_type=$(cat /sys/kernel/debug/sof/fw_profile/ipc_type); then
+	# "cat" was successful
+	case $ipc_type in
+	    0) return 1;; # IPC3 found
+	    1) return 0;; # IPC4 found. in shell, success = error code 0 = "true"
+	    *) die "invalid ipc_type in /sys/kernel/debug/sof/ipc_type=$ipc_type";;
+	esac
     fi
+
+    dlogw "using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter"
+    if ipc_type=$(cat /sys/module/snd_sof_pci/parameters/ipc_type); then
+	# "cat" was successful
+	case $ipc_type in
+	    -1)
+		dlogw "the ipc_type parameter is not set, assume IPC3 was used"
+		return 1;;
+	    0)	# the parameter is set, IPC3 was used.
+		return 1;;
+	    1) 	# the parameter is set, IPC4 was used.
+		return 0;; # in shell, success = error code 0 = "true"
+	    *)	die "invalid ipc_type in /sys/module/snd_sof_pci/parameters/ipc_type=$ipc_type";;
+	esac
+    fi
+
+    dlogw "/sys/module/snd_sof_pci/parameters/ipc_type does not exist, assuming IPC3 was used"
     return 1
 }
 


plbossart:
shellcheck error is a pre-existing issue that was obviously overlooked

````
In case-lib/lib.sh line 954:
        TGLU_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC_IPC4ZPH | MTLP_RVP_NOCODEC |ADLP_RVP_NOCODEC-ipc4 | TGLU_RVP_NOCODEC-ipc4 | MTLP_RVP_NOCODEC | MTLP_RVP_NOCODEC-multicore-2cores | MTLP_RVP_NOCODEC-multicore-3cores | LNLM_RVP_NOCODEC)
                                                              ^--------------^ SC22[21](https://github.com/thesofproject/sof-test/actions/runs/8544615585/job/23411051491?pr=1167#step:3:22) (warning): This pattern always overrides a later one on line 954.
                                                                                                                                ^--------------^ SC[22](https://github.com/thesofproject/sof-test/actions/runs/8544615585/job/23411051491?pr=1167#step:3:23)22 (warning): This pattern never matches because of a previous pattern on line 954.
````

MTLP_RVP_NOCODEC  is listed twice in the same OR.

File: case-lib/lib.sh [hunk 1]
```diff
@@ -810,9 +810,31 @@ is_firmware_file_zephyr()
...
+}
+
 is_ipc4()
 {
     local ipc_type
+
+    # First try with debugfs
+    if is_ipc4_debugfs; then
+	return 0
+    fi
```
marc-hb, line 835:
```shell
if ipc_type=$(cat /sys/kernel/debug/sof/ipc_type); then
   # "cat" was successful
   case $ipc_type in
   0) return ... # in shell, success = error code 0 = "true"
   1) return ...
   *) die "unexpected /sys/kernel/debug/sof/ipc_type=$ipc_type"
fi

# "cat" failed, fallback on ...
```

marc-hb (changes requested):
cc:
- https://github.com/thesofproject/linux/issues/3867

```
    # If /sys/module/snd_sof_pci/parameters/ipc_type does not exist
    # the DUT is running IPC3 mode
```
This shortcut and bad assumption should print a big fat `dlogw()` TODAY and be "upgraded" to a `die()` failure in the future. Please add the `dlogw()` now; I think that would have saved you time and it will save others'

case-lib/lib.sh
```diff
@@ -810,9 +810,31 @@ is_firmware_file_zephyr()
...
 }
 
+is_ipc4_debugfs()
+{
+    local ipc_type
+
+    ipc_file=/sys/kernel/debug/sof/ipc_type
+    ipc_type=$(cat $ipc_file) || {
+        return 1
+    }
```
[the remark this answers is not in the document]
plbossart: the overloading is already present in the initial code for is_ipc4().

And while I am at it it's just torture to see code a branch " if is_ipc4" taken when ipc4 is NOT used. 
Such mind-twisting is just not right. someone will screw-up when the value of 'true' is inverted.

case-lib/lib.sh
```diff
@@ -810,9 +810,31 @@ is_firmware_file_zephyr()
...
+
+    if [ "$ipc_type" -eq 1 ]; then
+        return 0
+    fi
+    return 1
```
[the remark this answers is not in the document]
plbossart: Or keep the syntax simple for the rest of us who just understand if then...

The simplest way to write a branch is the best... 

I vote to disallow any 'smart' bash construct which requires people like me to consult an online help. bash is a means to an end.

case-lib/lib.sh
plbossart: Good suggestion that anyone would understand, thanks.

marc-hb:
> MTLP_RVP_NOCODEC is listed twice in the same OR.

Fix submitted in #1168

case-lib/lib.sh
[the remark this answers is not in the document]
marc-hb: > the overloading is already present in the initial code for is_ipc4().

It's not because the higher level does something questionable that the lower level, internal implementation must align with it. Of course it shouldn't be different without any good reason either but there's no "strict" requirement there.

case-lib/lib.sh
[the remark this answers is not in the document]
marc-hb: I must differ here sorry.

I totally agree that bash is quirky and the confusion between booleans and error codes plays a large role. It is a real pain. However, learning about that particular confusion between booleans and error codes is unfortunately not optional. especially not in sof-TEST code, the main purpose of any test code being to deal with _errors_.

For instance, until a person can make sense of what I wrote in #312, then they're not allowed to have an informed opinion on this topic! I think that boolean/error code knowledge is really the minimum required. Note I just wrote "make sense" of #312, _not_ "remember it all".

So now back to this particular example: wrapping `test_something` into `if test_something; then return 0; else return 1` is:
- "useful" only for people who MISS this basic boolean/error code knowledge.
- very verbose because you'd have to do this wrapping pretty much everywhere
- distracting because it's exposing the fact that 0 = true instead of abstracting it away.


All this being said, a one-line comment never hurts. For example:

```shell
is_something()
{
   # stuff

   # return true if X exists
   test -e X
}
```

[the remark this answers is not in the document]
plbossart: You're entitled to your opinions.

I personally don't know *anything* to bash and everything I read about it gives me a migraine. I won't remember any of this next Monday anyways.

so if I can use if/else I will...

case-lib/lib.sh
```diff
@@ -813,22 +813,47 @@ is_firmware_file_zephyr()
...
-    # -1: DUT runs IPC3 mode, is_ipc4 return 1(false)
-    # 1: DUT runs IPC4 mode, is_ipc4 return 0(true)
-    if [ "$ipc_type" -eq 1 ]; then
-        return 0
+    # "cat" failed, backwards-compatible fallback with ipc_type kernel parameter
+    if ipc_type=$(cat /sys/module/snd_sof_pci/parameters/ipc_type); then
+	# "cat" was successful
+	case $ipc_type in
+	    -1)
+		# the parameter is not set, assume IPC3 was used.
```
marc-hb, line 843: Can you please "upgrade" this comment to a `dlogw`
```shell
dlogw "snd_sof_pci/parameters/ipc_type parameter is not set, assume IPC3 was used."
```

case-lib/lib.sh
```diff
@@ -813,22 +813,47 @@ is_firmware_file_zephyr()
...
-    # the DUT is running IPC3 mode
-    ipc_type=$(cat $ipc_file) || {
-        return 1
-    }
+    if ipc_type=$(cat /sys/kernel/debug/sof/fw_profile/ipc_type); then
+	# "cat" was successful
+	case $ipc_type in
+	    0)
+		return 1
+		;;
```
marc-hb, line 828: Nit: this is a lot of whitespace...

```shell
  0) return 1 ;;
```

case-lib/lib.sh
```diff
@@ -813,22 +813,47 @@ is_firmware_file_zephyr()
...
+		return 0 # in shell, success = error code 0 = "true"
+		;;
+	    *)
+		die "invalid ipc_type in /sys/module/snd_sof_pci/parameters/ipc_type=$ipc_type"
+		;;
+	esac
     fi
-    return 1
+
+    # If /sys/module/snd_sof_pci/parameters/ipc_type does not exist
```
marc-hb, line 860: upgrade this comment to a `dlogw` too.

case-lib/lib.sh
```diff
@@ -296,9 +296,15 @@ fake_kern_error()
...
-        local fw_path
-        fw_path=$(cat /sys/module/snd_sof_pci/parameters/fw_path)
+    local fw_path
+    local fw_path_info='/sys/kernel/debug/sof/fw_path'
+
+    if test ! -e $fw_path_info; then
+	fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
+    fi
+
+    if fw_path=$(cat $fw_path_info); then
```
marc-hb, line 306: This commit adds `if ... ; then` but no `fi` termination so I don't see how this can work?

case-lib/lib.sh
```diff
@@ -296,9 +296,15 @@ fake_kern_error()
...
 {
     local subdir='intel/sof' # default
-    if test -e /sys/module/snd_sof_pci/parameters/fw_path; then
-        local fw_path
-        fw_path=$(cat /sys/module/snd_sof_pci/parameters/fw_path)
+    local fw_path
+    local fw_path_info='/sys/kernel/debug/sof/fw_path'
+
+    if test ! -e $fw_path_info; then
+	fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
```
marc-hb, line 303: Nit: the shorter, "idiomatic" style is:
```shell
test -e $fw_path_info ||
    fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
```

It reads in plain English likes this: "either $fw_path_info exists, OR we redefine $fw_path_info and point it at the alternative"

marc-hb (commented):
Looks good except for some nits and one... missing `fi`?

case-lib/lib.sh
```diff
@@ -813,22 +819,47 @@ is_firmware_file_zephyr()
...
+	esac
+    fi
 
-    # If /sys/module/snd_sof_pci/parameters/ipc_type exists
-    # If the value of file ipc_type is:
-    # -1: DUT runs IPC3 mode, is_ipc4 return 1(false)
-    # 1: DUT runs IPC4 mode, is_ipc4 return 0(true)
-    if [ "$ipc_type" -eq 1 ]; then
-        return 0
+    # "cat" failed, backwards-compatible fallback with ipc_type kernel parameter
```
marc-hb, line 838: Some vague indication in the comment of which kernel version will be required would be useful.

case-lib/lib.sh
marc-hb: Never mind, now I see what happened.

case-lib/lib.sh
plbossart: done

case-lib/lib.sh
plbossart: done

case-lib/lib.sh
plbossart: done

marc-hb (approved):

marc-hb:
Lots of models missing but everything green for the ones present

kv2019i (approved):

marc-hb:
This does not work actually: it's at least missing a few `sudo`. I think you didn't notice because you run sof-test as root but that's not how it's usually tested.

https://sof-ci.01.org/softestpr/PR1167/build265/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=verify-kernel-module-load-probe

```
2024-04-05 14:58:59 UTC [REMOTE_INFO] Starting func_exit_handler(0)
cat: /sys/kernel/debug/sof/fw_profile/ipc_type: Permission denied
2024-04-05 14:58:59 UTC [REMOTE_WARNING] using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter
cat: /sys/kernel/debug/sof/fw_profile/ipc_type: Permission denied
2024-04-05 14:58:59 UTC [REMOTE_WARNING] using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter
2024-04-05 14:58:59 UTC [REMOTE_INFO] pkill -TERM -f mtrace-reader.py
2024-04-05 14:58:59 UTC [REMOTE_INFO] ktime=123 sof-test PID=3129: ending
2024-04-05 14:58:59 UTC [REMOTE_INFO] Test Result: PASS!
```

marc-hb:
Filed new issue:
- #1171

Issue #1163, opened 2024-03-23
[BUG] ignore i8042 probe error


plbossart:
https://sof-ci.01.org/linuxpr/PR4869/build1901/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=verify-kernel-boot-log

`````
[    2.470644] kernel: i8042: PNP: PS/2 Controller [PNP0303:PS2K] at 0x60,0x64 irq 1
[    2.470645] kernel: i8042: PNP: PS/2 appears to have AUX port disabled, if this is incorrect please boot with i8042.nopnp
[    2.484698] kernel: nvme nvme0: 21/0/0 default/read/poll queues
[    2.490918] kernel:  nvme0n1: p1 p2
[    3.171704] kernel: i8042: Can't read CTR while initializing i8042
[    3.171725] kernel: i8042 i8042: probe with driver i8042 failed with error -5
`````

no idea what an i8042 is and why we still have PS/2 support :-)

marc-hb:
Seen again in MTL daily test 39358

We could probably remove PS/2 support but some old systems might still have it?

More surprisingly, why is this showing only now?

marc-hb:
We've always ignored this error since 2021: d73423f5923c

But for some reason it's now 2 lines long?

plbossart:
Interesting, thanks @marc-hb. Maybe we should just ignore all of those errors - or blacklist those drivers since we don't care.

marc-hb:
I'll add the second line.

marc-hb:
Fix submitted in:
- #1168

marc-hb:
Error is gone

PR #1168, opened 2024-04-04
I8042 ignore

marc-hb:
Fixes:
- #1163


marc-hb:
Some models unavailable, everything else is green.

plbossart (approved):

Issue #1171, opened 2024-04-09
/sys/debug/   fw_profile/ipc_type: permission denied


marc-hb:
#1167 does not work: it's at least missing a few `sudo`. I think you didn't notice because you run sof-test as root but that's not how it's usually tested.

https://sof-ci.01.org/softestpr/PR1167/build265/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=verify-kernel-module-load-probe

```
2024-04-05 14:58:59 UTC [REMOTE_INFO] Starting func_exit_handler(0)
cat: /sys/kernel/debug/sof/fw_profile/ipc_type: Permission denied
2024-04-05 14:58:59 UTC [REMOTE_WARNING] using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter
cat: /sys/kernel/debug/sof/fw_profile/ipc_type: Permission denied
2024-04-05 14:58:59 UTC [REMOTE_WARNING] using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter
2024-04-05 14:58:59 UTC [REMOTE_INFO] pkill -TERM -f mtrace-reader.py
2024-04-05 14:58:59 UTC [REMOTE_INFO] ktime=123 sof-test PID=3129: ending
2024-04-05 14:58:59 UTC [REMOTE_INFO] Test Result: PASS!
```

_Originally posted by @marc-hb in https://github.com/thesofproject/sof-test/issues/1167#issuecomment-2046038705_
            

cc:
- #1166

marc-hb:
> I think you didn't notice because you run sof-test as root but that's not how it's usually tested.

You can borrow any CI device to test and fix, no need to change the configuration of yours.

plbossart:
I am a bit confused here. The two files have the same permissions and use the same cat command.

````
root@jf-cml-hel-rt5682-05:~# ls -l /sys/kernel/debug/sof/fw_profile/ipc_type
-r--r--r-- 1 root root 0 Apr  9 16:35 /sys/kernel/debug/sof/fw_profile/ipc_type
root@jf-cml-hel-rt5682-05:~# ls -l /sys/module/snd_sof_pci/parameters/ipc_type 
-r--r--r-- 1 root root 4096 Apr  9 21:22 /sys/module/snd_sof_pci/parameters/ipc_type
````
What causes one to complain and not the other?

marc-hb:
/sys/kernel/debug is a different filesystem / mount point.
[...]

PR #1172, opened 2024-04-09
case-lib/lib.sh: use sudo to access debugfs files

plbossart:
Apparently debugfs requires more care. Oh well.

Link: https://github.com/thesofproject/sof-test/issues/1171

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -304,7 +304,7 @@ get_ldc_subdir()
     test -e $fw_path_info ||
 	fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
 
-    if fw_path=$(cat $fw_path_info); then
+    if fw_path=$(sudo cat $fw_path_info); then
 	# "cat" was succesful
         if [ "$fw_path" != '(null)' ]; then
             subdir=${fw_path%/} # strip any trailing slash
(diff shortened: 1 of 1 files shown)


case-lib/lib.sh
```diff
@@ -831,7 +831,7 @@ is_ipc4()
     fi
 
     dlogw "using pre kernel 6.9 backwards-compatible fallback with ipc_type kernel parameter"
-    if ipc_type=$(cat /sys/module/snd_sof_pci/parameters/ipc_type); then
+    if ipc_type=$(sudo cat /sys/module/snd_sof_pci/parameters/ipc_type); then
```
marc-hb, line 834: This does not seem required. When running sof_insert/sof_remove I noticed `sudo` can be pretty slow, even slower on Fedora I think. Of course we should ideally cache the result of `is_ipc4()` but right now we don't.

marc-hb (approved):
Tests results look good. Very small change that gets rid of very confusing error messages: merging.

marc-hb:
@plbossart want to keep this a draft still? If not then please merge

plbossart:
I missed one:
````
grep: /sys/kernel/debug/sof/fw_profile/*: Permission denied
````

marc-hb (approved):

kv2019i (approved):

PR #1174, opened 2024-04-11
sof-tplgreader: parse rates from topology file

fredoh9:
I was able to test with https://github.com/thesofproject/sof/pull/9017

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,9 +146,12 @@ apl
 
 * tplgtool2.py
 <br> Dumps info from tplg binary file.
+     SOF CI uses this to generate a topology graph.
 
 * tplgtool.py
-<br> Dumps info from tplg binary file. (deprecated)
+<br> Dumps info from tplg binary file. sof-tplgreader.py still use this but new features
+     will go to tplgtool2.py. When all functions are migrated to tplgtool2.py
+     it will be deprecated.
 
 
 ## System configuration tips
--- a/tools/sof-tplgreader.py
+++ b/tools/sof-tplgreader.py
@@ -63,11 +63,13 @@ class clsTPLGReader:
                 # supported formats of playback pipeline in formats[0]
                 # supported formats of capture pipeline in formats[1]
                 formats = TplgFormatter.get_pcm_fmt(pcm)
+                rates = TplgFormatter.get_pcm_rate_list(pcm)
                 # if capture is present, pcm['capture'] = 1, otherwise, pcm['capture'] = 0,
                 # same thing for pcm['playback']
                 pipeline_dict['fmts'] = " ".join(formats[pcm['capture']])
                 # use the first supported format for test
                 pipeline_dict['fmt'] = pipeline_dict['fmts'].split(' ')[0]
+                pipeline_dict['rates'] = " ".join(rates[pcm['capture']])
                 pipeline_dict['rate_min'], pipeline_dict['rate_max'] = self._key2str(cap, 'rate')
                 pipeline_dict['ch_min'], pipeline_dict['ch_max'] = self._key2str(cap, 'channels')
                 # for pcm with both playback and capture capabilities, we can extract two pipelines.
@@ -95,7 +97,17 @@ class clsTPLGReader:
         # format pipeline, this change for script direct access 'rate' 'channel' 'dev' 'snd'
         for pipeline in self._pipeline_lst:
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
-            pipeline['rate'] = pipeline['rate_min'] if int(pipeline['rate_min']) != 0 else pipeline['rate_max']
+            # If rates is available get first sampling rate from the list.
+            # If rates list doesn't have any, then get it from rate_min or rate_max.
+            # If none of them is available, then set rate to 0
+            if pipeline['rates']:
+                pipeline['rate'] = pipeline['rates'].split(' ')[0]
+            elif int(pipeline['rate_min']) != 0:
+                pipeline['rate'] = pipeline['rate_min']
+            elif int(pipeline['rate_max']) != 0:
+                pipeline['rate'] = pipeline['rate_max']
+            else:
+                pipeline['rate'] = 0
             pipeline['channel'] = pipeline['ch_min']
             pipeline['dev'] = "hw:" + str(sofcard) + ',' + pipeline['id']
             # the character devices for PCM under /dev/snd take the form of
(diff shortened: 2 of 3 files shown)


tools/sof-tplgreader.py
```diff
@@ -95,7 +98,14 @@ def loadFile(self, filename, sofcard=0):
...
             #pipeline['fmt']=pipeline['fmt'].upper().replace('LE', '_LE')
-            pipeline['rate'] = pipeline['rate_min'] if int(pipeline['rate_min']) != 0 else pipeline['rate_max']
+            # set rate in order of rate_min, rate_max. If both are 0, get first
+            # sampling rate(rates[0]) from supported rates
+            if int(pipeline['rate_min']) != 0:
+                pipeline['rate'] = pipeline['rate_min']
+            elif int(pipeline['rate_max']) != 0:
+                pipeline['rate'] = pipeline['rate_max']
+            elif rates[0] != 0:
+                pipeline['rate'] = pipeline['rates'].split(' ')[0]
```
ranj063, line 108: I think the order should be reversed. check if rates is not empty and only then check min/max

ranj063: And also what if there's no rates or rate_min/rate_max in tplg? Can we catch that and throw an error is rate is 0?

tools/tplgtool.py
```diff
@@ -607,6 +607,31 @@ def _to_fmt_string(fmt):
...
+        if rate & (1 << 5) != 0:
+            rates.append("32000")
+        if rate & (1 << 6) != 0:
+            rates.append("44100")
+        if rate & (1 << 7) != 0:
+            rates.append("48000")
+        if rate & (1 << 8) != 0:
+            rates.append("64000")
+        if rate & (1 << 9) != 0:
+            rates.append("96000")
```
marc-hb, line 632: This is a lot of repetition; I think you don't need to "unroll" that loop:

```python
  bit_rates_dict = { 1 : 8000, 3 : 16000, ...}

  for bit in rates_dict:
     if rate & (1 << bit) != 0:
        rates.append(bit_rates_dict[bit])

```
Or even shorter

```python
  rates = [ bit_rates_dict[bit] for bit in bit_rates_dict
                   if rate & (1 << bit) != 0 ]
```

tools/tplgtool.py
```diff
@@ -618,6 +643,17 @@ def get_pcm_fmt(pcm):
...
+    # always return a list, pcm rates in the list in ascending order
+    @staticmethod
+    def get_pcm_rate_list(pcm):
+        #print("Fred: get_pcm_rates")
+        rate_list = []
+        caps = pcm["caps"]
+        #print(caps)
+        for cap in caps:
+            rate_list.append(TplgFormatter._to_rate_string(cap["rates"]))
+        return rate_list
```
marc-hb, line 655:
```python
  return [ TplgFormatter._to_rate_string(cap["rates"]) for cap in caps ]
```
https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions

tools/tplgtool.py
fredoh9: This is really good comment, but I can't get it to work. I put a TODO above. There are a couple of places where the same thing is applicable. Can I make this improvement as a separate PR?

ranj063 (approved):

tools/tplgtool.py
marc-hb: What is the error?

marc-hb: I tried to test this to help but the new functions are never called. Did you forget a part?

README.md
```diff
@@ -148,7 +148,8 @@ apl
 <br> Dumps info from tplg binary file.
 
 * tplgtool.py
-<br> Dumps info from tplg binary file. (deprecated)
+<br> Dumps info from tplg binary file. sof-tplgreader.py still use this but new features
+     will go to tplgtool2.py. Will be deprecated.
```
marc-hb, line 152: Can you please explain a little bit more which script uses which script and when?

marc-hb:
https://sof-ci.01.org/softestpr/PR1174/build297/devicetest/index.html, https://sof-ci.01.org/softestpr/PR1174/build298/devicetest/index.html and https://sof-ci.01.org/softestpr/PR1174/build296/devicetest/index.html are 100% fail. Switching to draft.

tools/tplgtool.py
marc-hb: > I tried to test this to help but the new functions are never called. Did you forget a part?

OK I figured it out: the "interactive" output of the parser is incomplete.


I just tested this below together with https://github.com/thesofproject/sof/pull/9017/files and it works perfectly fine:

```python
    @staticmethod
    def _to_rate_string(rate):
        # Note: intentionally put 48000 first to make first in the list
        bit_rates_dict = { 7 : 48000, 6 : 44100, }

        return [ str(bit_rates_dict[bit]) for
                 bit in bit_rates_dict if rate & (1 << bit) != 0 ]
```

I forgot the `str()` earlier, is that why it was failing for you?

README.md
marc-hb: The way this 3 scripts interact with each other is really not intuitive and it took me an unreasonable amount time to figure out. Since you're adding a new feature to the deprecated script (while leaving the newer script behind?), please add a couple more sentences describing the situation better. Right now this part of the README.md is worse than nothing.

tools/tplgtool.py
fredoh9: wow, it works nicely. I just updated again. Let me check the test results before remove draft

README.md
fredoh9: > The way this 3 scripts interact with each other is really not intuitive.
I agree this part.

I don't understand rest of your comment. Why do you think tplgtool.py is deprecated? README was misleading. When tplgtool2.py was introduced, (if it wants to replace tplgtool.py) it has to take care of all the functions in tplgtool.py. But it didn't. More likely it added a new file and put a deprecated notice on the currently being used script.

marc-hb (approved):

README.md
marc-hb: > Why do you think tplgtool.py is deprecated? 

Because it says so:
 
```
 warnings.warn("tplgtool.py is deprecated, use tplgtool2.py instead.", DeprecationWarning)

```

marc-hb: Glad I'm not the only one confused!

kv2019i:
This has some side-effects to tests with PCMs that have lower than 48000 rates:
https://github.com/thesofproject/sof/issues/9067#issuecomment-2069187188

marc-hb:
@fredoh9 should we revert temporarily while you design a fix?

fredoh9:
> @fredoh9 should we revert temporarily while you design a fix?

No, it will break every case. I will take a look shortly

PR #1156, opened 2024-01-25
kmod: fix hda-sdw-bpt dependency

plbossart:
We need to remove the SoundWire parts before removing hda-sdw-bpt

rmmod: ERROR: Module snd_sof_intel_hda_common is in use by: snd_sof_intel_hda_sdw_bpt

Link: https://sof-ci.01.org/linuxpr/PR4679/build1181/devicetest/index.html?model=MTLP_SDW_AIOC&testcase=check-kmod-load-unload

PR #1182, opened 2024-04-25
sof_remove: unload `snd_sof_intel_hda_common` before `soundwire_intel`

marc-hb:
Fixes this failure with older kernel like the 6.5 kernels distributed by Ubuntu:
```
 ERROR: Module soundwire_intel is in use by: snd_sof_intel_hda_common
```
Fixes commit 65951c4d0 ("kmod: fix hda-sdw-bpt dependency") which unloaded all soundwire drivers earlier as a block. This is not a problem with 6.9-ish kernels where snd_sof_intel_hda_common and soundwire_intel can apparently be unloaded in any order but it is a problem with older kernels.

So, unload `snd_sof_intel_hda_common` before soundwire modules to be compatible with any kernel.

See https://github.com/thesofproject/sof-test/pull/1156 for more info.

### DIFF
--- a/tools/kmod/sof_remove.sh
+++ b/tools/kmod/sof_remove.sh
@@ -120,6 +120,14 @@ remove_module snd_soc_hda_codec
 # platform drivers
 #-------------------------------------------
 remove_module snd_sof_intel_hda_generic
+# Attempt 1/2
+#
+# - In the SDW BRA future this will be "in use by
+#   snd_sof_intel_hda_sdw_bpt" and will fail.  Ignore that failure.
+# - In the 6.5-ish "present" 'soundwire_intel' depends on it so it must
+#   be removed first.
+# https://github.com/thesofproject/sof-test/pull/1182
+remove_module snd_sof_intel_hda_common || true
 
 #-------------------------------------------
 # SoundWire/SOF parts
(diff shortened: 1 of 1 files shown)


marc-hb:
Known ALSABAT failure in https://sof-ci.01.org/softestpr/PR1182/build356/devicetest/index.html

Everything else green.

tools/kmod/sof_remove.sh
```diff
@@ -120,6 +120,7 @@ remove_module snd_soc_hda_codec
 # platform drivers
 #-------------------------------------------
 remove_module snd_sof_intel_hda_generic
+remove_module snd_sof_intel_hda_common
```
plbossart, line 123: I am pretty sure this does not work if you have snd_sof_intel_hda_sdw_bpt in use.

See https://github.com/thesofproject/linux/pull/4679/commits/af779a3ca9bae844fb1691caaba7243e49afbe71#diff-117faa254cc1264b915358fff9c601a8b15e7545c3, the BPT stuff depends on the hda_common stuff.

so if you want to deal with older kernels, you can add this line and ignore errors, and keep it line 136 where it was.

plbossart (changes requested):

plbossart (approved):

marc-hb:
Pretty bad looking but isolated and unrelated crash of `ba-lnlm-rvp-sdw-01`  in https://sof-ci.01.org/softestpr/PR1182/build370/devicetest/index.html?model=LNLM_SDW_AIOC&testcase=multiple-pause-resume-50

Everything else 100% green.

PR #1183, opened 2024-04-26
lib.sh: catch some mtrace-reader failures to start thanks to "OSError"

marc-hb:
This makes tests fail at the end when mtrace-reader.py could be started, for instance because one is already running (see example below)

Reminder: you can use -s to ignore firmware logs.
```
  File "mtrace-reader.py", line 20, in <module>
    fd = os.open(MTRACE_FILE, os.O_RDONLY)
OSError: [Errno 16] Device or resource busy: '/sys/kernel/debug/sof/mtrace/core0'
```

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -527,7 +527,7 @@ check_error_in_fw_logfile()
     # -B 2 shows the header line when the first etrace message is an ERROR
     # -A 1 shows whether the ERROR is last or not.
     if (set -x
-        grep -B 2 -A 1 -i --word-regexp -e 'ERR' -e 'ERROR' -e '<err>' "$1"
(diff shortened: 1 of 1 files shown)


marc-hb:
Unrelated sof-logger issue in https://sof-ci.01.org/softestpr/PR1183/build360/devicetest/index.html

Everything 100% green.

kv2019i (approved):
Typo in commit message "could be started" -> "could not be started" ? .

But patch obvious enough.

marc-hb:
Thanks for the review! And for catching the typo. I would have pushed a fix for the commit title but I don't think it's worth it for one sentence in the commit message body; the risk of confusion seems low enough.

Issue #1112, opened 2023-10-11
[BUG] empty kernel logs on MAX_WAIT_FW_LOADING failures


marc-hb:
**Describe the bug**

#1059 added the ability to wait for the firmware to be loaded, which solved a number of issues, including some i915 timeout issues.

However sof-test does not collect kernel logs when this fails. This is very problematic because the only way to know why the firmware is not loaded is to look at kernel logs.

**To Reproduce**

- sudo rm -rf sof-tests/logs/* # do not delete the logs/ directory itself
- `sof-test/tools/kmod/sof_remove.sh`
- Run any test.
- wc -l logs/*/*/dmesg.txt

=> dmesg.txt is empty

**Expected behavior**

Kernel logs are collected.

**Detail Info**
1) Branch name and commit hash of the 3 repositories: sof (firmware/topology), linux (kernel driver) and sof-test (test case)
    * Kernel: any
    * SOF: any
    * SOF-TEST: e3d76647ec77
2) Test report ID (if you find it from test report)
    * ID: any. Recent examples in https://github.com/thesofproject/linux/pull/4628
3) Test DUT Model (or a brief discribtion about the device)
    * MODEL: any
4) Test TPLG
    * TPLG: any
5) Test case (what test script and how you run it)
    * TESTCASE: any

cc:
- #1036

marc-hb:
Seen again in https://sof-ci.01.org/sofpr/PR8550/build552/devicetest/index.html (#8550)

marc-hb:
We had a surge of FW failing to load for various reasons. We need to fix this, first (and small) part:
- #1170

marc-hb:
I just noticed #1036 by chance, possibly a duplicate

marc-hb:
Fix submitted: #1191

marc-hb:
> Only 2 CI TIMEOUTS, all other tests timed out after MAX_WAIT_FW_LOADING.

That's only because MTL still has an unreasonably long MAX_WAIT_FW_LOADING, otherwise there would have not been any CI timeout.

Fixed in:
- #1196

marc-hb:
#1196 merged, good enough.

Issue #1036, opened 2023-05-15
[BUG] No kernel boot log when SOF FW loading fails


fredoh9:
**Describe the bug**
No kernel log is captured in test result. This is side effect after https://github.com/thesofproject/sof-test/pull/1033 merged.
'start_test' will be called before the every test case, if it fails there, too early to capture the log

**To Reproduce**
This is one of example, FW loading is failed
https://sof-ci.01.org/sofpr/PR7584/build7585/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=verify-kernel-boot-log

**Expected behavior**
kernel log should be available on the failure

marc-hb:
> I think @lyakh found an issue with this new feature: kernel logs seem to be missing when the firmware does not boot. Example: https://sof-ci.01.org/sofpr/PR7584/build7585/devicetest/index.html

In jenkins logs generic_test/13429/
```
10:58:40  cmd: scp -r ubuntu@sh-adlp-rvp-sdw-01.sh.intel.com:~/sof-test/test-case/../logs/verify-firmware-presence/last/* results/2023-05-12-03:55:59/sh-adlp-rvp-sdw-01/verify-firmware-presence
10:58:40  return code: 1
10:58:40  cmd output: scp: /home/ubuntu/sof-test/test-case/../logs/verify-firmware-presence/last/*: No such file or directory
```

This is very likely because the new code runs "too early".

_Originally posted by @marc-hb in https://github.com/thesofproject/sof-test/issues/1033#issuecomment-1548052202_

marc-hb:
Fix submitted: #1191

marc-hb:
Should be good, will re-open if any issue.

PR #1187, opened 2024-05-01
Explicitly invoke start_test() in every test-case/*.sh

marc-hb:
2 commits, main one:

Stop unconditionally imposing start_test() to everything sourcing lib.sh

This makes it possible to source lib.sh without running anything which
can be useful for debugging.

For now this should be a no-op for all tests except two files in `tools/` which
were never really "tests" in the first place:

- tools/kmod/sof_remove.sh
- tools/sof-kernel-log-check.sh

In these, start_test() never made sense it never really run thanks the
(awkward) "is_subtest()" escape.

There are other tests where start_test() should not be invoked either,
most notably test-case/verify-kernel-boot-log.sh in order to fix
https://github.com/thesofproject/sof-test/issues/1112 but this is a huge
commit already; one change at a time.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -66,6 +66,12 @@ minvalue() { printf '%d' $(( "$1" < "$2"  ? "$1" : "$2" )); }
 #    concurrency 1% of the time, detecting it 99% of the time is much
 #    more than enough to spot reservation problems.
 #
+# 3. Register the func_exit_handler() EXIT trap that collects logs, kills
+#    loggers and prints test results.
+#
+# 4. Waits for the FW to be successfully booted (can be disabled)
+#
+#
 start_test()
 {
     if is_subtest; then
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -1024,4 +1024,3 @@ perf_analyze()
     fi
 }
 
-start_test
--- a/test-case/check-alsabat.sh
+++ b/test-case/check-alsabat.sh
@@ -73,6 +73,8 @@ frequency=${OPT_VAL['F']}
 sigmak=${OPT_VAL['k']}
 frames=${OPT_VAL['n']}
 
+start_test
+
 if [ "$pcm_p" = "" ]||[ "$pcm_c" = "" ];
 then
 	dloge "No playback or capture PCM is specified. Skip the alsabat test"
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -41,6 +41,8 @@ loop_cnt=${OPT_VAL['l']}
 func_pipeline_export "$tplg" "eq:any"
 sofcard=${SOFCARD:-0}
 
+start_test
+
 # Test equalizer
 func_test_eq()
 {
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -57,6 +57,7 @@ loop_cnt=${OPT_VAL['l']}
 out_dir=${OPT_VAL['o']}
 file_prefix=${OPT_VAL['f']}
 
+start_test
 logger_disabled || func_lib_start_log_collect
 
 setup_kernel_check_point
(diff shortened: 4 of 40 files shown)


marc-hb:
This is the one-before-last, big step towards fixing:
- https://github.com/thesofproject/sof-test/issues/1112

... and friends.

Known headset capture issue https://github.com/thesofproject/sof/issues/9018  in LNL https://sof-ci.01.org/softestpr/PR1187/build383/devicetest/index.html, everything else green.

One single suspend/resume failure in CAVS https://sof-ci.01.org/softestpr/PR1187/build381/devicetest/index.html,  everything else green.

stable-v2.2 all green https://sof-ci.01.org/softestpr/PR1187/build384/devicetest/index.html

case-lib/lib.sh
```diff
@@ -66,6 +66,12 @@ minvalue() { printf '%d' $(( "$1" < "$2"  ? "$1" : "$2" )); }
 #    concurrency 1% of the time, detecting it 99% of the time is much
 #    more than enough to spot reservation problems.
 #
+# 3. Register the func_exit_handler() EXIT trap that collects logs, kills
+#    loggers and prints test results.
+#
+# 4. Waits for the FW to be successfully booted (can be disabled)
```
ujfalusi, line 72: How can it be disabled?

case-lib/lib.sh
```diff
@@ -66,6 +66,12 @@ minvalue() { printf '%d' $(( "$1" < "$2"  ? "$1" : "$2" )); }
...
+#
+#
```
ujfalusi, line 74: nitpick: one blank line should be enough?

test-case/check-alsabat.sh
```diff
@@ -73,6 +73,8 @@ frequency=${OPT_VAL['F']}
 sigmak=${OPT_VAL['k']}
 frames=${OPT_VAL['n']}
 
+start_test
+
```
ujfalusi, line 77: In commit message, what is the meaning of the opening sentence in relation to moving the start_test from lib.sh to test cases?
`Stop unconditionally imposing start_test() to everything sourcing lib.sh`

test-case/check-audio-equalizer.sh
```diff
@@ -41,6 +41,8 @@ loop_cnt=${OPT_VAL['l']}
 func_pipeline_export "$tplg" "eq:any"
 sofcard=${SOFCARD:-0}
 
+start_test
```
ujfalusi, line 44: Isn't the `start_test` used to run at line 21?

test-case/check-capture.sh
```diff
@@ -57,6 +57,7 @@ loop_cnt=${OPT_VAL['l']}
 out_dir=${OPT_VAL['o']}
 file_prefix=${OPT_VAL['f']}
 
+start_test
```
ujfalusi, line 60: All of this churn is to keep the `systemctl_show_pulseaudio` in tools/kmod/sof_remove.sh instead of just dropping it from the sof_remove.sh along with sourcing the case-lib/lib.sh and be done?

marc-hb: > instead of just dropping it from the sof_remove.sh along with sourcing the case-lib/lib.sh 

The (non-mandatory) pulseaudio check is debatable. Including `lib.sh` is much less so. There's no reason to copy/paste routines when 

Anyway this most important and urgent benefit of this design fix is preparing the fix for #1036 and #1112. The same design fix just happens to help with some other scripts like sof_remove.sh.

test-case/verify-kernel-boot-log.sh
```diff
@@ -18,6 +18,11 @@ set -e
 # shellcheck source=case-lib/lib.sh
 source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
+start_test
+# TODO: replace start_test with this:
```
fredoh9, line 22: This TODO will be next PR?

case-lib/lib.sh
fredoh9: agree it's nitpick

test-case/verify-kernel-boot-log.sh
marc-hb: Yes immediately after this one. I want at least one run of daily tests between the two.

fredoh9 (approved):
Initially start_test from lib.sh was making sense. but it's not cover for all now.
Explicit invoking start_test() is making sense now.

marc-hb:
Thanks @fredoh9 !

PR #1191, opened 2024-05-08
verify-kernel-boot-log.sh: fix missing boot logs on failure

marc-hb:
Provide a new, boot_logs.txt file no matter what happens. Note that file now includes user-space logs, not just kernel logs.

Also: fix bug where script times out and does not run when the SOF
firmware is not loaded: no need to define NO_POLL_FW_LOADING anymore.

The trick is to stop calling `start_test()`. verify-kernel-boot-log.sh is not an audio test!

Fixes sof-test issues #1036 and #1112, find more details there.

### DIFF
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -18,10 +18,16 @@ set -e
 # shellcheck source=case-lib/lib.sh
 source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
 
-start_test
-# TODO: replace start_test with this:
-# trap 'print_test_result_exit $?' EXIT
-# https://github.com/thesofproject/sof-test/issues/1112
+trap 'boot_log_exit_handler $?' EXIT
+
+boot_log_exit_handler()
+{
+    local exit_code=$1
+
+    journalctl --boot > "$LOG_ROOT"/boot_log.txt
+
+    print_test_result_exit "$exit_code"
+}
 
 main()
 {
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -3,6 +3,9 @@
 SUDO_CMD=$(command -v sudo)
 
 # Overwrite other functions' exit to perform environment cleanup
+#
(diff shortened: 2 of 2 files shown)


marc-hb:
Wow, ALL hardware tests are 100% green!

The new `boot_log.txt` file is present in all of them.

test-case/verify-kernel-boot-log.sh
```diff
@@ -18,10 +18,16 @@ set -e
...
-# TODO: replace start_test with this:
-# trap 'print_test_result_exit $?' EXIT
-# https://github.com/thesofproject/sof-test/issues/1112
+trap 'boot_log_exit_handler $?' EXIT
+
+boot_log_exit_handler()
+{
+    local exit_code=$1
+
+    journalctl --boot > "$LOG_ROOT"/boot_log.txt
```
fredoh9, line 27: Normally I don't pay attention to user space. What can we expect more clues from user space log?

fredoh9 (approved):
Beauty of trap! Looks good.
Thanks a lot!

test-case/verify-kernel-boot-log.sh
marc-hb: > What can we expect more clues from user space log?

- Sound daemons errors (pulseaudio, pipewire,...)
- Display / auto-login issues: https://github.com/thesofproject/sof-test/discussions/964, https://github.com/thesofproject/sof-test/discussions/998
- Anything else that could give clues.

If you look at some examples, you'll see the added volume is low compared to SOF debug logs. One example on a fresh boot:

```
root@jf-lnlm-rvp-nocodec-3:~# journalctl -k | wc -l
1857
root@jf-lnlm-rvp-nocodec-3:~# journalctl -b | wc -l
2313
```

marc-hb: A couple more after looking at `diff -u <(journalctl -k) <(journalctl -b) | less`

- Network initialization
- NTP init
- sensors stuff (thermal)
- other systemd logs

PR #1192, opened 2024-05-10
verify-kernel-boot-log.sh: add "standalone" dmesg.txt back

marc-hb:
Interleaving could be useful when it's a problem with a sound daemon but tend to turn these off and most of the time it's just making SOF logs harder to read.

Keep boot_log.txt which is useful for all non-audio issues.

Fixes commit c8565e8b210d ("verify-kernel-boot-log.sh: fix missing boot logs on failure")


plbossart (approved):
yay @marc-hb

PR #1195, opened 2024-05-15
verify-kernel-boot-log.sh: cap boot logs to 6000 max_lines

marc-hb:
Protect storage in case this is not run first and immediately after boot as it should. Also makes the logs more relevant and usable: long files can make the user interface unusable, see example in https://github.com/thesofproject/sof/issues/8761

This also gives us --no-pager "for free" in case it was ever needed.

Fixes recent commits c8565e8b210d and 914c5c06e075.

### DIFF
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -24,10 +24,19 @@ boot_log_exit_handler()
 {
     local exit_code=$1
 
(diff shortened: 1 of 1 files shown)


marc-hb:
SOFCI TEST

marc-hb:
Everything green except the well known LNL alsabat issue in https://sof-ci.01.org/softestpr/PR1195/build451/devicetest/index.html

fredoh9 (approved):

PR #1199, opened 2024-05-21
[SKIP SOF-TEST] Two new, simple topology2 parsing Python scripts

marc-hb:
Does anyone find these useful? I did...


jsarha (approved):
Very nice demo how to use tplgtool2 package.  Should probably implement a tplg2 diff, that would compare two topologies and would show only their differences from SOF driver perspective.

PR #1200, opened 2024-05-23
verify-kernel-log.sh: guard "xrandr" with a short "timeout"

marc-hb:
xrandr can hang! Who could have known? See evidence in:

- https://github.com/thesofproject/linux/pull/5009#issuecomment-2122857477

Don't wait for CI to timeout and kill the whole script so we don't lose precious logs.

### DIFF
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -146,8 +146,9 @@ show_daemons_session_display()
      # a session-specific, /run/user/ tmpfs that cannot be seen over ssh (or from any
      # other session). Sometimes this shows XWAYLAND.
      # As of April 2024 there does not seem to be any Wayland equivalent for `xrandr`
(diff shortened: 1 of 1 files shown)


ujfalusi (approved):
Hrm, how does it locks up? anything in dmesg?
What else is going to be broken when xrandr hangs? Is it safe to test display audio for example?

marc-hb:
Thanks for the review!

> Hrm, how does it locks up? anything in dmesg?

There was nothing specific to xrandr but there were some i915 and drm errors, see my comments in https://github.com/thesofproject/linux/pull/5009#issuecomment-2122857477

I think they were related.

The main issue is: I had to use ssh to get these logs - because xrandr hung we were deprived of published logs = what this PR will hopefully fix.

> What else is going to be broken when xrandr hangs? Is it safe to test display audio for example?

It looks like the rest of the test in that linux PR passed but who knows - exactly why we need logs.

PR #1202, opened 2024-06-03
Fix check-sof-logger.sh

marc-hb:
2 commits

As usual, @fredoh9 clean up and simplification of the internal, Intel deployment scripts exposed some old bugs that were hidden so far because we have too many fallbacks in this area.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -309,11 +309,11 @@ get_ldc_subdir()
 {
     local subdir='intel/sof' # default
     local fw_path
-    local fw_path_info='/sys/kernel/debug/sof/fw_path'
+    local fw_path_info='/sys/kernel/debug/sof/fw_profile/fw_path'
 
     # either $fw_path_info exists, OR we redefine $fw_path_info with
     # the backwards-compatible alternative based on kernel parameter
-    test -e $fw_path_info ||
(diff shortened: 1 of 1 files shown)


case-lib/lib.sh
```diff
@@ -323,6 +325,8 @@ get_ldc_subdir()
             subdir=${subdir%/community}
             subdir=${subdir%/intel-signed}
             subdir=${subdir%/dbgkey}
+            subdir=${subdir%/dbgkey}
```
fredoh9, line 328: duplicated?

fredoh9:
Minor a comment but looks good

case-lib/lib.sh
marc-hb: Good catch thank you

marc-hb (commented):
Test results look good in LNL
https://sof-ci.01.org/softestpr/PR1202/build472/devicetest/index.html, MTL https://sof-ci.01.org/softestpr/PR1202/build469/devicetest/index.html and stable-v2.2 https://sof-ci.01.org/softestpr/PR1202/build471/devicetest/index.html

Will push again to address the review comment

fredoh9 (approved):
Looks good, safe to merge

PR #1203, opened 2024-06-04
[HACK] Skip Jack In PCM device

plbossart:
Test the hypothesis that the Bank switch issues are due to concurrency between Jack In and Jack Out usage - on the same link.


marc-hb:
https://sof-ci.01.org/softestpr/PR1203/build477/devicetest/index.html?model=LNLM_SDW_AIOC&testcase=check-playback-3times

> [  376.126183] kernel: soundwire sdw-master-0-2: Controller Timed out on bank switch

plbossart:
Very interesting, this is the first time I see the error for the Speakers, using link2
The problem is not related to the jack then.
This doesn't tell you what the issue is, but it tells us what this isn't related to. That's already progress.
@bardliao FYI

PR #1214, opened 2024-07-06
[DO-NOT-MERGE]temp workarounds for known issues

ssavati:
ignore : DRM error Unclaimed access detected prior to suspending
ignore : SDW Parity error detected
workaround : multiple-pipeline, use duration instead of pkill


tools/sof-kernel-log-check.sh
```diff
@@ -149,6 +149,12 @@ trap - EXIT
 # TODO explain why we ignore this one and where
 ignore_str='error: debugfs write failed to idle -16'
 
+#TWLignore DRM errors
+ignore_str="$ignore_str"'|i915 [[:digit:].:]+: \[drm\] \*ERROR\* Unclaimed access detected prior to suspending'
```
marc-hb, line 153: I've been deploying the following `modprobe.d` workaround and it's been working great:

`options i915 mmio_debug=-1`

https://github.com/intel-innersource/drivers.audio.ci.sof-framework/issues/573

I think you can drop this.

PR #1215, opened 2024-07-08
case-lib: remove check for log ldc file

kv2019i:
Checking for ldc (log dictionary) file existance as a way to decide whether to enable or disable logging during a test case, is no longer applicable as SOF supports multiple firmware logging solutions that do not use an ldc file.

Remove the check as it no longer makes sense.

Link: https://github.com/thesofproject/sof-test/issues/1216

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -874,13 +874,6 @@ is_ipc4()
 
 logger_disabled()
 {
-    local ldcFile
-    # Some firmware/OS configurations do not support logging.
-    ldcFile=$(find_ldc_file) || {
-        dlogi '.ldc dictionary file not found, SOF logs collection disabled'
-        return 0 # 0 is 'true'
-    }
-
     # Disable logging when available...
     if [ ${OPT_VAL['s']} -eq 0 ]; then
         return 0


kv2019i:
FYI @marc-hb @lyakh @ssavati @andyross

marc-hb (changes requested):
This code is useful for stable-v2.2

Please change to:

```shell
is_firmware_file_zephyr || ldcFile=$(find_ldc_file) || {

```

kv2019i:
Btw, this doesn't yet fix the check-sof-logger test case, but this does get FW logs back in other tests.

marc-hb (approved):
Ok you convinced me, let's remove this. In fact we still have two other ways to disable logging: `-s` in most tests and some environment variable, see 982f92db56e9a and #811 where this was discussed.

I think it was IPC4 migration time with more configurations and transitions that we have now.

I see check-sof-logger is still failing but now we have FW logs back thanks to this.

Just for the record, the loss of firmware logs (and other regressions) came from the big UUID PR which was merged broken

https://github.com/thesofproject/sof/pull/9287#issuecomment-2214017105

Issue #1216, opened 2024-07-08
[BUG] PR test fail to .ldc file not found!


kv2019i:
**Describe the bug**
Example PR test fail from today 8th July 2024 
https://sof-ci.01.org/sofpr/PR9186/build6326/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=check-sof-logger

Failure started to be seen after merging
https://github.com/thesofproject/sof/pull/9261

andyross:
Pushed up #1217 to (hopefully) address the CI problems with sof-test not having been updated to the new build behavior.

PR #1217, opened 2024-07-09
check-sof-logger.sh Predicate LDC search on !zephyr

andyross:
[I haven't completely figured out how this script runs in CI, so this patch is mostly thrown up blind for review.  But this is where the error in CI is coming from, I believe]

Zephyr builds don't produce .ldc files anymore, don't look for them and fail.

Fixes: #1216


andyross:
Can't add reviewers. @marc-hb @kv2019i @dbaluta and whoever else is bumping into this on their PRs will want to take a look.

marc-hb (approved):
Test results look good!

Thanks a lot, I was afraid it would be more complicated (this script is very convoluted in other places) but this seems enough, great!

PR #1218, opened 2024-07-09
Add new Expect script `tools/apause.exp`

marc-hb:
This all started as an effort to solve test failures caused by "MAX"
volume. See #931 and others linked from there for more context.

Another obvious issue was the duplication of `expect` code across two
tests (check-pause-resume.sh and multiple-pause-resume.sh)

After a bit of time actually "testing the tests" I realized the original
author did not really understand `expect` or the problem. So I rewrote
the entire `expect` part. The list of previous issues is a bit too long
not to forget any but here are some:

- The script could get "out of sync" with aplay and report that it was
  paused when it was actually resumed! And vice-versa.

- De-duplication; obviously.

- Moving to a separate script also solves the following problems:
  - Can be invoked, tested and debugged separately outside any shell script
  - Simplifies quoting
  - Unlocks editor features like syntax checking

- Proper understanding and handling of newlines.

- The expect script does not "sleep" anymore, which stops backpressuring
  and blocking the console output from aplay - with unknown side effects!

- Adding `log_user 0` makes the test logs readable at last

- Add decent logging for easier maintenance

- Logging timestamps demonstrate that the entire aproach is too slow for
  pause/resume cycles shorter than ~200 ms. Default values won't be
  changed yet but at least the problem is now obvious.

- Handle "MAX" volume! Not an error yet because it happens really across
  the board (MAX does usualy not happen long enough to timeout and
  _fail_ across the board) but the code is ready to upgrade the "MAX"
  warning to an ERROR with a one-line change.

- Report EOF and timeouts differently.

- Probably others.

### DIFF
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -17,8 +17,9 @@ set -e
 ##    no error happen for aplay/arecord
 ##
 
+TOPDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)
 # shellcheck source=case-lib/lib.sh
-source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
+source "$TOPDIR"/case-lib/lib.sh
 
 OPT_NAME['t']='tplg'     OPT_DESC['t']='tplg file, default value is env TPLG: $''TPLG'
 OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
--- a/test-case/check-pause-resume.sh
+++ b/test-case/check-pause-resume.sh
@@ -93,40 +93,19 @@ do
     dev=$(func_pipeline_parse_value "$idx" dev)
     snd=$(func_pipeline_parse_value "$idx" snd)
 
-    # expect is tcl language script
-    #   expr rand(): produces random numbers between 0 and 1
-    #   after ms: Ms must be an integer giving a time in milliseconds.
-    #       The command sleeps for ms milliseconds and then returns.
-    dlogi "Entering expect script with:
-      $cmd $SOF_ALSA_OPTS $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name -q"
+    ret=0
+    (set -x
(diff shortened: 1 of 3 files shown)


marc-hb:
stable-v2.2 results look good https://sof-ci.01.org/softestpr/PR1218/build598/devicetest/index.html

LNL failures in https://sof-ci.01.org/softestpr/PR1218/build600/devicetest/index.html are unrelated

marc-hb:
NOCODEC timeouts in https://sof-ci.01.org/softestpr/PR1218/build597/devicetest/index.html :-( HDA is OK. Test just taking too long?

marc-hb:
https://sof-ci.01.org/softestpr/PR1218/build632/devicetest/index.html?model=LNLM_RVP_NOCODEC&testcase=multiple-pause-resume-50 ends in some infinite loop of overruns (which are not firmware errors?) but that seems like a real issue, not a problem with the test.

Everything else is 100% green
https://sof-ci.01.org/softestpr/PR1218/build631/devicetest/index.html
https://sof-ci.01.org/softestpr/PR1218/build629/devicetest/index.html
https://sof-ci.01.org/softestpr/PR1218/build630/devicetest/index.html

marc-hb:
Good example of MAX saturation timeout that would be addressed by this:

https://sof-ci.01.org/softestpr/PR1220/build633/devicetest/index.html?model=LNLM_RVP_HDA&testcase=check-pause-resume-capture-100

test-case/check-pause-resume.sh
```diff
@@ -17,8 +17,9 @@ set -e
 ##    no error happen for aplay/arecord
 ##
 
+TOPDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)
```
kv2019i, line 20: Please remove the extra signed-off-by line.

kv2019i (approved):
Looks good. Can you fixup the one bogus signed-off-by line?

test-case/check-pause-resume.sh
marc-hb: Done. Great catch, I'm impressed :-)

PR #1223, opened 2024-07-19
check-suspend-resume: check process id in each loop for audio test case

fredoh9:
check-suspend-resume-with-audio.sh uses check-suspend-resume.sh. process check is in check-suspend-resume-with-audio.sh and check-suspend-resume.sh has the loop. Must check that the process id is available in each loop. TO do so, pass aplay/arecord pid to check-suspend-resume.sh.

### DIFF
--- a/test-case/check-suspend-resume-with-audio.sh
+++ b/test-case/check-suspend-resume-with-audio.sh
@@ -27,34 +27,34 @@ TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/" && pwd)
 source "${TESTDIR}"/../case-lib/lib.sh
 
 OPT_NAME['l']='loop'     OPT_DESC['l']='suspend/resume loop count'
-OPT_HAS_ARG['l']=1         OPT_VAL['l']=3
+OPT_HAS_ARG['l']=1       OPT_VAL['l']=3
 
 OPT_NAME['T']='type'     OPT_DESC['T']="suspend/resume type from /sys/power/mem_sleep"
-OPT_HAS_ARG['T']=1         OPT_VAL['T']=""
+OPT_HAS_ARG['T']=1       OPT_VAL['T']=""
 
 OPT_NAME['S']='sleep'    OPT_DESC['S']='suspend/resume command:rtcwake sleep duration'
-OPT_HAS_ARG['S']=1         OPT_VAL['S']=5
+OPT_HAS_ARG['S']=1       OPT_VAL['S']=5
 
 OPT_NAME['w']='wait'     OPT_DESC['w']='idle time after suspend/resume wakeup'
-OPT_HAS_ARG['w']=1         OPT_VAL['w']=5
+OPT_HAS_ARG['w']=1       OPT_VAL['w']=5
 
 OPT_NAME['r']='random'   OPT_DESC['r']="Randomly setup wait/sleep time, this option will overwrite s & w option"
-OPT_HAS_ARG['r']=0         OPT_VAL['r']=0
+OPT_HAS_ARG['r']=0       OPT_VAL['r']=0
 
 OPT_NAME['m']='mode'     OPT_DESC['m']='alsa application type: playback/capture'
-OPT_HAS_ARG['m']=1         OPT_VAL['m']='playback'
+OPT_HAS_ARG['m']=1       OPT_VAL['m']='playback'
 
 OPT_NAME['t']='tplg'     OPT_DESC['t']="tplg file, default value is env TPLG: $TPLG"
-OPT_HAS_ARG['t']=1         OPT_VAL['t']="$TPLG"
+OPT_HAS_ARG['t']=1       OPT_VAL['t']="$TPLG"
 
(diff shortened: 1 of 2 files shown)


fredoh9:
the main problem I'm trying to fix with this PR is...
For check-suspend-resume-with-audio.sh test, if loop count is 100, and the process is exit with input/output error, even with the error, testing keep going. hard to figure out the last minute of dmesg/mtrace for the error.

test-case/check-suspend-resume-with-audio.sh
```diff
@@ -134,6 +136,10 @@ do
         ps --ppid $$ -f
         exit 1
     }
+    # remove -p $process_id option, Need to unset twice
+    unset "opt_arr[${#opt_arr[@]}-1]"
+    unset "opt_arr[${#opt_arr[@]}-1]"
```
ranj063, line 141: why twice?

marc-hb (commented):
EDIT: just read your comment above.

This seems a bit complicated... why not just check the process status when the subtest returns a failure? I mean here:

```shell
"${TESTDIR}"/check-suspend-resume.sh "${opt_arr[@]}" || {
  
  # new code to check process status

  die "suspend resume failed"
} 
```

test-case/check-suspend-resume-with-audio.sh
marc-hb: 1. -p 2. PID

test-case/check-suspend-resume-with-audio.sh
```diff
@@ -119,6 +119,8 @@ do
         ps --ppid $$ -f
         exit 1
     }
+    # Pass -p $process_id option to check-suspend-resume.sh
+    opt_arr+=(-p "$process_id")
```
marc-hb, line 123:
```suggestion
    suspend_opts+=("${opt_arr[@]}" -p "$process_id")
    "${TESTDIR}"/check-suspend-resume.sh "${opt_arr[@]}"
```

test-case/check-suspend-resume-with-audio.sh
marc-hb: You don't need this if you simply use a copy, see `suspend_opts` above.

Every time you change a variable in place, consider making a copy instead. It's always better for correctness and concurrency (it's the default in functional languages)

test-case/check-suspend-resume.sh
```diff
@@ -44,6 +44,10 @@ OPT_HAS_ARG['w']=1         OPT_VAL['w']=5
 OPT_NAME['r']='random'   OPT_DESC['r']="Randomly setup wait/sleep time, range is [$random_min-$random_max], this option will overwrite s & w option"
 OPT_HAS_ARG['r']=0         OPT_VAL['r']=0
 
+# processid is set by check-suspend-resume-with-audio.sh for audio test case
+OPT_NAME['p']='processid'     OPT_DESC['p']='Process ID of aplay or arecord'
```
marc-hb, line 48: You didn't describe the main purpose/feature you're adding:
```suggestion
OPT_NAME['p']='processid'     OPT_DESC['p']='Fail immediately if this process dies'
```

Ideally, this should be generic and work with any process, not just aplay or arecord.

test-case/check-suspend-resume.sh
```diff
@@ -196,6 +200,17 @@ sleep_once()
     [ "$stats_success" -eq "$expected_stats_success" ] ||
         dump_and_die "/sys/power/suspend_stats/success is $stats_success, expected $expected_stats_success"
     check_suspend_fails || dump_and_die "some failure counts have changed"
+    # if OPT_VAL['p'] has process id, then check the process id is available in the system
+    if [ "${OPT_VAL['p']}" ]; then
+        dlogi "Check for the process status, pid: ${OPT_VAL['p']}"
+        sof-process-state.sh "${OPT_VAL['p']}" || {
+            #func_lib_lsof_error_dump "$snd"
```
marc-hb, line 207: Why commented out? This is a failure and it's very likely very hard to reproduce. We want as much information as possible. This will be called only once.

marc-hb (changes requested):
I think the idea is good, small changes requested.

test-case/check-suspend-resume.sh
fredoh9: $snd is only available in check-suspend-resume-with-audio.sh
I put a FIXME comment

marc-hb: So if this can never work then why have this commented out line at all?

marc-hb (commented):
Code change looks good but why:
- all these large, git-breaking whitespace changes?
- commented out code that can never work?

fredoh9:
split into two commits, first commit has no functionality change.

marc-hb (approved):

marc-hb:
A good few shellcheck warnings in https://github.com/thesofproject/sof-test/actions/runs/10084785339/job/27884307459?pr=1223

Most seem old and all seem harmless but it's a pity to include a cleanup commit in this PR and still have some many warnings :-(

PR #1224, opened 2024-07-22
Add and start using generic functions fw_reldir() and fw_relfilepath()

marc-hb:
Add two generic functions fw_reldir() and fw_relfilepath() which search in this order:

1. /sys/kernel/debug/sof/fw_profile/
2. /sys/module/snd_sof_pci/parameters/
3. Kernel logs.

The 3., slower, "legacy" option is kept because it's useful with either old kernels or when the modules have been unloaded.

Start using these two generic functions in `is_firmware_file_zephyr()` and `get_ldc_subdir()`.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -310,23 +310,14 @@ get_ldc_subdir()
 {
     local strip_arg="$1"
     local subdir='intel/sof' # default
+
     local fw_path
-    local fw_path_info='/sys/kernel/debug/sof/fw_profile/fw_path'
-
-    # either $fw_path_info exists, OR we redefine $fw_path_info with
-    # the backwards-compatible alternative based on kernel parameter
-    sudo test -e $fw_path_info ||
-	fw_path_info='/sys/module/snd_sof_pci/parameters/fw_path'
-
-    if fw_path=$(sudo cat $fw_path_info); then
-	# "cat" was succesful
-        if [ "$fw_path" != '(null)' ]; then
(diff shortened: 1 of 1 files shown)


marc-hb:
Usual LNL failures in https://sof-ci.01.org/softestpr/PR1224/build686/devicetest/index.html, everything else 100% green (even suspend/resume!)

case-lib/lib.sh
```diff
@@ -825,22 +825,30 @@ is_zephyr_ldc()
 # https://github.com/thesofproject/linux/issues/3867 and friends. Keep this existing
 # journalctl_cmd() code as a fallback for older kernels or when the firmware has been
```
fredoh9: journalctl_cmd() -> journalctl

case-lib/lib.sh
```diff
@@ -840,6 +840,45 @@ fw_reldir()
     printf '%s' "$(dirname "$from_klogs")"
 }
 
+# Prints the relative firmware filepath: e.g. `intel/sof-ipc4/tgl/community/sof-tgl.ri`
+#
+# Some of the code duplicates fw_reldir() above. Unfortunately can't re-use and invoke
```
fredoh9, line 845: I tried to put a comment of duplication and possible re-use, good comment here

fredoh9 (approved):
reviewed commit by commit. much make sense than reviewing entire change. Easy to follow.
Thank you for fixing the annoying error.

PR #1226, opened 2024-08-01
Catch aplay error messages in apause.exp and exit

marc-hb:
Always better to "fail fast".

See commit messages.

### DIFF
--- a/case-lib/apause.exp
+++ b/case-lib/apause.exp
@@ -81,7 +81,7 @@ proc rel_time_ms {} {
 
 proc press_space {} {
     global last_space_time
-    log 2 "Pressing SPACE"
+    log 1 "Pressing SPACE"
     send " "
     set last_space_time [rel_time_ms]
     log 3 "last_space_time set to $last_space_time"
@@ -237,12 +237,8 @@ expect {
 
     {=== PAUSE ===} {
         if {$state != "pause_requested"} {
-            # TODO: upgrade this to an ERROR if we want to fix pause bugs like
-            # https://github.com/thesofproject/linux/issues/5109
-            # As of July 2024, pause is rather being disabled:
-            # https://github.com/thesofproject/linux/pull/5041
-            log 0 "WARNING: received == PAUSE == while in state $state! Ignoring."
-            exp_continue
+            log 0 "ERROR: received == PAUSE == while in state $state!"
+            exit 1
         }
 
         set state paused
--- a/case-lib/apause.exp
+++ b/case-lib/apause.exp
@@ -137,7 +137,19 @@ set volume_always_zero true
 
 expect {
 
+    # `man re_syntax` claims that Tcl regular expressions are compliant
+    # with the basic and extended POSIX ones while adding a 3rd,
+    # extended flavor.  It's not clear which flavor `expect -re` uses
+    # but it's not the basic one.
+    # Use {} not "" to avoid quoting issues and backslash proliferation.
+
     # When multiple patterns match, first pattern wins.
+
+    -nocase -re {error.*\r|PAUSE.*no[[:blank:]]*hw[[:blank:]]*support.*\r} {
+        set buffer_with_lf "[cr_to_lf $expect_out(buffer)]"
+        log 0 "ERROR: $buffer_with_lf"
+        exit 1
+    }
 
     # Volume xx% or MAX line
     #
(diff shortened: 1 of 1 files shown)


marc-hb:
Perfect demonstration of the value of this PR:

https://sof-ci.01.org/softestpr/PR1226/build710/devicetest/index.html?model=LNLM_RVP_NOCODEC&testcase=check-pause-resume-playback-100
```
t=13242 ms: aplay: Pressing SPACE
t=14869 ms: aplay: (32/100) Found   === PAUSE ===  ,  pausing for 137 ms
t=14869 ms: aplay: ERROR: aplay: do_pause:1567: pause push error: File descriptor in bad state

```

marc-hb:
SOFCI TEST

case-lib/apause.exp
```diff
@@ -137,8 +137,20 @@ set volume_always_zero true
...
 
+    # `man re_syntax` claims that Tcl regular expressions are compliant
+    # with the basic and extended POSIX ones while adding a 3rd,
+    # extended flavor.  It's not clear which flavor `expect -re` uses
+    # but it's not the basic one.
+    # Use {} not "" to avoid quoting issues and backslash proliferation.
+
     # When multiple patterns match, first pattern wins.
 
+    -nocase -re {error.*\r|PAUSE.*no[[:blank:]]*hw[[:blank:]]*support.*\r} {
```
plbossart, line 148: laughing at my desk at 8am, no idea what this does haha.

case-lib/apause.exp
```diff
@@ -137,8 +137,20 @@ set volume_always_zero true
...
+        set buffer_with_lf "[cr_to_lf $expect_out(buffer)]"
+        log 0 "ERROR: $buffer_with_lf"
+        exit 1
+    }
+
```
plbossart, line 153: just to be clear on the commit message: PAUSE will be disabled unless we have a kernel parameter that keeps it enabled. All CI platforms SHALL keep the pause enabled.

plbossart (commented):
LGTM but the last time I used TCL was in 1996, can't really approve or reject.

marc-hb:
LNL https://sof-ci.01.org/softestpr/PR1226/build716/devicetest/index.html?model=LNLM_SDW_AIOC&testcase=multiple-pause-resume-50 has `pause push error: File descriptor in bad state` across three models.

case-lib/apause.exp
marc-hb: It took me only a couple days of practice to become productive in Tcl and expect (without any "stackoverflow garbage in, AI garbage out!); it's not so bad.

`{ }` is like the single quotes in shell except they can be nested.

The basic syntax of the `expect` instruction is just:
```
expect {
 { pattern1 } { action1 }
 { pattern2 } { action2 }
 ...
}
```
`-nocase -re` means the pattern is a case-insensitive regular expression (as opposed to globbing).

That regular expression has itself nothing specific to Tcl.

Thanks for taking a look!

plbossart:
"pause push error: File descriptor in bad state" is not necessarily a problem, we see it all the time when resuming and INFO_RESUME is not supported.

I've seen this log countless times and things were just fine, alsa-lib reinitialized whatever was never with a prepare.

Adding @ranj063 and @ujfalusi to make sure I am not completely mistaken.

case-lib/apause.exp
marc-hb: That's what I tried to say but more briefly and without getting into Intel CI specifics. Too briefly? I mean, should I change the commit message or is your comment here enough?

plbossart: thats' fine, just wanted to check we were aligned.

marc-hb:
> "pause push error: File descriptor in bad state" is not necessarily a problem, we see it all the time when resuming and INFO_RESUME is not supported.

Mmmm.... questions:

- why are we seeing this error message ONLY on LNL? Before this PR catches this error, LNL was also the only platform where the test failed (later).
- Can you please elaborate on "INFO_RESUME is not supported." is not supported? Is it the same as "pause not supported" that's coming with https://github.com/thesofproject/linux/pull/5041? If it's the same and it's not supported in some configuration, then the test should fail.

ranj063 (approved):

marc-hb:
> Can you please elaborate on "INFO_RESUME is not supported." ? Is it the same as "pause not supported" 

Yes it's the same according to @ranj063 .

Thanks for the reviews. This is a well tested and small change - easy to revert parts of it if needed so I'm merging it.

PR #1229, opened 2024-08-15
multiple pipelines: skip XTOS + NOCODEC configurations

marc-hb:
2 commits

Stop polluting test results daily.


marc-hb:
Test results all as expected.

fredoh9 (approved):
I don't like the idea of skipping test case(s) in the test case itself. And I'm afraid we might forget the known issue(s) and give impression of no known issue.

But we don't fix this issue for stable-v2.2. Without this known issue, it gives a lot easier tracking the test results.
So approved.

marc-hb:
Thanks! I don't like it either but this one will never get fixed.

PR #1230, opened 2024-09-07
tools: tplgtool2: Fix array size calculation for enum values

ranj063:
Convert the computed value to integer to set the values array size for parsing the enum controls properly.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -314,6 +314,7 @@ class TplgBinaryFormat:
         self._hw_config_max = 8 # SND_SOC_TPLG_HW_CONFIG_MAX
         self._tlv_size = 32 # SND_SOC_TPLG_TLV_SIZE
         self._id_name_maxlen = 44 # SNDRV_CTL_ELEM_ID_NAME_MAXLEN
+        self._id_name_maxint32 = 44 // 4
         self._num_texts = 16 # SND_SOC_TPLG_NUM_TEXTS
 
         self._block_header = Struct( # snd_soc_tplg_hdr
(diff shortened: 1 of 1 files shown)


tools/tplgtool2.py
```diff
@@ -470,7 +470,7 @@ def __init__(self):
             "mask" / Int32ul,
             "count" / Int32ul,
             "texts" / Array(self._num_texts, String(self._id_name_maxlen, "ascii")),
-            "values" / Array(self._num_texts * self._id_name_maxlen / 4, Int32ul),
+            "values" / Array(int(self._num_texts * self._id_name_maxlen / 4), Int32ul),
```
marc-hb, line 473: I think it's safer and clearer to avoid floats and conversions entirely:
```suggestion
            "values" / Array(self._num_texts * self._id_name_maxlen // 4, Int32ul),
```

Or even maybe better:
```suggestion
            self._id_name_maxint32 = 11
            ....
            "values" / Array(self._num_texts * self._id_name_maxint32, Int32ul),
```

ranj063: what is the problem with conversion?

marc-hb: Every floating point operation should immediately raise the "What is the rounding error this time?" red flag. Even when there's none, proving that there is none is a waste of mental energy. It's just weird to perform pure-integer arithmetic using floats; I bet most people would object to that in C.

```
python3
>>> 0.1+0.2
0.30000000000000004
>>>
```

It's also weird to repeatedly divide 44 by 4 every time.

marc-hb (commented):
BTW I'm really curious what `construct` was doing with this float... if not converting it to an `int` then what else? This sounds like a bug in `construct`, it should probably provide a better message

https://sof-ci.01.org/sofpr/PR9441/build7682/devicetest/index.html?model=MTLP_RVP_HDA&testcase=verify-tplg-binary

singalsu (approved):
Checked to fix the issue, thanks!

tools/tplgtool2.py
ranj063: fair enough. fixed now

marc-hb (approved):

PR #1231, opened 2024-09-10
LNL alsabat test to support both USB Codec/AudioPlug loopback setup

ssavati:


test-case/check-alsabat.sh
```diff
@@ -81,6 +81,37 @@ then
...
+	    pcm_c="hw:sofhdadsp,0"
+	    dlogi "Capture device changed to $pcm_c"
+        else
+	    pcm_p="hw:sofhdadsp,0"
+	    dlogi "Playback device changed to $pcm_p"
+	fi
+fi
+
+# Check device for AudioPlug Loopback enabled or not for LNL SDW
+if [ "$AUDIOPLUG_LOOPBACK" == "true" ] && [ "$MODEL" == "LNLM_SDW_AIOC" ]; then
```
plbossart, line 100: you probably want to avoid using the MODEL, we can use this test on previous generations as well.

marc-hb (commented):
This is a big hack with a lot of ugly hardcoding :-(

I think there is a much simpler, more flexible and more reliable way. Something like this (untested)

```shell
usb_audio_found()
{
   aplay -l | grep -q 'USB Audio'
}



if [[ "$pcm_p" =~ CODEC ]] && ! usb_audio_found; then
   dlogi "No USB audio found, discarding pcm_p=$pcm_p"
   unset pcm_p
fi

# <same thing for pcm_c>

# default values for pcm_p
if [ -n "$pcm_p" ]; then
   case $MODEL in
   *HDA)  pcm_p=...;;
    ...)
    
    *) die "No default pcm_p for MODEL=$MODEL"
    esac
fi

# <default values for pcm_c>



```

marc-hb:
BTW do you know why we have all these SKIPs? https://sof-ci.01.org/sofpr/PR9455/build7827/devicetest/index.html?model=LNLM_RVP_HDA&testcase=check-alsabat-nocodec-32bits-599

plbossart:
> BTW do you know why we have all these SKIPs? https://sof-ci.01.org/sofpr/PR9455/build7827/devicetest/index.html?model=LNLM_RVP_HDA&testcase=check-alsabat-nocodec-32bits-599

I am not sure why we have an alsa-bat test that's specific to the nocodec mode, but it makes sense that it would only be applied to nocodec platforms and skipped for both HDaudio and SoundWire platforms.

plbossart:
> ```shell
> usb_audio_found()
> {
>    aplay -l | grep -q 'USB Audio'
> }
> ```

This is also fragile in that it will fail if there's a USB device connected by accident.
I think it's better to rely on environment variables set by the CI setup to understand when we want to use the analog dongle or the USB loopback.

marc-hb:
> This is also fragile in that it will fail if there's a USB device connected by accident.

These loopback setups involve a fair amount of cables and are a bit tedious to install, it's not really something anyone has been connecting "by accident".

> I think it's better to rely on environment variables set by the CI setup to understand when we want to use the analog dongle or the USB loopback.

The main issue is: we don't have a reliable way to maintain per-device settings (in fact we don't even have a reliable way to maintain _consistent_ settings either). Unlike hardware configuration "accidents", software configuration accidents have been happening regularly.

Considering it's impossible for a device to be able to test both configurations, merely echoing the hardware configuration in a configuration file seems just error-prone with no value.

All this being said, `usb_audio_found()` can very easily be replaced by `TEST_USB_LOOPBACK=true` in my sample code. My suggestion was not just about that.

PR #1233, opened 2024-09-23
sof-kernel-log-check: ignore i915 firmware loading error on MTL

kv2019i:
Ignore FW load error on MTL. These don't affect audio test execution.


marc-hb:
Check the alternative, more flexible #1234.

marc-hb:
#1234 merged... @kv2019i close?

PR #1234, opened 2024-09-23
Add new local_config.bash files to solve device-specific issues

marc-hb:
Source additional files /etc/sof/local_config.bash and
sof-test/case-lib/local_config.bash.

This helps solving device-specific issues like https://github.com/thesofproject/sof-test/pull/1233 and many others
before that - for instance git blame the "ignore_str" variable in
sof-kernel-log-check.sh

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -427,18 +427,15 @@ else
 fi
 
 declare -p cmd
-# check priority err for error message
-if [[ "$ignore_str" ]]; then
-    err=$($cmd --priority=err | grep -vE "$ignore_str")
-else
-    err=$($cmd --priority=err)
-fi
 
-[[ -z "$err" ]] || {
+if err=$($cmd --priority=err |
+             grep -v -E -e "$ignore_str"); then
+
     type journalctl_cmd
     echo "$(date -u '+%Y-%m-%d %T %Z')" "[ERROR]" "Caught kernel log error"
     echo "===========================>>"
     echo "$err"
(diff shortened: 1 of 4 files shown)


marc-hb:
Expected failures.

tools/sof-kernel-log-check.sh
```diff
@@ -419,6 +419,18 @@ ignore_str="$ignore_str"'|nvme0: Admin Cmd\(0x[[:digit:]]+\), I/O Error \(sct 0x
...
 
+# This expects an array like for instance:
+#    sof_local_extra_kernel_ignores=(-e 'error 1' -e err2)
+# Arrays
+# - provide whitespace/quoting safety
+# - can be empty/optional
+# - can be arbitrarily complex. Best avoided but only on specific systems anyway.
+# shellcheck disable=SC2154
+if &>/dev/null declare -p sof_local_extra_kernel_ignores; then
+    dlogw "Ignoring extra errors on this particular system:"
```
kv2019i, line 430: One concern I have with this facility is that when adding local exceptions becomes easier, it can be used too much and people reading the test logs might not realize some DUTs have local exceptions others don't have. The global ignore list has drawbacks, but at least one needs to make a git commit with proper description and leave something that can be search through git history. OTOH, this dlogw does make local exception use clear in the log, which alleviates my concern. It really depends on how this will be used.

kv2019i (approved):
Looks good @marc-hb . I have some concerns, but I think they are more about the usage of this feature. Probably this should be merged to give the option, but just like to record my concern as a comment.

PR #1239, opened 2024-11-26
Create PTLP_RVP_SDW.sh

harajend:
Added Amixer setting for PTL SDW config

### DIFF
--- a/alsa_settings/PTLP_RVP_SDW.sh
+++ b/alsa_settings/PTLP_RVP_SDW.sh
@@ -1,0 +1,6 @@
+set -e
+
+# enable headset playback and capture
+amixer -c sofsoundwire cset name='Headset Mic Switch' on
+amixer -c sofsoundwire cset name='rt722 FU0F Capture Switch' on
+amixer -c sofsoundwire cset name='rt722 FU0F Capture Volume' 50


sofci:

marc-hb:
test this please

mohsinkh30 (commented):
verified

alsa_settings/PTLP_RVP_SDW.sh
```diff
@@ -0,0 +1,6 @@
+set -e
+
+# enable headset playback and capture
+amixer -c sofsoundwire cset name='Headset Mic Switch' on
+amixer -c sofsoundwire cset name='rt722 FU0F Capture Switch' on
+amixer -c sofsoundwire cset name='rt722 FU0F Capture Volume' 50
```
bardliao, line 6: Not sure if we need to set 'rt722 FU0F Capture Volume'. Maybe it is better to keep it as the default value?

bardliao (approved):

alsa_settings/PTLP_RVP_SDW.sh
harajend: For the default value of 75, the test fails with the error message 'over amplitude detected.' We reduced the capture volume and tested values between 30 and 50, which worked fine. Hence, we kept the same value here

kv2019i (approved):

PR #1240, opened 2024-12-16
Added "Kernel: ACPI" Error message to ignored string list

harajend:
Kernel and Boot log check test failed Due to ACPI bios error and its nothing harm to Audio functionality. Hence i added ACPI error as ignored string.


sofci:

marc-hb (commented):
The commit title and message sound like you're  ignoring all ACPI errors which would be excessive. Audio very much depends on ACPI. Fortunately, the actual code is more specific. They should align better with each other.

How do you know that "can't tag data node" can never affect audio on any device? What does that message mean?

If this error is specific to some device type, then it's better to add this line in the local configuration file of only these devices, see `sof_local_extra_kernel_ignores` and commit 83f5e4190e9f2 (this new feature should be better advertised,  maybe  in this file)

bardliao (approved):

marc-hb (approved):

bardliao (approved):

Issue #1243, opened 2024-12-17
[BUG] /proc/asound/sofprobes breaks "check-sof-logger" test


harajend:
**Describe the bug**
"check-sof-logger" test failed with error message "[ERROR] SOF ALSA card not available, driver probe fail?"

**Triage details**
Test case passed upon modified line 74 in the "check-sof-logger.sh" file
Original code: "test -e /proc/asound/sof*/id"
Modified code: "test -e /proc/asound/sofs*/id"
Cause: Above command line is looking for **"sofsoundwire"** directory, but in this case 2 directories were located for **"sof*"** string match

**To Reproduce**
1. Launch Terminal and Run "TPLG=/lib/firmware/intel/sof-ipc4-tplg/sof-ptl-rt722-4ch.tplg MODEL=PTLP_RVP_SDW SOF_TEST_INTERVAL=5 ~/sof-test/test-case/check-sof-logger.sh"

**Expected behavior**
- Test should PASS

**Detail Info**

1. Linux kernel and SOF Firmware details:
   * Linux Build Result: [2842](https://sof-ci.ostc.intel.com/#/buildresult/linuxbuildresult/2842)
   * Linux Branch: [topic/sof-dev](https://github.com/thesofproject/linux/tree/topic/sof-dev)
   * Linux Commit: [89772eb035b4](https://github.com/thesofproject/linux/commit/89772eb035b40a44c327e2d12de3a9d4450999e0)
   * KConfig Branch: [master](https://github.com/thesofproject/kconfig/tree/master)
   * KConfig Commit: [097c69ae4b81](https://github.com/thesofproject/kconfig/commit/097c69ae4b812778f64edc321b11d5b458357bb5)
   * SOF Build Result: [6015](https://sof-ci.ostc.intel.com/#/buildresult/sofbuildresult/6015)
   * SOF Branch: [main](https://github.com/thesofproject/sof/tree/main)
   * SOF Commit: [f4cd55c2e879](https://github.com/thesofproject/sof/commit/f4cd55c2e879e7f822cb793bd65449fa1da36b04)
   * Zephyr Commit: [7fa8bf17563b](https://github.com/zephyrproject-rtos/zephyr/commit/7fa8bf17563bc57d37605a724da61786f7759c28)
   * SOF TEST PR: [1240](https://github.com/thesofproject/sof-test/pull/1240)

2. Test report ID (if you find it from test report)
[...]

PR #1245, opened 2024-12-17
Updated check-sof-logger.sh and verify-kernel-boot-log.sh

harajend:
Updated SOF-specific codecs since the sof* regex also lists sofprobes.

### DIFF
--- a/test-case/check-sof-logger.sh
+++ b/test-case/check-sof-logger.sh
@@ -71,7 +71,10 @@ sof_alsa_card_found()
     # note: assumes SOF card names to start with "sof", e.g.
     #   - /proc/asound/sofsoundwire/id
     #   - /proc/asound/sofhdadsp/id
-    test -e /proc/asound/sof*/id
+    for i in /proc/asound/sof*/id; do
+        if test -e "$i"; then return 0; fi
+    done
+    return 1
 }
 
 wait_for_sof_alsa_card()
--- a/test-case/verify-kernel-boot-log.sh
+++ b/test-case/verify-kernel-boot-log.sh
@@ -115,7 +115,8 @@ dmic_switch_present()
     (set -x
      # name= is hardcoded in /usr/share/alsa/ucm2/*
      # This returns a non-zero error status on failure
-     amixer cget name='Dmic0 Capture Switch'
+     switch=$(aplay -l | head -2 | tail -1 | awk '{print $3}')
+     amixer -c "$switch" cget name='Dmic0 Capture Switch'
     )
 }
 


test-case/check-sof-logger.sh
```diff
@@ -71,7 +71,8 @@ sof_alsa_card_found()
     # note: assumes SOF card names to start with "sof", e.g.
     #   - /proc/asound/sofsoundwire/id
     #   - /proc/asound/sofhdadsp/id
-    test -e /proc/asound/sof*/id
+    test -e /proc/asound/sofsoundwire/id ||
+        test -e /proc/asound/sofhdadsp/id
```
marc-hb, line 75: The comment above does not make any sense after this change and must be removed (as shown in https://github.com/thesofproject/sof-test/issues/1243#issuecomment-2549132322)

marc-hb (changes requested):
- Are there any other possibilities besides these two? @kv2019i , @ranj063 , @bardliao can you answer?

- The commit message must link to the corresponding, e.g. "Fixes issue #1243 

- The commit title and message must describe the change much precisely. For instance: `check-sof-logger: remove wildcard from sof_alsa_card_found()`

See commit message examples in the git log.

Here are some good references, most of it apply to any open-source project so please take some time and take a look

https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-guidelines

https://wiki.openstack.org/wiki/GitCommitMessages

https://www.qemu.org/docs/master/devel/submitting-a-patch.html#write-a-meaningful-commit-message

https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes

marc-hb:
Simpler and probably good enough:

```sh
sof_alsa_card_found()
{
   for i in /proc/asound/sof*/id; do
      if test -e "$i"; then return 0; fi
   done
   return 1
}
```

marc-hb:
You need to use `git rebase -i` and squash the 2 git commits into a single one. We don't want to keep trial-and-error in the history; every git commit must work. Neither Zephyr nor SOF use Github in the "usual" way, they use it in the "traditional git" way https://github.com/zephyrproject-rtos/zephyr/issues/39194

Also, see advice above for a useful commit message.

marc-hb:
Jenkins did not test this at all :-(

marc-hb:
> Jenkins did not test this at all :-(

Still not running :-(

marc-hb:
Jenkins still not running; pushing new versions does not make any difference. Something is wrong on the Jenkins side. cc: @kv2019i

marc-hb:
There is still no testing.

Squashing looks good but the commit message of https://github.com/thesofproject/sof-test/pull/1245/commits/212a42023c04e3e0b92ca9c40a7204e74682b161  is still a mess:

```
Updated the test script with suggested changes

Signed-off-by: harajend <hariprasad.rajendra@intel.com>

Updated check-sof-logger.sh

Updated SOF-specific codecs since the sof* regex also lists sofprobes.

Signed-off-by: harajend <hariprasad.rajendra@intel.com>
```

Don't say "update", every commit is an update. Just describe the change (once).

test-case/verify-kernel-boot-log.sh
```diff
@@ -115,7 +115,8 @@ dmic_switch_present()
     (set -x
      # name= is hardcoded in /usr/share/alsa/ucm2/*
      # This returns a non-zero error status on failure
-     amixer cget name='Dmic0 Capture Switch'
+     switch=$(aplay -l | head -2 | tail -1 | awk '{print $3}')
+     amixer -c "$switch" cget name='Dmic0 Capture Switch'
```
kv2019i, line 119: Hmm, doesn't this still select a fixed card? 

E.g.
```
~$ aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: sofhdadsp [sof-hda-dsp], device 0: HDA Analog (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

... you'd get the same with "amixer -c 0" gets us.

test-case/check-sof-logger.sh
```diff
@@ -71,7 +71,10 @@ sof_alsa_card_found()
     # note: assumes SOF card names to start with "sof", e.g.
```
kv2019i, line 71: Change is good, but the commit message seems wrong. There are two signed offs. Can you amend the message and repush?

kv2019i (changes requested):
Mostly good, but some comments inline.

test-case/verify-kernel-boot-log.sh
marc-hb: `awk` is a "big boy" and can select lines itself. It's shorter AND more readable:

```shell
   switch=$(aplay -l |  awk 'NR == 2 {print $3}')
```

Of course `amixer -c 0` looks even better.

marc-hb:
test this please

marc-hb:
Jenkins is still not running :-( @harajend please try closing this one and re-submitting in a new PR?

test-case/verify-kernel-boot-log.sh
harajend: > @harajend But "head -2 | tail -1" will always pick the first card, so this is same as passing "amixer -c 0", right?

Yes, On functionality front both are same. Here I thought to pass "name/identifier" instead of "index number"

harajend:
Submitted new PR's for sof-logger and kernel-boot-log failures. Hence closing this (https://github.com/thesofproject/sof-test/pull/1245) pull request.
Sof-logger: https://github.com/thesofproject/sof-test/pull/1248
Kernel-boot-log: https://github.com/thesofproject/sof-test/pull/1249

PR #1248, opened 2025-01-13
Test-case: Updated check-sof-logger.sh with SOF-specific codec

harajend:
Updated SOF-specific codecs since the sof* regex also lists sofprobes


sofci:

marc-hb:
test this please

test-case/check-sof-logger.sh
```diff
@@ -71,7 +71,11 @@ sof_alsa_card_found()
     # note: assumes SOF card names to start with "sof", e.g.
     #   - /proc/asound/sofsoundwire/id
     #   - /proc/asound/sofhdadsp/id
-    test -e /proc/asound/sof*/id
+    #	SOF-specific codecs since the sof* regex also lists sofprobes
```
marc-hb, line 74: This comment is impossible to make sense of without knowing either the older code, or knowing the context described in bug #1243. But #1243 is referenced neither here nor in the commit message.

It is pretty obvious that the new code is designed to support multiple `sof*`, so please just drop this comment.

BTW `sof*` is NOT a regex, it's globbing (a regular expression would look like `sof.*`)

test-case/check-sof-logger.sh
marc-hb, line 1: The `Signed-off-by:` line must be the last one in the commit message. Some git scripts may look for it.

This is a follow up to #1245 which Jenkins stopped testing for some unknown reason. I see Jenkins testing this one now - great.

harajend: > The `Signed-off-by:` line must be the last one in the commit message. Some git scripts may look for it.
> 
> This is a follow up to #1245 which Jenkins stopped testing for some unknown reason. I see Jenkins testing this one now - great.

Yes, this is follow-up for #1245, i will close #1245 PR, since its not running Jenkins

marc-hb: #1245 had two (unrelated?) fixes, #1248 has only one for now.

test-case/check-sof-logger.sh
harajend: > This new comment is impossible to make sense of without knowing either the older code, or knowing the context described in bug #1243. But #1243 is referenced neither here nor in the commit message.
> 
> It is pretty obvious that the new code is designed to support multiple `sof*`, so please just drop this comment.
> 
> BTW `sof*` is NOT a regex, it's globbing (a regular expression would look like `sof.*`)

Added the changes

test-case/check-sof-logger.sh
```diff
@@ -71,7 +71,14 @@ sof_alsa_card_found()
     # note: assumes SOF card names to start with "sof", e.g.
     #   - /proc/asound/sofsoundwire/id
     #   - /proc/asound/sofhdadsp/id
-    test -e /proc/asound/sof*/id
+    # test -e /proc/asound/sof*/id
```
kv2019i, line 74: @harajend The git commit style is still a bit different than what is the established practise. Please check "git log" history for sof-test to see examples. E.g. there should be some prefix like "check-sof-logger: fix support for multiple SOF cards" on first line. Commit should also explain what is fixed and why, "check-sof-logger.sh fails to detect SOF in case mulitple SOF cards are present in the system. Fix the logic to check the SOF card."

test-case/check-sof-logger.sh
```diff
@@ -71,7 +71,14 @@ sof_alsa_card_found()
...
+    # - https://github.com/thesofproject/sof-test/issues/1243
+    # Internal issue #615
```
kv2019i, line 76: I'd avoid these internal references. At least should mention "Intel internal issue tracker".

kv2019i (approved):
No blockers, PR seems good, but I'd encourage to look up old git commits to get a feel of the style on conventions used in this repo.

test-case/check-sof-logger.sh
marc-hb: I don't think any reference or comment is required _in the source_ in this particular case. It's pretty obvious a `for` loop is designed to support multiple devices.

If you really want to add a comment, say something like `# we have multiple sof* devices every time we enable sofprobes`. When possible, it's better not to have to open a link to understand something. References in the source are for complicated problems that are too long to explain in place. Otherwise the commit message is enough: you can have as many references as you want in the git log.

marc-hb (approved):

ujfalusi (approved):

PR #1254, opened 2025-02-20
sof-kernel-log-check.sh: ignore ISH firmware missing error

harajend:
ignore ISH firmware missing error in PTL since it is not impact to audio test cases

### DIFF
--- a/tools/sof-kernel-log-check.sh
+++ b/tools/sof-kernel-log-check.sh
@@ -318,6 +318,9 @@ esac
 # on one CML_HEL_RT5682
 ignore_str="$ignore_str"'|kernel: i2c_hid_acpi i2c-GDIX0000:00'
 
+# ISH firmware missing failure can be ignored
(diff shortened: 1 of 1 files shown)


sofci:

kv2019i (approved):

redzynix (approved):

cgturner1 (approved):

marc-hb (commented):
Note there was already some ISH ignore in this regex. No one noticed because its size growing totally out of control. That's why it's now possible to add exclusions on a per device basis.

I wonder how long before this hits some maximum command line or regex size limit...

golowanow:
btw, is this error specific to ptl or is it on all platforms ?

marc-hb:
In my experience, this type of error is specific to... "the latest platform still under development" :-)

The ISH for PTL is probably still under development AND there is always a lag between internal and external releases - same as with SOF and everything else.

The "fix" is to be more aggressive and deploy internal releases on test devices, see discussion in internal issues 596, 572 and others.

harajend:
> btw, is this error specific to ptl or is it on all platforms ?

Currently we have seen this issue in PTL systems only!

PR #1255, opened 2025-02-25
tools: sof-ipc-timer.py: Timings of module init and config IPC loggings

jsarha:
The sof-ipc-timer collects module initialization and configuration timings from 'journalctl -k -o short-precise' output if SOF IPC debug is enabled.

### DIFF
--- a/tools/sof-ipc-timer.py
+++ b/tools/sof-ipc-timer.py
@@ -1,0 +1,318 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2025 Intel Corporation. All rights reserved.
+
+'''The sof-ipc-timer collects module initialization and configuration
+timings from 'journalctl -k -o short-precise' output if SOF IPC debug
+is enabled.
+
+Without any flags there is no output. If you want it all, just put
+everything on the command line: -t -i -c -m -p -s
+
+'''
+
+import re
+import sys
+import argparse
+from datetime import datetime
+
+class Component:
+    '''SOF audio component storage class'''
+    pipe_id: int
+    comp_id: int
+    wname: str
+    init_times: list
+    conf_times: list
+
+    def __init__(self, pipe_id, comp_id, wname):
+        self.pipe_id = pipe_id
+        self.comp_id = comp_id
+        self.wname = wname
+        self.init_times = []
+        self.conf_times = []
+
+    def __str__(self) -> str:
+        return f'{self.pipe_id}-{self.comp_id:#08x}'
+
+class Pipeline:
+    '''SOF audio pipeline storage class'''
+    pipe_id: int
+    comps: list
+    state_times: dict
+
+    def __init__(self, pipe_id):
+        self.pipe_id = pipe_id
+        self.comps = []
+        self.state_times = {}
+
+    def __str__(self) -> str:
+        return f'pipeline.{self.pipe_id + 1}'
+
+    def add_state_timing(self, state, usecs):
+        if self.state_times.get(state) is None:
+            self.state_times[state] = []
+        self.state_times[state].append(usecs)
+
+class LogLineParser:
+    def __init__(self, args, comp_data, pipe_data):
+        self.args = args
+        self.comp_data = comp_data
+        self.pipe_data = pipe_data
+
+class PipelineParser(LogLineParser):
+    '''Parse line of form
+
+    Feb 25 23:17:00.598919 lnlm-rvp-sdw kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-lnl 0000:00:1f.3: Create pipeline pipeline.1 (pipe 1) - instance 0, core 0
+
+    'pipe_id' from ' instance 0,'
+    '''
+    def __init__(self, args, comp_data, pipe_data):
+        super().__init__(args, comp_data, pipe_data)
+
+    def parse_line(self, line):
+        if match_obj := re.search(r" Create pipeline ", line):
+            match_end_pos = match_obj.span()[1]
+            line_split = line[match_end_pos + 1:].split()
+            pipe_id = int(line_split[5].rstrip(","))
+            if self.pipe_data.get(pipe_id) is None:
+                self.pipe_data[pipe_id] = Pipeline(pipe_id)
+            return True
+        return False
+
+class WidgetParser(LogLineParser):
+    '''Parse line of form:
+
+    Feb 25 23:17:00.599838 lnlm-rvp-sdw kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-lnl 0000:00:1f.3: Create widget host-copier.0.playback (pipe 1) - ID 4, instance 0, core 0
+
+    'widget_name' from 'host-copier.0.playback'
+    'pipi_id' integer form "(pipe 1)"
+    'widget_id' MSB integer from "instance 0," and LSB from "ID 4,"
+    '''
+    def __init__(self, args, comp_data, pipe_data):
+        super().__init__(args, comp_data, pipe_data)
+
+    def parse_line(self, line):
+        if match_obj := re.search(r" Create widget ", line):
+            match_end_pos = match_obj.span()[1]
+            line_split = line[match_end_pos:].split()
+            widget_name = line_split[0]
+            # pipeline instance id is one smaller than the pipeline number
+            pipe_id = int(line_split[2].rstrip(")")) - 1
+            module_instance_id = int(line_split[7].rstrip(","))
+            widget_id = int(line_split[5].rstrip(","))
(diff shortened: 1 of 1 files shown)


tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+                        help='Show module id')
+    parser.add_argument('-p', '--pipeline-id', action="store_true", default=False,
+                        help='Show pipeline id')
+
+    return parser.parse_args()
+
+def main():
+    args = parse_args()
+
+    with open(args.filename, 'r', encoding='utf8') as file:
```
marc-hb, line 142: IMHO a tool like this should really be "stdin-capable". Or even stdin-only which is flexible enough:
```
journalctl -k | tools/sof-ipc-timer.py
tools/sof-ipc-timer.py < capture.txt
```

```python
for line in sys.stdin:
```
``

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+    '''Pick ipc tx MOD_INIT_INSTANCE and MOD_LARGE_CONFIG_SET lines
+    from log, take time stamp from the first message and the 'done'
+    message, and store the difference of to either init_time or
+    conf_time in the component table.
+
+    '''
+    mod_id = None
+    start = None
+    for line in f:
+        if match_obj := re.search(r" ipc tx (     |reply|done )", line):
```
marc-hb, line 81: The massive length of the indented block is a sure sign it should split into separate functions. Small functions also make interactive testing possible, that's what the `"__main__"` idiom below is for:

```shell
ipython3

>> import sof-ipc-timer
>> small_function("some log line")
```

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self, ppln_id, comp_id, wname):
+        self.ppln_id = ppln_id
+        self.comp_id = comp_id
+        self.wname = wname
+
+    def __str__(self) -> str:
+        return f'{self.ppln_id}-{self.comp_id:#08x}'
+
+def read_comp_data(f):
+    '''Process the dmesg to get the component ID to component name mapping,
```
marc-hb, line 32:
```suggestion
    '''Process the kernel logs to get the component ID to component name mapping,
```
`dmesg` wraps around. `journalctl -k` does not.

marc-hb (commented):
I didn't look at any of the logic, just some general Python advice.

tools/sof-ipc-timer.py
lyakh: @marc-hb currently the file is read twice (at least AFAICS), wouldn't be very convenient with `stdin`

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+                        pipeline_id = "\tpipeline id: " + str(comp.ppln_id)
+                    if msg_part[2] == "MOD_INIT_INSTANCE":
+                        print("%s:\tinit done\t%d us%s%s" %
+			      (comp.wname, usecs - start, module_id, pipeline_id))
+                    elif msg_part[2] == "MOD_LARGE_CONFIG_SET":
+                        print("%s:\tconf done\t%d us%s%s" %
+			      (comp.wname, usecs - start, module_id, pipeline_id))
+                    mod_id = None
+                    start = None
+                    end = None
```
lgirdwood, line 123: Could also do with more inline comments describing each stage

lgirdwood (commented):
Just needs some more readability for me.

tools/sof-ipc-timer.py
marc-hb: > The massive length of the indented block is a sure sign it should split into separate functions.

@lgirdwood 's comment just reminded me of something: naming is hard (c), but when you find good function names they act like a Table of Contents that lets you quickly get the gist of what is happening without reading the whole code/book.

Functions are not just about code re-use, they're useful even when you invoke them only once; for testability _and_ readability.

jsarha:
I am actually thinking of turning this PR to draft, as it does not produce the accuracy needed to to show the DRAM offloading impact, so I do not think its worth my time to polish this the merge-able quality. Or maybe I come up with something extra to this script parsing FW logs too with some extra lines for the purpose, that can be turned on with a kcontrol option.

lgirdwood:
@jsarha ping

jsarha:
> @jsarha ping

There has been some further development since this version. I'll finish those developments first and then update and clean up everything according to comments here.

tools/sof-ipc-timer.py
jsarha: Its hard to split the contents of the loop into separate functions, parsing of each line make changes to the variables variables to be used on later line. AFAIK there is no way to pass strings as reference in python... well I can try to rewrite the whole thing in object oriented manner,  but I am not sure if it becomes more readable that way.

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+    for line in f:
+        if match_obj := re.search(r" ipc tx (     |reply|done )", line):
+            line_split = line.split()
+            dt_object = datetime.strptime(line_split[2], "%H:%M:%S.%f")
+            secs = dt_object.second + dt_object.minute * 60 + dt_object.hour * 3600
+            usecs = secs * 1000000 + dt_object.microsecond
+            span_end_pos = match_obj.span()[1]
+            # msg_type shoule be either "     ", "reply", or "done "
+            msg_type = line[match_obj.span()[0] + 8 : match_obj.span()[1]]
+            msg_part = line[span_end_pos:].split()
```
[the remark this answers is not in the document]
jsarha: Copilot obviously does not have a clue what the piece of code is doing and why.

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+        Create pipeline pipeline.3 (pipe 3) - instance 3, core 0
+
+    In practice, sof-test only capture kernel message and firmware trace generated during a test
+    case run. Mostly in manual tests, if the kernel message file contains multiple firmware runs
+    with overlapping information, the last one wins.
+    '''
+    comp_data = {}
+    ppln_id = None
+    for line in f:
+        if match_obj := re.search(r"Create (widget|pipeline)", line):
```
marc-hb, line 49:
```suggestion
        if not match_obj := re.search(r"Create (widget|pipeline)", line):
            continue
```

tools/sof-ipc-timer.py
```diff
@@ -0,0 +1,149 @@
...
+            line_split = line[span_end_pos + 1:].split()
+            widget_name = line_split[0]
+            # In the linux kernel, IDA is used to allocated pipeline widget instance ID,
+            # this ID later is used for pipeline creation, thus becomes pipeline ID in the
+            # firmware. Note that ppln_id variable will be assigned properly at pipeline widget
+            # creation, because it is always the first one to be created before all other
+            # widgets in the same pipeline and pipelines are created sequentially.
+            if widget_name.startswith('pipeline'):
+                # remove ending comma(,) with [:-1]
+                ppln_id = int(line_split[5][:-1])
```
marc-hb, line 60: To remove all leading and trailing commas whether there are some or not:
```suggestion
                ppln_id = int(line_split[5].strip(','))
```
This is also more readable, now you can drop the comment.

tools/sof-ipc-timer.py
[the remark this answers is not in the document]
marc-hb: I don't know whether it would actually be better here but the suggestion to use (more) regular expression(s) instead of hardcoded offsets makes sense to me.

tools/sof-ipc-timer.py
marc-hb: I think you could do at least something like this:
```python
    for line in f:
        if match_obj := re.search(r" ipc tx (     |reply|done )", line):
           (mod_id, and, other, global, state ) = big_line_parsing_function(mod_id, line, match_obj,...)
           # do something with (global, state) if needed
```
That should not require passing strings by reference or objects, should it?

tools/sof-ipc-timer.py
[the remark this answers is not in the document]
jsarha: Yes, but this is regular expression being used. The trouble is I need to make a distinction between token missing, just spaces in its place, "reply", or "done ". So I decided to target the correct place with regexp, and the take the exact spot where the spaces, "reply", or "done " should be. The static offset is skipping the known static part of the regexp match " ipc tx " = 8 chars, and its absolutely safe.

jsarha:
There has now been almost complete rewrite with added features. I took advice on the most of the suggestions where applicable, but on several places I restructured the code heavily to make it more maintainable.

I'll resolve all the past comments because its unlikely they are applicable any more.

marc-hb:
> There has now been almost complete rewrite with added features. 

Have you considered a fresh new PR and closing this one? We can still refer to this one when needed.

PR #1258, opened 2025-02-28
Test-case: check-audio-equalizers: Updates for IPC4 systems

singalsu:
This patch updates the test check-audio-equalizer.sh to work with IPC4 topologies with different naming for controls. The control name no more includes the component name, so the search for controls is changed. A tool topo_effect_kcontrols.py is used to find bytes controls for given component.

The blobs to test with sof-ctl are updated from SOF git repository. The ipc3 or ipc4 directory is chosen based on runtime check of DUT ipc version.

The sof-ctl usage is changed. The control is now referenced with the text control name instead of numid.


test-case/check-audio-equalizer.sh
```diff
@@ -49,8 +50,8 @@ func_test_eq()
     local id=$1
     local conf=$2
 
-    dlogc "sof-ctl -Dhw:$sofcard -n $id -s $conf"
-    sof-ctl -Dhw:"$sofcard" -n "$id" -s "$conf" || {
+    dlogc "sof-ctl -Dhw:$sofcard -c name='$id' -s $conf"
+    sof-ctl -Dhw:"$sofcard" -c name=\""$id"\"  -s "$conf" || {
```
marc-hb, line 54: This gives a quoted ID to `sof-ctl` itself. Does `sof-ctl` really needs to see double quotes and why? This deserves at least a comment IMHO.

If this is really what you want, you could also make it more explicit with an intermediate variable:
```shell
  local double_quoted_id=\""$id"\"

       ... -c name="$double_quoted_id"
```

test-case/check-audio-equalizer.sh
```diff
@@ -49,8 +50,8 @@ func_test_eq()
     local id=$1
     local conf=$2
 
-    dlogc "sof-ctl -Dhw:$sofcard -n $id -s $conf"
-    sof-ctl -Dhw:"$sofcard" -n "$id" -s "$conf" || {
+    dlogc "sof-ctl -Dhw:$sofcard -c name='$id' -s $conf"
```
marc-hb, line 53: Why the single quotes?

test-case/check-audio-equalizer.sh
```diff
@@ -64,19 +65,31 @@ func_test_eq()
...
+
+    if [[ ${Filterid^^} == *"IIR"* ]]; then
+	comp_dir="eq_iir"
+    elif [[ ${Filterid^^} == *"FIR"* ]]; then
+	comp_dir="eq_fir"
+    else
+	die "Not supported control: $Filterid"
+    fi
+
+    declare -a FilterList=($(ls -d "${my_dir}"/eqctl/$ipc_dir/$comp_dir/*.txt))
```
marc-hb, line 92: Don't parse ls: https://mywiki.wooledge.org/ParsingLs

You're iterating on it anyway, so iterate on it directly:

```shell
  nFilterList=$( ls -1 .../*.txt | wc -l) # counting without parsing
  ...

  for config in "${my_dir}"/eqctl/$ipc_dir/$comp_dir/*.txt; do

```

test-case/check-audio-equalizer.sh
```diff
@@ -125,10 +138,8 @@ do
 
     dlogi "eq_support= $eq_support"
     # if IIR/FIR filter is avilable, test with coef list
-    for filter_type in iir fir; do
-        if echo "$eq_support" | grep -q -i $filter_type; then
-            func_test_filter $filter_type
-        fi
+    for comp_id in $eq_support; do
```
marc-hb, line 141: eq_support should be a bash array `(  )`

lgirdwood (approved):

tools/topo_effect_kcontrols.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+    for vkc in bytes_kcontrols:
+        print(wname_prefix + vkc.hdr.name)
+
+
+# This could probably be moved to tplgtool2.py?
+def has_wname_prefix(widget):
+    """Is the kcontrol name prefixed with the widget name? ("PGAxx" or "Dmicxx")
+    Check SOF_TKN_COMP_NO_WNAME_IN_KCONTROL_NAME"""
```
marc-hb, line 58: That's a lot of duplication with `topo_vol_kcontrols.py`. Please you find some place for this in the already shared `tplgtool2.py` and move it there. That shouldn't be hard.

I will make this script 1/3 shorter and the other one 1/2 shorter!

EDIT: just realized I left this exact TODO in `topo_vol_kcontrols.py`!

test-case/check-audio-equalizer.sh
```diff
@@ -64,30 +66,41 @@ func_test_eq()
...
         die "can't find $testfilter"
     fi
 
-    declare -a FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))
-    nFilterList=${#FilterList[@]}
-    dlogi "$testfilter list, num= $nFilterList, coeff files= ${FilterList[*]}"
+    if is_ipc4; then
+	ipc_dir="ipc4"
+    else
+	ipc_dir="ipc3"
```
marc-hb, line 82: Strange indent. tab/spaces mix?

test-case/check-audio-equalizer.sh
```diff
@@ -64,30 +66,41 @@ func_test_eq()
...
+    fi
+
+    if [[ ${Filterid^^} == *"IIR"* ]]; then
+	comp_dir="eq_iir"
+    elif [[ ${Filterid^^} == *"FIR"* ]]; then
+	comp_dir="eq_fir"
+    else
+	die "Not supported control: $Filterid"
```
marc-hb, line 90: Same here

test-case/check-audio-equalizer.sh
```diff
@@ -64,30 +66,41 @@ func_test_eq()
...
-# param1 must be must be iir or fir
+# param1 must be must be component name
 func_test_filter()
 {
     local testfilter=$1
     dlogi "Get amixer control id for $testfilter"
-    # TODO: Need to match alsa control id with the filter in the pipeline,
-    #       currently the test discards EQ pipelines except first one.
-    Filterid=$(amixer -D hw:"$sofcard" controls | sed -n -e "/eq${testfilter}/I "'s/numid=\([0-9]*\),.*/\1/p' | head -1)
+    Filterid=`$my_dir/../tools/topo_effect_kcontrols.py $tplg $testfilter`
```
marc-hb, line 74: SC2006

There are some other shellcheck warnings, please take a look and fix all the easy ones.

marc-hb (commented):
This single, missing line is totally unrelated to your PR but it does get rid of most shellcheck warnings! Please add it.

```diff
--- a/test-case/check-audio-equalizer.sh
+++ b/test-case/check-audio-equalizer.sh
@@ -18,6 +18,7 @@ set -e
 
 # source from the relative path of current folder
 my_dir=$(dirname "${BASH_SOURCE[0]}")
+# shellcheck source=case-lib/lib.sh
 source "$my_dir"/../case-lib/lib.sh
 
 OPT_NAME['t']='tplg'     OPT_DESC['t']="tplg file, default value is env TPLG: $TPLG"
```

test-case/check-audio-equalizer.sh
```diff
@@ -64,30 +66,41 @@ func_test_eq()
...
     if [ -z "$Filterid" ]; then
         die "can't find $testfilter"
     fi
 
-    declare -a FilterList=($(ls -d "${my_dir}"/eqctl/eq_"${testfilter}"_*.txt))
```
marc-hb, line 79: Renaming all these files in a separate commit would have been nice from a git perspective...

test-case/check-audio-equalizer.sh
```diff
@@ -110,7 +123,7 @@ do
     rate=$(func_pipeline_parse_value "$idx" rate)
     fmt=$(func_pipeline_parse_value "$idx" fmt)
     type=$(func_pipeline_parse_value "$idx" type)
-    eq_support=$(func_pipeline_parse_value "$idx" eq)
+    eq_support=($(func_pipeline_parse_value "$idx" eq))
```
marc-hb, line 126: Maybe something like this? UNTESTED
```shell
 func_pipeline_parse_value "$idx" eq | read -d ' ' -a eq_support
```

Pipelines drop errors by default but in this case `func_pipeline_parse_value` would very likely have already failed earlier.

singalsu: It doesn't work but I don't know why. Just `func_pipeline_parse_value "$idx" eq` seems to return e.g.

`eqiir.21.1 eqfir.21.1` or `eqiir.41.0`

tools/topo_effect_kcontrols.py
singalsu: Done, and seems to work though not sure I did it right. I'm not familiar with python classes that most of the stuff in tplgtool2.py seems to be.

lgirdwood (approved):
Thanks @singalsu

test-case/check-audio-equalizer.sh
```diff
@@ -18,13 +18,14 @@ set -e
 
 # source from the relative path of current folder
 my_dir=$(dirname "${BASH_SOURCE[0]}")
+# shellcheck source=case-lib/lib.sh
```
marc-hb, line 21: This did get rid of a few shellcheck warnings (thx!) but fewer than I thought. Maybe this below would also help. Also off-topic sorry but also a tiny change/commit.

Warning: "latest" Ubuntu is not always the latest.

```diff
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -19,7 +19,7 @@ on: [pull_request, workflow_dispatch]
 jobs:
 
   shellcheck:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         with:
@@ -32,7 +32,7 @@ jobs:
                text/x-shellscript shellcheck -x
 
   pylint:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       # :-( https://github.community/t/support-for-yaml-anchors/16128
       - uses: actions/checkout@v4
```

test-case/check-audio-equalizer.sh
marc-hb: > There was no bisect to break since this was already broken,

That's a strong argument :-D

test-case/check-audio-equalizer.sh
marc-hb: Looks better!

It's still a pain but I found this to make it a bit less painful:
```
(defun tabs-toggle-marc()
  (interactive)
    (setq indent-tabs-mode  (not indent-tabs-mode)))
```

I also use `(require 'whitespace)` to show tabs.

tools/topo_effect_kcontrols.py
marc-hb: I don't know whether there is a better place or better way either, but at least not it's not duplicated and easy to make better if needed. Thanks!

test-case/check-audio-equalizer.sh
```diff
@@ -36,6 +37,7 @@ func_opt_parse_option "$@"
 tplg=${OPT_VAL['t']}
 duration=${OPT_VAL['d']}
 loop_cnt=${OPT_VAL['l']}
+ipc4="false"
```
marc-hb, line 40: I just noticed that this shellcheck warning points at a real bug!

test-case/check-audio-equalizer.sh
```diff
@@ -64,30 +67,41 @@ func_test_eq()
...
+
+    if [[ ${Filterid^^} == *"IIR"* ]]; then
+        comp_dir="eq_iir"
+    elif [[ ${Filterid^^} == *"FIR"* ]]; then
+        comp_dir="eq_fir"
+    else
+        die "Not supported control: $Filterid"
+    fi
+
+    nFilterList=$(ls -1 "${my_dir}"/eqctl/$ipc_dir/$comp_dir/*.txt | wc -l)
```
marc-hb, line 94: This is enough to please shellcheck:
```suggestion
    nFilterList=$(find "${my_dir}/eqctl/$ipc_dir/$comp_dir/" -name '*.txt' | wc -l)
```

singalsu: Yep

tools/topo_effect_kcontrols.py
```diff
@@ -0,0 +1,55 @@
...
+kcontrols of BYTES type, one per line.
+
+Pro tip: try using these commands _interactively_ with ipython3
+"""
+
+# Keep this script short and simple. If you want to get something else
+# from .tplg files, create another script.
+
+import sys
+from tplgtool2 import TplgBinaryFormat, TplgType, DapmType, SofVendorToken, has_wname_prefix
```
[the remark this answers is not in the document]
singalsu: Good catch!

test-case/check-audio-equalizer.sh
marc-hb: I thought it was a typo for another variable.

marc-hb (commented):
Code style LGTM. I didn't review the logic!

lgirdwood:
@golowanow good for you ?

golowanow:
> @golowanow good for you ?

I'm looking through it repeating ci checks.

test-case/check-audio-equalizer.sh
golowanow, line 126:
```suggestion
    IFS=" " read -r -a eq_support <<< "$(func_pipeline_parse_value "$idx" eq)" 
```
this way it makes shellcheck happy when I tried

singalsu: Excellent, that works!

tools/topo_effect_kcontrols.py
```diff
@@ -0,0 +1,55 @@
...
+    DAPMs = [
+        item for item in parsed_tplg if item.header.type == TplgType.DAPM_WIDGET.name
+    ]
+
+    for dapm in DAPMs:
+        effect_blocks = [b for b in dapm.blocks if b.widget.id == DapmType.EFFECT.name]
+
+        for gb in effect_blocks:
+            # debug
+            # print(f"{gb.widget.id}: {gb.widget.name}")
```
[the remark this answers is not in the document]
singalsu: Yep, removed now.

golowanow (approved):

PR #1259, opened 2025-03-10
ALSABAT: Added Alsabat mixer settings for PTL-SDW-RT712

harajend:


sofci:

lgirdwood (commented):
@harajend this is good for now, but long term we will have these settings in the UCM for each SKU.
Btw it looks like you have added an empty file ?

harajend:
> @harajend this is good for now, but long term we will have these settings in the UCM for each SKU. Btw it looks like you have added an empty file ?

@lgirdwood, I only added executable permission to the 'PTLP_RVP_SDW.sh' file, so it is showing as 'Empty file'. Earlier, this file did not have executable permission, so the SOF-Test framework threw some warnings during the test run

lgirdwood (approved):
@golowanow pls review/merge

alsa_settings/PTLP_RVP_SDW.sh
golowanow, line 1: this is already fixed by #1257

PR #1260, opened 2025-03-10
TinyAlsa: Add support for testing using tinycap and tinyplay.

gbarkadiusz:
Added option `SOF_ALSA_TOOL` to specify whether to use the `TinyASLA` or `ALSA` testing tool. For now, it is only for capturing and playback.
The default tool is ALSA.

This PR introduces also a generic wrapper  `tinyalsa-wrapper.sh` to define `SOF_ALSA_TOOL` in environments where setting environment variables directly is not possible. This is useful for frameworks or systems with limited configuration capabilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ expect alsa-utils python3 python3-construct python3-graphviz
 ```
 sudo apt install expect alsa-utils python3 python3-construct python3-graphviz
 ```
+If you would like to use tinyALSA for testing, install tinyALSA and SoX.
+- How to install tinyALSA: https://github.com/tinyalsa/tinyalsa
+- To install SoX run below command:
+```
+sudo apt-install sox
+```
 #### user group
 sudo adm audio
 
@@ -52,12 +58,14 @@ Usage: ./check-playback.sh [OPTION]
 2020-03-19 22:13:32 UTC [COMMAND] aplay -q --fatal-errors  -Dhw:0,0 -r 48000 -c 2 -f S16_LE -d 4 /dev/zero -v -q
     ...
 ```
+Some tests support these environment variables (work in progress):
+  - SOF_ALSA_TOOL is used to select the audio tool for testing.
+  Set this variable to 'alsa' (default value) or 'tinyalsa' to choose between the ALSA and TinyALSA toolsets.
+  - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
+  - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
+These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
 
-Some tests support SOF_ALSA_OPTS, SOF_APLAY_OPTS and SOF_ARECORD_OPTS,
-work in progress. Where supported, optional parameters in SOF_APLAY_OPTS
-and SOF_ARECORD_OPTS are passed to all aplay and arecord
-invocations. SOF_ALSA_OPTS parameters are passed to both aplay and
-arecord. Warning these environments variables do NOT support parameters
+Warning, these environment variables do NOT support parameters
 with whitespace or globbing characters, in other words this does NOT
 work:
 
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -752,17 +752,88 @@ func_lib_check_pa()
 # However, 1. arrays would complicate the user interface 2. ALSA does not
 # seem to need arguments with whitespace or globbing characters.
 
+# SOF_ALSA_TOOL:
+# This option is used for selecting tool for testing,
+# So far, supported tools are 'alsa' and 'tinyalsa'
+# To select appropriate tool, set SOF_ALSA_TOOL to one of above
+# before using 'aplay_opts' or 'arecord_opts' function.
+# Default is SOF_ALSA_TOOL='alsa'
+
+
+# Function to extract the card number and device number from $dev option (e.g., hw:0,10)
+parse_audio_device() {
+    # Extract the card number (e.g., "0" from hw:0,10)
+    card_nr=$(printf '%s' "$1" | cut -d ':' -f2 | cut -d ',' -f1)
+
+    # Extract the device number (e.g., "10" from hw:0,10)
+    dev_nr=$(printf '%s' "$1" | cut -d ',' -f2)
+}
+
+# Function to extract the numeric format value from the PCM sample formats
+# There is passes PCM sample format while using ALSA tool (arecord)
+# While using TinyALSA (tinycap -b) we need to convert PCM sample fomrat to bits.
+extract_format_number() {
+    # (e.g., extracting '16' from 'S16_LE')
+    printf '%s' "$1" | grep '[0-9]\+' -o
+}
+
+# Initialize the parameters using for audio testing.
+# shellcheck disable=SC2034
+initialize_audio_params()
+{
+    local idx="$1"
+
+    channel=$(func_pipeline_parse_value "$idx" channel)
+    rate=$(func_pipeline_parse_value "$idx" rate)
+    fmts=$(func_pipeline_parse_value "$idx" fmt)
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    pcm=$(func_pipeline_parse_value "$idx" pcm)
+    type=$(func_pipeline_parse_value "$idx" type)
+    snd=$(func_pipeline_parse_value "$idx" snd)
+
+    : "${SOF_ALSA_TOOL:="alsa"}"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+       parse_audio_device "$dev"
+    fi
+}
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # shellcheck disable=SC2154
+        if ! sox -n -r "$rate" -c "$channel" noise.wav synth "$duration" white; then
+            printf 'Error: sox command failed.\n' >&2
+	    return 1
+	fi
+        dlogc "tinyplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS -D $card_nr -d $dev_nr  -i wav noise.wav"
+	# shellcheck disable=SC2086
+        tinyplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS -D "$card_nr" -d "$dev_nr"  -i wav noise.wav
+    elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
+        # shellcheck disable=SC2086
+        aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    else
+        die "Unknown ALSA tool: $SOF_ALSA_TOOL"
+    fi
 }
+
 arecord_opts()
 {
-    dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
-    # shellcheck disable=SC2086
-    arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # shellcheck disable=SC2154
+        # Global variable "$fmt_elem" from check_capture.sh test script
+	format=$(extract_format_number "$fmt_elem")
+	dlogc "tinycap $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $file -D $card_nr -d $dev_nr -c $channel -t $duration -r $rate -b $format"
+	# shellcheck disable=SC2086
+        tinycap $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$file" -D "$card_nr" -d "$dev_nr" -c "$channel" -t "$duration" -r "$rate" -b "$format"
+    elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+        # shellcheck disable=SC2086
+        arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+    else
+        die "Unknown ALSA tool: $SOF_ALSA_TOOL"
+    fi
 }
 
 die()
@@ -952,7 +1023,7 @@ is_ipc4()
 logger_disabled()
 {
     # Disable logging when available...
-    if [ ${OPT_VAL['s']} -eq 0 ]; then
+    if [ "${OPT_VAL['s']}" -eq 0 ]; then
         return 0
     fi
 
@@ -1063,15 +1134,31 @@ set_alsa_settings()
 reset_sof_volume()
 {
     # set all PGA* volume to 0dB
-    amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
-    while read -r mixer_name
-    do
-        if is_ipc4; then
-            amixer -Dhw:0 -- sset "$mixer_name" 100%
-        else
-            amixer -Dhw:0 -- sset "$mixer_name" 0dB
-        fi
-    done
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
+
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                amixer -Dhw:0 -- sset "$mixer_name" 100%
+            else
+               amixer -Dhw:0 -- sset "$mixer_name" 0dB
+            fi
+        done
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        tinymix -D0 controls | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
+
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                tinymix -D0 set "$mixer_name" 100%
+            else
+                tinymix -D0 set "$mixer_name" 0dB
+            fi
+        done
+    else
+         echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
 }
 
 DO_PERF_ANALYSIS=0
--- a/test-case/check-capture.sh
+++ b/test-case/check-capture.sh
@@ -64,25 +64,20 @@ setup_kernel_check_point
 func_lib_check_sudo
 func_pipeline_export "$tplg" "type:capture & ${OPT_VAL['S']}"
 
-for round in $(seq 1 $round_cnt)
+for round in $(seq 1 "$round_cnt")
 do
     for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
     do
-        channel=$(func_pipeline_parse_value "$idx" channel)
-        rate=$(func_pipeline_parse_value "$idx" rate)
-        fmt=$(func_pipeline_parse_value "$idx" fmt)
-        dev=$(func_pipeline_parse_value "$idx" dev)
-        pcm=$(func_pipeline_parse_value "$idx" pcm)
-        type=$(func_pipeline_parse_value "$idx" type)
-        snd=$(func_pipeline_parse_value "$idx" snd)
 
-        if [ ${OPT_VAL['F']} = '1' ]; then
-            fmt=$(func_pipeline_parse_value "$idx" fmts)
+        initialize_audio_params "$idx"
+
+        if [ "${OPT_VAL['F']}" = '1' ]; then
+            fmts=$(func_pipeline_parse_value "$idx" fmts)
         fi
 
-        for fmt_elem in $fmt
+        for fmt_elem in $fmts
         do
-            for i in $(seq 1 $loop_cnt)
+            for i in $(seq 1 "$loop_cnt")
             do
                 dlogi "===== Testing: (Round: $round/$round_cnt) (PCM: $pcm [$dev]<$type>) (Loop: $i/$loop_cnt) ====="
                 # get the output file
@@ -90,12 +85,12 @@ do
                     dlogi "no file prefix, use /dev/null as dummy capture output"
                     file=/dev/null
                 else
-                    mkdir -p $out_dir
+                    mkdir -p "$out_dir"
                     file=$out_dir/${file_prefix}_${dev}_${i}.wav
                     dlogi "using $file as capture output"
                 fi
 
-                if ! arecord_opts -D"$dev" -r "$rate" -c "$channel" -f "$fmt_elem" -d $duration "$file" -v -q;
+                if ! arecord_opts -D"$dev" -r "$rate" -c "$channel" -f "$fmt_elem" -d "$duration" "$file" -v -q;
                 then
                     func_lib_lsof_error_dump "$snd"
                     die "arecord on PCM $dev failed at $i/$loop_cnt."
--- a/test-case/check-playback.sh
+++ b/test-case/check-playback.sh
@@ -71,25 +71,20 @@ setup_kernel_check_point
 func_lib_check_sudo
 func_pipeline_export "$tplg" "type:playback & ${OPT_VAL['S']}"
 
-for round in $(seq 1 $round_cnt)
+for round in $(seq 1 "$round_cnt")
 do
     for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
     do
-        channel=$(func_pipeline_parse_value "$idx" channel)
-        rate=$(func_pipeline_parse_value "$idx" rate)
-        fmts=$(func_pipeline_parse_value "$idx" fmt)
-        dev=$(func_pipeline_parse_value "$idx" dev)
-        pcm=$(func_pipeline_parse_value "$idx" pcm)
-        type=$(func_pipeline_parse_value "$idx" type)
-        snd=$(func_pipeline_parse_value "$idx" snd)
 
-        if [ ${OPT_VAL['F']} = '1' ]; then
+        initialize_audio_params "$idx"
+
+        if [ "${OPT_VAL['F']}" = '1' ]; then
             fmts=$(func_pipeline_parse_value "$idx" fmts)
         fi
 
         for fmt_elem in $fmts
         do
-            for i in $(seq 1 $loop_cnt)
+            for i in $(seq 1 "$loop_cnt")
             do
                 dlogi "===== Testing: (Round: $round/$round_cnt) (PCM: $pcm [$dev]<$type>) (Loop: $i/$loop_cnt) ====="
                 aplay_opts -D"$dev" -r "$rate" -c "$channel" -f "$fmt_elem" \
--- a/env-check.sh
+++ b/env-check.sh
@@ -83,6 +83,9 @@ out_str="" check_res=0
 printf "Checking for some OS packages:\t\t"
 func_check_pkg expect
 func_check_pkg aplay
+func_check_pkg sox
+func_check_pkg tinycap
+func_check_pkg tinyplay
 func_check_pkg python3
 # jq is command-line json parser
 func_check_pkg jq
--- a/test-case/tinyalsa-wrapper.sh
+++ b/test-case/tinyalsa-wrapper.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+##
+## Case Name: Wrapper to run a test case given with TinyALSA
+## Preconditions:
+##    TinyALSA and SoX are installed.
+## Description:
+##    This script serves as a wrapper to execute a test case script using TinyALSA.
+##    It expects the test case script file name (without path) as the first parameter,
+##    followed by other parameters required for that test case.
+## Case step:
+##    1. SOF_ALSA_TOOL environment variable is set to TinyALSA
+##    2. The test case script is executed.
+## Expect result:
+##    The test case script is executed using TinyALSA
+##
+
+set -e
+
+# Ensure the test case script file name is provided
+if [ -z "$1" ]; then
+    echo "Error: No test case script file name provided. Exiting..."
+    exit 1
+fi
+
+export SOF_ALSA_TOOL=tinyalsa
+
+TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+
+# shellcheck disable=SC2145
+[ -x "$TESTDIR/test-case/$(basename "$1")" ] && exec "$TESTDIR"/test-case/"$@"


test-case/check-playback.sh
```diff
@@ -91,12 +100,20 @@ do
...
+                    aplay_opts -D"$dev" -r "$rate" -c "$channel" -f "$fmt_elem" \
                       -d "$duration" "$file" -v -q || {
-                    func_lib_lsof_error_dump "$snd"
-                    die "aplay on PCM $dev failed at $i/$loop_cnt."
-                }
+                       func_lib_lsof_error_dump "$snd"
+                       die "aplay on PCM $dev failed at $i/$loop_cnt."
+                    }
+                else
+                    echo "Unknown tool"
```
golowanow, line 114: do this check for correct $tool earlier outside of the loop (e.g. L74)

marc-hb (commented):
This is not scalable at all. It's adding support in only 2 tests and there's already a lot of copy/paste/diverge.

The non-duplicated way is to make `aplay_opts` into a smarter function that implements all the logic in only one place.

case-lib/lib.sh
```diff
@@ -754,15 +754,43 @@ func_lib_check_pa()
...
+
+    # Check if the first argument is 'tinyalsa'
+    if [[ "$1" == "tinyalsa" ]]; then
+        use_tinyalsa=true
+    fi
+    shift
+    if $use_tinyalsa; then
+        card_nr=$(echo "$dev" | cut -d ':' -f2 | cut -d ',' -f1)
+        dev_nr=$(echo "$dev" | cut -d ',' -f2)
+        sox -n -r "$rate" -c "$channel" noise.wav synth "$duration" white
```
redzynix, line 767: IMHO I would declare channel, rate, duration, and other alsa related variables in lib.sh to avoid reference before the assignment.
https://github.com/thesofproject/sof-test/blob/main/test-case/check-playback.sh#L78C9-L84C52

marc-hb:
Avoid "spamming" the PR with force-pushes as this brings down limited hardware resources.

test-case/check-capture.sh
```diff
@@ -56,10 +59,16 @@ duration=${OPT_VAL['d']}
 loop_cnt=${OPT_VAL['l']}
 out_dir=${OPT_VAL['o']}
 file_prefix=${OPT_VAL['f']}
+tool=${OPT_VAL['T']}
 
 start_test
 logger_disabled || func_lib_start_log_collect
 
+if [[ "$tool" != "tinyalsa" ]] && [[ "$tool" != "alsa" ]]; then
```
golowanow, line 67: please consider just to add an optional environment variable (e.g. `SOF_ALSA_TOOL`) and make all the choices, checks, etc. at lib.sh only, so changing nothing in test cases.

golowanow:
> Avoid "spamming" the PR with force-pushes as this brings down limited hardware resources.

also, shellcheck can be run locally, before push to PR.

marc-hb:
Don't forget the `-x` option, as in: `shellcheck -x somefile.sh`, otherwise you get a lot of false positives. I wonder why `-x` is not ON by default.

If you forget which option it is, just look in either the CI logs or in `git grep shellcheck -- .github/` and run the same. True not just with shellcheck. That's one big advantage of GitHub actions over Jenkins and similar: no "CI secret".

(there can also be drawbacks having everything in the same git repo, I digress)

test-case/check-capture.sh
marc-hb: Agreed, this looks like it should be a global setting of some kind. I don't see why a user would need to keep jumping back and forth between tinyalsa and regular depending on which test they run. And if they really do, they could still prefix and override each run with the global variable:

```shell
./test-case/check-something.sh
SOF_TINY_ALSA=true ./test-case/check-something-else.sh
./test-case/check-another-thing.sh

```

case-lib/lib.sh
gbarkadiusz: Please take a look.

gbarkadiusz:
@marc-hb @golowanow I've changed the approach according to your suggestions. Now, only the `lib.sh` has been changed. To force the usage of `tinyalsa` as the testing tool, the testcase should be run with `SOF_TINY_ALSA=true` before the testcase.
e.g. `SOF_TINY_ALSA=true ./test-case/check-playback.sh`
Moreover, I've added the declaration of ALSA-related variables to `lib.sh` to avoid reference before assignment, as @redzynix mentioned.

case-lib/lib.sh
```diff
@@ -752,17 +752,37 @@ func_lib_check_pa()
 # However, 1. arrays would complicate the user interface 2. ALSA does not
 # seem to need arguments with whitespace or globbing characters.
 
+channel=$(func_pipeline_parse_value "$idx" channel)
+rate=$(func_pipeline_parse_value "$idx" rate)
+fmt_elem=$(func_pipeline_parse_value "$idx" fmt_elem)
+dev=$(func_pipeline_parse_value "$idx" dev)
+duration=$(func_pipeline_parse_value "$idx" duration)
```
marc-hb, line 759: There are 2 problems here that can be fortunately and easily solved with a single solution: moving these to a new "init" function explicitly invoked by each test that needs these. Whatever the question, a function is always the answer ;-)

1. These are run the moment `lib.sh` is `source`-d.
   1a. Individual tests have no control over when this is run
   1b. It is normally possible to `source case-lib/lib.sh` _from the command-line_ for interactive testing. To keep this functional, sourcing should run as little code as possible and have as few side effects as possible.

2. This code duplicates and clashes with what a lot of tests _already do_ with potential conflicts very difficult to debug. Just try `git grep func_pipeline_parse_value` and see.
 
Also, this addition is currently splitting `aplay_opts()` away from its comment/ "documentation".

By placing these in a new init function, you solve 1. and you can cleanup 2. for each test one by one. No need to do all of them at once, you can start with the few tests using `aplay_opts()` already.

case-lib/lib.sh
```diff
@@ -752,17 +752,37 @@ func_lib_check_pa()
...
+    fi
 }
 arecord_opts()
 {
-    dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
-    # shellcheck disable=SC2086
-    arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+    if [[ $SOF_TINY_ALSA == true ]]; then
+        card_nr=$(echo "$dev" | cut -d ':' -f2 | cut -d ',' -f1)
+        dev_nr=$(echo "$dev" | cut -d ',' -f2)
```
marc-hb, line 778: This is a copy of the code in `aplay_opts()`, move this to some new `set_tiny_alsa_params()` function.

case-lib/lib.sh
```diff
@@ -752,17 +752,37 @@ func_lib_check_pa()
...
+duration=$(func_pipeline_parse_value "$idx" duration)
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ $SOF_TINY_ALSA == true ]]; then
+        card_nr=$(echo "$dev" | cut -d ':' -f2 | cut -d ',' -f1)
+        dev_nr=$(echo "$dev" | cut -d ',' -f2)
```
marc-hb, line 765: When parsing like this, it really helps the reader to add an example in a one-line comment:
```shell
# example: dev=hw:0,3
```

Also, prefer `printf %s` to `echo`:

```suggestion
        dev_nr=$(printf '%s' "$dev" | cut -d ',' -f2)
```
`echo` has a lot of pitfalls. It should not matter in this case but it's good practice. Plus everyone knows `printf` from other languages.

https://www.shellcheck.net/wiki/SC2028
etc.

case-lib/lib.sh
marc-hb: > These are run the moment lib.sh is source-d.

A similar problem was fixed in 7c2ede03c9be1b99, take a look at #1187 for context.

case-lib/lib.sh
```diff
@@ -752,17 +752,37 @@ func_lib_check_pa()
...
+fmt_elem=$(func_pipeline_parse_value "$idx" fmt_elem)
+dev=$(func_pipeline_parse_value "$idx" dev)
+duration=$(func_pipeline_parse_value "$idx" duration)
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ $SOF_TINY_ALSA == true ]]; then
```
golowanow, line 763: just few additional suggestions from my side:
1.
```suggestion
    if [[ "$SOF_TINY_ALSA" = true ]]; then
```

2. `SOF_TINY_ALSA` - I'm not sure it is "scalable" enough as a bool variable in comparison to `SOF_ALSA_TOOL=tinyalsa` or whatever other/proprietary tool, a custom script tool etc. it might be in the future. 

3. anyway, document it: 
* [in comments](https://github.com/thesofproject/sof-test/blob/main/case-lib/lib.sh#L727-L753)
* [in README.md](https://github.com/thesofproject/sof-test/blob/main/README.md?plain=1#L56)

4. add `dlogc` similar to what the alsa branch does

gbarkadiusz: Regarding point number 2, I think that it will always force the use of the SOF_ALSA_TOOL parameter, e.g., SOF_ALSA_TOOL=alsa.

marc-hb: > SOF_ALSA_TOOL=alsa, which could break backward compatibility.

This can be a default value. Default values are often implemented like this:

```shell
: ${SOF_ALSA_TOOL:=alsa}
```

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

Word expansion -> parameter expansion -> default value.

case-lib/lib.sh
```diff
@@ -752,17 +752,84 @@ func_lib_check_pa()
...
+# So far, supported tools are 'alsa' and 'tinyalsa'
+# To select appropriate tool, set SOF_ALSA_TOOL to one of above
+# before using 'aplay_opts' or 'arecord_opts' function.
+# (e.g., SOF_ALSA_TOOL=alsa)
+
+
+# Function to extract the card number and device number from $dev option (e.g., hw:0,10)
+parse_audio_device() {
+    # Extract the card number (e.g., "0" from hw:0,10)
+    card_nr=$(printf '%s' "$dev" | cut -d ':' -f2 | cut -d ',' -f1)
```
marc-hb, line 766: Avoiding global variables is sometimes difficult in shell but not in this case. In this case there's no reason to use a global variable.

```suggestion
    card_nr=$(printf '%s' "$1" | cut -d ':' -f2 | cut -d ',' -f1)
```

case-lib/lib.sh
```diff
@@ -752,17 +752,84 @@ func_lib_check_pa()
...
+
+    # Extract the device number (e.g., "10" from hw:0,10)
+    dev_nr=$(printf '%s' "$dev" | cut -d ',' -f2)
+}
+
+# Function to extract the numeric format value from the $fmt_elem option
+extract_format_number() {
+    # (e.g., extracting '16' from 'S16_LE')
+    # shellcheck disable=SC2154
+    format=$(printf '%s' "$fmt_elem" | grep '[0-9]\+' -o)
```
marc-hb, line 776: Same as above, and now you can drop the `shellcheck disable=SC2154`.
```suggestion
    format=$(printf '%s' "$1" | grep '[0-9]\+' -o)
```

case-lib/lib.sh
```diff
@@ -752,17 +752,84 @@ func_lib_check_pa()
...
+}
+
+# Initialize the parameters using for audio testing.
+initialize_audio_params()
+{
```
marc-hb, line 781: Same as above
```suggestion
{
   local idx="$1"
   
```
This will temporarily shadow any global `idx` as expected.

test-case/check-capture.sh
```diff
@@ -68,19 +68,14 @@ for round in $(seq 1 $round_cnt)
...
     do
-        channel=$(func_pipeline_parse_value "$idx" channel)
-        rate=$(func_pipeline_parse_value "$idx" rate)
-        fmt=$(func_pipeline_parse_value "$idx" fmt)
-        dev=$(func_pipeline_parse_value "$idx" dev)
-        pcm=$(func_pipeline_parse_value "$idx" pcm)
-        type=$(func_pipeline_parse_value "$idx" type)
-        snd=$(func_pipeline_parse_value "$idx" snd)
+
+	initialize_audio_params
```
marc-hb, line 72:
```suggestion
	initialize_audio_params "$idx"
```

test-case/check-playback.sh
```diff
@@ -75,13 +75,8 @@ for round in $(seq 1 $round_cnt)
...
     do
-        channel=$(func_pipeline_parse_value "$idx" channel)
-        rate=$(func_pipeline_parse_value "$idx" rate)
-        fmts=$(func_pipeline_parse_value "$idx" fmt)
-        dev=$(func_pipeline_parse_value "$idx" dev)
-        pcm=$(func_pipeline_parse_value "$idx" pcm)
-        type=$(func_pipeline_parse_value "$idx" type)
-        snd=$(func_pipeline_parse_value "$idx" snd)
+
+        initialize_audio_params
```
marc-hb, line 79:
```suggestion
        initialize_audio_params "$idx"
```

case-lib/lib.sh
```diff
@@ -752,17 +752,84 @@ func_lib_check_pa()
...
+   rate=$(func_pipeline_parse_value "$idx" rate)
+   # shellcheck disable=SC2034
+   fmts=$(func_pipeline_parse_value "$idx" fmt)
+   # shellcheck disable=SC2034
+   dev=$(func_pipeline_parse_value "$idx" dev)
+   # shellcheck disable=SC2034
+   pcm=$(func_pipeline_parse_value "$idx" pcm)
+   # shellcheck disable=SC2034
+   type=$(func_pipeline_parse_value "$idx" type)
+   # shellcheck disable=SC2034
```
marc-hb, line 794: I don't get this warning, are you sure?

If you are sure:
- add it only once right before the function and it will apply to whole function. No need to duplicate.
- Explain very briefly why it's disabled briefly before it. No one knows what SC2034 is by heart.

case-lib/lib.sh
```diff
@@ -752,17 +752,84 @@ func_lib_check_pa()
...
+}
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        dlogc "tinyplay $*"
+	# shellcheck disable=SC2154
```
marc-hb, line 807: I don't see this warning. Did you use the `-x` option? `shellcheck -x ...`

marc-hb (commented):
Just a few minor issues left, otherwise looks good to me. Thanks for your persistence!

case-lib/lib.sh
marc-hb: I think I know what's happening! Cause it just happened to me in a totally unrelated context. You must run `shellcheck -x case-lib/lib.sh` IN THE TOP-LEVEL sof-test directory! That will hopefully silence a few warnings.

case-lib/lib.sh
gbarkadiusz: SC2034 (warning): pcm appears unused. Verify use (or export if used externally).

test-case/check-capture-tinyalsa.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
golowanow, line 2: 1) please add some comments here to explain the purpose of this file as a wrapper to execute a test case script using  tinyalsa. As its parameters it expects a test case script file name (without path) in the first parameter, followed by other parameters for that test.

2) also, add here some simple preconditions to check the tinyalsa executables are present on the DUT.

test-case/check-playback-tinyalsa.sh
golowanow, line 1: not needed with `check-tinyalsa.sh` wrapper

gbarkadiusz:
@golowanow I've added the changes for the volume-basic test. This test doesn't use lib.sh for executing the amixer command, so I've added if statements to make this test ready for testing mixers under TinyALSA. Please take a look and share your feedback.

test-case/volume-basic-test.sh
```diff
@@ -36,13 +36,19 @@ func_opt_parse_option "$@"
...
 start_test
 
 func_error_exit()
 {
     dloge "$*"
-    pkill -9 aplay
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        pkill -9 aplay
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+	pkill -9 tinyplay
```
marc-hb, line 48: indent

test-case/volume-basic-test.sh
```diff
@@ -36,13 +36,19 @@ func_opt_parse_option "$@"
...
+    else
+	echo "Unknowns asla tool $SOF_ALSA_TOOL"
```
marc-hb, line 50: indent + typo

This fallback does not seem necessary. This is a clean up function, if SOF_ALSA_TOOL is wrong then there must have been another error message already printed much earlier than this.

test-case/volume-basic-test.sh
```diff
@@ -56,20 +62,45 @@ rate=$(func_pipeline_parse_value 0 rate)
 fmt=$(func_pipeline_parse_value 0 fmt)
 dev=$(func_pipeline_parse_value 0 dev)
 
-dlogc "aplay -D $dev -c $channel -r $rate -f $fmt /dev/zero &"
-# play into background, this will wake up DSP and IPC. Need to clean after the test
-aplay -D "$dev" -c "$channel" -r "$rate" -f "$fmt" /dev/zero &
+if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
```
marc-hb, line 65: I don't understand: you did a great work making `aplay_opts()` generic, why not use it here?

test-case/volume-basic-test.sh
```diff
@@ -83,10 +114,17 @@ do
         dlogi "$volctrl"
 
         for vol in "${volume_array[@]}"; do
-            dlogc "amixer -c$sofcard cset name='$volctrl' $vol"
-            amixer -c"$sofcard" cset name="$volctrl" "$vol" > /dev/null ||
+            if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
```
marc-hb, line 117: Same question: why not use the new, upgraded and generic reset_volume() here?

test-case/volume-basic-test.sh
```diff
@@ -96,8 +134,15 @@ do
                       func_error_exit "dmesg has errors!"
 done
 
-#clean up background aplay
-pkill -9 aplay || true
+if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
```
marc-hb, line 137: Extract the killing part of `func_error_exit()` above into a new function and invoke it here.

marc-hb (commented):
> I've added the changes for the volume-basic test. This test doesn't use lib.sh for executing the amixer command, so I've added if statements ...

Why? You've done a great work upgrading aplay_opts(), making it generic, but at the last minute just before this PR was about to be all approved and merged, you add the same feature in volume-basic-test.sh while NOT using your brand new and shiny aplay_opts()? Strange?

Maybe there is a reason not to use aplay_opts() in volume-basic-test.sh. Maybe there is none. Either way, don't change volume-basic-test.sh in this PR and get this PR over with. Discuss volume-basic-test.sh in a new, separate PR.

test-case/check-tinyalsa.sh
```diff
@@ -0,0 +1,34 @@
...
+    else
+        echo "$1 is not installed. Exiting..."
+        exit 1
+    fi
+}
+
+# Preconditions: Check if tinyalsa executables are present on the DUT
+check_command "sox"
+check_command "tinycap"
+check_command "tinyplay"
```
marc-hb, line 21: I'm not convinced this wrapper is worth the added lines of code. If one of these commands are missing, `aplay_opts()` or equivalent should already fail with a decent "not found" error message. Is it not already the case? It should be.

Either way this is not strictly required, independent and can be moved to a future PR. This one has already lasted for a very long time :-)

marc-hb:
I suspect some are older than this PR but most of the few warnings left in https://github.com/thesofproject/sof-test/actions/runs/14243977340/job/39920223082?pr=1260 seem very easy to fix.

README.md
```diff
@@ -53,11 +53,13 @@ Usage: ./check-playback.sh [OPTION]
     ...
 ```
 
-Some tests support SOF_ALSA_OPTS, SOF_APLAY_OPTS and SOF_ARECORD_OPTS,
+Some tests support SOF_ALSA_TOOL, SOF_ALSA_OPTS, SOF_APLAY_OPTS and SOF_ARECORD_OPTS,
 work in progress. Where supported, optional parameters in SOF_APLAY_OPTS
 and SOF_ARECORD_OPTS are passed to all aplay and arecord
 invocations. SOF_ALSA_OPTS parameters are passed to both aplay and
-arecord. Warning these environments variables do NOT support parameters
+arecord. SOF_ALSA_TOOL is used to select tool for testing.
```
golowanow, line 60: TinyAlsa should also use SOF_(ALSA|APLAY|ARECORD)_OPTS assuming that they follow SOF_ALSA_TOOL value.
Please document this assumption here as well.

case-lib/lib.sh
```diff
@@ -752,17 +752,80 @@ func_lib_check_pa()
...
+# Function to extract the card number and device number from $dev option (e.g., hw:0,10)
+parse_audio_device() {
+    # Extract the card number (e.g., "0" from hw:0,10)
+    card_nr=$(printf '%s' "$1" | cut -d ':' -f2 | cut -d ',' -f1)
+
+    # Extract the device number (e.g., "10" from hw:0,10)
+    dev_nr=$(printf '%s' "$1" | cut -d ',' -f2)
+}
+
+# Function to extract the numeric format value from the $fmt_elem option
```
golowanow, line 772: not clear what is `$fmt_elem` here ?

case-lib/lib.sh
```diff
@@ -752,17 +752,80 @@ func_lib_check_pa()
 # However, 1. arrays would complicate the user interface 2. ALSA does not
 # seem to need arguments with whitespace or globbing characters.
 
+# SOF_ALSA_TOOL:
+# This option is used for selecting tool for testing,
+# So far, supported tools are 'alsa' and 'tinyalsa'
+# To select appropriate tool, set SOF_ALSA_TOOL to one of above
+# before using 'aplay_opts' or 'arecord_opts' function.
+# (e.g., SOF_ALSA_TOOL=alsa)
```
golowanow, line 760:
```suggestion
Default is `SOF_ALSA_TOOL=`alsa`
```

test-case/volume-basic-test.sh
golowanow: I have the same question

test-case/check-tinyalsa.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# This script serves as a wrapper to execute a test case script using tinyalsa.
```
golowanow, line 2: please follow the same comment style here as in other test case scripts.

golowanow (commented):
some general asks from my side:
1) better to introduce TinyAlsa in one (this) PR to make all these changes consistent.
2) split changes into logically ordered commits.

gbarkadiusz:
@marc-hb restored to the previous version. I'll add `tinymix` in the separate pr.

gbarkadiusz:
> I suspect some are older than this PR but most of the few warnings left in https://github.com/thesofproject/sof-test/actions/runs/14243977340/job/39920223082?pr=1260 seem very easy to fix.

These warnings don't appear locally.
`shellcheck -x case-lib/lib.sh`

marc-hb:
> These warnings don't appear locally.
> shellcheck -x case-lib/lib.sh

Most of them are in check-playback.sh and check-capture.sh.

There's one in case-lib/lib.sh though and it's most basic type of warning so it's strange you don't see it.

Anyway they are all just missing quotes, super easy to fix. Just add double quotes! shellcheck even prints the correct code.

https://mywiki.wooledge.org/Quotes?highlight=%28quote%29#I.27m_Too_Lazy_to_Read.2C_Just_Tell_Me_What_to_Do

marc-hb:
> There's one in case-lib/lib.sh though and it's most basic type of warning so it's strange you don't see it.

I may have solved this mystery. I don't see these missing quotes with shellcheck 0.8 (Ubuntu 22.04) but I do see them with shellcheck 0.9 (Ubuntu 24.04). So, shellcheck bug fixes. The Ubuntu runner was recently updated (94bbe9d50b41)

What shellcheck -V ersion do you have?

gbarkadiusz:
@marc-hb Could you take a look, please.

README.md
gbarkadiusz: Done.

case-lib/lib.sh
```diff
@@ -1063,15 +1129,31 @@ set_alsa_settings()
...
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                tinymix -D0 set "$mixer_name" 100%
+            else
+                tinymix -D0 set "$mixer_name" 0dB
+            fi
+        done
+    else
+         echo "Unknowns alsa tool $SOF_ALSA_TOOL"
```
marc-hb, line 1155: Typo. Does not matter.

marc-hb (approved):
As I wrote before, I'm not convinced the wrapper adds much value but AFAICT it's optional. Hope it stays like this :-)

I have NOT looked at the test results (which are all red...) Obviously, this should not be merged if it introduces any regression.

Everything else looks good.

case-lib/lib.sh
gbarkadiusz: Has changed.

golowanow:
I'm looking through it again with recent probe run results.

README.md
```diff
@@ -53,11 +53,15 @@ Usage: ./check-playback.sh [OPTION]
     ...
 ```
 
-Some tests support SOF_ALSA_OPTS, SOF_APLAY_OPTS and SOF_ARECORD_OPTS,
-work in progress. Where supported, optional parameters in SOF_APLAY_OPTS
-and SOF_ARECORD_OPTS are passed to all aplay and arecord
-invocations. SOF_ALSA_OPTS parameters are passed to both aplay and
-arecord. Warning these environments variables do NOT support parameters
+Some tests support the use of environment variables:
+SOF_ALSA_TOOL, SOF_ALSA_OPTS, SOF_APLAY_OPTS and SOF_ARECORD_OPTS (work in progress)
```
golowanow, line 57:
```suggestion
Some tests support these environment variables (work in progress):
```

README.md
```diff
@@ -53,11 +53,15 @@ Usage: ./check-playback.sh [OPTION]
...
+  - SOF_ALSA_TOOL is used to select the audio tool for testing.
+  Set this variable to 'alsa' or 'tinyalsa' to choose between the ALSA and TinyALSA toolsets.
```
golowanow, line 59:
```suggestion
  Set this variable to 'alsa' (default value) or 'tinyalsa' to choose between the ALSA and TinyALSA toolsets.
```

README.md
```diff
@@ -53,11 +53,15 @@ Usage: ./check-playback.sh [OPTION]
...
+  - SOF_ALSA_OPTS contains optional parameters passed to both aplay and arecord.
+  - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed specifically to aplay and arecord.
+These options aply to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL
```
golowanow, line 62:
```suggestion
These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
```

golowanow (changes requested):
@gbarkadiusz - why is shellcheck (binary!) included with its .txt files in this PR ?!
I'll continue review for the rest of changes, but -1 for now.

marc-hb:
>  why is shellcheck (binary!) included with its .txt files in this PR ?!

At first I felt ashamed I missed that and approved, then I realized this just came in the most recent force-push; _after_ I approved.

The reason is probably this above:
- https://github.com/thesofproject/sof-test/pull/1260#issuecomment-2786258632

@gbarkadiusz you should simply upgrade the entire distribution in your WSL container; not just shellcheck.

https://repology.org/project/shellcheck/versions

test-case/check-tinyalsa.sh
golowanow, line 1: I think `check-tinyalsa.sh` should be moved to another commit - "Add support for TinyALSA"

case-lib/lib.sh
```diff
@@ -752,17 +752,83 @@ func_lib_check_pa()
...
+    card_nr=$(printf '%s' "$1" | cut -d ':' -f2 | cut -d ',' -f1)
+
+    # Extract the device number (e.g., "10" from hw:0,10)
+    dev_nr=$(printf '%s' "$1" | cut -d ',' -f2)
+}
+
+# Function to extract the numeric format value from the PCM sample formats
+# There is passes PCM sample format while using ALSA tool (arecord)
+# While using tiny asla (tinycap) we need to convert PCM sample fomrat to bites
+# suporrted by tiny alsa -b parameter.
```
golowanow, line 775:
```suggestion
# While using TinyALSA (tinycap -b) we need to convert PCM sample fomrat to bits.
```
please use 'TinyALSA' and 'ALSA' naming consistently in this PR (where it is not a variable name, but some description)

golowanow:
>  ... came in the most recent force-push; _after_ I approved.

perhaps this repo/branch has (default?) setting of keep (no dismiss) approvals on force-push ?

test-case/check-tinyalsa.sh
```diff
@@ -0,0 +1,34 @@
...
+# It expects the test case script file name (without path) as the first parameter,
+# followed by other parameters required for that test case.
+
+set -e
+
+# Function to check if a command is available
+check_command() {
+    if command -v "$1" &> /dev/null; then
```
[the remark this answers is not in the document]
marc-hb: `type "$1"` does it all. I don't understand why some people use `command`.

https://lwn.net/Articles/874049/

test-case/check-tinyalsa.sh
marc-hb: > I think check-tinyalsa.sh should be moved to another commit - "Add support for TinyALSA"

Yes, please, please @gbarkadiusz move this wrapper to a separate PR. So we can finally close this PR. It was not an easy one and it deserved time but no need to add new stuff. Why add it so late? 

This is a bit ironic: it's common for submitters to complain that reviewers keep "moving the goal post". This time it's the exact way around! :-)

https://github.com/zephyrproject-rtos/zephyr/pull/83839/files

test-case/check-tinyalsa.sh
marc-hb: The LICENSE is especially important. But: separate PR.

marc-hb (commented):
> perhaps this repo/branch has (default?) setting of keep (no dismiss) approvals on force-push ?

Yes, feel free to change it (if you can find it!)

There is no "good" setting IMHO:

- auto-dismissal increases review overhead, especially when submitters don't "batch" force-pushes but do it immediately after each review.

- Preserving the approval state runs the risk of submitters "sneaking" large changes after approval - intentionally or not!

test-case/check-tinyalsa.sh
golowanow: > > I think check-tinyalsa.sh should be moved to another commit - "Add support for TinyALSA"
> 
> Yes, please, please @gbarkadiusz move this wrapper to a separate PR. So we can finally close this PR. 

There are two commits currently in this PR, and for `check-tinyalsa.sh` either the first commit ("Add support for TinyALSA") looks better, or just one more commit here for that wrapper script.

IMO, having TinyALSA feature in one PR is better/easier to keep its implementation consistent and tested.

Let's finalize this PR with minimal changes to what's already done here and discussed.

gbarkadiusz:
I accidently added shellcheck to that pr with last push force.
I've applied all above suggestions. @golowanow Please take a look again.

golowanow:
@gbarkadiusz you still have `shellcheck` in this PR: added by the first commit 4df594cbea4543f41 and removed in the last commit 5e56a0aba0a

marc-hb:
> @gbarkadiusz you should simply upgrade the entire distribution in your WSL container; not just shellcheck.

In the meantime:
- Don't install shellcheck inside your clone
- Do not use "blind" git commands like `git commit -a` or `git add *`

Either of these would have avoided the issue.

golowanow (commented):
I think with introducing TinyALSA, it needs 

1) check `tinyalsa` and `sox` packages at https://github.com/thesofproject/sof-test/blob/main/env-check.sh

and

2) update https://github.com/thesofproject/sof-test/blob/main/README.md?plain=1#L15

case-lib/lib.sh
```diff
@@ -752,17 +752,82 @@ func_lib_check_pa()
...
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        dlogc "tinyplay $*"
+        # shellcheck disable=SC2154
+        sox -n -r "$rate" -c "$channel" noise.wav synth "$duration" white
```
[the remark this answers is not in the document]
marc-hb: `set -e` is very useful, but it does not work in every situation. There is a number of documented exceptions / situations where it does nothing. Like in C, it's good to explicitly check the exit status of all "important" test commands like this:

```shell
    important_test_command ... || die 'important command ...."
```

More context in:
- https://github.com/thesofproject/sof-test/issues/312

case-lib/lib.sh
```diff
@@ -752,17 +752,82 @@ func_lib_check_pa()
...
+    fi
+}
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        dlogc "tinyplay $*"
```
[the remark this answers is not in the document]
marc-hb: Here's a logging trick that I use a lot:

```shell
  ( set -x
    sox -n -r "$rate" -c "$channel" noise.wav syn
  )
```

It has some limitations but it works  fine in the majority of cases.

case-lib/lib.sh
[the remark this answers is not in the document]
golowanow: better to add an explicit check for `sox` result here

marc-hb:
I really don't understand the insistence on including `check-tinyalsa.sh` in this PR. It's a different file so postponing it does not even require any git rebase skillz...

golowanow:
> I really don't understand the insistence on including `check-tinyalsa.sh` in this PR. It's a different file so postponing it does not even require any git rebase skillz...

we already use this wrapper script to run TinyALSA tests on CI with this PR

marc-hb:
> we already use this wrapper script to run TinyALSA tests on CI with this PR

It's convoluted to prefix with a wrapper script when prefixing with the variable is enough; strange.

This "check-tinyalsa.sh" can be a separate (and smaller) test that runs first = exactly what its name sounds. Similar to the sanity checks found in `test-case/verify-kernel-boot-log.sh`: they are run once at the start, not every time.

golowanow:
@gbarkadiusz please let me know when all my comments are addressed to review again, and I hope, it would be the last round in this review.

> > we already use this wrapper script to run TinyALSA tests on CI with this PR
> 
> It's convoluted to prefix with a wrapper script when prefixing with the variable is enough; strange.

@marc-hb, this way, as a wrapper script, it allows for both alsa and tinyalsa test case flavors to coexist in our current CI environment's configuration.

README.md
```diff
@@ -52,12 +58,14 @@ Usage: ./check-playback.sh [OPTION]
...
-invocations. SOF_ALSA_OPTS parameters are passed to both aplay and
-arecord. Warning these environments variables do NOT support parameters
+Some tests support these environment variables (work in progress):
+  - SOF_ALSA_TOOL is used to select the audio tool for testing.
+  Set this variable to 'alsa' (default value) or 'tinyalsa' to choose between the ALSA and TinyALSA toolsets.
+  - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
+  - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
+These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
+
+Warning these environments variables do NOT support parameters
```
redzynix, line 68:
```suggestion
Warning, these environment variables do NOT support parameters
```

marc-hb:
> @marc-hb, this way, as a wrapper script, it allows for both alsa and tinyalsa test case flavors to coexist in our current CI environment's configuration.

I still don't get it sorry. Anywhere you write:

```
check-tinyalsa.sh sometest.sh test args
```

You can write instead:

```
SOF_ALSA_TOOL=tinyalsa sometest.sh test args
```

... and the effect is the same - but with fewer indirections and complexity.

(The environment variable can also be used in other ways that the wrapper can't do)

marc-hb (changes requested):
> I still don't get it sorry.

Thanks @golowanow for the private discussion and explanation. Trying to summarize it here:

sof-test has always been invoked by an internal, higher level, Intel test framework that manages test plans and test runs. This framework has some shortcomings, probably the most important (and somewhat relevant here) being sof-framework/issues/602.

That framework does NOT allow environment variables when defining test plans. The interface only lets the user input a test script and script arguments. So, this wrapper script is a way to disguise and shoehorn an environment variable as a test script, sneaking the real test script as a parameter.

Considering this situation, I cannot think of any better to achieve what this wrapper does. BUT, I think the actual purpose of this script should be much clearer. So interactive users and better test frameworks do NOT start using it when they do not need to.

- The wrapper should not be called `check-tinyalsa.sh` which sounds like something good. It should be called something like `tinyalsa-wrapper.sh`.
- The comments at the top of the wrapper should clearly state something like "This is a wrapper that allows to define SOF_ALSA_TOOL in frameworks that cannot define environment variables. Do not use this script if you can just define SOF_ALSA_TOOL.
- The "checking bits" are useful and a very different feature! They should be in a separate "check-tinyalsa.sh` script that performs ONLY the checks and that everyone can use. The former script can call the latter script.

golowanow (approved):

gbarkadiusz:
@marc-hb Please take a look.

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+    # Extract the device number (e.g., "10" from hw:0,10)
+    dev_nr=$(printf '%s' "$1" | cut -d ',' -f2)
+}
+
+# Function to extract the numeric format value from the PCM sample formats
+# There is passes PCM sample format while using ALSA tool (arecord)
+# While using TinyALSA (tinycap -b) we need to convert PCM sample fomrat to bits.
+extract_format_number() {
+    # (e.g., extracting '16' from 'S16_LE')
+    format=$(printf '%s' "$1" | grep '[0-9]\+' -o)
```
marc-hb, line 777: `format` is a very common name for a global variable.

This function returns a single value so it could just print it instead and avoid a global?

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+{
+    local idx="$1"
+
+    channel=$(func_pipeline_parse_value "$idx" channel)
+    rate=$(func_pipeline_parse_value "$idx" rate)
+    fmts=$(func_pipeline_parse_value "$idx" fmt)
+    dev=$(func_pipeline_parse_value "$idx" dev)
+    pcm=$(func_pipeline_parse_value "$idx" pcm)
+    type=$(func_pipeline_parse_value "$idx" type)
+    snd=$(func_pipeline_parse_value "$idx" snd)
```
marc-hb, line 792: That's a lot of global variables with very short names...

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+}
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # shellcheck disable=SC2154
+	if ! sox -n -r "$rate" -c "$channel" noise.wav synth "$duration" white; then
```
marc-hb, line 804: Indentation? Did you use tabs? I think Github does not like them (I don't either)

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+	fi
+        dlogc "tinyplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS -D $card_nr -d $dev_nr  -i wav noise.wav"
+	# shellcheck disable=SC2086
+        tinyplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS -D "$card_nr" -d "$dev_nr"  -i wav noise.wav
+    elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
+        # shellcheck disable=SC2086
+        aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    else
+        printf '%s' "Unknown alsa tool: $SOF_ALSA_TOOL "
```
marc-hb, line 816: This should be fatal, using `die` or similar.

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+
 aplay_opts()
 {
-    dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-    # shellcheck disable=SC2086
-    aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # shellcheck disable=SC2154
+	if ! sox -n -r "$rate" -c "$channel" noise.wav synth "$duration" white; then
+            printf '%s' "Error: sox command failed." >&2
```
marc-hb, line 805:
```suggestion
            printf 'Error: sox command failed.\n' >&2
```
or
```suggestion
            dloge 'Error: sox command failed." 
```

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+
 arecord_opts()
 {
-    dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
-    # shellcheck disable=SC2086
-    arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+
+    if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # shellcheck disable=SC2154
+        extract_format_number "$fmt_elem"
```
marc-hb, line 825: Could you add a comment explaining where $fmt_elem comes from? Shellcheck seems to make a good point here....

More global variables :-(

case-lib/lib.sh
```diff
@@ -752,17 +752,87 @@ func_lib_check_pa()
...
+	dlogc "tinycap $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $file -D $card_nr -d $dev_nr -c $channel -t $duration -r $rate -b $format"
+	# shellcheck disable=SC2086
+        tinycap $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$file" -D "$card_nr" -d "$dev_nr" -c "$channel" -t "$duration" -r "$rate" -b "$format"
+    elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+        # shellcheck disable=SC2086
+        arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+    else
+        printf '%s' "Unknown alsa tool: $SOF_ALSA_TOOL "
```
marc-hb, line 834: Should be fatal using `die` or similar.

case-lib/lib.sh
```diff
@@ -1063,15 +1133,31 @@ set_alsa_settings()
...
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                amixer -Dhw:0 -- sset "$mixer_name" 100%
+            else
+               amixer -Dhw:0 -- sset "$mixer_name" 0dB
+            fi
+        done
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        tinymix -D0 controls | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
```
marc-hb, line 1148: Nit: later, it would be nice to de-duplicate the sed+grep bit with a function. No need to do it now.

marc-hb (commented):
I keep finding issues sorry. I don't which it is:
- I didn't see them before
- The code has changed
- I reported them before but my comments got "lost"

Nothing major.

case-lib/lib.sh
gbarkadiusz: Fixed.

case-lib/lib.sh
gbarkadiusz: Fixed.

case-lib/lib.sh
gbarkadiusz: Done.

case-lib/lib.sh
gbarkadiusz: True, Fixed.

case-lib/lib.sh
gbarkadiusz: I know, but I only moved this section from:
https://github.com/thesofproject/sof-test/pull/1260/files#diff-ac523ac7afa0a9a96eb0ee6c918c10f317d1b201b7124bf10bcf98b611b7324dL78
https://github.com/thesofproject/sof-test/pull/1260/files#diff-4c371b42f9f03814333bc7e33dce34faf542f27087fcf0b5702029f55f630edbL71
to an additional function in lib.sh file.

case-lib/lib.sh
gbarkadiusz: Fixed.

marc-hb (approved):

PR #1261, opened 2025-03-12
CODEOWNERS: Add maintainers explicitly

golowanow:
Add maintainers as code owners explicitly to resolve 'unknown owner' error which appears in forks.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,10 +2,10 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-*                       @thesofproject/test-maintainers
+*                       @thesofproject/test-maintainers @lgirdwood @marc-hb @golowanow
 
 # Order is important. The last matching pattern has the most precedence.
-# So if a pull request only touches javascript files, only these owners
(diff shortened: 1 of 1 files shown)


lgirdwood:
@marc-hb if you are still interested pls approve, if not we can remove your name from the PR.

marc-hb (approved):
Thanks, I don't mind the notifications, there are not that many :-) I will remove myself if that changes.

marc-hb:
> Add maintainers as code owners explicitly to resolve 'unknown owner' error which appears in forks.

The change LGTM but I don't understand and I'm curious about this GitHub error message... could you elaborate and do you have an example you can share?

golowanow:
> I'm curious about this GitHub error message... could you elaborate and do you have an example you can share?

well, an example is this PR itself - currently it shows only the test-maintainers group as the CODEOWNER's owner (and myself as the group member). GH should also show a error here if there is an owner which doesn't exist or have no access to the repo.

image

marc-hb:
> ... to resolve 'unknown owner' error which appears in forks.

> GH should also show a error here if ...

These look like two different issues. Nevermind.

PR #1263, opened 2025-03-14
ALSABAT:fix bug: updated alsabat mixer settings

mohsinkh30:
alsa_settings/PTLP_RVP_SDW.sh: Modified alsabat mixer settings with respect to UCM2 file for PTL SDW RT722"

### DIFF
--- a/alsa_settings/PTLP_RVP_SDW.sh
+++ b/alsa_settings/PTLP_RVP_SDW.sh
@@ -1,6 +1,16 @@
 set -e
 
+#Enable "SoundWire microphones"
+amixer -c sofsoundwire cset name='rt722 FU1E Capture Switch' 1
+
+#Enable Speaker Switch
+amixer -c sofsoundwire cset name='Speaker Switch' on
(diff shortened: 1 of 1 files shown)


sofci:

golowanow:
@mohsinkh30 do we have a CI probe run with this change already ?

kv2019i (approved):
Code change looks good.

The git commit message is not following common style. The summary line should be with 72 characters and there should be a paragraph explaining the change. Your git user name is also not entirely correct.

Not blocker for sof-test, but please note in future PRs.

kv2019i:
@golowanow It's true the volume is very low:

```
Channel 1 - Checking for target frequency 599.00 Hz
Amplitude: 175.2; Percentage: [0]
Detected peak at 599.12 Hz of 38.21 dB
```

Still better than clipping, so I'm ok to fine-tune later.

bardliao (approved):

golowanow (approved):

PR #1267, opened 2025-04-10
TinyAlsa: Add support for tinymix command.

gbarkadiusz:
Refactor the volume_basic_test to use lib.sh instead of calling ALSA commands directly, and add support for the tinymix command from TinyAlsa.

Note: This PR depends on [sof-test#1260](https://github.com/thesofproject/sof-test/pull/1260) and should only be merged after it has been merged.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -952,7 +952,7 @@ is_ipc4()
 logger_disabled()
 {
     # Disable logging when available...
-    if [ ${OPT_VAL['s']} -eq 0 ]; then
+    if [ "${OPT_VAL['s']}" -eq 0 ]; then
         return 0
     fi
 
@@ -1060,18 +1060,63 @@ set_alsa_settings()
     esac
 }
 
+set_sof_volume()
+{
+    local device=$1
+    local control_name=$2
+    local value=$3
+
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        dlogc "amixer -c$device' -- cset 'name=$control_name' '$value'"
+        amixer -c"$device" cset name="$control_name" "$value" > /dev/null
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        dlogc "tinymix -D'$device' set '$control_name' '$value'"
+        tinymix -D"$device" set "$control_name" "$value"  > /dev/null
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
+}
+
+get_sof_controls()
+{
+    local sofcard=$1
+
+    if [[  "$SOF_ALSA_TOOL" = "alsa" ]]; then
+            amixer -c"$sofcard" controls
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+            tinymix --card "$sofcard" controls
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
+
+}
+
 reset_sof_volume()
 {
     # set all PGA* volume to 0dB
-    amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
-    while read -r mixer_name
-    do
-        if is_ipc4; then
-            amixer -Dhw:0 -- sset "$mixer_name" 100%
-        else
-            amixer -Dhw:0 -- sset "$mixer_name" 0dB
-        fi
-    done
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        amixer -Dhw:0 scontrols | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                amixer -Dhw:0 -- sset "$mixer_name" 100%
+            else
+                amixer -Dhw:0 -- sset "$mixer_name" 0dB
+            fi
+        done
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        tinymix -D 0 contorls | sed -e "s/^.*'\(.*\)'.*/\1/" |grep -E 'PGA|gain' |
+        while read -r mixer_name
+        do
+            if is_ipc4; then
+                tinymix -D 0 set "$mixer_name" 100%
+            else
+                tinymix -D 0 set "$mixer_name" 0dB
+            fi
+        done
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
 }
 
 DO_PERF_ANALYSIS=0
--- a/test-case/volume-basic-test.sh
+++ b/test-case/volume-basic-test.sh
@@ -39,11 +39,39 @@ maxloop=${OPT_VAL['l']}
 
 start_test
 
-func_error_exit()
+check_alsa_tool_process()
 {
-    dloge "$*"
-    pkill -9 aplay
-    exit 1
+    sleep 1
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        # Check if the aplay process is running
+        if [[ ! $(pidof aplay) ]]; then
+            die "aplay process is terminated too early"
+        fi
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        # Check if the tinyplay process is running
+        if [[ ! $(pidof tinyplay) ]]; then
+            die "tinyplay process is terminated too early"
+        fi
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
+}
+
+func_exit()
+{
+
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        pkill -9 aplay
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        pkill -9 tinyplay
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
+
+    if [[ "$1" -eq 1 ]]; then
+        dloge "$*"
+        exit 1  # error
+    fi
 }
 
 [[ -z $tplg ]] && die "Missing tplg file needed to run"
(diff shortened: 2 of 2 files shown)


test-case/volume-basic-test.sh
```diff
@@ -39,41 +39,63 @@ maxloop=${OPT_VAL['l']}
...
+
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
+        pkill -9 aplay
+    elif [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        pkill -9 tinyplay
+    else
+        echo "Unknown alsa tool $SOF_ALSA_TOOL"
+    fi
+
+    if [[ "$1" -eq 1 ]]; then
```
marc-hb, line 71: All SOF_ALSA_TOOL logic should be in the same central place: `lib.sh`. I suspect it should also search for `arecord`

`func_exit 1` is cryptic and error-prone. It's much simpler and clearer to have two distinct functions. Then every test can do what it wants and it's always clear what it does.

```shell
kill_play_record()
{
  if "$SOF_ALSA_TOOL" 
     kill ...
}

cleanup_and_fail() # better name if you can think of one
{
   kill_play_record
   die "$1"
}

```

test-case/volume-basic-test.sh
```diff
@@ -39,41 +39,63 @@ maxloop=${OPT_VAL['l']}
...
 start_test
 
-func_error_exit()
+check_alsa_tool_process()
 {
-    dloge "$*"
-    pkill -9 aplay
-    exit 1
+    sleep 1
+    if [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
```
marc-hb, line 45: Nit:
```suggestion
    case "$SOF_ALSA_TOOL" in
      'alsa') ...  ;;
      
      *) die ...
```

case-lib/lib.sh
```diff
@@ -797,6 +797,86 @@ initialize_audio_params()
...
+    kill_play_record
+    die "$1"
+}
+
+check_alsa_tool_process()
+{
+    case "$SOF_ALSA_TOOL" in
+       "alsa")
+           # Check if the aplay process is running
+            if [[ ! $(pidof aplay) ]]; then
```
marc-hb, line 864: Always prefer the exit status to the standard output when possible.
```suggestion
            pidof -q aplay ||
                  die ...
```
Avoiding negations is also more readable when possible. In this case it translates to plain English like "still_running or die".

Shorter _and_ more readable!

test-case/volume-basic-test.sh
```diff
@@ -51,29 +51,24 @@ func_pipeline_export "$tplg" "type:playback"
...
-dev=$(func_pipeline_parse_value 0 dev)
-
-dlogc "aplay -D $dev -c $channel -r $rate -f $fmt /dev/zero &"
+duration=50 # Duration of playing white noise while testing under tinyalsa
+initialize_audio_params "0"
 # play into background, this will wake up DSP and IPC. Need to clean after the test
-aplay -D "$dev" -c "$channel" -r "$rate" -f "$fmt" /dev/zero &
-
+# shellcheck disable=SC2154
+aplay_opts -D "$dev" -c "$channel" -r "$rate" -f "$fmts" -d "$duration" /dev/zero >/dev/null &
```
golowanow, line 58: `-d $duration` looks confusing to me: is it also passed to aplay for 'alsa' ? is it correct ?

test-case/volume-basic-test.sh
```diff
@@ -51,29 +51,24 @@ func_pipeline_export "$tplg" "type:playback"
...
 
 [[ $PIPELINE_COUNT -eq 0 ]] && die "Missing playback pipeline for aplay to run"
-channel=$(func_pipeline_parse_value 0 channel)
-rate=$(func_pipeline_parse_value 0 rate)
-fmt=$(func_pipeline_parse_value 0 fmt)
-dev=$(func_pipeline_parse_value 0 dev)
-
-dlogc "aplay -D $dev -c $channel -r $rate -f $fmt /dev/zero &"
+duration=50 # Duration of playing white noise while testing under tinyalsa
+initialize_audio_params "0"
```
golowanow, line 55: `dlogc "aplay        ` is missed here as it was before.

case-lib/lib.sh
```diff
@@ -797,6 +797,86 @@ initialize_audio_params()
...
+                die "aplay process is terminated too early"
+            fi
```
golowanow, line 866:
```suggestion
            [[ ! $(pidof aplay) ]] && die "aplay process is terminated too early"
```
as it was before, it is shorter and self documented, imo

test-case/volume-basic-test.sh
```diff
@@ -83,21 +78,20 @@ do
...
+                              cleanup_and_fail "mixer return error, test failed"
         done
     done
 
     sleep 1
 
     dlogi "check dmesg for error"
     sof-kernel-log-check.sh "$KERNEL_CHECKPOINT" ||
-                      func_error_exit "dmesg has errors!"
+                      cleanup_and_fail "dmesg has errors!"
```
golowanow, line 90:
```suggestion
                      kill_play_record || die diecleanup_and_fail "dmesg has errors!"
```

test-case/volume-basic-test.sh
```diff
@@ -83,21 +78,20 @@ do
         dlogi "$volctrl"
 
         for vol in "${volume_array[@]}"; do
-            dlogc "amixer -c$sofcard cset name='$volctrl' $vol"
-            amixer -c"$sofcard" cset name="$volctrl" "$vol" > /dev/null ||
-                              func_error_exit "amixer return error, test failed"
+            set_sof_volume "$sofcard" "$volctrl" "$vol" ||
+                              cleanup_and_fail "mixer return error, test failed"
```
golowanow, line 82:
```suggestion
                              kill_play_record || die "mixer return error, test failed"
```

case-lib/lib.sh
```diff
@@ -797,6 +797,86 @@ initialize_audio_params()
...
 
+set_sof_volume()
+{
+    local device=$1
+    local control_name=$2
+    local value=$3
+
+    case "$SOF_ALSA_TOOL" in
+        'alsa')
+            dlogc "amixer -c$device' -- cset 'name=$control_name' '$value'"
```
golowanow, line 808:
```suggestion
            dlogc "amixer -c$device' cset 'name=$control_name' '$value'"
```

case-lib/lib.sh
```diff
@@ -797,6 +797,86 @@ initialize_audio_params()
...
+            amixer -c"$device" cset name="$control_name" "$value" > /dev/null
+       ;;
+        'tinyalsa')
+           dlogc "tinymix -D'$device' set '$control_name' '$value'"
```
golowanow, line 812:
```suggestion
            dlogc "tinymix -D'$device' set '$control_name' '$value'"
```

test-case/volume-basic-test.sh
```diff
@@ -51,29 +51,24 @@ func_pipeline_export "$tplg" "type:playback"
 logger_disabled || func_lib_start_log_collect
 
 [[ $PIPELINE_COUNT -eq 0 ]] && die "Missing playback pipeline for aplay to run"
-channel=$(func_pipeline_parse_value 0 channel)
-rate=$(func_pipeline_parse_value 0 rate)
-fmt=$(func_pipeline_parse_value 0 fmt)
-dev=$(func_pipeline_parse_value 0 dev)
-
-dlogc "aplay -D $dev -c $channel -r $rate -f $fmt /dev/zero &"
+duration=50 # Duration of playing white noise while testing under tinyalsa
```
golowanow, line 54: `duration=50` looks like a magic value, and if it is tinyalsa only it should be in the appropriate `case` statement for tinyalsa.

test-case/volume-basic-test.sh
marc-hb: > kill_play_record || die ...

It could be changed, but  right now the exit status of `kill_play_record` does not like it can be trusted for doing this.

test-case/volume-basic-test.sh
gbarkadiusz: It was, but I refactored `volume_basic_test` to use `lib.sh` and `aplay_opts`, so the log now comes from that function.
https://github.com/thesofproject/sof-test/blob/main/case-lib/lib.sh#L812

test-case/volume-basic-test.sh
gbarkadiusz: TinyALSA doesn't support continuous playback, which is why the test needs to provide that value. I'll add an `if` statement to avoid setting this variable when testing with ALSA.

golowanow (commented):
Please change commit/PR description: this PR adds tinymix support at one test only.

marc-hb (commented):
I did not carefully scrutinize every single line but it Looks Good To Me.

case-lib/lib.sh
```diff
@@ -797,9 +797,92 @@ initialize_audio_params()
...
+    esac
+}
+
 aplay_opts()
 {
     if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+        if [[ "$SOF_ALSA_TOOL" = "tinyalsa" ]]; then
+            # Duration of playing white noise while testing under tinyalsa
+            duration=50
+        fi
```
golowanow, line 885: it is already under if tinyalsa, isn't it ?
```suggestion
        # Duration of playing white noise while testing under tinyalsa
        duration=50        
```

case-lib/lib.sh
```diff
@@ -1133,34 +1214,21 @@ set_alsa_settings()
 
 reset_sof_volume()
 {
+    level_db=$(is_ipc4 && echo "100%" || echo "0%")
```
golowanow, line 1217:
```suggestion
    level_db=$(is_ipc4 && echo "100%" || echo "0dB")
```

case-lib/lib.sh
```diff
@@ -1133,34 +1214,21 @@ set_alsa_settings()
 
 reset_sof_volume()
 {
+    level_db=$(is_ipc4 && echo "100%" || echo "0dB")
```
marc-hb, line 1217:
```suggestion
    level_db=$(if is_ipc4; then echo "100%"; else echo "0dB"; fi)
```

`A && B || C` is a very dangerous construct and bad example. There is no bug in this particular case but there very easily can be, compare:

```shell
if true; then grep notfound /etc/passwd ; else 'ELSE is not run'; fi
true && grep notfound /etc/passwd || echo 'ELSE is also run!!'
```

gbarkadiusz: Thank you for that hint.

case-lib/lib.sh
```diff
@@ -797,9 +797,90 @@ initialize_audio_params()
...
+get_sof_controls()
+{
+    local sofcard=$1
+
+    case  "$SOF_ALSA_TOOL" in
+        'alsa')
+            amixer -c"$sofcard" scontrols
+       ;;
+        'tinyalsa')
+            tinymix --card "$sofcard" controls
```
golowanow, line 830: this difference `scontrols` vs. `controls` is confusing and needs its explanation in a comment

gbarkadiusz: The `tinymix` doesn't support `scontrols` its default command to get controls is `controls`.
From `tinymix -h`
```
usage: tinymix [options] <command>
options:
        -h, --help               : prints this help message and exits
        -v, --version            : prints this version of tinymix and exits
        -D, --card NUMBER        : specifies the card number of the mixer

commands:
        get NAME|ID              : prints the values of a control
        set NAME|ID VALUE(S) ... : sets the value of a control
                VALUE(S): integers, percents, and relative values
                        Integers: 0, 100, -100 ...
                        Percents: 0%, 100% ...
                        Relative values: 1+, 1-, 1%+, 2%+ ...
        controls                 : lists controls of the mixer
        contents                 : lists controls of the mixer and their contents
```

gbarkadiusz: According to my knowledge, it should still be `scontrols` as before. Using `controls` causes issues on some platforms when trying to access unavailable controls during setting.

PR #1270, opened 2025-05-12
[DNM] Add alsa-info debug

LukaszMrugala:
WIP

Print alsa-info to the logs.


sofci:

case-lib/lib.sh
```diff
@@ -1030,10 +1030,20 @@ re_enable_ntp_sync()
     sudo timedatectl set-ntp true
 }
 
+# Get alsa-info of DUTs in CI
+print_alsa_info()
+{
+    OUTPUT_FILE="/tmp/alsa-info.txt"
+    alsa-info --output "$OUTPUT_FILE" --with-aplay --with-amixer --with-alsactl --with-configs
+    ALSA_INFO_TEXT=$(cat "$OUTPUT_FILE")
+    dlogi "$ALSA_INFOTEXT"
```
marc-hb, line 1039: This is convoluted. Just do `alsa-info --stdout ...` and you can drop all the other lines. In fact, you can even drop the function and inline the call.

marc-hb (commented):
Replacing imperative with declarative is usually a good idea but why is this almost 10 lines longer!? Is it really equivalent?

For sure the script approach was INCREMENTAL: changing some settings and leaving other in place. Can `alsactl` do that too or does it save/restore a FULL state every time? In the latter case, this is not just refactoring with a different approach but a big functional change. Which may or may not be better, I don't know that but it should be very clear what it is.

By the way, incremental approaches tend to support (ALSA) upgrades better, will a "FULL restore" approach be compatible across different ALSA versions?

> Creating such files for DUTs no longer in CI is a tough nut to crack.

It shouldn't be hard to have some transition period that supports both approaches. That's even more important if the two approaches are not equivalent to each other.

LukaszMrugala:
> For sure the script approach was INCREMENTAL: changing some settings and leaving other in place. Can `alsactl` do that too or does it save/restore a FULL state every time? In the latter case, this is not just refactoring with a different approach but a big functional change. Which may or may not be better, I don't know that but it should be very clear what it is.

This approach should only modify the values mentioned in the `.state` file. So while a typical machine would have a couple dozen values, only the few specified in the `.state` file will be changed. That should mean this approach is equivalent.

It's most clear in the `MTLP_SDW_AIOC.state` case. The `.state` file specifies controls 1, 3, 5, 8, 35 and 36. The same controls were set by the `.sh` scripts. All the others should be left as-is.

LukaszMrugala:
> > Replacing imperative with declarative is usually a good idea but why is this almost 10 times longer!?
> 
> I spent more than 10s looking at this and realized most of the extra verbosity comes from `comment` clauses. Most of which do... not look like English comments but like code instead. I'm confused.

They inform the user or, more like, someone who wants to modify the values set there, which are possible. Let's take an arbitrary control as an example, firstly without a `comment` block:

```
control.1 {
	iface MIXER
	name 'rt711 FU05 Playback Volume'
	value.0 80
	value.1 80
}
```

Say that my tests come out too low (volume too low). I'd like to max out this volume. What is the maximum value? I cannot know without accessing the machine. Now with the `comment` block:
```
control.1 {
	iface MIXER
	name 'rt711 FU05 Playback Volume'
	value.0 80
	value.1 80
	comment {
		access 'read write'
		type INTEGER
		count 2
		range '0 - 87'
		dbmin -6525
		dbmax 0
		dbvalue.0 -525
		dbvalue.1 -525
	}
}
```

OK, maximum value is 87 and I can change `80` to `87` and be sure that it's max. I do not need to worry what type the values are, how many there are, etc. - it's all in the comment block.

alsa_settings/alsactl/README.md
```diff
@@ -0,0 +1,7 @@
+# Creating a .state file
+
+1. Gain access to a machine you'd like to create a state file for.
```
golowanow, line 3: How all the *.state files in this PR were obtained ? (platform in default state, kernel, version, topology, whatever else matters)

case-lib/lib.sh
```diff
@@ -1030,6 +1030,42 @@ re_enable_ntp_sync()
     sudo timedatectl set-ntp true
 }
 
+# Get alsa-info of DUTs in CI
+print_alsa_info()
+{
+    dlogi "----------- ↓ alsa-info ↓ ------------"
+    alsa-info --stdout --with-aplay --with-amixer --with-alsactl --with-configs --no-upload
+    dlogi "----------- ↑ alsa-info ↑ ------------"
```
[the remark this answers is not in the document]
marc-hb: @golowanow I recommended this change. In the first version, the output was directed to a file AND that file was 1. read in a shell variable 2. printed using `dlogi`.

I didn't realize it could be "huge".

If it is indeed "huge" (how many lines typically?) then I agree it should not be logged entirely to the standard output. Shell variable and dlogi arguments should never be "huge" either.

I find "----" rulers very helpful when dealing with huge outputs but better not have huge outputs in the first place.

Maybe a smaller subset could be useful? Something like `alsa-info --stdout | grep -C 5 something_useful`

case-lib/lib.sh
```diff
@@ -1048,7 +1084,19 @@ set_alsa_settings()
         ;;
         TGLU_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC-ipc4 | TGLU_RVP_NOCODEC-ipc4 | MTLP_RVP_NOCODEC | MTLP_RVP_NOCODEC-multicore-2cores | MTLP_RVP_NOCODEC-multicore-3cores | LNLM_RVP_NOCODEC)
             dlogi "Use reset_sof_volume function to set amixer setting."
-	;;
+        ;;
+        LNLM_RVP_HDA | LNLM_SDW_AIOC | MTLP_RVP_HDA | MTLP_RVP_SDW | MTLP_SDW_AIOC | PTLH_HDA_AIOC | PTLH_SDW_RT712 | PTLP_RVP_SDW)
```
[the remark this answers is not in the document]
marc-hb: Sounds great for backwards compatibility and not having to touch legacy configurations.

marc-hb (commented):
> The comment section should not be parsed by the machine but used by humans. It looks like a machine comment, because ALSA generates the comment section when dumping its state, so the end user doesn't have to write it by themselves.

That's unusual and not intuitive. Even worse if not documented by ALSA. I think that deserves an explanation somewhere, maybe some comments  in alsactl/README.md and lib.sh?

Also, can you make  these comments look LESS like they are parsed and used by ALSA? Changing the wording, order, other, dunno.

Or maybe I'm still missing something.

alsa_settings/alsactl/README.md
LukaszMrugala: sof and linux builds were taken from that day's CI relevant run. Then I've run the PR with alsa-info print added in. This gave me the full .state file I could then adjust.

case-lib/lib.sh
```diff
@@ -1030,6 +1030,42 @@ re_enable_ntp_sync()
...
+# Check whether a relevant .state file exists
+# If not, save current state for future use
+# param1: script home path
+# param2: platform name
+get_alsactl_state_or_create()
+{
+    ALSACTL_STATE_FILE_PATH="$1"/alsa_settings/alsactl/"$2".state
+    # .state file does not exist. Creating one from current setup.
+    if [ ! -f "$ALSACTL_STATE_FILE_PATH" ]; then
+        alsactl store --file="$ALSACTL_STATE_FILE_PATH"
```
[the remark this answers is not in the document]
LukaszMrugala: That shouldn't be a problem, right?

case-lib/lib.sh
```diff
@@ -1048,9 +1095,11 @@ set_alsa_settings()
         ;;
         TGLU_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC-ipc4 | TGLU_RVP_NOCODEC-ipc4 | MTLP_RVP_NOCODEC | MTLP_RVP_NOCODEC-multicore-2cores | MTLP_RVP_NOCODEC-multicore-3cores | LNLM_RVP_NOCODEC)
             dlogi "Use reset_sof_volume function to set amixer setting."
-	;;
+        ;;
         *)
             # if script name is same as platform name, default case will handle all
+            if [ -f "$SCRIPT_HOME"/alsa_settings/alsactl/"$PNAME".state ]; then
```
golowanow, line 1101: please reuse `get_alsactl_state()` - the new function introduced in this PR, isn't it for this exact purpose ?

alsa_settings/alsactl/README.md
golowanow: as I understand the idea behind this PoC proposal, these `*.state` files are transitional having their content specific to the SOF/Linux components currently deployed on a DUT, thus in most cases they should be explicitly created at some early stage of a test plan execution to keep the appropriate defaults to reset before each test case, where it is important.

LukaszMrugala: I do not seem to understand. They necessarily cannot be transitional (temporary in their existence), as they are the definition of the DUT state we want to achieve, just as the previously-used `.sh` files. I would not call the `.sh` files transitional as I wouldn't call the `.state` files transitional.

I also do not understand had would we create the `.state` files during testplan execution, given that those files are our source of truth when it comes to out desired state. It would be akin to creating a MAVEN file in Java for every compilation - the benefits would be slim and the process unnecessarily onerous.

case-lib/lib.sh
LukaszMrugala: You're right; done

alsa_settings/alsactl/README.md
golowanow: > .. given that those files are our source of truth when it comes to out desired state

right, my question is about these *.state files which are included into this PR, and what the 'desired state' they set.

If the basic goal is to reset the state before each test to defaults, then there is no need to keep these files in the sof-test repo as a 'golden source', but call `get_alsactl_state_or_create()` once at the test plan beginning right after the DUT deployment stage, and then call `restore_settings_via_alsactl()` before each test it needs. Isn't it enough ? Did you try/consider to call `alsactl init` for the same purpose ?

When there is a need to do some custom/experimental state change, it can be done e.g. by a CI script deploying a .state file here from some other repo/source with the platform-specific BKC, or a state file placed here by some test case script during its test plan execution.

marc-hb:
> I don't think it's worth the huge number of lines increase.
> It's confusing because it's mixing input versus output and code versus documentation in the same file.

In fact, if you had submitted `*.state` files with the same content than the `*.sh` files you're removing, without even explaining where and how you created them, any reviewer would have immediately noticed that they are strictly equivalent to the `*.sh` files, the entire confusion and discussion about "incremental" versus not would have not happened and this PR might have been already merged :-)

alsa_settings/alsactl/README.md
LukaszMrugala: @marc-hb Yes, that's exactly what I do - full `.state` files can have over 60 controls specified, while we oftentimes need less than 10, we I trim them to just what the `.sh` had specified. 

@golowanow The full `.state` file has all the controls and their current values. However, the current `.sh` files specify only some controls to some values that can be different from what currently exist on the machine. That's why we do not create them during the runs - we'd be restoring the current machine state rather than restoring the desired machine state.

`alsactl init` is just `alsactl restore` on a specific file in a predetermined place. Why obfuscate the files to use rather than keep them in the same repo that uses them? Those aren't multi-gigabyte behemoths to be wary of downloading them unnecessarily.

LukaszMrugala:
> I think these comment blocks should be removed. I see the convenience but:
> 
> * I don't think it's worth the huge number of lines increase.
> * It's confusing because it's mixing input versus output and code versus documentation in the same file.
> * People should not just blindly "max out" without accessing the system and testing interactively. When a test is failing on some specific system, people should generally debug on  that specific system.

I can accept that logic. I've removed the comment lines inside of the `.state` files.

> Speaking of number of lines, does this work?
> 
> ```
> state.sofsoundwire {
> 	control.1 {
> 		iface MIXER  name 'rt711 FU05 Playback Volume'
> 		value.0 80   value.1 80
>       }
>     ...
> ```

I don't think so and it makes readability much worse - it breaks the `KEY VALUE` per-line segmentation common in configuration files.

I've implemented the changes and if no further problems are inputted, I'll squash the commits, rebase and be ready for merging (Rebase should only follow after the discussions' resolution).

marc-hb:
While I have not scrutinized every single, this LGTM now.

> (Rebase should only follow after the discussions' resolution).

That's an interesting suggestion in a very old and complex debate :-)

- https://github.com/zephyrproject-rtos/zephyr/issues/39194

LukaszMrugala:
All commits squashed, PR rebased onto main to implement the RT712 changes, ready for review.

alsa_settings/alsactl/README.md
LukaszMrugala: > Why we expect that after alsactl init the DUT's state is wrong and we need some handmade *.state file to set correct defaults ?

As we have experienced tests failing in our current CI operation and we have never used `alsactl init`'s files contained by default in `/usr/share/alsa/init/` to restore the DUT state to our liking, why should we assume that `alsactl init` will bring the DUT to a state we desire? We have no information that would indicate that ALSA's default state is congruent with our desired state. Thusly, we are to use custom `.state` files, as that's the status quo.

> I think this PR should evolve from its current 'proof-of-concept' to some runnable test case, then it will be easier to see how it solves the problem and ready for merge.

In my understanding, this repo houses tests for ALSA configurations and tools to that end. Separate tests for the tools inside the `sof-test` repo seem missing and it would be awkward to task this PR specifically with establishing a base for tests of `sof-test` itself.

case-lib/lib.sh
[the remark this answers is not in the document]
LukaszMrugala: > It will look different to what is in the handcrafted *.state files which are for only one card at the moment.

It will look different, but that specific function is unused in the automatic code and is left as a tool for new `.state` file creation.

> Can you try your PR on a NOCODEC dut ?

The current code does not touch NOCODEC DUTs. The run No. 53589 contains a NOCODEC platform.

case-lib/lib.sh
```diff
@@ -1119,10 +1166,12 @@ set_alsa_settings()
...
         TGLU_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC_IPC4ZPH | ADLP_RVP_NOCODEC-ipc4 | TGLU_RVP_NOCODEC-ipc4 | MTLP_RVP_NOCODEC | MTLP_RVP_NOCODEC-multicore-2cores | MTLP_RVP_NOCODEC-multicore-3cores | LNLM_RVP_NOCODEC)
             dlogi "Use reset_sof_volume function to set amixer setting."
-	;;
+        ;;
         *)
             # if script name is same as platform name, default case will handle all
-            if [ -f "$SCRIPT_HOME"/alsa_settings/"$PNAME".sh ]; then
+            if get_alsactl_state "$SCRIPT_HOME" "$PNAME"; then
+                restore_settings_via_alsactl "$PNAME"
+            elif [ -f "$SCRIPT_HOME"/alsa_settings/"$PNAME".sh ]; then
```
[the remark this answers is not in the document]
LukaszMrugala: > This statement - .. uses it if able. If not .. in the commit description is confusing: it sounds like when the new method is unable to restore from the .state file, then the .sh script is expected to run.

There's no need to overcomplicate - anyone confused by the commit message can verify that the mentioned switch-case does not have any try-catch mechanism. Additionally, "if able" connects with the previous line's "checks whether a relevant .state file exists".

> Also, with all current *.sh scripts removed, no fallback will be possible, so it is better to keep them.

New .state files are supposed to replace the old .sh files. We should not assume that `alsactl` break by itself, lest we try to have `ksh` fallbacks in case `bash` breaks. Keeping the equivalent `.sh` files is superfluous with new equivalent `.state` files present.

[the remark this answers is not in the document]
golowanow: > There's no need to overcomplicate

right, let's remove alss_settings/*.sh files describing the change as `..  whether a relevant alsa_settings/alsactl/$MODEL.state file exists, or alsa_settings/$MODEL.sh is executed instead, if present.`

> We should not assume that alsactl break by itself,

We should consider it is possible for `alsactl restore` to fail when the .state file is incorrect, and then exit the test case with FAIL to avoid running it in some unexpected state.

golowanow (commented):
@LukaszMrugala, assuming `alsactl init` is used, please check do we really need to provide explicit *.state files with this PR if there is no difference there wrt the default state established by init.

LukaszMrugala:
We need more than just the default initialisation of configuration, as evidenced by the fact that we already do that via the `.sh` files. Different hardware and codec setups can have different e.g. amplifying power, which leads necessarily to different configurations on top of the default.
We really shouldn't cast into doubt our whole configuration setup in this PR.

golowanow:
> We need more than just the default initialisation of configuration, as evidenced by the fact that we already do that via the `.sh` files. 

that's should be easy to check: 

1. take a DUT $MODEL, reboot it, 
2. `alsactl init`, 
3. `alsactl store --file=foo_0.state`
4. `$MODEL.sh`
5. `alsactl store --file=foo_`.state`
6. `diff foo_0.state foo_1.state`

as a result it will be clear - do we actually need to have $MODEL.state or not

case-lib/lib.sh
```diff
@@ -1101,10 +1101,59 @@ re_enable_ntp_sync()
...
+
+    alsactl restore --file="$ALSACTL_STATE_FILE_PATH" --pedantic --no-init-fallback --no-ucm
+}
+
 # check-alsabat.sh need to run optimum alsa control settings
 # param1: platform name
 set_alsa_settings()
 {
+    # This will bring the machine ALSA state to a common known point - a good baseline
+    alsactl init
```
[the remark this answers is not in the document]
golowanow: I think that we even need it this way to reset before applying MODEL.state or MODEL.sh:
```
    alsactl init
    # restores ALSA defaults from /var/lib/alsa/asound.state
    sudo alsactl restore
```

marc-hb:
> We really shouldn't cast into doubt our whole configuration setup in this PR.

I will nuance this.

AFAIK, this PR does not try to make any functional change. It only tries to "convert" some .sh files to their .state equivalent. I think that's a good thing. It does not try to fundamentally change how devices are configured. That's perfectly fine. This PR can be merged.

-------

However, this PR is also an interesting time to discuss a bit the high level design of device configuration. Maybe that design discussion would have been better in an issue instead of a PR (like for instance #1275?). This PR did not link to any issue :-(


> We need more than just the default initialisation of configuration, as evidenced by the fact that we already do that via the .sh files. 

That's a bit circular :-) I mean because it being done does not make it automatically a good thing.

> Different hardware and codec setups can have different e.g. amplifying power, which leads necessarily to different configurations on top of the default.

... but there is no way `alsactl init` could ideally learn these differences and address this in the first place? Naive question sorry. Again the question is: if sof-test has a "better" configuration, why not share it with all ALSA users? Why keep it "secret"? Maybe there is a good reason not too (like: not deafening users maybe?) but it should be clearly explicit somewhere.

I'm asking all this because "CI and validation secrets" is unfortunately a very common practice in the industry and I've seen A LOT of them in my SOF time. So, when a test fails in some configuration, instead of spending time filing a product bug and waiting for it to be fixed maybe, the validation team performs a quick validation hack instead and ta-da: the tests are passing! One particular type of "green failures" to make managers happier. Not customers but who cares?

https://github.com/thesofproject/sof-test/labels/False%20Pass%20%2F%20green%20failure  + a lot more internally in sof-framework

case-lib/lib.sh
```diff
@@ -1180,6 +1185,11 @@ set_alsa_settings()
             fi
         ;;
     esac
+
+    # START DEBUG
+    alsactl store --file /tmp/after_set_alsa_settings_to_diff
+    diff /tmp/after_alsactl_init_to_diff /tmp/after_set_alsa_settings_to_diff
```
marc-hb, line 1191:
```suggestion
    diff -u tmp/after_alsactl_init_to_diff /tmp/after_set_alsa_settings_to_diff
```

marc-hb (commented):
Good idea!

I don't find the `_to_diff` suffixes useful. Everything that evolves over time can be "diffed". Name suffixes should describe what the file type and that's all. `.alsa_state` or something?

LukaszMrugala:
> 2. Simplify .state files and drop what duplicates `alsactl init`.

That's exactly what I have just done.

> Keep in mind different people test different ALSA versions.

If different ALSA versions have different `init` files and thus the initial common state is different between them, then the whole idea of using `alsactl init` as a common initial state is useless. In that case we cannot trim the `.state` files at all, as we have no guarantee of anything in the `init` configuration and we need to carry all of the configuration information that ensures the state for the tests on a given machine inside the `.state` files. Thus, if different ALSA versions have different `init`s, then we are back to where we were two weeks ago.

LukaszMrugala:
> If different ALSA versions have different `init` files and thus the initial common state is different between them, then the whole idea of using `alsactl init` as a common initial state is useless.

I've checke din the [repo](https://github.com/alsa-project/alsa-utils/tree/master/alsactl/init) and the last change to the `init` files has occured nine years ago. I'd say that's stable enough for us.

> It doesn't solve https://github.com/thesofproject/sof-test/issues/1275 which needs some MODEL_default.state with full and consistent parameters to restore it on that particular kind of DUT.

PR #1279 solves that.

golowanow:
>  I have trimmed the `.state` files so that they contain only the changes that are not in the `alsactl init`.

ok, thank you

> Keep in mind different people test different ALSA versions.

That's the point why we need reset to ALSA defaults to have it consistent with the current version.
The 'trimmed' MODEL.state files are just another way of what is done by MODEL.sh files to align with the DUT's harness.

@LukaszMrugala, please also address these comments - https://github.com/thesofproject/sof-test/pull/1270/files#r2166950950

case-lib/lib.sh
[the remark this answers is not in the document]
golowanow: This PR will already affect two tests, so the changes need to be aligned.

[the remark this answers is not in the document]
LukaszMrugala: I've created a consisted `setup_alsa` and added it to all tests using ALSA. Run tests before and after that change.

redzynix (approved):
Good job Lukasz, changes looks really good!

redzynix:
SOFCI TEST

case-lib/lib.sh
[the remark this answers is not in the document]
golowanow: > I think that we even need it this way to reset before applying MODEL.state or MODEL.sh:
> 
> ```
>     alsactl init
>     # restores ALSA defaults from /var/lib/alsa/asound.state
>     sudo alsactl restore
> ```

IMO it is essential to execute `alsactl restore` to reset properly.

golowanow (changes requested):

redzynix:
> test this please

test this please

LukaszMrugala:
test this please

marc-hb:
> test this please

@LukaszMrugala you cannot approve yourself, this is a security feature. You really need to request membership to https://github.com/orgs/thesofproject/teams/sof-developers

PR #1271, opened 2025-05-15
Fix: Set SOF_ALSA_TOOL to default value globally.

gbarkadiusz:
Set SOF_ALSA_TOOL variable to a default value globally.
Justification:
After tinyalsa support was added ([PR #1260](https://github.com/thesofproject/sof-test/pull/1260)), the use of SOF_ALSA_TOOL is now required for all tests before calling the aplay_opt or arecord_opt functions from lib.sh.

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -759,6 +759,7 @@ func_lib_check_pa()
 # before using 'aplay_opts' or 'arecord_opts' function.
 # Default is SOF_ALSA_TOOL='alsa'
 
+: "${SOF_ALSA_TOOL:="alsa"}"
 
 # Function to extract the card number and device number from $dev option (e.g., hw:0,10)
 parse_audio_device() {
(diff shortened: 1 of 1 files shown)


gbarkadiusz:
> > Refactor the multiple-pipline.sh to use initialize_audio_params function.
> > Justification: After tinyalsa support was added ([PR #1260](https://github.com/thesofproject/sof-test/pull/1260)), tests must set the $SOF_ALSA_TOOL variable before calling the aplay_opt or arecord_opt functions from lib.sh. The default value (alsa) for $SOF_ALSA_TOOL is set in the initialize_audio_params function.
> 
> Please don't mix 'fix' and 'refactor' purposes.
> 
> As a fix it is better to treat $SOF_ALSA_TOOL not defined as default 'alsa', or set $SOF_ALSA_TOOL 'alsa' default globally. This way it should work for any test case not 'refactored' yet.

WIP.

redzynix (approved):

PR #1272, opened 2025-05-22
[SKIP SOF-TEST] test-case: latency-baseline: introduce latency tests

golowanow:
Introduce latency tests using `JACK` Audio Connection Kit:
run `jackd` audio service, connect ports with loopback and run `jack_iodelay` to collect latency measurements, calculate and save statistics in a json report file.


test-case/latency-baseline.sh
```diff
@@ -0,0 +1,228 @@
...
+dlogi "Check Jack server can be started"
+check_jackd_configured   # "$@"
+
+dlogi "Start collecting latency data"
+collect_latency_data   # "$@"
+
+if [ ! -f "${LOG_ROOT}/metrics.json" ] || [ "$(grep -ce 'metrics' "${LOG_ROOT}/metrics.json")" -lt 1 ];
+then
+  dloge "No metrics collected"
+  exit 2
```
marc-hb, line 220: There is a `skip_test` function.

test-case/latency-baseline.sh
```diff
@@ -0,0 +1,228 @@
...
+  dlogi "Latency data collection"
+  wait ${jackdPID} ${iodelayPID}
+  dlogi "Latency data collection completed."
+}
+
+# TODO: should we set volume to some pre-defined level (parameterized)
+# reset_sof_volume
+# set_alsa_settings
+
+start_test
```
marc-hb, line 207: https://github.com/thesofproject/sof-test/issues/740 - everything in a function.

test-case/latency-baseline.sh
```diff
@@ -0,0 +1,228 @@
...
+    printf("\"min_frames\":%0.3f, \"max_frames\":%0.3f, ", min(latency_frames), max(latency_frames))
+    printf("\"avg_ms\":%0.3f, ", (length(latency_ms) ? sum(latency_ms) / length(latency_ms) : 0))
+    printf("\"min_ms\":%0.3f, \"max_ms\":%0.3f, ", min(latency_ms), max(latency_ms))
+    printf("\"stdev_frames\":%0.6f, \"stdev_ms\":%0.6f }", stddev(latency_frames), stddev(latency_ms))
+    printf("]")
+    fflush()
+  }
+}
+'
+opts_json()
```
marc-hb, line 131: Should this be in case-lib/lib.sh?


```suggestion
opts2json()
```

test-case/latency-baseline.sh
```diff
@@ -0,0 +1,228 @@
...
+    printf("{\"name\":\"roundtrip latency\", ")
+    printf("\"probes\":%d, ", length(latency_frames))
+    printf("\"avg_frames\":%0.3f, ", (length(latency_frames) ? sum(latency_frames) / length(latency_frames) : 0))
+    printf("\"min_frames\":%0.3f, \"max_frames\":%0.3f, ", min(latency_frames), max(latency_frames))
+    printf("\"avg_ms\":%0.3f, ", (length(latency_ms) ? sum(latency_ms) / length(latency_ms) : 0))
+    printf("\"min_ms\":%0.3f, \"max_ms\":%0.3f, ", min(latency_ms), max(latency_ms))
+    printf("\"stdev_frames\":%0.6f, \"stdev_ms\":%0.6f }", stddev(latency_frames), stddev(latency_ms))
+    printf("]")
+    fflush()
+  }
```
marc-hb, line 128: This looks plenty big enough to deserve its separate `jack_metrics.awk` file for:
- no quoting nightmare
- easier testing: can be tested separately
- better git logs
- re-usable in other tests
- syntax highlighting in most editors!

test-case/latency-baseline.sh
```diff
@@ -0,0 +1,228 @@
...
+
+OPT_NAME['v']='verbose'         OPT_DESC['v']='Verbose logging.'
+OPT_HAS_ARG['v']=0              OPT_VAL['v']=0
+
+func_opt_parse_option "$@"
+
+setup_kernel_check_point
+
+# shellcheck disable=SC2016
+AWK_SCRIPT='
```
marc-hb, line 60: This looks PLENTY long enough to deserve its separate `jack_metrics.awk` file for:
- no quoting nightmare
- easier testing: can be tested separately
- better git logs
- syntax highlighting in most editors!

Similarly, I hesitated for too long to extract the new .exp file in commit https://github.com/thesofproject/sof-test/pull/1218/files. I shouldn't have hesitated: it literally changed my file for all the reasons above. It was night and day.

Inlining a different language is OK for less than ~10 lines. If it's longer then forget it.

golowanow: right, better to split already

test-case/latency-baseline.sh
golowanow: moved to `case-lib/opt.sh`

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,156 @@
...
+RUN_PERIOD_MAX="$((run_period + 30))s"
+JACKD_TIMEOUT="$((run_period + 15))s"
+JACKD_OPTIONS="${run_verbose} --realtime --temporary"
+JACKD_BACKEND="alsa"
+JACKD_BACKEND_OPTIONS="-d ${alsa_device} -r ${rate} ${alsa_shorts}"
+WAIT_JACKD="2s"
+IODELAY_TIMEOUT="${run_period}s"
+WAIT_IODELAY="2s"
+
+if [ "$port_playback" == "" ] || [ "$port_capture" == "" ];
```
marc-hb, line 80: Codestyle nit: I think having `then` on the same line is more "standard". Like this:


```suggestion
if [ "$port_playback" == "" ] || [ "$port_capture" == "" ]; then
```

BTW most people don't know this but bash has a built-in formatter. If you do this:

```
some_function()
{
  blabla
}
type some_function
exit 1
```

Then `type some_function` will reformat and print `some_function`. The output uses quoting correctly and can be re-used as it. You may or may not like that default style but it's generally pretty good.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,156 @@
...
+then
+  skip_test "No playback or capture Jack port is specified. Skip the test."
+fi
+
+setup_kernel_check_point
+
+start_test
+
+logger_disabled || func_lib_start_log_collect
```
marc-hb, line 89: I know it hasn't been the usual practice, but I think these should be moved to `main()` too. Once they are moved to main(), you can very quickly comment out the final call to `main()` and source the entire file interactively to test any function interactively. For instance you can invoke `options2json` only.

case-lib/jack_iodelay_metrics.awk
```diff
@@ -0,0 +1,31 @@
...
+  if (length(latency_frames) !=0 && length(latency_ms) != 0) {
+    printf("\"metrics\":[")
+    printf("{\"name\":\"roundtrip latency\", ")
+    printf("\"probes\":%d, ", length(latency_frames))
+    printf("\"avg_frames\":%0.3f, ", (length(latency_frames) ? sum(latency_frames) / length(latency_frames) : 0))
+    printf("\"min_frames\":%0.3f, \"max_frames\":%0.3f, ", min(latency_frames), max(latency_frames))
+    printf("\"avg_ms\":%0.3f, ", (length(latency_ms) ? sum(latency_ms) / length(latency_ms) : 0))
+    printf("\"min_ms\":%0.3f, \"max_ms\":%0.3f, ", min(latency_ms), max(latency_ms))
+    printf("\"stdev_frames\":%0.6f, \"stdev_ms\":%0.6f }", stddev(latency_frames), stddev(latency_ms))
+    printf("]")
```
marc-hb, line 28: That's a lot of `printf`! You can have multi-lines strings in awk, see below. This also saves a lot of double-quotes, of which there are also very many!

```awk
printf("\n\
\"metrics\":[                              \n\
   {\"name\":\"roundtrip latency\",        \n\
    \"probes\":%d,                         \n\
    \"avg_frames\":%0.3f, ...              \n",
    length(latency_frames), ... )

```
I'm not suggesting to put EVERYTHING in a single `printf`, there's likely a middle ground.

test-case/latency-metrics.sh
marc-hb: BTW the `main()` definition should IMHO be near the top, with only the `main "$@"` call at the bottom. This hasn't been the case in sof-test so far but this brand new test is a great opportunity to set a new standard :-)

case-lib/opt.sh
```diff
@@ -26,6 +26,16 @@ add_common_options()
...
 }
 
+# Convert options to a json dictionary.
+options2json()
+{
+  local items_=()
+  for idx_ in "${!OPT_NAME[@]}" ; do
+    items_+=("\"${OPT_NAME[$idx_]}\":\"${OPT_VAL[$idx_]}\"")
+  done
+  echo "$(IFS=',' ; printf "%s" "${items_[*]}")"
```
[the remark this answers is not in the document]
golowanow: with '@' it doesn't work properly

golowanow:
25/May changes:
* detect and report `xruns`;
* some renames for better integration with CI;
* coding style improvements.

golowanow:
26/May changes (should be the last significant change in this PR):
* add ALSA pcm-s parameters to run on SDW.
* add 'trial' mode: repeat latency measurements gradually reducing buffer size (frames) to find the smallest latency achievable.
* set alsa parameters to the same as `check-alsabat.sh` does.
* allow to set a max number of ignored 'xruns' (per each run).
* run mode with loopback set at the JACK to estimate its latency.

case-lib/opt.sh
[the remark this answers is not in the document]
golowanow: I did double check and a comment added: with '@' array expansion the IFS value is not used, but with `*`.

golowanow:
27/May changes: 
after more experiments with 'trial' mode and reproduceable xrun's, I've recognized that these errors are actually from jackd, and it needs its own log parser (in the very early versions of this test jackd and iodelay had the same output stream).

case-lib/opt.sh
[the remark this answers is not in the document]
marc-hb: ok it makes sense to me now, thanks for adding the comment.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+report_start()
+{
+  dlogi "Compose ${RESULT_JSON}"
+  echo -n "{\"options\":{$(options2json)}, " > "${RESULT_JSON}"
+  echo -n  "\"metrics\":[" >> "${RESULT_JSON}"
+}
+
+report_metric()
+{
+  echo -n  "${next_sep}{\"period_frames\":${jackd_period}, \"periods\":${jackd_nperiods}, \"rate\":${rate}, " >> "${RESULT_JSON}"
```
marc-hb, line 215: This time it should be easy to avoid the backslashes, no "challenge"

Also, `%d` will fail if the input is not a number; a good thing.

```suggestion
  printf '%s{"%d":"%d", ...'  "${next_sep}" "period_frames" "${jackd_period}"  >> "${RESULT_JSON}"
```

OK, maybe just ONE backslash newline :-)

```suggestion
  printf '%s{"%d":"%d", ...'  \
     "${next_sep}" "period_frames" "${jackd_period}"  >> "${RESULT_JSON}"
```

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+
+  if [ ! -f "${METRICS_JSON}" ] || [ "$(grep -ce 'metric_name' "${METRICS_JSON}")" -lt 1 ]; then
+    echo -n "\"probes\":0}" >> "${RESULT_JSON}"
```
marc-hb, line 218: See printf above.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+  JACKD_BACKEND_OPTIONS="-P ${pcm_p} -C ${pcm_c} ${JACKD_BACKEND_OPTIONS}"
+fi
+
+if [ "${set_loopback}" == 1 ]; then
+  port_playback="${set_loopback}"
+  port_capture="${set_loopback}"
+fi
+
+if [ -z "${port_playback}" ] || [ -z "${port_capture}" ]; then
+  skip_test "No playback or capture JACK port is specified."
```
marc-hb, line 141: Most/all the above should be in the `main` function (#740), which can/should be moved here.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+run_verbose=$([ "${OPT_VAL['v']}" -eq 1 ] && echo '--verbose' || echo '')
+max_xruns=${OPT_VAL['X']}
+set_loopback=${OPT_VAL['L']}
+jackd_frames=${OPT_VAL['f']}
+jackd_period=${jackd_frames}
+jackd_nperiods=${OPT_VAL['n']}
+trial_mode=${OPT_VAL['t']}
+trial_until=${OPT_VAL['T']}
+
+next_sep=""
```
marc-hb, line 110: This should be close to the single function where it is used.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+  skip_test "No playback or capture JACK port is specified."
+fi
+
+set_alsa()
+{
+  reset_sof_volume
+
+  # If MODEL is defined, set proper gain for the platform
+  if [ -z "$MODEL" ]; then
+    dlogw "NO MODEL is defined. Please define MODEL to run alsa_settings/MODEL.sh"
```
marc-hb, line 150: No `return 1`, no `die`?

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+  else
+    set_alsa_settings "$MODEL"
+  fi
+}
+
+check_jackd_configured()
+{
+  dlogi "Check JACK audio server can be started"
+  dlogc "jackd ${JACKD_OPTIONS} -d ${JACKD_BACKEND} ${JACKD_BACKEND_OPTIONS} -p ${jackd_period}"
+  # shellcheck disable=SC2086
```
marc-hb, line 160: Is it for `${JACKD_OPTIONS}` or `${JACKD_BACKEND_OPTIONS}`, other? If you quote all the other ones then it will become clear.

Can you not make it an array?

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,287 @@
...
+  wait ${jackdPID} ${iodelayPID}
+
+  dlogi "Latency data collection completed for period=${jackd_period} frames."
+}
+
+report_start()
+{
+  dlogi "Compose ${RESULT_JSON}"
+  echo -n "{\"options\":{$(options2json)}, " > "${RESULT_JSON}"
+  echo -n  "\"metrics\":[" >> "${RESULT_JSON}"
```
marc-hb, line 210: See `printf` below, no need for backslashes.

case-lib/jack_iodelay_metrics.awk
```diff
@@ -0,0 +1,30 @@
...
+
+END {
+  if (length(latency_frames) !=0 && length(latency_ms) != 0) {
+    printf("\"metric_name\":\"roundtrip latency\", ")
+    printf("\"probes\":%d, ", length(latency_frames))
+    printf("\"avg_frames\":%0.3f, ", (length(latency_frames) ? sum(latency_frames) / length(latency_frames) : 0))
+    printf("\"min_frames\":%0.3f, \"max_frames\":%0.3f, ", min(latency_frames), max(latency_frames))
+    printf("\"avg_ms\":%0.3f, ", (length(latency_ms) ? sum(latency_ms) / length(latency_ms) : 0))
+    printf("\"min_ms\":%0.3f, \"max_ms\":%0.3f, ", min(latency_ms), max(latency_ms))
+    printf("\"stdev_frames\":%0.6f, \"stdev_ms\":%0.6f", stddev(latency_frames), stddev(latency_ms))
```
marc-hb, line 27: multi-lines string, see https://github.com/thesofproject/sof-test/pull/1272#discussion_r2105104540

marc-hb (commented):
Again, some audio expert should review the logic.

test-case/latency-metrics.sh
golowanow: moved close to these two functions where it is used

case-lib/jack_iodelay_metrics.awk
golowanow: I think it is easier to read and maintain with multiple printf-s having format part and value part closer to each other.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,305 @@
...
+report_metric()
+{
+  printf '%s{"period_frames":%d, "periods":%d, "rate":%d, ' \
+    "${json_next_sep}" "${jackd_period}" "${jackd_nperiods}" "${rate}" >> "${RESULT_JSON}"
+
+  if [ ! -f "${METRICS_JSON}" ] || [ "$(grep -ce 'metric_name' "${METRICS_JSON}")" -lt 1 ]; then
+    printf '"probes":0, "xruns":0}' >> "${RESULT_JSON}"
+    [ -f "${METRICS_JSON}" ] && rm "${METRICS_JSON}"
+    [ -f "${EVENTS_JSON}" ] && rm "${EVENTS_JSON}"
+    [ "${trial_mode}" -eq 0 ] && skip_test "No latency metrics collected"
```
marc-hb, line 238: Nit:
```suggestion
    [ "${trial_mode}" -ne 0 ] || skip_test "No latency metrics collected"
```
See above and #312

marc-hb (commented):
LGTM, some audio expert must review.

case-lib/jack_iodelay_metrics.awk
marc-hb: It does not have to be a single `printf` for everything. Two or maybe three would IMHO be the sweet spot.

kv2019i (approved):
Thanks @golowanow , looks great! Good comments in code, this will be good basis to build on other tests.

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,305 @@
...
+  TIMEOUT_JACKD_RUN=("${TIMEOUT_RUN[@]}" "${JACKD_TIMEOUT}" "${JACKD_RUN[@]}")
+}
+
+check_jackd_configured()
+{
+  dlogi "Check JACK audio server can be started"
+
+  compose_commands
+
+  dlogc "${TIMEOUT_JACKD_RUN[@]}"
```
marc-hb, line 175: Does dlogc support multiple arguments? If it does, that's only by chance...

```suggestion
  dlogc "${TIMEOUT_JACKD_RUN[*]}"
```

test-case/latency-metrics.sh
```diff
@@ -0,0 +1,305 @@
...
+  "${TIMEOUT_JACKD_RUN[@]}" & jackdPID=$!
+
+  sleep ${WAIT_JACKD}
+
+  dlogc "jack_lsp"
```
marc-hb, line 180:
```suggestion
  dlogc "jack_lsp -AclLpt"
```
Or
```suggestion
 (set -x; jack_lsp -AClLpt)
```

test-case/latency-metrics.sh
marc-hb: Merge first and fix that later if needed.

test-case/latency-metrics.sh
golowanow: `dlogc` has its timestamp and easier to find in the log.

Issue #1038, opened 2023-05-18
[FEATURE] Support reboot cycles/stress-tests


plbossart:
While debugging on MTL_RVP_HDA, I encountered multiple cases of stuck reboots, which can only be solved with a power-cycle.
It's my understanding that we do not test reboot on our CI and daily test devices, so we don't know what we don't know.

I am not sure if this can be implemented with sof-test 'run-all-test', this may require an external device to send the reboot commands and check what the state is. Or this may be a script in sof-test that executes on a host device, with commands sent over ssh to the DUT?

Comments and ideas welcome @keqiaozhang @marc-hb @fredoh9 @ranj063

marc-hb:
Testing a device that does up and down is not easy, it requires relatively smart outside control and sof-test is not that because sof-test runs on the device.

In fact we already had some discussions about this in internal issue 272. They were focused on suspend/resume which is easier but not 100% reliable either. So I think the next step is to extract from 272 information and pointers relevant to this.

plbossart:
It's not easy, that's a given. But I am really worried about the cost of doing nothing. probably time to think about the scripts and whether the premise that they all run locally is correct. we could have parts of sof-test that are dedicated to remote control, instead of having this in some random CI hidden repo.

PR #1279, opened 2025-06-16
alsa: Restore test settings after each test

LukaszMrugala:
Currently, some tests change ALSABAT settings during their execution, but do not restore them to their prior state.

This causes tests that occur later in the execution order to intermittently fail.

To remedy this, all tests that use ALSA now use a common `save_alsa_state` function, which shall save current ALSA state to a file and then restore that state on the bash script EXIT signal.


sofci:

test-case/check-alsabat.sh
```diff
@@ -74,6 +74,7 @@ sigmak=${OPT_VAL['k']}
 frames=${OPT_VAL['n']}
 
 start_test
+save_alsa_state "${0##*/}".state
```
marc-hb, line 77: Rather than have every caller append `.state`, you can let `save_alsa_state` do it.

case-lib/lib.sh
```diff
@@ -1187,3 +1187,42 @@ perf_analyze()
...
+
+# Save the machine state to a file.
+# Should be used at the start of
+# every test touching ALSA.
+# Coupled with restore_machine_state on an exit signal.
+# param1: file name
+save_alsa_state()
+{
+    dlogi "save_alsa_state called with ${1}"
+    trap "restore_alsa_state '${1}'" EXIT
```
marc-hb, line 1222: There is already an EXIT handler: `func_exit_handler()`. You can't have both; how did you test this?

You can likely have `func_exit_handler()` invoke `restore_alsa_state` when the conditions are right.

marc-hb:
Isn't this a followup to #1270?

test-case/check-alsabat.sh
LukaszMrugala: I wanted to separate the concerns of handling ALSA and handling of the arbitrary filename. This approach was used to ensure no collisions between scripts.

case-lib/lib.sh
```diff
@@ -1187,3 +1187,39 @@ perf_analyze()
...
+save_alsa_state()
+{
+    dlogi "save_alsa_state called with ${1}"
+    # Prepend the restore_alsa_state to the preexisting EXIT trap.
+    cur_exit_trap_code=$(trap -p EXIT | awk -F\' '{print $2}')
+    if [[ -z "${cur_exit_trap_code// }" ]]; then
+        trap "restore_alsa_state ${1}" EXIT
+    else
+        trap "restore_alsa_state ${1}; ${cur_exit_trap_code}" EXIT
+    fi
```
LukaszMrugala, line 1223: This approach works perfectly in my local tests, but generates TIMEOUTs in the CI in runs that previously were FAILs.
I didn't want to dump the ALSA traps inside the `func_exit_handler`, as it's a generic solution for every test and not every test requires ALSA setup and teardown.
I welcome comments as to why this would cause `func_exit_handler` to not run.

LukaszMrugala:
> Isn't this a followup to #1270?

It's connected, but it's not a follow-up per se. It aims to resolve the specific problem of ALSA tests changing machine's ALSA settings and messing with the following test executions. Note that it does nothing to e.g. prevent people from connecting to the machines and messing around in settings on an ad hoc basis.

golowanow:
> Currently, some tests change ALSABAT settings during their execution, but do not restore them to their prior state.
> This causes tests that occur later in the execution order to intermittently fail.

Do we really need to strive for ALSA state restore _on_exit_ ?

Isn't it easier and just enough for our CI's purposes to restore the default state _before_ at each test case where it is needed ?

This way it solves both these cases:

> .. the specific problem of ALSA tests changing machine's ALSA settings and messing with the following test executions. 

> .. people from connecting to the machines and messing around in settings on an ad hoc basis.

case-lib/lib.sh
```diff
@@ -1187,3 +1187,39 @@ perf_analyze()
...
+restore_alsa_state()
+{
+    status=$?
+    dlogi "restore_alsa_state called with ${1}"
+    if [ -f "$1" ]; then
+        dlogi "restore_alsa_state found relevant state file."
+        alsactl restore --file /var/tmp/"$1" --pedantic --no-ucm --no-init-fallback || dlogi "alsactl state restoration failed!"
+        rm /var/tmp/"$1" || dlogi "Old state file removal failed!"
+    fi
+    return "$status"
```
marc-hb, line 1206: I'm not a fan of this... for instance it might print a misleading stack trace if the previous command failed.

case-lib/lib.sh
marc-hb: > I didn't want to dump the ALSA traps inside the func_exit_handler, as it's a generic solution for every test and not every test requires ALSA setup and teardown.

`func_exit_handler()` easily can be "smart" and run `restore_alsa_state()` only when `save_alsa_state()` was run before; this does not have to be "dumped".

> I welcome comments as to why this would cause func_exit_handler to not run.

I think my previous "it does not work" comment was on some earlier and quite different version.

But even now, I don't like "stacking" trap functions like this because it scatters the exit handling code across multiple places and it becomes more difficult to see what it does in any given test and situation and in which order. Who says that `restore_alsa_state` should run first on EXIT? Maybe it shouldn't be first...

marc-hb (commented):
> Yes, that's the ideal solution, given the assumption that we have a perfect description of the "default state" every time we run a test for every machine we run a test on.

That "assumption" is not really optional: if test results depend on some out-of-control configuration, then there is a test setup bug that must be fixed somewhere (BTW see https://github.com/thesofproject/sof-test/labels/area%3AAnsible and the same label internally)

Test cleaning up after themselves does not hurt but it's less important than comprehensive test setup. Also, cleanup must be easy to turn off when the test is failing. Anything that runs after a failure makes debugging harder - whether that's in interactive use or when staring at logs from automated runs.

case-lib/lib.sh
LukaszMrugala: No TIMEOUT problems anymore after I've moved the called inside the `func_exit_handler`.

case-lib/hijack.sh
```diff
@@ -185,6 +185,11 @@ function func_exit_handler()
         fi
     }
 
+    # Restore ALSA settings after execution if state file exists
+    if [ -f /var/tmp/"${SCRIPT_NAME##*/}".state ]; then
```
marc-hb, line 189: I've been wanting to reduce SCRIPT_NAME usage in #546 for   a few years :-)

marc-hb:
```suggestion
    if [ -f /tmp/"${SCRIPT_NAME##*/}".state ]; then
```

While different distros do things differently, `/tmp` is in general on `tmpfs` ("RAM disk") whereas `/var/tmp` is persistent across reboots - which I'm pretty sure we do not WANT.

BTW did you consider using the process ID in the name? I mean `$$`

There is also https://en.wikipedia.org/wiki/TMPDIR, it's complicated :-(

marc-hb: Here's another idea:
```
   SOF_TEST_ALSA_STATE=/tmp/"${SCRIPT_NAME##*/}".$$.state
```

... and then at restore time you can `test -n SOF_TEST_ALSA_STATE`? So the filename is defined in only one place.

case-lib/lib.sh
marc-hb: I just remembered that the Jenkins runner waits forever for PASS or FAIL or SKIP to be printed, see commit 4395b600e8477d3 and others. I'm afraid Jenkins does not care about the test process exiting and does not care about the exit value either. That's likely a (very old) design mistake.

This would tend to prove that the earlier version did not restore `func_exit_handler()` properly.

BTW can you set a trap from a trap? Dunno.

This discussion is obsolete and just for completeness, sorry.

case-lib/hijack.sh
LukaszMrugala: I chose `/var/tmp/` at first because I remembered that the DUT is rebooted in our CI during testing, but you're right - that reboot occurs before running the test case `.sh` file, so `/tmp` should be used instead.
I've applied your last suggestion and squashed the commits.
I've taken a look at the failed shellcheck and it all seemed like problems that already existed in the codebase, rather than something newly introduced.
PR checked in CI run No. 54699

marc-hb: > but you're right - that reboot occurs before running the test case .sh file, so /tmp should be used instead.

There is no reboot test AFAIK right now:
- https://github.com/thesofproject/sof-test/issues/1038

PS: there are some suspend/resume tests but they are not very reliable (see https://github.com/thesofproject/sof-test/discussions/1017 and internal sof-framework/issues/408).

LukaszMrugala:
test this please

redzynix:
SOFCI TEST

PR #1278, opened 2025-06-16
test-case: add new test case for check mic privacy

arikgreen:
Add new test case to check hardware mic privacy mode enablement feature.
For switching mic privacy state is using a USB relay.
https://github.com/darrylb123/usbrelay

PR #1280, opened 2025-07-07
test-case: add test jack detection playback

arikgreen:
Add a new test case to test jack detection during playback.
Verify jack detection functionality by simulating plugging and unplugging of audio jack using a USB relay switch.
The unplugging and plugging of the jack will be during playback operations to ensure the system can handle jack detection events correctly.
The test will check if the jack detection status is updated correctly when the jack is plugged in and unplugged.

For unplug/plug headset jack is using a USB relay. https://github.com/darrylb123/usbrelay

### DIFF
--- a/case-lib/relay.sh
+++ b/case-lib/relay.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2021-2025 Intel Corporation. All rights reserved.
+
+# needs usbrelay package: https://github.com/darrylb123/usbrelay
+# param1: --debug | switch name
+# param2: switch state
+usbrelay_switch()
+{
+    if [[ "$1" == "--debug" ]]; then
+        dlogi "Debug mode: Current status of all relays:"
+        usbrelay || {
+            die "Failed to get usbrelay status.
+            The usbrelay hw module is not responding or no relays detected.
+            Check hardware connection."
+        }
+    fi
+
+    # Declare a constant for the relay settle time
+    local USBRELAY_SETTLE_TIME=0.5
+
+    local switch_name=$1
+    local state=$2
+
+    dlogi "Setting usbrelay switch $switch_name to $state."
+    usbrelay "$switch_name=$state" --quiet || {
+        die "Failed to set usbrelay switch $switch_name to $state.
+        The usbrelay hw module is not responding or no relays detected.
+        Check hardware connection."
+    }
+
+    # wait for the switch to settle
+    sleep "$USBRELAY_SETTLE_TIME"
+
+    # Display current state of the switch
+    current_state=$(usbrelay | awk -F= -v name="$switch_name" '$1 == name { print $2 }')
+
+    # Check if current_state is equal to the requested state
+    [[ "$current_state" == "$state" ]] || {
+        die "usbrelay switch $switch_name failed to set to $state (current: $current_state)"
+    }
+
+    case "$current_state" in
+        '1') dlogi "Current state of $switch_name is: on";;
+        '0') dlogi "Current state of $switch_name is: off";;
+        *) die "Invalid state for $switch_name: $current_state";;
+    esac
+}


test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,153 @@
...
+# 3. Run aplay/arecord command to play/record audio.
+# 4. Check the jack detection status via amixer. The status should indicate **on**.
+# 5. Set the USB relay switch to state on (1), simulate unplugging the headset from the jack.
+# 6. Check the jack detection status via amixer. The status should indicate **off**.
+# 7. Check dmesg for any unexpected errors.
+# 
+# Repeat for both headphone and headset jacks.
+# Repeat for all pipelines.
+
+source "$(dirname "${BASH_SOURCE[0]}")"/../case-lib/lib.sh
```
marc-hb, line 36: Missing this line, see other scripts.
```bash
shellcheck source=...
```

case-lib/lib.sh
```diff
@@ -1187,3 +1187,39 @@ perf_analyze()
...
+# test-mic_privacy.sh needs to control mic privacy settings (on/off)
+# needs usbrelay package: https://github.com/darrylb123/usbrelay
+# param1: switch name
+# param2: switch state
+usbrelay_switch()
+{
+    local switch_name=$1
+    local state=$2
+    dlogi "Setting usbrelay switch $switch_name to $state."
+    usbrelay "$switch_name=$state" >/dev/null 2>&1 || {
```
marc-hb, line 1199: Discarding stderr is still very wrong:

https://github.com/thesofproject/sof-test/pull/1278#discussion_r2202016464

marc-hb (commented):
I didn't look at the code much but there is seems to be come duplication with #1278?

case-lib/lib.sh
arikgreen: yes, you right Marc I missed it

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,164 @@
...
+{
+    # Check the state of the switch using amixer.
+    # The switch name is passed as the first argument, and the expected state (on/off)
+    # is passed as the second argument.
+    # Returns 0 if the state matches, 1 otherwise.
+    local switch_name="$1"
+    local expected_switch_state="$2"
+    local switch_state
+
+    switch_state=$(echo -e $(amixer -c 0 contents | grep -i "$switch_name .* *jack" -A 2) | sed -n '1s/^.*values=//p')
```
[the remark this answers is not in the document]
marc-hb: What does the `echo -e` achieve?

BTW: https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

But better use none if none needed in the first place.

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,164 @@
...
+
+    # check if usbrelay tool is installed
+    command -v usbrelay || {
+        skip_test "usbrelay command not found. Please install usbrelay to control the mic privacy switch."
+    }
+
+    dlogi "Reset - plug jack audio"
+    usbrelay_switch "$relay" 0
+
+    for round in $(seq 1 "$round_cnt")
```
marc-hb, line 106: You can use `while` or `shellcheck disable=SC2034`

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,164 @@
...
+                die "Plug headset jack failed."
+            }
+
+            check_control_switch_state "headphone" "on" || {
+                die "Plug headphone jack failed."
+            }
+        done
+    done
+
+    sof-kernel-log-check.sh "$KERNEL_CHECKPOINT"
```
marc-hb, line 159: Maybe check after each iteration to "fail fast"?

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,164 @@
...
+        skip_test "usbrelay command not found. Please install usbrelay to control the mic privacy switch."
+    }
+
+    dlogi "Reset - plug jack audio"
+    usbrelay_switch "$relay" 0
+
+    for round in $(seq 1 "$round_cnt")
+    do
+        for idx in $(seq 0 $((PIPELINE_COUNT - 1)))
+        do
```
marc-hb, line 109: It would be more readable to move this to some new `test_one_idx()`  function (better name wanted), that would avoid a lot of indentation and a lot of distance between the start and end of the for loops.

test-case/test-jack-detection-playback-capture.sh
marc-hb: Do you never use shellcheck?

case-lib/lib.sh
```diff
@@ -791,7 +791,7 @@ extract_format_number() {
     else
         die "Error: Unknown format: %s\n"
     fi
-}  
+}
```
marc-hb, line 794: To keep git clean, don't mix pure whitespace changes with code changes; especially not when it's far away in a completely different file.

case-lib/relay.sh
```diff
@@ -0,0 +1,42 @@
...
+usbrelay_switch()
+{
+    # Declare a constant for the relay settle time
+    USBRELAY_SETTLE_TIME=0.5
+
+    local switch_name=$1
+    local state=$2
+
+    dlogi "Setting usbrelay switch $switch_name to $state."
+    usbrelay "$switch_name=$state" > /dev/null || {
```
marc-hb, line 19: How verbose is `usbrelay`, can you share an example of what is getting discarded here?

This is test code; as long as their size stays reasonable, logs files should have as much information as possible. Imagine trying to make sense of test logs capture on a system you don't have access to.

BTW `usbrelay` has a `--quiet` option, why not just use it instead of /dev/null ? Only if needed.

case-lib/relay.sh
```diff
@@ -0,0 +1,42 @@
...
+        die "Failed to set usbrelay switch $switch_name to $state.
+        The usbrelay hw module is not responding or no relays detected.
+        Check hardware connection."
+    }
+
+    # wait for the switch to settle
+    sleep "$USBRELAY_SETTLE_TIME"
+
+    # Display current state of the switch
+    current_state=$(usbrelay | grep "$switch_name" | awk -F= '{print $2}')
```
marc-hb, line 30: `awk` has `grep` built-in.
```suggestion
    current_state=$(usbrelay | grep  awk -F= "/$switch_name/ {print \$2}")
```

Or, if you prefer
```suggestion
    current_state=$(usbrelay | awk -F= "/$switch_name/"' {print $2}')
```

test-case/test-jack-detection-playback-capture.sh
marc-hb: Very strange: in https://github.com/thesofproject/sof-test/actions/runs/16506744373/job/46679195194 shellcheck complained like this:

```
In test-case/test-jack-detection-playback-capture.sh line 108:
    for round in $(seq 1 "$round_cnt")
    ^-^ SC2034 (warning): round appears unused. Verify use (or export if used externally).
```

But now it's gone in newer https://github.com/thesofproject/sof-test/actions/runs/16507901117/job/46683162054?pr=1280

(The dubious `echo` on line 79 is flagged in both)

case-lib/relay.sh
arikgreen:
```
ubuntu@gk-ptlh-rvp-05:~$ usbrelay HURTM_1=1
HURTM_1=0
HURTM_2=0
```
the verbose is unnecessary for us and it is confused because display previous state of the relay.

arikgreen:
Thank you @marc-hb for your review, it was very helpful. I hope the code now meets your requirements.

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,206 @@
...
+    check_control_switch_state "headset" "on" || {
+        die "Plug headset jack failed."
+    }
+
+    check_control_switch_state "headphone" "on" || {
+        die "Plug headphone jack failed."
+    }
+
+    kill -9 $pid_playback > /dev/null 2>&1
+    wait $pid_playback 2>/dev/null || true
```
marc-hb, line 158: I hope we already have some code for this, maybe in `case-lib/`

(every script tries to re-invent this)

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    dlogi "Plug jack audio."
+    usbrelay_switch "$relay" 0
+
+    # Wait for a short period to allow the system to detect the plug event
+    sleep $TWO_SECONDS
+
+    # check if the aplay process is still running after unplugging the jack
+    ps -p "$pid_playback" > /dev/null || {
```
marc-hb, line 144: I hope we already have some code for this, maybe in case-lib/

marc-hb (commented):
> it was very helpful. 

My pleasure.

> I hope the code now meets your requirements.

I don't have any requirement; I just try to spot common shell mistakes (after shellcheck caught most of them).

The shell is quirky, everyone keeps making the same mistakes.

case-lib/relay.sh
arikgreen: the current state is verified in this line
[[ "$current_state" == "$state" ]]

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,209 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright(c) 2025 Intel Corporation. All rights reserved.
+
+##
+## Preconditions
+# 1. Runtime PM status is on.
```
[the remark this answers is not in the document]
arikgreen: PM is power management but this point was required in previous version. Should be removed.

test-case/test-jack-detection-playback-capture.sh
```diff
@@ -0,0 +1,209 @@
...
+# 2. aplay (playback) is running.
+# 3. USB relay switch is available and configured.
+#    Jack detection header should be connected to the USB relay switch
+#    to the port HURTM_2 (NC) connector.
+
+## Test Description
+# Verify jack detection functionality by simulating plugging and unplugging
+# of audio jack using a USB relay switch.
+# The unplugging and plugging of the jack will be during playback operations
```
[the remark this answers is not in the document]
arikgreen: this scenario will test in a different test - test_jack_detection_during_suspend

test-case/test-jack-detection-playback-capture.sh
arikgreen: IMO ps -p is more explicit about checking for a process

redzynix (approved):

marc-hb:
This has the same `relay.sh` file as #1278 but it's not added in a separate commit.

KamilxPaszkiet (approved):

PR #1281, opened 2025-07-07
[SKIP SOF-TEST] test-case: Add ALSA conformance tests

golowanow:
Add ALSA conformance tests from ChromeOS Audio Test package.

The new test case `check-alsa-conformance.sh` executes `alsa_conformnance_test` and compose its results into a JSON file.

### DIFF
--- a/test-case/check-alsa-conformance.sh
+++ b/test-case/check-alsa-conformance.sh
@@ -1,0 +1,372 @@
+#!/bin/bash
+
+# Copyright(c) 2025 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+
+##
+## Case Name: Execute ALSA conformance tests.
+##
+## Preconditions:
+##    - ChromeOS Audio Test package is installed
+##      https://chromium.googlesource.com/chromiumos/platform/audiotest
+##
+## Description:
+##    Run `alsa_conformance_test.py` for the playback devices
+##    and the capture devices with the test suite paramenters given.
+##    Compose resulting JSON reports.
+##
+##    To select PCMs use either -d, or -p with or without -c parameters.
+##    If a PCM id has no device id (e.g. 'hw:sofnocodec' instead of 'hw:sofnocodec,0')
+##    then all devices on that card will be selected for the test run.
+##    To select all available PCMs omit any -d, -p, -c parameters.
+##
+##    Pass multiple values of the test parameters -d, -p, -c, -r, -F enclosing them
+##    in quotes, eg. `-F 'U8 S16_LE'` or `-p 'sofnocodec,1 sofnocodec,2'`
+##
+## Case steps:
+##    0. Set ALSA parameters.
+##    1. For each PCM selected:
+##    1.1 Try to start `alsa_conformance_test` in device info mode.
+##    1.2 Start `alsa conformance_test.py` for playback devices.
+##    1.3 Start `alsa conformance_test.py` for capture devices.
+##    2. Compose the resulting JSON report.
+##
+## Expect result:
+##    ALSA conformance results collected and saved in `test_result.json` file.
+##    Exit status 0.
+##    In case of errors this test tries to continue and have its JSON report correctly structured.
+##
+
+TESTDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+TESTLIB="${TESTDIR}/case-lib"
+
+# shellcheck source=case-lib/lib.sh
+source "${TESTLIB}/lib.sh"
+
+OPT_NAME['d']='device'          OPT_DESC['d']='ALSA pcm device for playback and capture. Example: hw:0'
+OPT_HAS_ARG['d']=1              OPT_VAL['d']=''
+
+OPT_NAME['p']='pcm_p'           OPT_DESC['p']='ALSA pcm device for playback only. Example: hw:soundwire,0'
+OPT_HAS_ARG['p']=1              OPT_VAL['p']=''
+
+OPT_NAME['c']='pcm_c'           OPT_DESC['c']='ALSA pcm device for capture only. Example: hw:soundwire,1'
+OPT_HAS_ARG['c']=1              OPT_VAL['c']=''
+
+OPT_NAME['r']='rates'           OPT_DESC['r']='Sample ratis to try. Default: check all available rates.'
+OPT_HAS_ARG['r']=1              OPT_VAL['r']=''
+
+OPT_NAME['F']='formats'         OPT_DESC['F']='Data formats to try. Default: check all available formats.'
+OPT_HAS_ARG['F']=1              OPT_VAL['F']=''
+
+OPT_NAME['s']='sof-logger'      OPT_DESC['s']="Open sof-logger trace the data will store at $LOG_ROOT"
+OPT_HAS_ARG['s']=0              OPT_VAL['s']=1
+
+OPT_NAME['v']='verbose'         OPT_DESC['v']='Verbose logging.'
+OPT_HAS_ARG['v']=0              OPT_VAL['v']=0
+
+OPT_NAME['E']='rate-diff'       OPT_DESC['E']="ALSA conformance --rate-criteria-diff-pct (difference, %)."
+OPT_HAS_ARG['E']=1              OPT_VAL['E']=''
+
+OPT_NAME['e']='rate-err'        OPT_DESC['e']="ALSA conformance --rate-err-criteria (max rate error)."
+OPT_HAS_ARG['e']=1              OPT_VAL['e']=''
+
+OPT_NAME['a']='avail-delay'     OPT_DESC['a']="ALSA conformance --avail-delay"
+OPT_HAS_ARG['a']=0              OPT_VAL['a']=0
+
+OPT_NAME['T']='test-suites'     OPT_DESC['T']="ALSA conformance --test-suites (Default: all)."
+OPT_HAS_ARG['T']=1              OPT_VAL['T']=''
+
+OPT_NAME['t']='timeout'         OPT_DESC['t']="ALSA conformance --timeout (Default: none)."
+OPT_HAS_ARG['t']=1              OPT_VAL['t']=''
+
+OPT_NAME['A']='allow-channels'  OPT_DESC['A']="ALSA conformance --allow-channels (Default: all)."
+OPT_HAS_ARG['A']=1              OPT_VAL['A']=''
+
+OPT_NAME['S']='skip-channels'   OPT_DESC['S']="ALSA conformance --skip-channels (Default: none skipped)."
+OPT_HAS_ARG['S']=1              OPT_VAL['S']=''
+
+func_opt_parse_option "$@"
+
+# Options for the ALSA conformance test script call
+CMD_OPTS=()
+
+# Recompose OPT_VAL[$1] option as ALSA test script option $2
+add_cmd_option()
+{
+  local opt_val="${OPT_VAL[$1]}"
+  local prefix=$2
+
+  if [ -n "${opt_val}" ]; then
+    # Split list parameters to separate values
+    opt_val=("${opt_val//[ ,]/ }")
+    # shellcheck disable=SC2206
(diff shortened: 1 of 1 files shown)


ujfalusi (approved):

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,251 @@
...
+alsa_device=${OPT_VAL['d']}
+pcm_p=${OPT_VAL['p']}
+pcm_c=${OPT_VAL['c']}
+rates=$([ -z "${OPT_VAL['r']}" ] && echo '' || echo "--allow-rates ${OPT_VAL['r']}")
+formats=$([ -z "${OPT_VAL['F']}" ] && echo '' || echo "--allow-formats ${OPT_VAL['F']}")
+run_verbose=$([ "${OPT_VAL['v']}" -eq 1 ] && echo '--log-file /dev/stdout' || echo '')
+rate_diff=$([ -z "${OPT_VAL['E']}" ] && echo '' || echo "--rate-criteria-diff-pct ${OPT_VAL['E']}")
+rate_err=$([ -z "${OPT_VAL['e']}" ] && echo '' || echo "--rate-err-criteria ${OPT_VAL['e']}")
+avail_delay=$([ "${OPT_VAL['a']}" -eq 1 ] && echo '--avail-delay' || echo '')
+test_suites=$([ -z "${OPT_VAL['T']}" ] && echo '' || echo "--test-suites ${OPT_VAL['T']}")
```
marc-hb, line 80: Any reason why this is not in a function?

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,251 @@
...
+OPT_NAME['T']='test-suites'     OPT_DESC['T']="ALSA conformance --test-suites (Default: all)."
+OPT_HAS_ARG['T']=1              OPT_VAL['T']=''
+
+func_opt_parse_option "$@"
+
+alsa_device=${OPT_VAL['d']}
+pcm_p=${OPT_VAL['p']}
+pcm_c=${OPT_VAL['c']}
+rates=$([ -z "${OPT_VAL['r']}" ] && echo '' || echo "--allow-rates ${OPT_VAL['r']}")
+formats=$([ -z "${OPT_VAL['F']}" ] && echo '' || echo "--allow-formats ${OPT_VAL['F']}")
```
marc-hb, line 75: The `echo ''` is not needed.
```bash
rates=$([ -z "${OPT_VAL['r']}" ] || echo "--allow-rates ${OPT_VAL['r']}")
```

Also, this repetition is calling for some new `prefix_if_not_empty()` and `prefix_if_1()` functions. I know they would be useful outside this new script.

```bash
formats=$(prefix_if_not_empty --allow-formats "${OPT_VAL['F']}")
```

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,251 @@
...
+run_verbose=$([ "${OPT_VAL['v']}" -eq 1 ] && echo '--log-file /dev/stdout' || echo '')
```
marc-hb, line 76: `echo ''` not needed here either.

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,251 @@
...
+    set_alsa_settings "$MODEL"
+  fi
+}
+
+alsa_conformance_device_info()
+{
+  local run_cmd="${ALSA_CONFORMANCE_TEST} ${PLAYBACK_OPTIONS} ${CAPTURE_OPTIONS} --dev_info_only"
+  dlogc "${run_cmd}"
+  local rc
+  bash -c "${run_cmd}" ; rc=$?
```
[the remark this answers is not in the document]
golowanow: @marc-hb, thank you for your comments, I think all are addressed, except this one.

I'd keep it this way (calling from another bash) for these reasons:
1) in case we need to call the test with timeout.
2) to compose OPTIONS with multiple values (-T, -r, -F)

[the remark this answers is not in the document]
marc-hb: 1. I don't see why `timeout ... "${run_cmd[@]}"` would not work
2. "composing" sounds even more risky = needing arrays even more
3. Whatever the reasons are, they definitely deserve a comment in the source.

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,255 @@
...
+OPT_HAS_ARG['T']=1              OPT_VAL['T']=''
+
+func_opt_parse_option "$@"
+
+init_globals()
+{
+  alsa_device=${OPT_VAL['d']}
+  pcm_p=${OPT_VAL['p']}
+  pcm_c=${OPT_VAL['c']}
+  rates=$([ "${OPT_VAL['r']}" ] && echo "--allow-rates ${OPT_VAL['r']}")
```
marc-hb, line 76: I'm really not a fan of testing for whitespace like this because there is always a risk of the variable starting with a dash or something. `[ -n ...` and `[ -z ...` are your friends.

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,255 @@
...
+alsa_conformance_test_capture()
+{
+  if [ "${CAPTURE_OPTIONS}" ]; then
+    local run_cmd="PATH=${ALSA_CONFORMANCE_PATH}:${PATH} "
+    run_cmd+="${ALSA_CONFORMANCE_TEST}.py ${ALSA_CONFORMANCE_OPTIONS[*]} ${CAPTURE_OPTIONS} "
+    run_cmd+="--json-file ${AUDIOTEST_OUT}_capture.json ${run_verbose}"
+    dlogc "${run_cmd}"
+    local rc=1
+    bash -c "${run_cmd}" ; rc=$?
+    [[ "${rc}" -ne 0 ]] && dloge "Failed CAPTURE TESTS, rc=${rc}"
```
marc-hb, line 169:
```suggestion
    bash -c "${run_cmd}" || dloge "Failed CAPTURE TESTS, rc=$?"
```
BTW shouldn't this fail the test? `dloge` is not enough AFAIK.

test-case/check-alsa-conformance.sh
golowanow: good point

golowanow:
13/Jul/2025  PCMs selection for the test run is extended - now it is possible to
* choose several PCMs;
* run for all devices on a card;
* run for all PCMs on the DUT.

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,359 @@
...
+  dlogi "pcm_c=(${pcm_c[*]})"
+
+  local p_dev_expanded=()
+  PLAYBACK_DEVICES=()
+
+  for p_dev in "${pcm_p[@]}"
+  do
+    mapfile -t p_dev_expanded < <(get_card_devices 'playback' "${p_dev}")
+    # shellcheck disable=SC2206
+    PLAYBACK_DEVICES+=( ${p_dev_expanded[@]} )
```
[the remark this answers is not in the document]
golowanow: right, here it is not needed

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,357 @@
...
+
+  local alsa_list=''
+  local res_devs=("${arg_pcm}")
+
+  if [ "${mode}" == 'playback' ]; then
+    alsa_list=('aplay' '-l')
+  elif [ "${mode}" == 'capture' ]; then
+    alsa_list=('arecord' '-l')
+  else
+    return
```
marc-hb, line 150: Not a failure? A `dlogi` maybe?

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,357 @@
...
+  fi
+
+  if [ -n "${arg_pcm}" ]; then
+    # check is only card name is given or exact device
+    if [ "${arg_pcm}" == "${arg_pcm##*,}" ]; then
+      # strip 'hw:' prefix
+      arg_pcm="${arg_pcm#*:}"
+      # shellcheck disable=SC2016
+      local gawk_script='match($0, /^card [0-9]+: ('"${arg_pcm}"') .+ device ([0-9]+): /, arr) { print "hw:" arr[1] "," arr[2] }'
+      mapfile -t res_devs < <( "${alsa_list[@]}" | gawk "${gawk_script}" )
```
marc-hb, line 160: I feel like we may have this sort of parsing already somewhere in `tools/` - or that we keep re-inventing it in every script... it's OK, it's short enough but every time someone (re-)writes this sort of screen-scraping, a kitten dies!

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,357 @@
...
+  [ "${mode}" == 'capture' ] && opt=("-C" "${device}")
+  [ -z "${opt[*]}" ] && die "No ALSA PCM parameter."
+
+  local run_prefix=("export" "PATH=${ALSA_CONFORMANCE_PATH}:${PATH}")
+  local run_cmd=()
+  run_cmd+=("${ALSA_CONFORMANCE_TEST}.py" "${CMD_OPTS[@]}" "${opt[@]}")
+  run_cmd+=("--json-file" "${AUDIOTEST_OUT}_${mode}.json")
+  dlogc "${run_cmd[@]}"
+  local rc=0
+  "${run_prefix[@]}" && "${run_cmd[@]}" || rc=$?
```
marc-hb, line 253: I think this is correct but it's really tricky; not for the faint of heart.

Too many people think `A && B || CC` is the same as `if A; then B; else C; fi` and that they can save typing 15 characters that way. But it's not the same! (#312)
```
# Generally NOT what you want!
true && false || echo 'else_is_also_run!!'
```

HOWEVER, I think this example is correct but it is in any case way too subtle and bad role-modelling IMHO.

There are (at least) two better ways:

```suggestion
  "${run_prefix[@]}" # export can hardly fail
   "${run_cmd[@]}" || rc=$?
```

Maybe simpler and better IMHO: you don't need `run_prefix` at all if  the current process does not need to be affected:
```suggestion
"PATH=${ALSA_CONFORMANCE_PATH}:${PATH}" "${run_cmd[@]}" || rc=?
```

marc-hb (commented):
Sorry I keep finding minor problems but that's because the code keep changing :-)

test-case/check-alsa-conformance.sh
golowanow: There is similarity in `parse_audio_device()` and related topology-parsing functions, but here, for the test case based on an external tool, I've decided to use output from alsa-utils,  also to simplify multiple selections.
We can move this piece of code into `tools/` later, If  it becomes useful for other test cases and save kittens.

test-case/check-alsa-conformance.sh
golowanow: I've tried it with different combinations. And the prefix part is to make clear/shorter separation of what is the preparational part (eg. other env. variables).

test-case/check-alsa-conformance.sh
```diff
@@ -0,0 +1,357 @@
...
+  if [ -n "${arg_pcm}" ]; then
+    # check is only card name is given or exact device
+    if [ "${arg_pcm}" == "${arg_pcm##*,}" ]; then
+      # strip 'hw:' prefix
+      arg_pcm="${arg_pcm#*:}"
+      # shellcheck disable=SC2016
+      local gawk_script='match($0, /^card [0-9]+: ('"${arg_pcm}"') .+ device ([0-9]+): /, arr) { print "hw:" arr[1] "," arr[2] }'
+      mapfile -t res_devs < <( "${alsa_list[@]}" | gawk "${gawk_script}" )
+    fi
+    echo "${res_devs[@]}"
```
marc-hb, line 162: For anything non trivial I would always prefer `printf`

marc-hb:
> yeah, I need to use this test already, so it evolves while in review ..

Not ideal but typical for brand new tests and other big changes.

Actually, I had my own requests in mind and not this but it's probably both :-)

> > Sorry I keep finding minor problems but that's because the code keep changing :-)

To be clear: this is IMHO good enough to merge. As a brand new test, it cannot regress anything :-) On other hand, "fix later" == "fix never" more often than not....

PR #1285, opened 2025-07-23
tools: Add a new test for pause/resume

ranj063:
The test runs the specified number of pause/resume iterations on a combination of a playback PCM and a capture PCM looping through all available playback and capture PCM devices available in the SOF audio card. It skips the deep buffer playback PCMs for the time being.

### DIFF
--- a/tools/pause_resume_test.c
+++ b/tools/pause_resume_test.c
@@ -1,0 +1,386 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <alsa/asoundlib.h>
+#include <unistd.h>
+#include <pthread.h>
+#include <stdbool.h>
+
+// This program requires ALSA library and a Linux environment with ALSA support.
+// sudo apt install libasound2-dev
+// To compile: gcc -o pcm_pause pause_test.c -lasound
+// To run: ./pcm_pause [pause_count] or ./pcm_pause [pause_count] [playback_device_id] [capture_device_id]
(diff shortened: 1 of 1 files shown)


lyakh:
@ranj063 do you plan to add building and running the test to this PR? I think it would be important to see its results in the same PR

marc-hb (commented):
> I could be wrong but there is currently no support to compile C code in sof-test before deployment.

> I wonder if some existing tool could be used as well -- could be easier to integrate to sof-test runs in CI (versus starting to build C tools and deploy them to DUTs).

Actually, doesn't this new test belong to https://github.com/alsa-project rather than sof-test? Please correct me but the only SOF-specific bit in this PR seems to be where the test ignores non-SOF cards (why?) Which cards to test looks like it should be a command line argument, not hard-coded.

Hosting this test in ALSA would solve the deployment problem since ALSA already needs a C compiler for deployment.

cc: @golowanow , @cgturner1 

Now you can argue that a lot of sof-test is not that SOF-specific either and could/should be moved to https://github.com/alsa-project too. True but there are important differences:

1. sof-test/test-case/ code is not brand new; it's much more difficult to move stuff that already exists
2. it's shell and Python and not C code and does not have a C deployment issue. Not sure ALSA maintainers would even accept higher level languages much.
3. it would be a challenge to disentangle the non-SOF specific bits while maintaining some stable interface with ALSA. Making sure the combinations of sof-test + ALSA versions deployed stay compatible with each other would be "fun".

Issue #1120, opened 2023-10-27
[FEATURE] use --fatal-errors for aplay/arecord tests


plbossart:
**Is your feature request related to a problem? Please describe.**
When we reach an xrun, we should stop sof-test scripts. xruns need to be recoverable, but they should never happen. we don't want to miss these issues because the recovery quicks-in and hides problems.

**Describe the solution you'd like**
Add --fatal-errors to aplay/arecord command-line options.

Example run with --test-position (needs aplay patch to dump the last result)

````
Status(R/W) (standalone avail=36 delay=924):
  state       : RUNNING
  trigger_time: 2045.504937
  tstamp      : 2190.754602
  delay       : 924
  avail       : 36
  avail_max   : 444
Status(R/W) (standalone avail=44 delay=912):
  state       : RUNNING
  trigger_time: 2045.504937
  tstamp      : 2190.754852
  delay       : 912
  avail       : 48
  avail_max   : 48
underrun!!! (at least 471.161 ms long)
Status:
  state       : XRUN
  trigger_time: 2190.786234
  tstamp      : 2191.257392
  delay       : 0
  avail       : 1412
  avail_max   : 1412
aplay: xrun:1690: fatal underrun: Success
````

**Describe alternatives you've considered**
none that I can think of

**Additional context**
Not sure if we want to add this in sof-test proper, or add this in the environment variables with "APLAY_OPTS" that are already supported.

plbossart:
related PR: https://github.com/alsa-project/alsa-utils/pull/242

plbossart:
btw this PR https://github.com/alsa-project/alsa-utils/pull/242 was merged so most likely we can "just" add this --fatal-errors option to tests?

marc-hb:
Yes, work in "progress": b4523e0f3ae8

marc-hb:
> btw this PR https://github.com/alsa-project/alsa-utils/pull/242 was merged

We don't have a way to upgrade ALSA on test devices AFAIK

marc-hb:
I lost track of why this is not a duplicate of
- #489

PR #1287, opened 2025-07-24
case-lib: apause: Make sure the PCM is running before pausing

ranj063:
This change addresses the following error seen when doing pause with the multiple-pause-resume test:
aplay: do_pause:1586: pause push error: File descriptor in bad state

When an xrun happens during the test, the application tries to recover from the xrun by preparing and restarting the stream. There could be a race between when this happens and when the script tries to pause the stream. To avoid this, make sure that the stream state is RUNNING before going ahead with the pause.

### DIFF
--- a/case-lib/apause.exp
+++ b/case-lib/apause.exp
@@ -107,6 +107,29 @@ proc cr_to_lf {arg} {
 #
 #    arecord $cmd_opts -D $dev -r $rate -c $channel -f $fmt -vv -i $file_name ...
 log 0 "$argv0 spawning: $argv"
+
+set device [lindex $argv 2]
+set command [lindex $argv 0]
+
+# Determine direction based on the command
+if {$command eq "aplay"} {
+    set direction "p"  ; # Playback
+} else {
+    set direction "c"  ; # Capture
+}
+
+# parse the card number and device
+if {![regexp {hw:(\d+),(\d+)} $device match card_number device_id]} {
+    log 0 "ERROR: Failed to parse hw string: $hw_string"
+}
+
+set pcm_status_file "/proc/asound/card${card_number}/pcm${device_id}${direction}/sub0/status"
+
+if {![file exists $pcm_status_file]} {
+    log 0 "ERROR: PCM status file not found: $pcm_status_file"
+    exit 1
+}
+
 spawn  {*}$argv
 set start_time_ms [clock milliseconds]; # re-adjust
 set last_space_time 0 ; # could not resist that name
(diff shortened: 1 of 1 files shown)


marc-hb (commented):
This PR deals with an xrun that happened _when_ in the pause/resume cycle?

Which xrun did you observe, underrun or overrun or both? Can you elaborate?

This looks like a fairly complex issue, I feel like a new https://github.com/thesofproject/sof-test/issues/new/choose would not hurt.

Shouldn't the test fail anyway when an xrun happened? Ideally, with an error message less cryptic than "file descriptor in bad state" but still fail? Doesn't this PR hide the xrun and make the test pass when it shouldn't?

marc-hb:
Thanks, good to know.

The code change makes sense to me.

I still think it would be useful to have a short description of when the xrun happens and why. If not in a new sof-test bug, then in the commit message.

marc-hb:
What is (or... should be) the "canonical" way to detect xruns? If it is for instance scanning logs (kernel or firmware), then there could be some sort "universal" toggle that works the same across all tests.

Doesn't this test scan logs already? Most do.

marc-hb:
> but given we don't have --fatal-errors in other tests at the moment

Actually: you can already do this with some tests: #489, #1120, SOF_ALSA_OPTS

Work in progress.

kv2019i:
@marc-hb The application can detect xruns via ALSA API and it's upto the app to decide whether it's lethal or not (stremaing stopped, or does it proceed with a possible glitch in played/captured audio).

I agree this PR is in effect is reducing the coverage of this test case and it can hide fails we had earlier flagged as fails. OTOH, the xruns are a smaller impact issue compared to an IPC timeout (= one might need to reboot the DUT to get any audio functionality back). In practise we have disabled pause in the SOF drivers towards applications and we primarily use this test case to stress test the pipeline state machines and root out IPC timeout scenarios. In this context, I think this PR makes sense as it will filter out the xruns (that we anyways are not fixing in context of pause), and will get more reliable pass/fail w.r.t. to IPC timeouts.

redzynix (approved):
Looks good.

case-lib/apause.exp
```diff
@@ -218,6 +241,26 @@ expect {
                 log 1 "($pauses_counter/$repeat_count) Found volume ### | __%, active for $_record_for ms"
 
                 set _delay [substract_time_since_last_space $_record_for]
+
+                # wait 50ms for the PCM status to be RUNNING before pausing
+                # this is to make sure that in the case of an xrun the application
+                # successfully recovers and restarts the stream.
```
marc-hb:
```suggestion
                # successfully recovers and restarts the stream.
                # change `max_attempts` to zero when observing xruns is desired.
```

PR #1289, opened 2025-07-31
Add reboot wrapper function

PLangowski:
The `reboot_wrapper` function prints `Reboot command` to stdout and reboots the platform


marc-hb:
The context in #1038 and internal framework issues 57 and 272 seem relevant.

I don't know what this is for (and I don't need to know), but whatever you do, there is already context for it. Always search and link to it.

"Those who do not learn history are doomed to repeat it"

PR #1294, opened 2025-08-08
lib.sh: Set ALSA settings from a state file

golowanow:
#### This PR is to continue with #1284 and PoC #1270
----

Set ALSA settings from a state file and migrate PTL configurations to the new mechanism:

1. `set_alsa_settings()` function now uses `alsactl restore` command, to change ALSA controls from a custom state file first, if it is present, and next - execute a bash script with amixer commands, if it is also present.
The state file is expected as `./alsa_settings/${PLATFORM}.state` whereas the bash script remains as `./alsa_settings/${PLATFORM}.sh`
The custom state should be compatible with the platform's default state, see `set_alsa()` function.

2. Replace PTL platform ALSA settings done as bash script commands by their equivalent ALSA state files for:
 * PTLH_HDA_AIOC
 * PTLH_SDW_RT712
 * PTLP_RVP_SDW
  * MTLP_RVP_HDA
  * MTLP_SDW_AIOC
  * MTLP_RVP_SDW
  * LNLM_RVP_HDA
  * LNLM_SDW_AIOC

The parameters in the state files are explicitly set even if the default values at the appropriate DUT are currently the same.

3. Log difference between the current ALSA state and the restored ALSA state.

4. Add `README.md` documentation for the SOF test ALSA settings.


marc-hb (commented):
No time to review and it's still a draft but this looks very useful!

alsa_settings/MTLP_SDW_AIOC.state
```diff
@@ -0,0 +1,38 @@
+state.sofsoundwire {
```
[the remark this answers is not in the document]
golowanow: > Why this file is different from MTLP_RVP_SDW.state if last configuration was marked as duplicated ?

ALSA controls for MTLP_RVP_SDW and MTLP_SDW_AIOC are different (see .9 and .10 vs. .36 and CODEC respectively).

marc-hb:
For some unknown reason I just remembered that UCM was supposed to be tested interface, not ALSA states...
- https://github.com/thesofproject/linux/issues/2211

Quoting internal issue sof-framework 635:

> Please stop using asound.state and start using UCM files.
> If there are no UCM files let's create them
> We have to stop the duplication of work, and asound.state files are EVIL.

:-)

golowanow:
> > Please stop using asound.state and start using UCM files.
> > If there are no UCM files let's create them

@marc-hb, AFAIK this choice (state vs. ucm) has been discussed several months ago, and #1270 PoC was started focusing on the state method. At this point we need to sort out the currently used mechanism first, with its limitations known.
The UCM method needs its attention too as a next step.

golowanow:
> > > Please stop using asound.state and start using UCM files.
> > > If there are no UCM files let's create them

@marc-hb, AFAIK this choice (state vs. ucm) has been discussed several months ago, and #1270 PoC was started focusing on the state method. At this point we need to sort out the currently used mechanism first, with its limitations known. The UCM method needs its attention too as a next step.
btw, this PR has been already used to identify issues with misaligned configurations, so its merge will improve test execution on our CI.

alsa_settings/MTLP_RVP_SDW.state
```diff
@@ -0,0 +1,35 @@
+state.sofsoundwire {
```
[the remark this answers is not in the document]
majunkier: In that case, comments should be move to README file to preserve info why specific config was set

golowanow:
@majunkier - do you have any other blocking comments ?

alsa_settings/MTLP_RVP_SDW.state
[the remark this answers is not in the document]
marc-hb: A good git log is important but it should be focused and explaining why things CHANGED. It should ideally not be required to explain why things ARE the way they are.

alsa_settings/README.md
```diff
@@ -0,0 +1,73 @@
+# ALSA settings
+
+SOF test case execution requires certain DSP and mixer configurations being set
+on the DUT (Device Under Test). An important part of this configuration is the
+signal levels adjusted to the hardware setup which runs the tests, including its
+codec and loopback devices.
+
+The `./alsa_settings/` directory contains custom ALSA configurations which
+are currently applied at the SOF CI pipeline's hardware platforms.
```
marc-hb, line 9: Why is the _default_ hardware configuration different and not enough? If this is by design (in other words, testing requires some "special sauce"), very briefly state why. If this is not design, clearly state that all these files are really "workarounds".

Either way, it's not good to be testing something different from regular users. So this deserves a bit of justification in either case.

alsa_settings/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+2. `PLATFROM.sh` - a shell script to configure the ALSA driver parameters,
+   e.g. calling `amixer` tool.
+
+The custom state should be compatible with the platform's default state.
+
+Currently, the custom settings include all these default ALSA settings the SOF
+tests are expecting at the appropriate DUT.
+In the future most of them should be moved to the DUT's default config
+out of the test case scope, so only the test-specific settings will remain here.
```
marc-hb, line 36: Move this at the top because this is the "raison d'etre" of this whole thing.

case-lib/lib.sh
```diff
@@ -1218,13 +1218,24 @@ set_alsa_settings()
...
+            if [ -f "${ALSA_SETTINGS_FILE}.state" ]; then
+                dlogc "alsactl restore --file=${ALSA_SETTINGS_FILE}.state --no-init-fallback --no-ucm"
+                alsactl restore --file="${ALSA_SETTINGS_FILE}".state --no-init-fallback --no-ucm 2>&1 || rc=$?
+                [[ "${rc}" -ne 0 ]] && dloge "alsactl restore error=${rc}"
+                # With `--pedantic` alsactl returns error=2, e.g. "No state is present for card CODEC"
+            fi
+            if [ -f "${ALSA_SETTINGS_FILE}.sh" ]; then
+                dlogc "${ALSA_SETTINGS_FILE}.sh"
+                "${ALSA_SETTINGS_FILE}".sh 2>&1 || rc=$?
+                [[ "${rc}" -ne 0 ]] && dloge "ALSA settings error=${rc}"
```
marc-hb, line 1235: I think you should reset `rc` to zero first, otherwise this `.sh` message could be triggered by the _previous_ . `.state` error despite the `.sh` running fine...

alsa_settings/README.md
```diff
@@ -0,0 +1,79 @@
...
+as a reference.
+
+These configurations include ALSA settings which are considered essential
+for the SOF tests and expected at the appropriate DUT by default. They allow
+ALSA settings tweaking in along with SOF and test developmnent, if needed.
+
+Please note that in some cases the settings are tuned to CI DUT 'harnesses':
+external codecs with their versions, audio cards at loopback connections, etc.
+In the future, most of these settings should be moved to the DUT's default config
+out of the test case scope, so only the test-specific settings will remain here.
```
marc-hb, line 20: There should be a simple way to identify which setting is part of which category in each file using using some "standardized" comment convention and/or sections.

alsa_settings/MTLP_RVP_SDW.state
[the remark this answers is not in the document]
majunkier: Agree, git log should focus on explaining why something was changed. But there are cases where leaving comment in code is also important, for changes that could rise questions in future, or like here proposing whole new approach, comment in file can be quicker to find and understand than digging in git history.

case-lib/lib.sh
golowanow: right, added

golowanow:
`check-alsabat-` test cases are passed on all CI configurations

SzymonRichert (commented):
LGTM

lgirdwood (approved):

PR #1296, opened 2025-08-13
tests: add option to test with pipewire

ekurdybx:
Added option to use pipewire instead of ALSA direct mode in tests, that can be turned on by setting TEST_WITH_PIPEWIRE=true. Added pipewire-wrapper, to run in SOF CI.
Currently only test that can run on pipewire is check-perfomance.sh (runs aplay/arecord on one of each types of sinks/sources).

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Some tests support these environment variables (work in progress):
   - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
   - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
 These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
+  - SOF_TEST_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Supported by a limited number of tests,
+    try 'git grep SOF_TEST_PIPEWIRE' to find them.
 
 Warning, these environment variables do NOT support parameters
 with whitespace or globbing characters, in other words this does NOT
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -12,6 +12,8 @@ function func_exit_handler()
 
     dlogi "Starting func_exit_handler($exit_status)"
 
+    func_lib_check_and_disable_pipewire
+
     # call trace
     if [ "$exit_status" -ne 0 ] ; then
         dloge "Starting ${FUNCNAME[0]}(), exit status=$exit_status, FUNCNAME stack:"
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -83,6 +83,10 @@ minvalue() { printf '%d' $(( "$1" < "$2"  ? "$1" : "$2" )); }
 #
 start_test()
 {
+    if [ "$SOF_TEST_PIPEWIRE" == true ]; then
+        func_lib_enable_pipewire
+    fi
+
     if is_subtest; then
         return 0
     fi
@@ -571,6 +575,45 @@ func_lib_check_sudo()
     }
 }
 
+func_lib_enable_pipewire()
+{
+    dlogi "Starting Pipewire..."
+    
+    sudo systemctl --global unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user unmask pipewire{,-pulse}.{socket,service}
+    sudo systemctl --global unmask wireplumber.service 
+    systemctl --user unmask wireplumber.service
+
+    systemctl --user start pipewire{,-pulse}.{socket,service}
+    systemctl --user start wireplumber.service
+
+    systemctl --user daemon-reload
+
+    systemctl is-active --user --quiet pipewire{,-pulse}.{socket,service} && dlogi "Pipewire started"
+    systemctl is-active --user --quiet wireplumber.service && dlogi "Wireplumber started"
+}
+
+func_lib_check_and_disable_pipewire()
+{
+    dlogi "Check if pipewire is running"
+    
+    if systemctl is-active --user --quiet pipewire{,-pulse}.{socket,service}; then func_lib_disable_pipewire; else dlogi "Pipewire not running"; fi
+}
+
+func_lib_disable_pipewire()
+{
+    dlogi "Stopping Pipewire..."
+
+    systemctl --user stop wireplumber.service
+    systemctl --user stop pipewire{,-pulse}.{socket,service}
+
+    sudo systemctl --global mask wireplumber.service
+    sudo systemctl --global mask pipewire{,-pulse}.{socket,service}
+
+    if systemctl is-active --user --quiet wireplumber.service; then dlogi "Wireplumber not stopped"; else dlogi "Wireplumber stopped"; fi
+    if systemctl is-active --user --quiet pipewire{,-pulse}.{socket,service}; then dlogi "Pipewire not stopped"; else dlogi "Pipewire stopped"; fi
+}
+
 systemctl_show_pulseaudio()
 {
     printf '\n'
@@ -908,9 +951,15 @@ aplay_opts()
 	# shellcheck disable=SC2086
         tinyplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS -D "$card_nr" -d "$dev_nr"  -i wav noise.wav
     elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
-        dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
-        # shellcheck disable=SC2086
-        aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+        if [[ "$SOF_TEST_PIPEWIRE" == true ]]; then
+            dlogc "timeout -k $duration $duration aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"  # option -d doesn't work with pipewire so we need timeout
+            # shellcheck disable=SC2086
+            timeout -k "$duration" "$duration" aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+        else
+            dlogc "aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS $*"
+            # shellcheck disable=SC2086
+            aplay $SOF_ALSA_OPTS $SOF_APLAY_OPTS "$@"
+        fi
     else
         die "Unknown ALSA tool: ${SOF_ALSA_TOOL}"
     fi
@@ -927,12 +976,36 @@ arecord_opts()
 	# shellcheck disable=SC2086
         tinycap $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$file" -D "$card_nr" -d "$dev_nr" -c "$channel" -t "$duration" -r "$rate" -b "$format"
     elif [[ "$SOF_ALSA_TOOL" = "alsa" ]]; then
-        dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
-        # shellcheck disable=SC2086
-        arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+        if [[ "$SOF_TEST_PIPEWIRE" == true ]]; then
+            dlogc "timeout -k $duration $duration arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"  # option -d doesn't work with pipewire so we need timeout
+            # shellcheck disable=SC2086
+            timeout -k "$duration" "$duration" arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+        else
+            dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+            # shellcheck disable=SC2086
+            arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+        fi  
     else
         die "Unknown ALSA tool: ${SOF_ALSA_TOOL}"
     fi
+}
+
+# Get the ID of the first sink/source of a given type, e.g. "Speaker" or "Headphones". Print an empty line if ID not found.
+get_id_of_pipewire_endpoint()
+{
+    # $ wpctl status returns list of all endpoints managed by wireplumber. We filter by given sink/source type, which returns something like this: 
+    # │  *   48. sof-soundwire Headphones            [vol: 0.40]  (or without the * when it's not the current default)
+    # We filter out everything but ID, and only take the first line of the output (if there's more that one object of that type we ignore the rest)
+
+    local object_name="$1"
+    object_id=$(wpctl status | awk -v name="$object_name" 'tolower($0) ~ tolower(name) { sub(/\*/,""); sub(/\./,"",$2); print $2; exit }')
+
+    # Check if object_id is a number
+    re='^[0-9]+$'
+    if [[ "$object_id" =~ $re ]] ; then
+        printf '%s' "$object_id"
+    fi
+
 }
 
 die()
--- a/test-case/check-performance.sh
+++ b/test-case/check-performance.sh
@@ -20,6 +20,9 @@
 
 set -e
 
+PIPEWIRE_OUTPUTS=("Speaker" "Headphones" "HDMI" "Stereo")
+PIPEWIRE_INPUTS=("Digital Microphone" "DMIC" "Headset Microphone" "SoundWire microphone" "Stereo")
+
 # It is pointless to perf component in HDMI pipeline, so filter out HDMI pipelines
 # shellcheck disable=SC2034
 NO_HDMI_MODE=true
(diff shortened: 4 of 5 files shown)


README.md
```diff
@@ -64,6 +64,7 @@ Some tests support these environment variables (work in progress):
   - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
   - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
 These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
+  = TEST_WITH_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Currently only works with check-performance.sh.
```
marc-hb, line 67:
```suggestion
  = SOF_TEST_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Currently only works with check-performance.sh.
```
Use a prefix / "namespace"

test-case/pipewire-wrapper.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+##
+## Case Name: Wrapper to run a test case given with Pipewire
```
marc-hb, line 4:
```suggestion
## Case Name: Wrapper to run a test case given with Pipewire in setup that cannot set the environment variable. Keep this script as simple as possible and avoid additional layers of indirections when possible.
```

test-case/check-performance.sh
```diff
@@ -51,27 +54,80 @@ func_pipeline_export "$tplg" "type:any"
...
         fi
-done
+        dlogi "Setting default source to $source_id: $source_type"
+        wpctl set-default "$source_id"
+        arecord_opts -Ddefault /dev/zero -q &
+        arecord_num=$((arecord_num+1))
+    done
+}
+
+if [ "$SOF_TEST_PIPEWIRE" == true ]; then
```
marc-hb, line 95: This should be a `main()` function.

test-case/check-performance.sh
marc-hb, line 1: Right now there's very little code shared between `SOF_TEST_PIPEWIRE` versus not, it's like two different tests concatenated in one. Either you split these two different tests into two different files (and then maybe have the wrapper call one or other), or you refactor the code to have more in common. 

The two different files can still share more code in `lib.sh`; for instance the end of  the script that counts the number of processes should likely be in `lib.sh`, it does not look test-specific at all (and I wouldn't be surprised if it has already been duplicated).

case-lib/lib.sh
```diff
@@ -571,6 +575,55 @@ func_lib_check_sudo()
...
+func_lib_enable_pipewire()
+{
+    dlogi "Starting Pipewire..."
+    # Sometimes unmasking does not work as expected, and we need to manually delete /dev/null links and try again
+    start_pipewire_services || true
+    pipewire_unmask_manually
+    start_pipewire_services
+
+    # Wait for pipewire to start
+    sleep 5s
```
marc-hb, line 612: Here again, please share some evidence (logs, etc.) that this is required.

case-lib/lib.sh
```diff
@@ -571,6 +575,55 @@ func_lib_check_sudo()
...
+        pw_service_path=${pw_service_path/"FragmentPath="/}
+        pw_service_link=$(ls -l "$pw_service_path")
+        if [[ "$pw_service_link" == *"/dev/null" ]]; then  # Delete link if it points to /dev/null
+            dlogi "Found ->/dev/null link:  $pw_service_path, removing it"
+            sudo rm "$pw_service_path"
+        fi
+    done
+}
+
+start_pipewire_services()
```
marc-hb, line 593:
```suggestion
unmask_start_pipewire_services()
```

marc-hb (commented):
Don't share fixup commits

https://docs.zephyrproject.org/latest/contribute/guidelines.html#contribution-workflow

case-lib/lib.sh
```diff
@@ -571,6 +575,55 @@ func_lib_check_sudo()
...
+    systemctl --user unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user start pipewire{,-pulse}.{socket,service}   
+    systemctl --user unmask wireplumber.service
+    systemctl --user start wireplumber.service
+}
+
+func_lib_enable_pipewire()
+{
+    dlogi "Starting Pipewire..."
+    # Sometimes unmasking does not work as expected, and we need to manually delete /dev/null links and try again
```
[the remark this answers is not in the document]
ekurdybx: Okay so after looking at it again I saw that the issue is fixed by unmasking pipewire globally and then for user. So the workaround isn't needed after all.

case-lib/lib.sh
```diff
@@ -927,14 +965,32 @@ arecord_opts()
...
+get_id_of_pipewire_endpoint()
+{
+    # $ wpctl status returns list of all endpoints managed by wireplumber. We filter by given sink/source type, which returns something like this: 
+    # │  *   48. sof-soundwire Headphones            [vol: 0.40]  (or without the * when it's not the current default)
+    # We filter out everything but ID, and only take the first line of the output (if there's more that one object of that type we ignore the rest)
+
+    local object_name="$1"
+    object_id=$(wpctl status | grep -i "$object_name" | tr -d '*' | awk '{print $2}' | tr -d '.' | head -n 1)
+    echo $object_id 
+}
```
marc-hb, line 992: Because the parsing is relatively complex, there should be some sanity check. For instance, you could verify that `object_id` is a number like this:

```shell
test "$object_id" -eq "$object_id" || die "$object_id is not a number"
echo "$object_id"
```
or if you prefer:

```shell
test "$object_id" -eq "$object_id" || {
  dlogw "$object_id is not a number"
  return 1
}
echo "$object_id"
```

Try this bug for instance:

```shell
object_name="not existent"
object_id=(wpctl status | grep -i "$object_name" | tr -d '*' | awk '{print $2}' | tr -d '.' | head -n 1)
echo "$object_id"
```
=> empty output

Or this one:

```shell
object_name="" # empty or missing arg
object_id=(wpctl status | grep -i "$object_name" | tr -d '*' | awk '{print $2}' | tr -d '.' | head -n 1)
echo "$object_id"
```

=> 'pipewire-0'

case-lib/lib.sh
```diff
@@ -927,14 +965,32 @@ arecord_opts()
...
 
+# Get the ID for a given sink or source type, e.g. "Speaker" or "Headphones"
+get_id_of_pipewire_endpoint()
+{
+    # $ wpctl status returns list of all endpoints managed by wireplumber. We filter by given sink/source type, which returns something like this: 
+    # │  *   48. sof-soundwire Headphones            [vol: 0.40]  (or without the * when it's not the current default)
+    # We filter out everything but ID, and only take the first line of the output (if there's more that one object of that type we ignore the rest)
+
+    local object_name="$1"
+    object_id=$(wpctl status | grep -i "$object_name" | tr -d '*' | awk '{print $2}' | tr -d '.' | head -n 1)
```
marc-hb, line 990: The output of `wpctl status --name` looks a bit more "deterministic" and machine parsable but I don't know pipewire.

case-lib/lib.sh
marc-hb: You could/should also "sanitize your input" with something like this. Not mutually exclusive:

```shell
wpctl status | grep "$object_name" || {
   ... $object_name not found error ...
} 
```

case-lib/lib.sh
marc-hb: A single `awk` invocation can do all that (and more)
```suggestion
    object_id=$(wpctl status | awk "/$object_name/"'{ sub("*",""); sub("\\.","", $2); print $2; exit 0; }')
```

test-case/check-performance.sh
```diff
@@ -41,52 +44,112 @@ func_opt_parse_option "$@"
...
+    real_aplay_num=$(ps --no-headers -C aplay | wc -l)
+    real_arecord_num=$(ps --no-headers -C arecord | wc -l)
+    if [ "$real_aplay_num" != "$aplay_num" ] || [ "$real_arecord_num" != "$arecord_num" ];
+    then
+        dlogi "Number of aplay/arecord process running: $real_aplay_num, $real_arecord_num"
+        die "aplay/arecord process exit unexpectedly"
+    fi
+
+    dlogi "Waiting for aplay/arecord process to exit"
+    sleep $((duration + 2))
```
marc-hb, line 146: What if they get stuck because of some bug/issue and never exit?

Actually wait for them, either using `aPID=$!; ... wait $aPID` or re-use some code from `lib.sh` or elsewhere. Waiting for these is a common pattern and it is already done in some other tests.

case-lib/lib.sh
```diff
@@ -571,6 +575,34 @@ func_lib_check_sudo()
...
+{
+    dlogi "Starting Pipewire..."
+    
+    sudo systemctl --global unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user start pipewire{,-pulse}.{socket,service}
+    sudo systemctl --global unmask wireplumber.service 
+    systemctl --user unmask wireplumber.service
+    systemctl --user start wireplumber.service
+    systemctl --user daemon-reexec
```
marc-hb, line 588: Please don't just "throw everything at it". It's called "programming by coincidence".

From `man systemctl`

> Reexecute the systemd manager. This will serialize the manager state, reexecute the process and deserialize the state again. This command is of little use except for debugging and package upgrades. Sometimes, it might be helpful as a heavy-weight daemon-reload.

Either `daemon-reload` is enough and you can drop this, or you know it's not enough for some documented reason (which you mention in a comment) and then you use only `daemon-reexec`. But not both.

case-lib/lib.sh
```diff
@@ -571,6 +575,34 @@ func_lib_check_sudo()
...
+func_lib_enable_pipewire()
+{
+    dlogi "Starting Pipewire..."
+    
+    sudo systemctl --global unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user unmask pipewire{,-pulse}.{socket,service}
+    systemctl --user start pipewire{,-pulse}.{socket,service}
+    sudo systemctl --global unmask wireplumber.service 
+    systemctl --user unmask wireplumber.service
+    systemctl --user start wireplumber.service
```
marc-hb, line 587: I would unmask everything first to get back to the normal OS state, and then start.

case-lib/lib.sh
```diff
@@ -571,6 +575,34 @@ func_lib_check_sudo()
...
+}
+
+func_lib_disable_pipewire()
+{
+    dlogi "Stopping Pipewire..."
+
+    systemctl --user stop wireplumber.service
+    sudo systemctl --global mask wireplumber.service
+    systemctl --user stop pipewire{,-pulse}.{socket,service}
+    sudo systemctl --global mask pipewire{,-pulse}.{socket,service}
```
marc-hb, line 601: Same as above, I would stop everything first, and then mask.

Mmm... or maybe both at the same time in case there is a race where a client causes a restart between stopping and masking?

case-lib/lib.sh
marc-hb: You can even do this:

```shell
local object_id
object_id=$(wpctl status |
   awk "/$object_name/"' { sub("*",""); sub("\\.","", $2); print $2; found=1; exit 0 } 
         END { if (!found) exit 1 }') || {
  error $object_name not found ...
  ...
} 

```

case-lib/lib.sh
```diff
@@ -927,14 +968,39 @@ arecord_opts()
...
+    # │  *   48. sof-soundwire Headphones            [vol: 0.40]  (or without the * when it's not the current default)
+    # We filter out everything but ID, and only take the first line of the output (if there's more that one object of that type we ignore the rest)
+
+    local object_name="$1"
+    object_id=$(wpctl status | awk -v name="$object_name" 'tolower($0) ~ tolower(name) { sub(/\*/,""); sub(/\./,"",$2); print $2; exit }')
+
+    # Check if object_id is a number
+    re='^[0-9]+$'
+    if [[ $object_id =~ $re ]] ; then
+        # shellcheck disable=SC2086
```
marc-hb, line 998: I don't get this, why would you not quote this?

case-lib/lib.sh
```diff
@@ -927,14 +968,39 @@ arecord_opts()
...
+            dlogc "arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS $*"
+            # shellcheck disable=SC2086
+            arecord $SOF_ALSA_OPTS $SOF_ARECORD_OPTS "$@"
+        fi  
     else
         die "Unknown ALSA tool: ${SOF_ALSA_TOOL}"
     fi
 }
 
+# Get the ID of the first sink/source of a given type, e.g. "Speaker" or "Headphones". Return nothing if ID not found.
```
marc-hb, line 985:
```suggestion
# Get the ID of the first sink/source of a given type, e.g. "Speaker" or "Headphones". Print an empty line if ID not found.
```
Printing is not the same as `return`. The former is stdout/stderr, the latter is the exit status.

I see no obvious reason why, but if you really want to print nothing then don't use `echo`, use `printf` instead.

In fact, there's never any reason to use `echo` in scripts.

case-lib/hijack.sh
```diff
@@ -12,6 +12,10 @@ function func_exit_handler()
 
     dlogi "Starting func_exit_handler($exit_status)"
 
+    if [ "$SOF_TEST_PIPEWIRE" == true ]; then
+        func_lib_disable_pipewire
```
marc-hb, line 16: I imagine you're doing this to restore the system in either the previous state, or in a specific state, "pipewire off", correct?

- to always restore in the "pipewire off" state, there is no need look at  "$SOF_TEST_PIPEWIRE". I mean if the current test ran the plain ALSA test code but with pipewire already ON by accident, then that it is not a good reason to keep doing that. Better stop it now.

- If you want to restore the system in the previous pipewire state (whichever that was), then you would first need to remember it. 

Either way, merely basing the disable decision on `"$SOF_TEST_PIPEWIRE" == true` does not seem correct.

Maybe a simpler solution is to either enable or disable at test start? I mean never leaving the system "as is" at test start.

case-lib/lib.sh
marc-hb: You should be able to remove the comment now.

marc-hb (commented):
Only one concern left about the "state machine", otherwise the codestyle looks good. Someone from the SOF team must review the actual logic.

test-case/check-performance.sh
```diff
@@ -41,52 +44,112 @@ func_opt_parse_option "$@"
...
+    func_lib_check_sudo
+    func_pipeline_export "$tplg" "type:any"
+
+    aplay_num=0
+    arecord_num=0
+
+    if [ "$SOF_TEST_PIPEWIRE" == true ]; then
+
+        dlogi "Running aplays"
+        run_aplays "${PIPEWIRE_OUTPUTS[@]}"
```
marc-hb, line 101: It's generally more convenient to have the `for` loop _outside_ the function, I mean like this:

```shell
    for source_type in...
         run_(one)_aplay

```

This makes it easier to copy/paste `run_aplay()` in the interactive shell for interactive testing.

README.md
```diff
@@ -64,6 +64,7 @@ Some tests support these environment variables (work in progress):
   - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
   - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
 These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
+  = SOF_TEST_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Currently only works with check-performance.sh.
```
PLangowski, line 67:
```suggestion
  - SOF_TEST_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Currently only works with check-performance.sh.
```

test-case/pipewire-wrapper.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Ensure the test case script file name is provided
+if [ -z "$1" ]; then
+    echo "Error: No test case script file name provided. Exiting..."
+    exit 1
+fi
+
+export SOF_TEST_PIPEWIRE=true
+
+TESTDIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
```
redzynix, line 28: Sorry for double comments on the same line but I've noticed that realpath without ```-e``` will not fail if the path does not exist, and to maintain the same behaviour as ```cd {path}``` it needs -e
```suggestion
TESTDIR=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")
```

marc-hb:
You have just pushed 2 merge commits:

Screenshot 2025-09-15 at 08 35 43

ekurdybx:
> You have just pushed 2 merge commits:
> 
> Screenshot 2025-09-15 at 08 35 43

My mistake, fixed that

README.md
```diff
@@ -64,6 +64,7 @@ Some tests support these environment variables (work in progress):
   - SOF_ALSA_OPTS contains optional parameters passed on both play and record.
   - SOF_APLAY_OPTS and SOF_ARECORD_OPTS contain optional parameters passed additionally on play and record respectively.
 These options are applied to the selected tool (alsa or tinyalsa) based on the value of SOF_ALSA_TOOL 
+  - SOF_TEST_PIPEWIRE (default: false) allows to use pipewire instead of ALSA direct mode. Currently only works with check-performance.sh.
```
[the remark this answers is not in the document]
ekurdybx: Done

case-lib/hijack.sh
marc-hb: tl;dr: drop the `if [ "$SOF_TEST_PIPEWIRE" == true ]`

>  So there is no case when pipewire would be already running.

OK, so it looks like you're assuming case 1) which is: the desired state is for pipewire to always be masked between tests - whatever gets tested and how. This should be documented somewhere.

However this is just a desire. By definition, a lot of bad things always happen in test environments and tests must aggressively cleanup before _and_ after themselves when they can and when it does not take too long.

It's easy to think of cases where pipewire would be running when it shouldn't. The most obvious examples are:
- some earlier pipewire test run  stopped in some unexpected place and did not clean up and mask again
- someone borrowed the system for interactive testing and does not even know what state the system should be left in - assuming they care in the first place (experience shows most people don't care).

So you can simply drop the `if [ "$SOF_TEST_PIPEWIRE" == true ]` and mask unconditionally.

marc-hb: > > Pipewire is masked on all of our CI platforms after it's being installed.

> It's easy to think of cases where pipewire would be running when it shouldn't. The most obvious examples: ...

Forgot the most obvious example: sof-test is not just for Intel CI.

case-lib/lib.sh
```diff
@@ -571,6 +575,48 @@ func_lib_check_sudo()
...
+    systemctl is-active --user --quiet pipewire{,-pulse}.{socket,service} && dlogi "Pipewire started"
+    systemctl is-active --user --quiet wireplumber.service && dlogi "Wireplumber started"
+}
+
+func_lib_check_and_disable_pipewire()
+{
+    dlogi "Check if pipewire is running"
+    
+    # shellcheck disable=SC2015
+    systemctl is-active --user --quiet pipewire{,-pulse}.{socket,service} && func_lib_disable_pipewire || dlogi "Pipewire not running"
```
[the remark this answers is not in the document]
ekurdybx: Fixed that

ekurdybx:
> There could some minor issues left (including at least one below) but they can be fixed later.
> 
> On the other hand, this should really be run in jenkins. @ekurdybx your [github.com/orgs/thesofproject/teams/sof-developers](https://github.com/orgs/thesofproject/teams/sof-developers) membership is way overdue. Please ask someone in [github.com/orgs/thesofproject/teams/sof-developers?query=role%3Amaintainer](https://github.com/orgs/thesofproject/teams/sof-developers?query=role%3Amaintainer) . You probably can't see that link but most of your coworkers can.

This was run in jenkins, I tested this on lnlm, ptlh and mtlp platforms, it works as expected.

marc-hb:
>This was run in jenkins, I tested this on lnlm, ptlh and mtlp platforms, it works as expected.

I just realized none of the sof-test Pull Requests show Jenkins results any more, it's not specific to this one. I mean, there is no sof-test Pull Request that has test results like this:

https://sof-ci.01.org/sofpr/PR10251/build15809/devicetest/index.html

@ekurdybx , @cgturner1 could you investigate this?

PR #1297, opened 2025-08-19
[SKIP SOF-TEST] sof_perf_analyzer: fix float_format not callable error

golowanow:
Fix float_format string passed to pandas to_csv() and to_html() to be a callable object resolving 'TypeError: 'str' object is not callable' observed (with pandas==2.1.4)


golowanow:
SOFCI TEST

PR #1298, opened 2025-09-02
alsa_settings: WCL platform settings

majunkier:
Create WCL platform ALSA settings state files for:

 * WCL_HDA_AIOC
 * WCL_SDW_RT712
 * WCL_RVP_SDW

The parameters in the state files are explicitly set even if the default values at the appropriate DUT are currently the same.

### DIFF
--- a/alsa_settings/WCL_HDA_AIOC.state
+++ b/alsa_settings/WCL_HDA_AIOC.state
@@ -1,0 +1,36 @@
+state.sofhdadsp {
+	control.1 {
+		iface MIXER
+		name 'Headphone Playback Volume'
+		value.0 60
+		value.1 60
+	}
+	control.2 {
+		iface MIXER
+		name 'Headphone Playback Switch'
+		value.0 true
+		value.1 true
+	}
+	control.6 {
+		iface MIXER
+		name 'Capture Volume'
+		value.0 30
+		value.1 30
+	}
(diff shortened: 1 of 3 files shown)


marc-hb (commented):
As usual I wonder why these are not already the default.

> The parameters in the state files are explicitly set even if
the default values at the appropriate DUT are currently the same.

Does that mean tests should pass even when all other users have bad settings?

KamilxPaszkiet (approved):

marc-hb:
>  In summary, some tests may modify ALSA settings without restoring defaults, which can lead to failures

There are (at least) two different questions:

1. **Who** cleans up and **when**. 
2. **What** cleanup is performed and **how**

The general answer to 1. is easy and well-known: in general each test should CLEAN-TEST-CLEAN. The first clean is for itself, the second clean is for other users: either interactive users or to help "bad" tests (which should be fixed eventually).

I don't think 1. is relevant to this PR. This PR is about 2.

> i believe most discussion was in https://github.com/thesofproject/sof-test/pull/1294

Yes, and based on the discussion there, why is `alsactl restore` not enough? Every time sof-test hides some "private" ALSA settings like this, it means people and machines who do not use sof-test will miss them. Sometimes it's necessary but it should ideally be temporary and always justified and documented.

PR #1303, opened 2025-09-22
latency-metrics.sh: Add latency threshold parameter

PLangowski:
Add parameter which sets a threshold for the average latency value. If `avg_ms` exceeds this value, the test will fail.

### DIFF
--- a/test-case/latency-metrics.sh
+++ b/test-case/latency-metrics.sh
@@ -40,53 +40,56 @@ TESTLIB="${TESTDIR}/case-lib"
 # shellcheck source=case-lib/lib.sh
 source "${TESTLIB}/lib.sh"
 
-OPT_NAME['R']='running-time'    OPT_DESC['R']='Test running time (in seconds) to collect latency probes.'
-OPT_HAS_ARG['R']=1              OPT_VAL['R']="30"
+OPT_NAME['R']='running-time'        OPT_DESC['R']='Test running time (in seconds) to collect latency probes.'
+OPT_HAS_ARG['R']=1                  OPT_VAL['R']="30"
 
-OPT_NAME['d']='device'          OPT_DESC['d']='ALSA pcm device for playback and capture. Example: hw:0'
-OPT_HAS_ARG['d']=1              OPT_VAL['d']=''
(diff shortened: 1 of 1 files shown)


PLangowski:
Tested - works

test-case/latency-metrics.sh
```diff
@@ -88,6 +88,9 @@ OPT_HAS_ARG['t']=0              OPT_VAL['t']=0
 OPT_NAME['T']='trial-until'     OPT_DESC['T']="Trial mode: repeat until this value."
 OPT_HAS_ARG['T']=1              OPT_VAL['T']=2
 
+OPT_NAME['l']='latency_threshold'   OPT_DESC['l']='latency threshold in ms'
+OPT_HAS_ARG['l']=1         OPT_VAL['l']=85
```
redzynix, line 92: Align all opts in lines starting from top.

test-case/latency-metrics.sh
```diff
@@ -40,53 +40,56 @@ TESTLIB="${TESTDIR}/case-lib"
...
+OPT_HAS_ARG['X']=1                  OPT_VAL['X']=0
 
-OPT_NAME['L']='loopback'        OPT_DESC['L']="Set internal loopback at JACK instead of ports."
-OPT_HAS_ARG['L']=0              OPT_VAL['L']=0
+OPT_NAME['L']='loopback'            OPT_DESC['L']="Set internal loopback at JACK instead of ports."
+OPT_HAS_ARG['L']=0                  OPT_VAL['L']=0
 
-OPT_NAME['t']='trial'           OPT_DESC['t']="Trial mode: repeat measurements decreasing buffer."
-OPT_HAS_ARG['t']=0              OPT_VAL['t']=0
+OPT_NAME['t']='trial'               OPT_DESC['t']="Trial mode: repeat measurements decreasing buffer."
```
majunkier, line 85: remove in all lines extra space

majunkier (changes requested):

test-case/latency-metrics.sh
majunkier: ok, right, now i see purpose of this adjustment

majunkier (approved):

PR #1305, opened 2025-10-01
latency: Fail when xruns exceed threshold

PLangowski:
Before this change, the test was being skipped after the threshold was exceeded. It is better to fail the test in this case.


kv2019i (approved):

marc-hb (approved):

marc-hb:
test this please

redzynix (approved):

PR #1307, opened 2025-10-06
latency: Fail test on unsuccessful option validation

PLangowski:
Tests will fail instead of skipping if wrong latency parameters are provided.


marc-hb (approved):

marc-hb:
test this please

PR #1333, opened 2025-11-28
tools: tplgtool2.py: Differentiate PCM and COMPR devices in pcm list

ujfalusi:
At the moment there is no information available about the type of the device (PCM or compr) in the pcm list.
Use the compress parameter of the PCM to differentiate the two types of devices.


redzynix (approved):

ujfalusi:
No conflicts still, can we merge this as it will help to see which ALSA device is PCM and which is for COMPR.

lgirdwood (approved):

PR #1345, opened 2026-02-19
test: add compressed audio socwatch test

ekurdybx:
Add test that plays MP3 sound and performs Socwatch analysis to check if dut enters and stays in PC10 state for the expected % of time (default expected value: 80% in PC10.2 state)
Note: this test will only pass when:
1) the changes for playing compressed audio will be merged
2) the currently known pc10 state issue will be resolved


tools/analyze-pc-states.py
```diff
@@ -0,0 +1,45 @@
...
+                pc_state = "PC"+str(pc_state_nr)
+                value = float(m.group(2))
+                expected_value = expected_results.get(pc_state)
+                if not expected_value:
+                    continue
+                if not compare_values(value, float(expected_value)):
+                    print(f"Incorrect value: {pc_state} time % was {value}, expected {expected_value}")
+                    failures += 1
+
+    return 0 if failures == 0 else 1
```
redzynix: It's cool but I think keeping it easier to read would be better.

```suggestion
    if failures == 0:
        return 0
    return 1    
```

redzynix (approved):

tools/analyze-pc-states.py
ekurdybx: done

tools/analyze-pc-states.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def analyze_pc_states(pc_states_file, expected_results):
+    pattern = re.compile(r'^PC(\d+)\s*,\s*(\d+(?:\.\d+)?)\s*,\s*(\d+(?:\.\d+)?)\s*$')
+    failures = 0
+
+    with open(pc_states_file, encoding="utf-8") as file:
+        for line in file:
+            m = pattern.match(line)
+            if m:
```
marc-hb: "early return" pattern
```suggestion
            if not m:
                continue
```
Same as you already did below but saves a lot more indentation (and mental load).

tools/analyze-pc-states.py
```diff
@@ -0,0 +1,47 @@
...
+                pc_state = "PC"+str(pc_state_nr)
+                value = float(m.group(2))
+                expected_value = expected_results.get(pc_state)
+                if not expected_value:
+                    continue
+                if not compare_values(value, float(expected_value)):
+                    print(f"Incorrect value: {pc_state} time % was {value}, expected {expected_value}")
+                    failures += 1
+
+    if failures:
```
marc-hb, line 37:
```suggestion
    return failures
```

Then you may not want the main script to exit the number of failures too because some values may have special meaning - but for sure this _function_ has nothing better to return than the number of failures.

tools/analyze-pc-states.py
```diff
@@ -0,0 +1,47 @@
...
+    if abs(real-expected) <= ACCEPTANCE_BUFFER:
+        return True
+    return False
+
+
+def analyze_pc_states(pc_states_file, expected_results):
+    pattern = re.compile(r'^PC(\d+)\s*,\s*(\d+(?:\.\d+)?)\s*,\s*(\d+(?:\.\d+)?)\s*$')
+    failures = 0
+
+    with open(pc_states_file, encoding="utf-8") as file:
```
marc-hb, line 21: You could move the `with open()` to the caller and save even more indentation.

tools/analyze-pc-states.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def compare_values(real, expected):
+    if abs(real-expected) <= ACCEPTANCE_BUFFER:
+        return True
+    return False
+
+
+def analyze_pc_states(pc_states_file, expected_results):
+    pattern = re.compile(r'^PC(\d+)\s*,\s*(\d+(?:\.\d+)?)\s*,\s*(\d+(?:\.\d+)?)\s*$')
```
marc-hb, line 15: Regexps are notoriously cryptic. What really helps in such cases is to give a few input examples in a comment.

test-case/test-compressed-audio.sh
```diff
@@ -0,0 +1,106 @@
...
+    if [ ! -f "$audio_filename" ]; then
+        dlogi "Generating audio file for the test..."
+        generate_mp3_file "$audio_filename" "$duration" "$channels_p"
+    fi
+}
+
+check_cplay_command()
+{
+    dlogi "${play_command[@]}"
+    if ! "${play_command[@]}"; then
```
marc-hb: Style nit:

```suggestion
    "${play_command[@]}" ||
          die ...
```

test-case/test-compressed-audio.sh
```diff
@@ -0,0 +1,106 @@
...
+{
+    pc_states_file="$LOG_ROOT/pc_states.csv"
+    touch "$pc_states_file"
+    results=$(grep "Platform Monitoring Technology CPU Package C-States Residency Summary: Residency" -A 10 < "$socwatch_output".csv)
+    echo "$results" | tee "$pc_states_file"
+
+    expected_results="{\"PC10.2\":$pc10_threshold}"
+
+    # Analyze if the % of the time spent in given PC state was as expected
+    if python3 "$SCRIPT_HOME"/tools/analyze-pc-states.py "$pc_states_file" "$expected_results"; then
```
marc-hb, line 57: `chmod a+x analyze-pc-states.py `
I'm afraid sof-test is not going to work on Windows anytime soon!
```suggestion
    if "$SCRIPT_HOME"/tools/analyze-pc-states.py "$pc_states_file" "$expected_results"; then
```

marc-hb (commented):
Very clean and easy to read! Minor suggestions below

PLangowski (approved):

tools/analyze-pc-states.py
ekurdybx: This func returns the result of the analysis - which is true or false.

test-case/test-compressed-audio.sh
ekurdybx: done

tools/analyze-pc-states.py
ekurdybx: done

tools/analyze-pc-states.py
ekurdybx: done

PR #1352, opened 2026-03-09
test-case: Implement one common method for kill process

SzymonRichert:
This PR replaces all direct usages of kill -9 with a new helper function, kill_process().
The new function introduces a safer and more controlled process‑termination flow, using graceful shutdown attempts before falling back to a forced kill.

Function was tested and it works:
image

### DIFF
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -945,6 +945,30 @@ check_alsa_tool_process()
     esac
 }
 
+kill_process()
+{
+    local pid="$1"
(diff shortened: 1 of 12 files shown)


arikgreen (approved):

marc-hb (commented):
This was badly needed indeed!

It's really sad Linux forces everyone to reinvent this https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html

redzynix (approved):

PR #1357, opened 2026-04-07
fix: tools: sof_perf_analyzer: ignore errors when decoding mtrace file

ekurdybx:
Add ignore errors option when decoding mtrace file in sof_perf_analyzer.py script.

### DIFF
--- a/case-lib/hijack.sh
+++ b/case-lib/hijack.sh
@@ -169,10 +169,7 @@ function func_exit_handler()
 
     # Only do performance analysis on test passed
     [ "x$exit_status" != "x0" ] || {
-        # On performance analysis return error, set exit code
-        if ! perf_analyze; then
-           exit_status=1
-        fi
+        perf_analyze || dlogw 'perf_analyze failed in exit handler'
     }
 
     # Only run when changing topology was initiated during the test
--- a/case-lib/lib.sh
+++ b/case-lib/lib.sh
@@ -1496,11 +1496,11 @@ perf_analyze()
         fi
         dlogc "$perf_cmd"
         eval "$perf_cmd" || {
-            dloge "SOF component performace analysis tool exit with error"
+            dlogw "SOF component performace analysis tool exit with error"
             return 1
         }
     else
-        dloge "Firmware trace file not found"
+        dlogw "Firmware trace file not found"
         return 1
     fi
 }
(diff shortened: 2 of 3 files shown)


tools/sof_perf_analyzer.py
```diff
@@ -173,7 +173,7 @@ def process_trace_file():
     #
     # pylint: disable=C0103
     ts_shift = 0
-    with open(args.filename, 'r', encoding='utf8') as file:
+    with open(args.filename, 'r', encoding='utf8', errors='ignore') as file:
```
marc-hb, line 176:
```suggestion
    with open(args.filename, 'r', encoding='utf8', errors='backslashreplace') as file:
```
`ignore` = "ignore bugs".

case-lib/hijack.sh
```diff
@@ -169,10 +169,7 @@ function func_exit_handler()
 
     # Only do performance analysis on test passed
     [ "x$exit_status" != "x0" ] || {
-        # On performance analysis return error, set exit code
-        if ! perf_analyze; then
-           exit_status=1
-        fi
+        perf_analyze
```
marc-hb:
```suggestion
        perf_analyze || dlogw 'perf_analyze failed in exit handler'
```

case-lib/lib.sh
```diff
@@ -1496,12 +1496,10 @@ perf_analyze()
         fi
         dlogc "$perf_cmd"
         eval "$perf_cmd" || {
-            dloge "SOF component performace analysis tool exit with error"
-            return 1
+            dlogw "SOF component performace analysis tool exit with error"
```
marc-hb, line 1499: It's more flexible and future-proof to leave that decision to the caller.

marc-hb (commented):
A commit message would not hurt?

tools/sof_perf_analyzer.py
ekurdybx: backslashreplace is only supported for writing to the file

case-lib/hijack.sh
ekurdybx: Ok done

case-lib/lib.sh
ekurdybx: Ok, done

redzynix (approved):

PLangowski (approved):

tools/sof_perf_analyzer.py
marc-hb: > backslashreplace is only supported for writing to the file

I just tried it and it worked4me. There is also `replace` and a few others that don't hide corrupted input -> much more time-consuming if something goes wrong.

PR #1360, opened 2026-05-07
Function tplg support

majunkier:
This commits adds multi-topology support for function tplg feature for SDW. Tests can accept colon- or comma-separated lists of topology files, and fixes several issues.

- **pipeline.sh** — new func_tplg_parse_and_validate() normalises TPLG separators, applies NO_HDMI/BT/DMIC_MODE filters, resolves paths and exports TPLG_FILES / TPLG_COUNT. func_pipeline_export() updated to pass multi-file list to sof-tplgreader.py, deduplicate pipelines by PCM device, and sort by numeric id.
- **hijack.sh** — cleanup leftover aplay/arecord/tinyplay/tinycap processes at exit to prevent EBUSY errors in subsequent tests.
- **lib.sh** — fix logger_disabled() guard on OPT_VAL['s'] to prevent "unary operator expected" crash in tests that don't declare -s.

**Test cases:**

- **verify-tplg-binary.sh / verify-pcm-list.sh** — accept multi-file TPLG input and deduplicate output 
- **simultaneous-playback-capture.sh** — rewritten pipeline detection to correctly identify both pipelines across multiple topology files.
- **volume-basic-test.sh** — merge and deduplicate kcontrols from all topology files before testing.
- **multiple-pipeline.sh** — track tested PCMs in associative array to avoid testing same physical device more than once.
- **check-sof-logger.sh** — fix logger binary selection when multiple copies exist in PATH; handle timeout exit code 124 and 125.
- **check-audio-equalizer.sh** — updated to support multi-topology input.
- **check-alsabat.sh** — drop duplicated initial block


case-lib/hijack.sh
```diff
@@ -12,6 +12,34 @@ function func_exit_handler()
...
     dlogi "Starting func_exit_handler($exit_status)"
 
+    # Cleanup audio processes (aplay/arecord/tinyplay/tinycap) that may be left running
+    # This prevents "Device or resource busy" errors in subsequent tests
+    case "$SOF_ALSA_TOOL" in
+        'alsa')
+            if pgrep aplay >/dev/null 2>&1 || pgrep arecord >/dev/null 2>&1; then
+                dlogi "Cleaning up leftover aplay/arecord processes"
+                pkill -9 aplay 2>/dev/null || true
+                pkill -9 arecord 2>/dev/null || true
```
marc-hb:
```suggestion
                pkill -9 arecord || true
```

case-lib/hijack.sh
```diff
@@ -12,6 +12,34 @@ function func_exit_handler()
 
     dlogi "Starting func_exit_handler($exit_status)"
 
+    # Cleanup audio processes (aplay/arecord/tinyplay/tinycap) that may be left running
```
marc-hb, line 15: Déjà vu...I feel like there are a couple instances of such cleanup functions around already? Sorry if not.

kv2019i (approved):
Thanks @majunkier !

case-lib/hijack.sh
```diff
@@ -12,6 +12,34 @@ function func_exit_handler()
...
+    # This prevents "Device or resource busy" errors in subsequent tests
+    case "$SOF_ALSA_TOOL" in
+        'alsa')
+            if pgrep aplay >/dev/null 2>&1 || pgrep arecord >/dev/null 2>&1; then
```
marc-hb:
```suggestion
            if pgrep aplay >/dev/null || pgrep arecord >/dev/null ; then
```

marc-hb (commented):
Discarding `stderr` in test code is really the "anti-pattern", an entire cattery dies every time :-(

- If everything goes well, there should be no leftover processes and this cleanup code will not even run so no need to discard anything.

- If there _are_ leftover processes, then a problem  happened somewhere and the entire purpose of test code is to find and show problems, not hide them.


Moreover, test logs are always big and ugly and painful to read anyway, so discarding one needle from that haystack is never going to make them look good.

case-lib/hijack.sh
majunkier: i dont look for other cleanup code block, there was some issues that require cleanup, i have tested this changes for some time and  i dont notice anything wrong. ofc there could be place for improvement to rewrite some functions but i dont feel it should be done with this PR

case-lib/hijack.sh
majunkier: done

case-lib/hijack.sh
majunkier: done

SzymonRichert (approved):
LGTM

PR #1361, opened 2026-06-05
tools: tplgtool2: handle graph-only feature topology nodes

ujfalusi:
Feature topologies may contain graph endpoints that do not map to a DAPM widget (for example, mixout.21.1).

Make tplgtool2 robust to these graph-only nodes instead of raising KeyError while building graph metadata. Keep the issue visible by emitting a warning and by rendering graph-only nodes with a distinct style and label in graph output.

This keeps parsing and rendering usable for feature topologies while still surfacing broken/incomplete graph<->widget mappings.

### DIFF
--- a/tools/tplgtool2.py
+++ b/tools/tplgtool2.py
@@ -787,7 +787,11 @@ class TplgGraph:
         names = {}
         for edge in graph_list:
             for ename in [edge["source"], edge["sink"]]:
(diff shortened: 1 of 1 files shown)


ujfalusi:
Closing in favor of https://github.com/thesofproject/sof-test/pull/1362, which extends on this

